PR #22, opened 2020-10-19
Support any built-in controller for boids

JVillella:
cc @Nickniyihan 

This PR,

* Updates some of the interfaces/plumbing between vehicle index and bubble management to allow any of the built-in controllers to be used for boids (e.g. TrajectoryTracking which was the initial ask)
* Pushes Visdom client down into SMARTS (from the envs) so that they're usable for social agents
* Supports visualizing boid agent in Visdom

---

Demo using trajectory tracking controller w/ boids,

trajectory-controller-boids

### DIFF
--- a/smarts/core/bubble_manager.py
+++ b/smarts/core/bubble_manager.py
@@ -315,8 +315,9 @@ class BubbleManager:
         mission_planner = MissionPlanner(
             sim.scenario.waypoints, sim.scenario.road_network
         )
+        is_boid = isinstance(social_agent_actor, BoidAgentActor)
         vehicle = sim.vehicle_index.prepare_for_agent_control(
-            sim, vehicle_id, agent_id, interface, mission_planner
+            sim, vehicle_id, agent_id, interface, mission_planner, boid=is_boid
         )
 
         # Setup mission (also used for observations)
@@ -353,8 +354,9 @@ class BubbleManager:
         self._log.debug(f"Hijack vehicle={vehicle_id} with actor={social_agent_actor}")
         agent_id = BubbleManager._make_social_agent_id(vehicle_id, social_agent_actor)
 
+        is_boid = isinstance(social_agent_actor, BoidAgentActor)
         vehicle = sim.vehicle_index.switch_control_to_agent(
-            sim, vehicle_id, agent_id, recreate=False
+            sim, vehicle_id, agent_id, boid=is_boid, recreate=False
         )
         self._update_cursor(vehicle_id, vehicle=vehicle)
 
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -39,6 +39,8 @@ from panda3d.core import (
 from sklearn.metrics.pairwise import euclidean_distances
 
 from envision import types as envision_types
+from envision.client import Client as EnvisionClient
+from .utils.visdom_client import VisdomClient
 
 from . import glsl
 from . import models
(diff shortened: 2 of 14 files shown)


smarts/env/rllib_hiway_env.py
```diff
@@ -37,8 +37,6 @@ class RLlibHiWayEnv(MultiAgentEnv):
                 used to specify a specific sumo port (default None)
             timestep_sec:
                 the step length for all components of the simulation (default 0.1)
-            visdom:
```
liamchzh, line 40: Why removed this comment?

liamchzh (approved):

smarts/env/rllib_hiway_env.py
JVillella: We weren't actually supporting this parameter at this point.

smarts/core/bubble_manager.py
```diff
@@ -353,8 +354,9 @@ def _hijack_social_vehicle_with_social_agent(
         self._log.debug(f"Hijack vehicle={vehicle_id} with actor={social_agent_actor}")
         agent_id = BubbleManager._make_social_agent_id(vehicle_id, social_agent_actor)
 
+        is_boid = isinstance(social_agent_actor, BoidAgentActor)
```
Gamenot, line 357: Same here as above.

smarts/core/smarts.py
```diff
@@ -267,6 +272,9 @@ def reset(self, scenario: Scenario):
         observations, _, _, _ = self._agent_manager.observe(self)
         observations_for_ego = self._agent_manager.reset_agents(observations)
 
+        # Visualization
+        self._try_emit_visdom_obs(observations)
```
Gamenot, line 276: If we are doing a visualisation here, is there any reason why envision should not also try?

smarts/core/bubble_manager.py
```diff
@@ -315,8 +315,9 @@ def _airlock_social_vehicle_with_social_agent(
         mission_planner = MissionPlanner(
             sim.scenario.waypoints, sim.scenario.road_network
         )
+        is_boid = isinstance(social_agent_actor, BoidAgentActor)
```
Gamenot, line 318: I think it might be best to leave the check for being a boid to the agent manager.

smarts/core/vehicle_index.py
```diff
@@ -71,6 +75,18 @@ def social_vehicle_ids(self):
         ]["vehicle_id"]
         return set(vehicle_ids)
 
+    def actor_is_boid(self, actor_id):
```
Gamenot, line 78: I would suggest that it should be the role of the agent manager to keep track of if the actor is a boid. I can see why this would be useful though.

Gamenot (changes requested):

Nickniyihan (approved):
Tested with Trajectory controller, lgtm

smarts/core/smarts.py
JVillella: We could, out of scope for the PR at this time. Some minor work would need to be required to support running Envision w/ missing fields (e.g. no scores or rewards).

smarts/core/bubble_manager.py
JVillella: That is the case, but routing through the vehicle index which tracks boid state. Here we're working off of SStudio types, which `AgentManager` knows nothing about.

Gamenot: Ok, that seems reasonable then.

Gamenot (approved):

PR #28, opened 2020-10-20
Fix twistd process leaking

### LINKED ISSUE #24: Social agent zoo leaking twistd process
After building a scenario with agents come from social agent zoo, `twistd` is still running.

https://github.com/huawei-noah/SMARTS/blob/master/cli/studio.py#L49
```
proc = subprocess.Popen("twistd", shell=True)
proc.terminate()
```

In this case, 2 processes will be spawned, one is the shell process and the other is the real twistd process (the shell's child process).

When `terminate` is called, only the shell process gets terminated.

liamchzh:
close #24

### DIFF
--- a/cli/studio.py
+++ b/cli/studio.py
@@ -47,8 +47,7 @@ def _build_single_scenario(clean, scenario):
             # Serve policies through the static file server, then kill after the
             # scenario has been created
             proc = subprocess.Popen(
-                f"twistd -n web --path {path}",
-                shell=True,
+                ["twistd", "-n", "web", "--path", path],
                 # Hide output to keep display simple
                 stdout=subprocess.DEVNULL,
                 stderr=subprocess.STDOUT,


cli/studio.py
```diff
@@ -46,8 +46,11 @@ def _build_single_scenario(clean, scenario):
         with pkg_resources.path(zoo.policies, "") as path:
             # Serve policies through the static file server, then kill after the
             # scenario has been created
+            cmd = f"twistd -n web --path {path}"
             proc = subprocess.Popen(
-                f"twistd -n web --path {path}",
+                # By adding `exec`, `Popen` returns the actual twistd process,
+                # which will be terminated later to avoid leaking.
+                "exec " + cmd,
                 shell=True,
```
davidrusu, line 54: How about we remove the `shell=True` instead?

```python
Popen(["twistd", "-n", "web", "--path", path],...)
```

liamchzh: Thanks. It works.

davidrusu (approved):
:ship:

Gamenot (approved):

PR #30, opened 2020-10-20
Add a copyright header to files

liamchzh:
Add a header to all *.py and *.js under cli, envision and smarts.

Examples and test folders were excluded.

### DIFF
--- a/envision/web/src/math.js
+++ b/envision/web/src/math.js
@@ -1,3 +1,23 @@
+// Copyright (C) 2020. Huawei Technologies Co., Ltd. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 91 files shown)


envision/web/src/math.js
```diff
@@ -1,3 +1,23 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
 // These standard set operations are taken from Mozilla's web documentation for developers:
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
```
Gamenot, line 22: I am slightly unsure what to do with this file since it is not ours to copyright.

edit:
I checked their website and it says it falls under public domain (CC0) in https://developer.mozilla.org/en-US/docs/MDN/About#Copyrights_and_licenses since even the oldest revision of the source is after the August 10th, 2012 cutoff. It can be preceded with `Any copyright is dedicated to the Public Domain. http://creativecommons.org/publicdomain/zero/1.0/`.

Gamenot (approved):

envision/web/src/math.js
junluo-huawei: Thanks very much @Gamenot for catching this one.

liamchzh: Yes. Thanks. This one should be excluded.

PR #43, opened 2020-10-21
Fix leak case where scenario cycle iterator caches scenario

Gamenot:
Python docs describes how the cycle iterator caches iterations to be reused once the base iterations have exausted: https://docs.python.org/3.7/library/itertools.html#itertools.cycle

For us, this meant that the scenario objects were being cached and that is a problem when given an unbounded number of scenarios.

### DIFF
--- a/smarts/core/scenario.py
+++ b/smarts/core/scenario.py
@@ -208,10 +208,8 @@ class Scenario:
                 scenario_roots.extend(Scenario.discover_scenarios(root))
         np.random.shuffle(scenario_roots)
 
-        return cycle(
-            Scenario.variations_for_all_scenario_roots(
-                scenario_roots, agents_to_be_briefed
-            )
+        return Scenario.variations_for_all_scenario_roots(
+            cycle(scenario_roots), agents_to_be_briefed
         )
 
     @staticmethod


smarts/core/scenario.py
```diff
@@ -208,10 +208,8 @@ def scenario_variations(
                 scenario_roots.extend(Scenario.discover_scenarios(root))
         np.random.shuffle(scenario_roots)
 
-        return cycle(
-            Scenario.variations_for_all_scenario_roots(
-                scenario_roots, agents_to_be_briefed
-            )
+        return Scenario.variations_for_all_scenario_roots(
+            cycle(scenario_roots), agents_to_be_briefed
```
aurorahcx, line 212: Does this mean variations_for_all_scenario_roots() will get run repeatedly? Is it costly? If it is, is there a way to first check whether the number of scenarios is unbounded then do this special handling?

Otherwise, this looks good to me!

liamchzh: When will we have an unbounded number of scenarios?

Gamenot: When I say unbounded I mean that there is no defined upper limit on the number of scenarios that can be passed to this method. I know that Ultra has the potential to load 100k or more different scenarios. This means that all of those scenarios get cached and quickly leads to SMARTS running out of memory.

Gamenot: > Does this mean variations_for_all_scenario_roots() will get run repeatedly? Is it costly? If it is, is there a way to first check whether the number of scenarios is unbounded then do this special handling?

The cycle in the outer loop was saving the elements and expanding to the same number of scenarios as was provided to cycle.

```python
def cycle(iterable):
    # cycle('ABCD') --> A B C D A B C D A B C D ...
    saved = []
    for element in iterable:
        yield element
        saved.append(element)
    while saved:
        for element in saved:
              yield element
```

Changing it to cycle through the scenario roots instead passes to the following method:

```python
@staticmethod
def variations_for_all_scenario_roots(scenario_roots, agents_to_be_briefed):
    for scenario_root in scenario_roots:
        ...
        yield Scenario(...)
```

So the result is that `variations_for_all_scenario_roots` will loop through all the provided scenarios and repeatedly construct the scenario objects without caching. I think some optimisation could be done here by caching some parts of the scenario `map.net.xml` data in the case that maps are reused.

davidrusu (approved):
looks good :ship:

aurorahcx (approved):

PR #44, opened 2020-10-21
Remove redundant TraCI connecting

### LINKED ISSUE #42: Remove redundant TraCI connecting/loading
After initializing TraCI connection, SMARTS updates a few params by using `traci_conn.load(load_params)`, which causes connecting SUMO twice. This will cause  some overhead costs especially when running a large map.

liamchzh:
close #42


liamchzh:
@JVillella `test_boids` is not deterministic, it failed due to different rand num.

Gamenot (approved):
This is an odd bug but I could understand why it might cause a refresh of the SUMO gui.

davidrusu (approved):

PR #48, opened 2020-10-22
Departing agent vehicles have precedence in SUMO departure

Gamenot:
This fixes a current issue with default spawning that can cause a vehicle to spawn in on top of a departing ego vehicle.

### DIFF
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -254,7 +254,9 @@ class AgentInterface:
                 action=ActionSpaceType.ActuatorDynamic,
             )
         elif requested_type == AgentType.Laner:  # The lane-following agent
-            interface = AgentInterface(waypoints=True, action=ActionSpaceType.Lane,)
+            interface = AgentInterface(
+                waypoints=True, rgb=True, ogm=True, action=ActionSpaceType.Lane,
+            )
         # The lane-following agent with speed and relative lane change direction
         elif requested_type == AgentType.LanerWithSpeed:
             interface = AgentInterface(
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -254,9 +254,7 @@ class AgentInterface:
                 action=ActionSpaceType.ActuatorDynamic,
             )
         elif requested_type == AgentType.Laner:  # The lane-following agent
-            interface = AgentInterface(
-                waypoints=True, rgb=True, ogm=True, action=ActionSpaceType.Lane,
-            )
+            interface = AgentInterface(waypoints=True, action=ActionSpaceType.Lane,)
         # The lane-following agent with speed and relative lane change direction
         elif requested_type == AgentType.LanerWithSpeed:
             interface = AgentInterface(
--- a/smarts/core/vehicle_index.py
+++ b/smarts/core/vehicle_index.py
@@ -237,7 +237,9 @@ class VehicleIndex:
             #       vehicle recreation seems to address the problem. Ideally we discover
             #       the underlaying problem and can go back to the preferred implementation
             #       of simply swapping control of a persistent vehicle.
-            return self._switch_control_to_agent_recreate(sim, agent_id, vehicle_id, boid)
+            return self._switch_control_to_agent_recreate(
+                sim, agent_id, vehicle_id, boid
+            )
 
         vehicle = self._vehicles[vehicle_id]
         ackermann_chassis = AckermannChassis(pose=vehicle.pose, bullet_client=sim.bc)
@@ -251,34 +253,6 @@ class VehicleIndex:
                 actor_id=agent_id,
                 shadow_actor_id="",
                 is_boid=boid,
-            )
-        )
-
-        return vehicle
-
-    def relinquish_agent_control(self, sim, vehicle_id, social_vehicle_id):
-        self._log.debug(
-            f"Relinquishing agent control v_id={vehicle_id} sv_id={social_vehicle_id}"
-        )
-        vehicle = self._vehicles[vehicle_id]
-        box_chassis = BoxChassis(
-            pose=vehicle.chassis.pose,
-            speed=vehicle.chassis.speed,
-            dimensions=vehicle.chassis.dimensions,
-            bullet_client=sim.bc,
-        )
-        vehicle.swap_chassis(box_chassis)
-        Vehicle.detach_all_sensors_from_vehicle(vehicle)
-
-        v_index = self._controlled_by["vehicle_id"] == vehicle_id
-        entity = self._controlled_by[v_index][0]
-        entity = _ControlEntity(*entity)
-        self._controlled_by[v_index] = tuple(
-            entity._replace(
-                actor_type=_ActorType.Social,
-                actor_id=social_vehicle_id,
-                shadow_actor_id="",
-                is_boid=False,
             )
         )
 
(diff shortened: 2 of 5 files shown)


smarts/core/vehicle_index.py
```diff
@@ -230,29 +230,34 @@ def switch_control_to_agent(
...
+            self._controlled_by[v_index] = tuple(
+                entity._replace(
+                    actor_type=_ActorType.Agent,
+                    actor_id=agent_id,
+                    shadow_actor_id="",
+                    is_boid=boid,
+                )
+            )
+        sim._traffic_sim.reserve_traffic_location_for_vehicle(
+            vehicle_id, vehicle.chassis.to_polygon
```
davidrusu, line 259: When `recreate=True`, vehicle will be None, this line should fail right? since we're trying to read `chassis` on `None`

smarts/core/vehicle_index.py
```diff
@@ -230,29 +230,34 @@ def switch_control_to_agent(
...
-
-        v_index = self._controlled_by["vehicle_id"] == vehicle_id
-        entity = _ControlEntity(*self._controlled_by[v_index][0])
-        self._controlled_by[v_index] = tuple(
-            entity._replace(
-                actor_type=_ActorType.Agent,
-                actor_id=agent_id,
-                shadow_actor_id="",
-                is_boid=boid,
+        else:
```
davidrusu, line 241: I wouldn't nest this in the `else`, you've introduced a `vehicle=None` above because you reserve space in the traffic sim below (line 258), but vehicle is only set to a non-None value in this `else` branch.

Either de-indent back to how it was or move the traffic sim reservation + `return vehicle` into the else branch. Either way we need to remove the `vehicle=None` above

Gamenot: The intention here was to forward the vehicle from `_switch_control_to_agent_recreate()` and accidentally got lost in a merge resolution.

I realise now this still did not accomplish what I was intending. A entry location should not be reserved when just taking over control of an existing vehicle.

smarts/core/vehicle_index.py
```diff
@@ -393,6 +388,11 @@ def build_agent_vehicle(
             boid,
         )
 
+        # Reserve space inside the traffic sim
+        sim._traffic_sim.reserve_traffic_location_for_vehicle(
```
Gamenot, line 392: It is better to reserve the vehicle location in SUMO inside building the vehicle because this should be the case for any time an ego vehicle is added.

davidrusu: @Gamenot You don't actually call this `build_agent_vehicle` from within `_switch_control_to_agent_recreate`. You have `Vehicle.build_agent_vehicle` instead of `self.build_agent_vehicle`. Is this intentional? It seems like we'd be skipping the traffic sim reservation step.

smarts/core/agent_interface.py
```diff
@@ -254,7 +254,9 @@ def from_type(requested_type: AgentType, **kwargs):
                 action=ActionSpaceType.ActuatorDynamic,
             )
         elif requested_type == AgentType.Laner:  # The lane-following agent
-            interface = AgentInterface(waypoints=True, action=ActionSpaceType.Lane,)
+            interface = AgentInterface(
+                waypoints=True, rgb=True, ogm=True, action=ActionSpaceType.Lane,
```
JVillella, line 258: We don't want to merge this change in (rgb and ogm should not be set to `True`)

smarts/core/vehicle_index.py
```diff
@@ -231,11 +231,38 @@ def switch_control_to_agent(
     ):
         self._log.debug(f"Switching control of {agent_id} to {vehicle_id}")
         if recreate:
+            # TODO: Fix the SUMO lane changing issue
+            # There exists a SUMO connection error bug that occurs
```
JVillella, line 235: Please fix the indentation

smarts/core/vehicle_index.py
```diff
@@ -372,7 +342,17 @@ def build_agent_vehicle(
             initial_speed,
         )
 
-        waypoint_paths = sim.waypoints.waypoint_paths_at(vehicle.position, lookahead=1)
+        if vehicle_exists:
+            old_vehicle = self._vehicles[vehicle_id]
```
JVillella, line 346: This code is a much lower level of abstraction, we should aim to keep the same abstraction levels for each line within a method. IOW, we should move this out into a sep. method.

smarts/core/vehicle_index.py
JVillella: This cross-talk between the vehicle and the traffic sim is breaking API boundaries. Traffic sim was private intentionally.

JVillella (changes requested):
Left a few design-related concerns.

smarts/core/vehicle_index.py
Gamenot: I realize now that if the departure ordering works out I might be able to remove this part and have it only in the traffic sim.

smarts/core/vehicle_index.py
```diff
@@ -343,6 +284,37 @@ def relinquish_agent_control(self, sim, vehicle_id, social_vehicle_id):
 
         return vehicle
 
+    def recreate_agent_vehicle(
```
JVillella, line 287: Why is this a public method?

smarts/core/vehicle_index.py
```diff
@@ -372,7 +347,17 @@ def build_agent_vehicle(
...
+            old_vehicle = self._vehicles[vehicle_id]
+            sensor_state = self._sensor_states[vehicle_id]
+            mission_planner = mission_planner or sensor_state.mission_planner
+            # Apply the physical values from the old vehicle chassis to the new one
+            vehicle.chassis.inherit_physical_values(old_vehicle.chassis)
+
+            # Remove the old vehicle
+            self.teardown_vehicles_by_vehicle_ids([vehicle_id])
+            # HACK: Directly remove the vehicle from the traffic provider
+            sim._traffic_sim.remove_traffic_vehicle(vehicle_id)
```
JVillella, line 360: We really need to avoid these hacks it breaks down the boundaries between classes. Instead, we should revise our interfaces as necessary to accomplish what was intended here. `VehicleIndex` is now directly talking to a provider. It should be going through the provider interface. At the very least, we could hub-and-spoke through SMARTS. But we certainly don't want to introduce touch points between the index and providers directly.

JVillella (changes requested):
See my comments about interfaces and API boundaries.

smarts/core/vehicle_index.py
Gamenot: This line is from:
https://github.com/huawei-noah/SMARTS/pull/48/files#diff-6113417ab844aea7f78d69bb51a162a0faecc6323fd5222cdaf55867386bc186L302

The main pain point here is that when trying to recreate is the timing of how the vehicles can be added. Currently it takes two steps for a provider added vehicle to get into the simulation and one step to delete a SUMO vehicle. It is also not possible to add a vehicle with the same id while another one exists. This is a problem because if trying to negotiate the problem through the provider it will take three steps total: two steps for the provider vehicle to reach the traffic sim, and then another step to delete the traffic vehicle. That extra delay has adverse effects elsewhere.

I think it might be possible to remove the `reserve_traffic_location_for_vehicle` call to the traffic sim but it perhaps needs more discussion as to if SMARTS should have the ability to call to remove a traffic vehicle.

smarts/core/vehicle_index.py
Gamenot: I reverted it back to the original private method.

Gamenot:
I think what I will end up doing with the vehicle index `add_vehicle` is bump out the provider-like aspect from here and move it to a proper provider that accesses the `VehicleIndex`.

smarts/core/vehicle_index.py
JVillella: Want to talk about this in our sync tonight?

PR #69, opened 2020-10-26
OpEn Agent packaged with pre-compiled solver (linux only)

davidrusu:
The open-agent wheel now contains a pre-compiled solver, eliminating the Rust run-time dependency.

Performance is now also much better, it's hard to do an apples to apples comparison since I've changed the cost equations since switching to bindings, but I'm seeing a solid 15-20fps increase now that we've switched to generating bindings directly.

The gains that are set as  defaults in the package have been tested on most scenarios that we include in SMARTS by default, they're not perfect, we still see occasional crashes or off-road, but overall they are pretty robust.


liamchzh (approved):

Gamenot (approved):

zoo/policies/open-agent/setup.py
```diff
@@ -1,11 +1,36 @@
...
 
-from open_agent.policy import VERSION
+from open_agent.version import VERSION
+
+try:
+    from open_agent.policy import compile_solver
+
+    compiled_extensions = compile_solver("./build")
+except ImportError:
+    compiled_extensions = []
```
Gamenot, line 11: If this is important it might be useful to add a warning here.

davidrusu: Done!

PR #81, opened 2020-10-27
Memory growth tests

Gamenot:
Some primitive tests to test for memory growth.


smarts/core/tests/test_smarts_memory.py
```diff
@@ -0,0 +1,190 @@
...
+    except Exception:
+        # Not testing for crashes
+        pass
+
+    gc.collect()
+    all_objects = muppy.get_objects()
+    end_size = muppy.get_size(all_objects)
+
+    # Check for a major leak
+    assert end_size - initial_size < 5e4
```
JVillella, line 118: (The other ones are 5e5)

JVillella (approved):
I like the idea

smarts/core/tests/test_smarts_memory.py
```diff
@@ -0,0 +1,190 @@
+import gc
+import time
+
+import gym
+from pympler import tracker, muppy, summary, classtracker
```
liamchzh, line 5: Have `tracker` `summary` & `classtracker` been used?

BTW, move this `from ... import ...` after `import pytest` will look better

smarts/core/tests/test_smarts_memory.py
```diff
@@ -0,0 +1,190 @@
...
+    _memory_buildup(
+        agent_id, seed, primative_scenarios, 100, agent_params[1], agent_params[2]
+    )
+
+    gc.collect()
+    initial_size = muppy.get_size(muppy.get_objects())
+
+    gc.collect()
+    try:
+        _memory_buildup(agent_id, seed, primative_scenarios, *agent_params)
```
liamchzh, line 108: when the training finishes (after the `env.close()`), the leaking objects will be destroyed?

Gamenot: They should be, from what I am seeing there is still a leak, although much less significant compared to what was happening earlier with the scenarios.

Gamenot:
From what I can tell, the tests are doing their job properly but we still have memory leaks. but at a rate significantly less than the size of the previous leaks.

The main leak is in the cleanup between using separate SMARTS instances rather than in the reset of episodes.

liamchzh (approved):

Gamenot:
These tests are catching memory leaks but the root memory leaks have not been tracked down so it should not go in yet.

smarts/core/remote_agent_buffer.py
```diff
@@ -34,12 +34,13 @@ def __init__(self, buffer_size=1):
...
         self._replenish_thread = None
         self._start_replenish_thread()
-        atexit.register(self.__del__)
+        atexit.register(self.destroy)
 
     def __del__(self):
         self.destroy()
 
     def destroy(self):
+        atexit.unregister(self.destroy)
```
Gamenot, line 43: `atexit` was a source of memory leak since it would capture the instance.

smarts/core/sensors.py
```diff
@@ -486,6 +486,11 @@ def _agent_vehicle_is_not_moving(cls, sim, agent_id):
 
         return False
 
+    @classmethod
+    def clean_up(cls):
+        cls._oncoming_traffic_edge.cache_clear()
+        cls._vehicle_off_route_info.cache_clear()
```
Gamenot, line 492: While `@lru_cache` is nice it looks like it can be useful to clear it in some cases.

davidrusu (approved):
Looks good, lets try it out

PR #83, opened 2020-10-27
Issue 68: ogm/drivable area are blank in the first few steps of agents

aurorahcx:
- Re-introduce Panda3D task manager to ensure update and render threads are synched every frame
- Update camera position before calling observe() to render grid maps
- Check that window resources are not leaking

Before vs. After (grey squares indicate vehicle positions from observations):
observations


JVillella:
We do _not_ want to be re-introducing the task manager at the sensor level. This was intentionally removed because it's a leaky abstraction. Furthermore, it complicates our step loop. We now have code that operates out of flow of the top-level loop.

Did you try the solution we were brainstorming as a team? If that's not working could you share why and we can consider another approach?

JVillella (changes requested):
We do not want to be re-introducing the task manager at the sensor level. This was intentionally removed because it's a leaky abstraction. Furthermore, it complicates our step loop. We now have code that operates out of flow of the top-level loop.

Did you try the solution we were brainstorming as a team? If that's not working could you share why and we can consider another approach?

smarts/core/smarts.py
```diff
@@ -226,6 +226,9 @@ def _step(self, agent_actions):
         # Agents
         self._agent_manager.step_agent_sensors(self)
 
+        # Panda3D
+        self.taskMgr.mgr.poll()  # update camera position in the render pipeline
+
```
aurorahcx, line 231: This is the change we brainstormed as a team, I thought we agreed that we had to use task manager to be able to synch update and render threads per frame?
If we do not use task manager to synch per frame, there can be mismatch across multiple steps.
@JVillella

Gamenot:
I never did figure out why the resources were not cleaned up. Perhaps it is the use of `partial()` here? If `panda3d` for some reason assumes that it needs to cache some methods it might not clean up the camera reference.

aurorahcx:
> I never did figure out why the resources were not cleaned but that the task manager was involved. Perhaps it is the use of `partial()` here? If `panda3d` for some reason assumes that it needs to cache some methods it might not clean up the camera reference in the task manager even after calling to remove the task.
> 
> Edit:
> 
> image
> 
> Otherwise it might have to do with the fact that the task manager is a global object.

Is this SMARTS code?

Gamenot:
Please also check that window resources are not leaking before merging this PR.

I did so before by using  `xwininfo -root -children | wc -l` or `xrestop`.

aurorahcx:
> We do not want to be re-introducing the task manager at the sensor level. This was intentionally removed because it's a leaky abstraction. Furthermore, it complicates our step loop. We now have code that operates out of flow of the top-level loop.
> 
> Did you try the solution we were brainstorming as a team? If that's not working could you share why and we can consider another approach?
> 
> (EDIT: Meant to request changes, not close)

Sorry I misunderstood, I thought I had to add back the task manager at the sensor level to make `self.taskMgr.mgr.poll()` work in smarts.py, I realized I don't need it to do this later. I will close this PR and open a new one.

Issue #72, opened 2020-10-26
Register vehicles type in traffic provider for other co-simulator


liamchzh:
When SMARTS integrate with other simulators, other simulators might need to register/create vehicles upfront.

The context is that when SMARTS integrate with another simulator, we need to supply a *.rou.xml to SUMO to register a vehicle type and a route with a specific id so that simulator can create a vehicle with the type.

A *.rou.xml looks like:
```
<routes xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://sumo.dlr.de/xsd/routes_file.xsd">
	<vType accel="0.8" color="0,0,1" decel="4.5" guiShape="passenger" height="1.5" id="IntelligentVehicle" length="5" maxSpeed="3" minGap="2.5" sigma="0.5" width="2"/>
	<route edges="A2-B2" id="ego_rou"/>
</routes>
```

This file is used to specify the vehicle type and route. No vehicle is created.

liamchzh:
vType:
* ID
* Dimension(length, width, height)

liamchzh:
Actually no need to supply `vType` upfront because SUMO will give the vehicle a default vehicle type. Even if KyberSim needs to change the color or dimension of the vehicle, also can do it via TraCI client.

PR #92, opened 2020-10-29
Register predefined vehicles type in scenario studio

liamchzh:


davidrusu (approved):

smarts/sstudio/types.py
```diff
@@ -559,9 +559,41 @@ class _ActorAndMission:
     mission: Union[Mission, EndlessMission, LapMission]
 
 
+@dataclass(frozen=True)
+class VehicleType:
```
Gamenot, line 563: How do you imagine the use of this this vehicle is distinguished from the use of traffic actors in scenario studio?

Gamenot (changes requested):
Just a comment to address before merging.

smarts/sstudio/types.py
liamchzh: Yes, they look similar because they are both leaking the SUMO's implementation details. For now, `VehicleType` is transparent to SMARTS, which means SMARTS just pass everything to the underlying implementation.

Gamenot (approved):

smarts/sstudio/types.py
Gamenot: OK, The duplication requires another pass on the scenario studio types, I think this can go in with another issue raised addressing these noted problems.

JVillella:
Spoke over chat about more appropriate abstractions to support external sub-sims for Scenario Studio.

JVillella (changes requested):
Spoke over chat about more appropriate abstractions to support sub-sim traffic definitions via Scenario Studio.

liamchzh:
From group discussion:

> Julian:
> I think this is a true hack just to get ChenDong's case working. Meanwhile, there is a real use case around this problem. In fact, someone at the lunch and learn actually asked for this.
> 
> Approaches to solve this: we agree that scenario studio should support 1) external 2) internal sub sim specific traffic (my vote) and we use the proper abstractions to do so. 3) sub sim external traffic support is done out of band (I.e. smarts and sstudio doesn't care or get involved)
> 
> I think we could do something like reuse the traffic language but specify who is providing said traffic and we have sub sim specific translators. E.g. how to go from scenario studio DSL to Carla files.
> 
> But the concept of a sumo pass through is setting us up for broken window theory. It's going to be easy to tack on other features using this mechanism and breaking our abstractions.
> 
> The sumo params presently there for vehicle properties is another hack but it was fairly localized and can be swapped out with some translation easily.

liamchzh:
close this PR(see https://github.com/huawei-noah/SMARTS/issues/72#issuecomment-721950462)

PR #94, opened 2020-10-29
WIP: Marl model zoo

KornbergFresnel:
register MARL agents to the model zoo.


KornbergFresnel:
@mg2015started have a check.

Issue #51, opened 2020-10-22
Ensure equal spacing along the path for waypoints


iman512003:

iman512003:
![waypoints](https://user-images.githubusercontent.com/27947083/97707967-6e1cee00-1a8e-11eb-8544-49728f4b4fac.png)
I noticed that also for  some locations where the lanes are coming together(where netedit modify the connections to become curves) the shape waypoints are stack together due to the fact that the lane segment is less than `spacing`.

PR #98, opened 2020-10-30
Remove cvx and quadprog from controller

iman512003:
This is to address #51.

### DIFF
--- a/smarts/core/controllers/trajectory_tracking_controller.py
+++ b/smarts/core/controllers/trajectory_tracking_controller.py
@@ -25,8 +25,6 @@ from numpy.linalg import matrix_power
 import numpy as np
 
 from scipy import signal
-from cvxopt import matrix, solvers
-from qpsolvers import solve_qp
 
 from smarts.core.chassis import AckermannChassis
 from smarts.core.utils.math import (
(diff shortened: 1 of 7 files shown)


smarts/core/controllers/trajectory_tracking_controller.py
```diff
@@ -531,25 +534,20 @@ def MPC(
...
-            matrix(np.eye(matrix_H.shape[0])),
-            matrix(np.ones((matrix_H.shape[0], 1))),
-        )
+        # This is the solution to the unconstrained optimization problem
+        # for the MPC.
+        unconstrained_optimal_solution = np.matmul(
+            np.linalg.inv(2 * matrix_H),
+            np.matmul(matrix_F, np.array([[lateral_error], [0], [heading_error], [0]]))
+            + np.reshape(matrix_F1, (-1, 1)),
+        )[0][0]
```
davidrusu, line 551: that's it? cool

davidrusu (approved):

smarts/core/controllers/trajectory_tracking_controller.py
```diff
@@ -25,8 +25,6 @@
 import numpy as np
 
 from scipy import signal
-from cvxopt import matrix, solvers
-from qpsolvers import solve_qp
```
Gamenot, line 29: Remember to also remove this from `setup.py` and `requirements.txt`.

Gamenot (changes requested):

smarts/core/controllers/trajectory_tracking_controller.py
liamchzh: There are other usages in `lane_following_controller.py` and `actuator_dynamic_controller.py`. @iman512003 Are you going to eliminate them in one PR?

iman512003: > Remember to also remove this from `setup.py` and `requirements.txt`.

Thanks, sure.

iman512003: > There are other usages in `lane_following_controller.py` and `actuator_dynamic_controller.py`. @iman512003 Are you going to eliminate them in one PR?

Oh, Thanks, sure. I removed them from other controller files as well.

smarts/core/controllers/trajectory_tracking_controller.py
iman512003: Thanks, Yes, basically, the MPC formulation results in the following QP:

min (1/2)x^T*H*x+F^T*X, subject to A*X<=B
  x

Solving the above full QP requires handling constraint(something like interior point method) but for unconstrained formulation(i.e ignoring A*X<=B) we can simply take the derivative which results in:

H*x+F=0 hence X=-H^-1 * F
Also, Autoware has a internal unconstrained solver similar to the above solution:

https://github.com/Autoware-AI/core_planning/blob/master/mpc_follower/src/qp_solver/qp_solver_unconstr.cpp

liamchzh (approved):

Gamenot (approved):

Issue #90, opened 2020-10-28
benchmark/agents/common  function cal_lane_its_info seems to have errors


feidieufo:
Even if there are many cars in intersections, the flag **intersection_distance** and **closest_its_nv_rel_speed**  is no different
I think line 624  has errors,         
```
line_heading = (
                (np.arctan2(nv_to_ego_vec[:, 1], nv_to_ego_vec[:, 0]) * 180 / np.pi)
                - 90
            ) % 360
 nv_to_line_heading = (nv_heading - line_heading) % 360
 ego_to_line_heading = (ego_heading - line_heading) % 360
```
I replace -90 to +90, the flag **intersection_distance** and **closest_its_nv_rel_speed** can have a effect
I think there may be other angle calculation errors in this file

Gamenot:
Taking a quick look, it seems like the `heading_to_degree` and `heading_to_vec` conversions have an errors converting from north facing heading to east facing degrees. Counter clockwise rotation should have the conversions subtracting 0.5 pi.

Hyperion-shuo:
> Taking a quick look, it seems like the `heading_to_degree` and `heading_to_vec` conversions have an errors converting from north facing heading to east facing degrees/radians. Counter-clockwise rotation should have the conversions subtracting 0.5 pi.
> 
> There is someone looking into it.

In line 68, it says heading_to_degree function takes 0 as up direction , but actually 0 means down direction.See code and picture blow:

    def heading_to_degree(heading):

    ### +y = 0 rad. Note the 0 means up direction ###

    return np.degrees((heading + math.pi) % (2 * math.pi))

<img width="412" alt="WechatIMG1251" src="https://user-images.githubusercontent.com/55169296/97670865-1c8e5680-1ac2-11eb-9320-15b138b3a49c.png">
<img width="239" alt="WechatIMG1252" src="https://user-images.githubusercontent.com/55169296/97670874-20ba7400-1ac2-11eb-839f-23572c3e8d96.png">

So line 301 minus 90 degree to convert **heading_to_vec** from east to north because it thinks **heading_to_degree** is north facing。Actually heading_to_degree is south facing, is should add 90.

mg2015started:
@feidieufo @Gamenot @Hyperion-shuo yes, this is a bug.

feidieufo:
@mg2015started  I found another problem,in common.py,when I use cruising.yaml, in get_reward_adapter function，
```
[...]

PR #100, opened 2020-10-31
fix heading to degree error(north 0, counterclockwise to 360)

mg2015started:
Fix heading_to_degree error. see issue #90

### DIFF
--- a/benchmark/agents/common.py
+++ b/benchmark/agents/common.py
@@ -92,7 +92,7 @@ def proximity_detection(OGM):
 
 def heading_to_degree(heading):
     # +y = 0 rad. Note the 0 means up direction
-    return np.degrees((heading + math.pi) % (2 * math.pi))
+    return np.degrees(heading % (2 * math.pi))
 
 
 def heading_to_vec(heading):


benchmark/agents/common.py
```diff
@@ -92,7 +92,7 @@ def proximity_detection(OGM):
 
 def heading_to_degree(heading):
     # +y = 0 rad. Note the 0 means up direction
-    return np.degrees((heading + math.pi) % (2 * math.pi))
+    return np.degrees(heading % (2 * math.pi))
```
Gamenot, line 95: Just to be clear: you want the result of the heading to degrees to also be in the `+y = 0` rather than `+x = 0` direction? If the rest of the code operates on that assumption it is OK but I think that might also be confusing.

benchmark/agents/common.py
```diff
@@ -92,7 +92,7 @@ def proximity_detection(OGM):
...
 
 
 def heading_to_vec(heading):
```
Gamenot, line 98: For this method if the intention is for this to convert a north bound heading to `x: right` and `y: up` the correction will need to be `- pi/2` or the phase will be off by `pi`.

Gamenot (changes requested):

Gamenot:
Feel free to merge if you think these changes address the problem.

benchmark/agents/common.py
mg2015started: yes, i do the calculation based on +y = 0 and clockwise increase to 360 degree. The old heading is degree in this way, so when i do migration, i just keep this way.

benchmark/agents/common.py
mg2015started: Converting to `x:right`, `y:up` should be `heading + pi/2`, since now is `x: up`, `y:left`.
Now we assume Up is 0 for heading and increase to 360 degree counterclockwise, following in a way as old heading calculation way.

mg2015started: Temporarily using this way, later i think a refactorization will be done to keep the calculation to align with  xy-coordinate system。

davidrusu:
@mg2015started Please also sign the CLA linked in https://github.com/huawei-noah/SMARTS/pull/100#issuecomment-719910825

Gamenot (approved):

benchmark/agents/common.py
Gamenot: OK

PR #105, opened 2020-11-02
Restart sumo process only when map file changes

liamchzh:
SMARTS forces the reinitialization of TRACI and the SUMO process to prevent the simulation from freezing. Though this change causes overhead costs, especially when switching between large maps, it solved the problem. 

A better way to decide whether to restart the SUMO process or not is based on the map. Ideally, we restart SUMO only if the map changes, otherwise, just reset everything.

### DIFF
--- a/smarts/core/sumo_traffic_simulation.py
+++ b/smarts/core/sumo_traffic_simulation.py
@@ -219,14 +219,19 @@ class SumoTrafficSimulation:
             "Can't setup twice, %s, see teardown()" % self._is_setup
         )
 
+        # restart sumo process only when map file changes
+        if self._scenario and self._scenario.net_filepath == scenario.net_filepath:
+            restart_sumo = False
+        else:
+            restart_sumo = True
+
         self._scenario = scenario
         self._log_file = scenario.unique_sumo_log_file()
 
-        # We force reinitialization of TRACI and the underlying SUMO process.
-        # This incurs ~50ms of overhead on an i9 Ubuntu desktop. But it prevents
-        # the simulation from freezing at a call to self._traci_conn.simulationStep()
-        # after dozens of scenario swapping. Issue #353.
-        self._initialize_traci_conn()
+        if restart_sumo:
+            self._initialize_traci_conn()
+        else:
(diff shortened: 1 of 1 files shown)


smarts/core/sumo_traffic_simulation.py
```diff
@@ -266,7 +271,6 @@ def teardown(self):
             return
 
         assert self._is_setup
-        self._close_traci_and_pipes()
```
Gamenot, line 269: If you are removing the traci close guarantee here I would suggest adding a destroy method to make sure that the pipes get closed.

Gamenot (changes requested):

smarts/core/sumo_traffic_simulation.py
liamchzh: Actually `initialize_traci_conn` will call `_close_traci_and_pipes` so I think the pipes are not leaking.

JVillella:
Just write as

```python
# restart sumo process only when map file changes
if self._scenario and self._scenario.net_filepath == scenario.net_filepath:
    self._traci_conn.load(self._base_sumo_load_params())
else:
    self._initialize_traci_conn()
```
?

JVillella (approved):
Seems fine 👍 (btw, can we use PR description to provide context?)

liamchzh:
> Seems fine  (btw, can we use PR description to provide context?)

For sure.

liamchzh:
> Just write as
> 
> ```python
> # restart sumo process only when map file changes
> if self._scenario and self._scenario.net_filepath == scenario.net_filepath:
>     self._traci_conn.load(self._base_sumo_load_params())
> else:
>     self._initialize_traci_conn()
> ```

Now I remember why I need a `restart_sumo`, because the `self._scenario` needs to be assigned first, so either an extra variable, 

or 
```
        if self._scenario and self._scenario.net_filepath == scenario.net_filepath:
            self._scenario = scenario
            self._traci_conn.load(self._base_sumo_load_params())
        else:
            self._scenario = scenario
            self._initialize_traci_conn()
```

Gamenot (approved):

PR #106, opened 2020-11-02
Iterating on Social Agent Zoo and packaged OpEn Agent for macOS

davidrusu:
When trying to integrate the OpEn agent into scenarios as a social agent, I ran into some rough spots with the SocialAgentActor::from_zoo() api.

I reworked it to using a `requirements.txt` in the scenario directory to specify any agent dependency a scenario may have. Then we can use the standard locator's as we do with other social agents


Gamenot:
I can see from the CI that smarts/core/remote_agent.py:86 is throwing an exception.

Gamenot:
Tried running the branch on `loop` and got the following: 
[crash.txt](https://github.com/huawei-noah/SMARTS/files/5483876/crash.txt)

smarts/zoo/run_agent.py
```diff
@@ -36,28 +36,36 @@
...
-import numpy
-import sklearn
-import shapely
-import scipy
-import trimesh
-import panda3d
-import gym
-import ray
+modules = [
+    "pybullet",
```
JVillella, line 51: I understand why we need this, but it's a bit unfortunate.

JVillella (approved):
Besides the CI crash, code looks 👍

Gamenot (approved):

Issue #73, opened 2020-10-26
 Restart the SUMO process based on map change


liamchzh:
A few months ago, we made a change that forces the reinitialization of TRACI and the SUMO process to prevent the simulation from freezing. This change causes overhead costs, especially when switching between large maps.

A better way to decide whether to restart the SUMO process or not is based on the map. Ideally, we restart SUMO only if the map changes, otherwise, just reset everything.

PR #127, opened 2020-11-05
Restart sumo process only when map file hash changes

liamchzh:
A further PR for #73.

In #105, the approach is to compare the `net_filepath`, which is bound with the scenario path. However, different scenarios can have the same maps, in this case, SUMO doesn't have to restart either, so use file hash to compare is more robust.

### DIFF
--- a/smarts/core/utils/file.py
+++ b/smarts/core/utils/file.py
@@ -76,12 +76,9 @@ def path2hash(file_path: str):
     return m.hexdigest()
 
 
-def have_same_hash(file_a: str, file_b: str) -> bool:
-    hasher_a = hashlib.md5()
-    hasher_b = hashlib.md5()
-    with open(file_a) as f:
-        hasher_a.update(f.read().encode())
-    with open(file_b) as f:
-        hasher_b.update(f.read().encode())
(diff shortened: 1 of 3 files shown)


aurorahcx (approved):

aurorahcx:
The change LGTM. It might not be related to this PR, but please note that the test is failing.

liamchzh:
Or cache the hash in `Scenario` instance

smarts/core/utils/file.py
```diff
@@ -74,3 +74,11 @@ def path2hash(file_path: str):
     m = hashlib.md5()
     m.update(bytes(file_path, "utf-8"))
     return m.hexdigest()
+
+
+def file_md5_hash(file_path: str) -> str:
```
JVillella, line 79: lrucache this? (learned that trick from @Gamenot)

JVillella (approved):

Gamenot (approved):

smarts/core/utils/file.py
liamchzh: I thought about this but I don't think caching is helping because scenario folders are not duplicated in most cases.

PR #136, opened 2020-11-09
Add accelerometer sensor

Gamenot:
Adds acceleration and jerk to the observations so that stateless actions have access to it without having to deal with the problem through the environment.

### DIFF
--- a/envision/client.py
+++ b/envision/client.py
@@ -160,6 +160,7 @@ class Client:
 
         def on_error(ws, error):
             self._log.warning("Unable to connect to Envision")
+            self._log.debug(error)
 
         def on_open(ws):
             setattr(threadlocal, "connection_established", True)
--- a/envision/client.py
+++ b/envision/client.py
@@ -159,8 +159,7 @@ class Client:
             self._log.debug("Connection to Envision closed")
 
         def on_error(ws, error):
-            self._log.warning("Unable to connect to Envision")
-            self._log.debug(error)
+            self._log.error(f"Connection to Envision terminated with: {error}")
 
         def on_open(ws):
             setattr(threadlocal, "connection_established", True)
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -97,6 +97,11 @@ class NeighborhoodVehicles:
     radius: float = None
 
 
+@dataclass
+class Accelerometer:
+    ...
+
+
 class AgentType(IntEnum):
     Buddha = 0
     """Agent sees nothing and does nothing"""
@@ -202,6 +207,8 @@ class AgentInterface:
     The choice of action space, this action space also decides the controller that will be enabled.
     """
 
+    accelerometer: Union[Accelerometer, bool] = True
+
     def __post_init__(self):
         self.neighborhood_vehicles = AgentInterface._resolve_config(
             self.neighborhood_vehicles, NeighborhoodVehicles
@@ -216,6 +223,9 @@ class AgentInterface:
         self.ogm = AgentInterface._resolve_config(self.ogm, OGM)
         self.rgb = AgentInterface._resolve_config(self.rgb, RGB)
         self.lidar = AgentInterface._resolve_config(self.lidar, Lidar)
+        self.accelerometer = AgentInterface._resolve_config(
+            self.accelerometer, Accelerometer
+        )
 
     @staticmethod
     def from_type(requested_type: AgentType, **kwargs):
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -99,7 +99,7 @@ class NeighborhoodVehicles:
 
 @dataclass
 class Accelerometer:
-    ...
+    pass
 
 
 class AgentType(IntEnum):
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -192,10 +192,10 @@ class Sensors:
         ego_vehicle_state = vehicle.state
 
         acceleration_params = {
-            "linear_acceleration": 0.0,
-            "angular_acceleration": 0.0,
-            "linear_jerk": 0.0,
-            "angular_jerk": 0.0,
+            "linear_acceleration": None,
+            "angular_acceleration": None,
+            "linear_jerk": None,
+            "angular_jerk": None,
         }
         if vehicle.subscribed_to_accelerometer_sensor:
             acceleration_values = vehicle.accelerometer_sensor(
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -968,15 +968,13 @@ class AccelerometerSensor(Sensor):
         self.linear_accelerations = deque(maxlen=3)
         self.angular_accelerations = deque(maxlen=3)
 
-    list()
-
     def __call__(self, linear_velocity, angular_velocity):
-        if not linear_velocity is None:
+        if linear_velocity is not None:
             self.linear_accelerations.append(linear_velocity)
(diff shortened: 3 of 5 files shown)


smarts/core/agent_interface.py
```diff
@@ -202,6 +207,8 @@ class AgentInterface:
     The choice of action space, this action space also decides the controller that will be enabled.
     """
 
+    accelerometer: Union[Accelerometer, bool] = True
```
liamchzh, line 210: If `Accelerometer` has no attributes, maybe just use a boolean flag, i.e., skip the `_resolve_config`

smarts/core/sensors.py
```diff
@@ -930,3 +961,36 @@ def paths_for_lane(self, lane, overflow_offset=None):
...
+        if len(self.linear_accelerations) < 3 or len(self.angular_accelerations) < 3:
+            return (linear_acc, angular_acc, None, None)
+
+        linear_jerk = linear_acc - last_linear_acc
+        angular_jerk = angular_acc - last_angular_acc
+
+        return (linear_acc, angular_acc, linear_jerk, angular_jerk)
+
+    def teardown(self):
+        ...
```
liamchzh, line 996: Use `pass` instead? I suppose it is really rare to use `...` in Python

smarts/core/sensors.py
```diff
@@ -930,3 +961,36 @@ def paths_for_lane(self, lane, overflow_offset=None):
...
     def teardown(self):
         pass
+
+
+class AccelerometerSensor(Sensor):
+    def __init__(self, vehicle, sim):
+        self.linear_accelerations = deque(maxlen=3)
+        self.angular_accelerations = deque(maxlen=3)
+
+    list()
```
liamchzh, line 971: What is this `list()` used for?

smarts/core/sensors.py
```diff
@@ -930,3 +961,36 @@ def paths_for_lane(self, lane, overflow_offset=None):
...
+
+    def __call__(self, linear_velocity, angular_velocity):
```
liamchzh, line 973: @iman512003 could you take a look at this?

smarts/core/sensors.py
```diff
@@ -930,3 +961,36 @@ def paths_for_lane(self, lane, overflow_offset=None):
...
+        if not linear_velocity is None:
```
liamchzh, line 974: -> `if linear_velocity is not None`

same below

smarts/core/sensors.py
```diff
@@ -930,3 +961,36 @@ def paths_for_lane(self, lane, overflow_offset=None):
...
+            self.linear_accelerations.append(linear_velocity)
+        if not angular_velocity is None:
+            self.angular_accelerations.append(angular_velocity)
+
+        if len(self.linear_accelerations) < 2 or len(self.angular_accelerations):
+            return (None, None, None, None)
+
+        linear_acc = self.linear_accelerations[0] - self.linear_accelerations[1]
+        last_linear_acc = self.linear_accelerations[1] - self.linear_accelerations[2]
```
liamchzh, line 983: Do we need to ensure `self.linear_accelerations` has at least 3 items?

liamchzh (changes requested):

envision/client.py
```diff
@@ -160,6 +160,7 @@ def on_close(ws):
 
         def on_error(ws, error):
             self._log.warning("Unable to connect to Envision")
+            self._log.debug(error)
```
JVillella, line 163: Remove this? Or log the error as an error?

smarts/core/agent_interface.py
```diff
@@ -97,6 +97,11 @@ class NeighborhoodVehicles:
     radius: float = None
 
 
+@dataclass
+class Accelerometer:
+    ...
```
JVillella, line 102: This is intentionally left blank, or you plan to add fields?

JVillella: Also ellipses are fine and cute, but project has been using `pass` until now.

smarts/core/sensors.py
JVillella: +1

smarts/core/sensors.py
```diff
@@ -71,6 +71,10 @@ class EgoVehicleObservation(NamedTuple):
     mission: Mission
     linear_velocity: np.ndarray
     angular_velocity: np.ndarray
+    linear_acceleration: np.ndarray
```
JVillella, line 74: I'm tempted to suggest we combine linear and angular into one field or velocity, accel, jerk in one field. But this may be a bad temptation :)

smarts/core/sensors.py
```diff
@@ -186,6 +190,31 @@ def observe(sim, agent_id, sensor_state, vehicle):
...
+        if vehicle.subscribed_to_accelerometer_sensor:
+            (
+                linear_acc,
+                angular_acc,
+                linear_jerk,
+                angular_jerk,
+            ) = vehicle.accelerometer_sensor(
+                ego_vehicle_state.linear_velocity, ego_vehicle_state.angular_velocity
+            )
+            acceleration_params.update(
```
JVillella, line 209: Another (bad?) temptation :)


```python
acceleration_params.update(dict(zip(
    ["linear_acceleration", "angular_acceleration", "linear_jerk", "angular_jerk"],
	vehicle.accelerometer_sensor(
		ego_vehicle_state.linear_velocity,
		ego_vehicle_state.angular_velocity
    )
))
```

smarts/core/agent_interface.py
Gamenot: It is intentionally blank.

envision/client.py
Gamenot: I will upgrade the warning to an error and merge the two logs as an error. The current without the error being included is misleading.

smarts/core/agent_interface.py
Gamenot: In this case it is an option that could easily have more attributes in the future. I think it is better here as is.

smarts/core/sensors.py
Gamenot: I thought I had but it does now.

smarts/core/sensors.py
Gamenot: I'll meet half way. :)

liamchzh (approved):

smarts/core/sensors.py
```diff
@@ -930,3 +961,31 @@ def paths_for_lane(self, lane, overflow_offset=None):
...
+        self.angular_accelerations = deque(maxlen=3)
+
+    def __call__(self, linear_velocity, angular_velocity):
+        if linear_velocity is not None:
+            self.linear_accelerations.append(linear_velocity)
+        if angular_velocity is not None:
+            self.angular_accelerations.append(angular_velocity)
+
+        if len(self.linear_accelerations) < 3 or len(self.angular_accelerations) < 3:
+            return (None, None, None, None)
```
davidrusu, line 978: default to 0.0, it'll simplify policy code

davidrusu (approved):
Just please change the response from `None -> 0.0` when we don't yet have enough data to compute accel/jerk

smarts/core/sensors.py
```diff
@@ -930,3 +961,31 @@ def paths_for_lane(self, lane, overflow_offset=None):
 
     def teardown(self):
         pass
+
+
+class AccelerometerSensor(Sensor):
+    def __init__(self, vehicle, sim):
```
aurorahcx, line 967: Are vehicle and sim used in this sensor?

PR #139, opened 2020-11-09
Resolve CI issues

Gamenot:
The summary of this is that we do not want to copy in the local file because it will add in a set version of SMARTS and whatever else is in the creator's SMARTS directory(whether they intend to or not.) The CI automatically pulls in the HEAD of master branch anyway and sets the entry-point to that directory.


aurorahcx (approved):

Dockerfile
```diff
@@ -63,14 +63,6 @@ ENV SUMO_HOME /usr/share/sumo
 COPY requirements.txt /tmp/requirements.txt
 RUN pip install --no-cache-dir -r /tmp/requirements.txt
 
-ENV PYTHONPATH=/src
-COPY . /src
-RUN cd /usr/local/lib/python3.7 && \
-    mkdir site-packages && \
-    cd site-packages && \
-    python /src/setup.py develop
-WORKDIR /src
```
davidrusu, line 72: @JVillella Didn't you add this to resolve some recursion limit crash?

JVillella (changes requested):
This was added specifically this way so that we can mount our source volume locally and have our changes automatically picked up via the editable pip package.

If it's not done this way, then mounting the source will override the egg links.

Gamenot:
OK, we are going to need to try something else then because we are copying in the source repository.

The current issue with this version is a static version of SMARTS is worked into the image and set-up at that point and so it fails when it goes through CI.

JVillella:
I also prefer _not_ to copy in the source directory. But why is that actually a problem? Isn't GH cloning the new latest source anyways?

Possible solutions,
* uninstall the 0.4.2-pre installed smarts in CI
* clone to a diff directory
* split between a base image (to provide dependencies) and a build image
* do nothing; are we sure this is actually the problem?

.github/workflows/ci-requirements.yml
```diff
@@ -0,0 +1,34 @@
+name: SMARTS CI
```
davidrusu, line 1: rename this file: `ci-formatting-and-dependencies.yaml`

davidrusu (approved):
Just a nit on file naming

.github/workflows/ci-requirements.yml
liamchzh: +1

PR #141, opened 2020-11-09
Add timeout to Envision client theads while waiting termination

### LINKED ISSUE #128: hiway_env close block
use SMARTS master version
in zoo/evaluation direcotry use run_evaluation.py to run multiple scenarios and metrics, but block in first scenario env close.

liamchzh:
close #128.

@charleschen6 reported the teardown of SMARTS instance was blocked by the hanging of Envision client thread. Add timeout to `thread.join()` to force terminating the thread.

@charleschen6 If you can find out what caused the hanging, would help to improve the WebSocket connection.

### DIFF
--- a/envision/client.py
+++ b/envision/client.py
@@ -214,9 +214,9 @@ class Client:
         self._state_queue.put(Client.QueueDone())
         self._logging_queue.put(Client.QueueDone())
         if self._thread:
-            self._thread.join()
+            self._thread.join(timeout=3)
             self._thread = None
 
         if self._logging_thread:
-            self._logging_thread.join()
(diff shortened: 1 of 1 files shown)


JVillella (approved):

envision/client.py
```diff
@@ -214,9 +214,9 @@ def teardown(self):
         self._state_queue.put(Client.QueueDone())
         self._logging_queue.put(Client.QueueDone())
         if self._thread:
-            self._thread.join()
+            self._thread.join(timeout=3)
```
Gamenot, line 217: Do the threads just hang around indefinitely if they timeout?

Gamenot (approved):

PR #143, opened 2020-11-09
Equalize the distance between waypoints

iman512003:
This is to address #51 . 
col5


smarts/core/waypoints.py
```diff
@@ -256,7 +256,81 @@ def _waypoints_starting_at_waypoint(
...
+        normalized_distant = np.linspace(0, 1, lookahead + 1)
+
+        (
+            ref_waypoints_coordinates["ref_wp_positions_x"],
+            ref_waypoints_coordinates["ref_wp_positions_y"],
+            ref_waypoints_coordinates["ref_wp_headings"],
+        ) = (
+            ref_wp_x_interp(normalized_distant),
+            ref_wp_y_interp(normalized_distant),
+            ref_wp_heading_interp(normalized_distant),
```
davidrusu, line 303: There's probably a bug here, this is a linear interpolation along the heading. ie.

if the first waypoing has a heading of 0.5 degrees and the next waypoint has a heading of 359.4, the interpolated heading will do a 360.

smarts/core/waypoints.py
```diff
@@ -256,7 +256,81 @@ def _waypoints_starting_at_waypoint(
...
+                            [
+                                ref_waypoints_coordinates["ref_wp_positions_x"][idx],
+                                ref_waypoints_coordinates["ref_wp_positions_y"][idx],
+                            ]
+                        ),
+                        heading=Heading(
+                            ref_waypoints_coordinates["ref_wp_headings"][idx]
+                        ),
+                        lane_width=waypoint.wp.lane_width,
+                        speed_limit=waypoint.wp.speed_limit,
```
davidrusu, line 324: I think we need to interpolate these values as well.

smarts/core/waypoints.py
```diff
@@ -256,7 +256,81 @@ def _waypoints_starting_at_waypoint(
...
+        )
+
+        ref_waypoints = [[]]
+
+        for idx, waypoint in enumerate(waypoint_paths[0]):
+
+            ref_waypoints[0].append(
+                LinkedWaypoint(
+                    wp=Waypoint(
+                        id=waypoint.wp.id,
```
davidrusu, line 313: Do you know what waypoint id's are used for? maybe we can just get rid of this field

smarts/core/waypoints.py
```diff
@@ -256,7 +256,81 @@ def _waypoints_starting_at_waypoint(
...
+                            ]
+                        ),
+                        heading=Heading(
+                            ref_waypoints_coordinates["ref_wp_headings"][idx]
+                        ),
+                        lane_width=waypoint.wp.lane_width,
+                        speed_limit=waypoint.wp.speed_limit,
+                        lane_id=waypoint.wp.lane_id,
+                        lane_index=waypoint.wp.lane_index,
+                        right_of_way=waypoint.wp.right_of_way,
```
davidrusu, line 327: I don't think anyone uses the `right of way` signal. Lets delete it cause I don't trust it to be accurate here.

smarts/core/waypoints.py
```diff
@@ -256,7 +256,81 @@ def _waypoints_starting_at_waypoint(
...
+                                ref_waypoints_coordinates["ref_wp_positions_y"][idx],
+                            ]
+                        ),
+                        heading=Heading(
+                            ref_waypoints_coordinates["ref_wp_headings"][idx]
+                        ),
+                        lane_width=waypoint.wp.lane_width,
+                        speed_limit=waypoint.wp.speed_limit,
+                        lane_id=waypoint.wp.lane_id,
+                        lane_index=waypoint.wp.lane_index,
```
davidrusu, line 326: this also feels a bit unsafe to do, maybe we can map a cumulative distance along a path to a lane_id/lane_index?

smarts/core/waypoints.py
```diff
@@ -256,7 +256,81 @@ def _waypoints_starting_at_waypoint(
...
+            ref_wp_y_interp(normalized_distant),
+            ref_wp_heading_interp(normalized_distant),
+        )
+
+        ref_waypoints = [[]]
+
+        for idx, waypoint in enumerate(waypoint_paths[0]):
+
+            ref_waypoints[0].append(
+                LinkedWaypoint(
```
davidrusu, line 311: Do we need to wrap in a `LinkedWaypoint`? if we can avoid that, then we don't need to fill the `nexts` field.

davidrusu (changes requested):
nice work, I think this raises some questions about what fields we should have on our waypoints. We can use this opportunity to simplify the api surface here.

smarts/core/waypoints.py
```diff
@@ -256,7 +256,81 @@ def _waypoints_starting_at_waypoint(
...
+        for waypoint in waypoint_paths[0]:
+            ref_waypoints_coordinates["ref_wp_positions_x"].append(waypoint.wp.pos[0])
+            ref_waypoints_coordinates["ref_wp_positions_y"].append(waypoint.wp.pos[1])
+            ref_waypoints_coordinates["ref_wp_headings"].append(
+                waypoint.wp.heading.as_bullet
+            )
+
+        # To ensure that the distance between waypoints are equal, we used
+        # interpolation approach inspired by:
+        # https://stackoverflow.com/questions/51512197/python-equidistant-points-along-a-line-joining-set-of-points/51515357
```
liamchzh, line 274: https://stackoverflow.com/a/51515357 this one is shorter :smiley:

davidrusu:
From discussion:
* Remove `Waypoint::id`
* Remove `Waypoint::right_of_way`

Gamenot:
The data replay failure I saw may raise a bit of caution.

smarts/core/waypoints.py
iman512003: Thanks, sure, I added np.unwrap before the interpolation to resolve this issue.

smarts/core/waypoints.py
iman512003: Thanks, sure.

smarts/core/waypoints.py
iman512003: Thanks, I removed id from waypoints.

smarts/core/waypoints.py
iman512003: Thanks, I removed `right of way `from waypoints.

smarts/core/waypoints.py
iman512003: Thanks, I added a map with cumulative distance to both lane_id/lane_index.

smarts/core/waypoints.py
iman512003: Oh Thanks, sure. I think I reached to that page from another question.

smarts/core/waypoints.py
iman512003: Thanks, I wanted to return it the same way as before since we also using it in the controller and we might need them somewhere so I tried to keep the return structure the same.

davidrusu: my concern is that the `nexts` is meaningless here.

If you look at the call sites of this function you can see that we immediately unlink the waypoints. Lets change the API  of this function to return unlinked waypoints and then update the call sites to work with the unlinked waypoints.

smarts/env/tests/test_determinism.py
```diff
@@ -138,7 +138,7 @@ def check_callback(rewards, agent_obs, dones, index):
             agent_obs.waypoint_paths, orig_agent_obs.waypoint_paths
         ):
             for now, prev in zip(w_paths_now, w_paths_prev):
-                assert now.id == prev.id
+                assert hash(now) == hash(prev)
```
davidrusu, line 141: Implement `__eq__` on Waypoint instead, that way we can simply compare waypoints i.e. `now == prev` and you can remove all the remaining equality checking code below since it's just checking what `__eq__` will check.

davidrusu (changes requested):
A few more rounds of changes needed, but I think we're getting close

smarts/core/sensors.py
```diff
@@ -826,7 +826,7 @@ def append_waypoint_if_new(self, new_wp):
             or wp_edge in self._mission_planner.route.edges
         )
 
-        if most_recent_wp.id != new_wp.id and should_count_wp:
+        if hash(most_recent_wp) != hash(new_wp) and should_count_wp:
```
davidrusu, line 829: impl `Waypoint::__eq__()` and use that instead

davidrusu (approved):
:ship:

iman512003 (commented):
> ship

Thanks @davidrusu  for all the comments.

smarts/core/waypoints.py
```diff
@@ -76,6 +74,30 @@ def signed_lateral_error(self, p):
     def pose(self):
         return Pose.from_center(tuple(self.pos), self.heading)
 
+    def __hash__(self):
```
JVillella, line 77: Is the built-in hash and equal from `dataclass(...)` not sufficient?

JVillella: Ideally it is so we don't need to maintain our own.

JVillella (approved):
Looks good, excited about getting this in :)

liamchzh (approved):

smarts/core/waypoints.py
iman512003: Thanks, If we only implement hash then for some of the tests assertions, we need to use expressions like `hash(some_wp)==hash(other_wp)` as it is mentioned by @davidrusu , by defining `eq` then those expression simplifies to `some_wp==other_wp`.

iman512003: Thanks, unfortunately, the use of numpy array in the waypoints results in ambiguity error for the built in hash, that's why I had to implement our version of the hash.

iman512003: > Looks good, excited about getting this in :)

Thank you so much, @JVillella

PR #144, opened 2020-11-09
History provider adopts the dimensions provided by the history data

### LINKED ISSUE #135: Support vehicle size customization
In the NGSIM dataset, the length and width of each vehicle are given and vary.

The current version of SMARTS supports only fixed-sized vehicles.

I think we need to support the creation of vehicles with specified length and width.

liamchzh:
Close #135 

In the past, the history provider uses the passenger vehicle's dimensions as the default dimensions. I left a TODO in the previous MR, at that time I was thinking to have a way to map the vehicle type and dimensions of the history data to SMARTS internal vehicle type. Turns out I was overthinking the problem, I found that the other providers (like SUMO) and observation will honor the dimensions that the history provider provides in the `VehicleState`.

### DIFF
--- a/smarts/core/traffic_history_provider.py
+++ b/smarts/core/traffic_history_provider.py
@@ -96,8 +96,22 @@ class TrafficHistoryProvider:
                         [*vehicle_state["position"][:2], 0,],
                         Heading(vehicle_state["heading"]),
                     ),
-                    # TODO: specify dimensions
-                    dimensions=VEHICLE_CONFIGS[vehicle_type].dimensions,
+                    dimensions=BoundingBox(
+                        length=vehicle_state.get(
+                            "vehicle_length",
+                            VEHICLE_CONFIGS[vehicle_type].dimensions.length,
+                        ),
+                        width=vehicle_state.get(
+                            "vehicle_width",
+                            VEHICLE_CONFIGS[vehicle_type].dimensions.width,
+                        ),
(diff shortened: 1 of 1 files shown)


Gamenot (approved):

smarts/core/traffic_history_provider.py
```diff
@@ -96,8 +96,22 @@ def step(self, provider_actions, dt, elapsed_sim_time) -> ProviderState:
...
+                        ),
+                        width=vehicle_state.get(
+                            "vehicle_width",
+                            VEHICLE_CONFIGS[vehicle_type].dimensions.width,
+                        ),
+                        # Note: Neither NGSIM nor INTERACTION provide
+                        #       the height of the vehicles.
+                        height=vehicle_state.get(
+                            "vehicle_height",
+                            VEHICLE_CONFIGS[vehicle_type].dimensions.height,
```
Gamenot, line 112: The one thing I can question is the usefulness of the defaults here. Perhaps it might be better to just assert.

davidrusu (approved):

smarts/core/traffic_history_provider.py
liamchzh: But `height` is not provided in the dataset. So for the length, width, height, I wanted the assigning to be consistent, if not provided, use a default value.

PR #147, opened 2020-11-10
Add support for "persistent boids"

### LINKED ISSUE #25: Support for persistent boids
Split from: https://github.com/huawei-noah/SMARTS/issues/17

The intended behavior with this feature would be for boid agents to get created once per episode and stay alive regardless of all vehicles exiting the boid. For discussion: if we want to make this behavior optional, or the only option.

JVillella:
closes #25 

cc @Nickniyihan 

Persistent boids are spun up a the beginning of the episode and are "kept alive" until the end of the episode. This way vehicles coming in and out of the bubble are controlled by the same agent.

### DIFF
--- a/scenarios/loop/scenario.py
+++ b/scenarios/loop/scenario.py
@@ -51,6 +51,7 @@ gen_scenario(
                     open_agent_actor, mission_group="all"
                 ),
                 follow_offset=(-7, 10),
+                keep_alive=True,
             ),
         ],
     ),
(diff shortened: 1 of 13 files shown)


Gamenot:
Please note all tests are failing.

JVillella:
Yes, a few things need small reworking. Doing that now. The code won't change much.

JVillella:
Tests are passing locally but `test_boids.py` seems to be a bit finicky. I'll monitor; it may not be deterministic, but certainly is brittle.

Nickniyihan (approved):
keep alive feature works fine with me. However, it appears that from the beginning of an episode, the boid agent is stepped every smarts step, even if there is no vehicle in the bubble. Is this intended?

PR #148, opened 2020-11-10
Disable model cache when loading road network glb

davidrusu:
Hopefully this helps with the memory leaks we were noticing.

NB. I didn't profile this, but perhaps we can get this patch tried out to see if it helps things? @Gamenot


Gamenot (approved):
Seems reasonable, does it help?

aurorahcx (approved):

davidrusu:
> Seems reasonable, does it help?

hope so, we'll see

PR #149, opened 2020-11-10
112 add trap default emit speed

Gamenot:
This change gives an option for some departing speed even if the trap manager defaults to emit a vehicle when it cannot find a vehicle to hijack.


smarts/core/trap_manager.py
```diff
@@ -342,15 +333,16 @@ def _mission2trap(self, road_network, mission, default_zone_dist=6):
                 n_lanes=1,
             )
 
-        out_config = TrapConfig(
```
liamchzh, line 345: Remove `TrapConfig`? Even the whole `trap_config.py` ?

smarts/core/chassis.py
```diff
@@ -509,9 +515,11 @@ def step(self, current_simulation_time):
 
     def inherit_physical_values(self, other: BoxChassis):
         self.set_pose(other.pose)
-        # bullet wants m/s
-        velocity = radians_to_vec(self.pose.heading) * other.speed
-        self.speed = other.speed
+        self.reset_speed(other.speed)
+
+    def reset_speed(self, speed: float):
```
liamchzh, line 520: Is `set_speed` a better naming? And since the chassis already has `speed.setter`, why adding a `reset_speed`?

smarts/core/trap_manager.py
```diff
@@ -174,7 +163,7 @@ def largest_vehicle_plane_dimension(vehicle):
                 vehicle = vehicles[v_id]
                 point = Point(vehicle.position)
 
-                if any(v.startswith(prefix) for prefix in trap.exclusion_prefixes):
+                if any(v_id.startswith(prefix) for prefix in trap.exclusion_prefixes):
```
liamchzh, line 166: Finally fix this :smile:

smarts/core/trap_manager.py
Gamenot: Certainly.

smarts/core/chassis.py
Gamenot: In this case the reason is that this method should only be called before the chassis sees any operation on it.

liamchzh (approved):

smarts/core/chassis.py
Gamenot: I ended up removing it.

davidrusu (approved):
Looks good :ship:

smarts/core/chassis.py
```diff
@@ -509,9 +509,11 @@ def step(self, current_simulation_time):
...
         self.set_pose(other.pose)
-        # bullet wants m/s
-        velocity = radians_to_vec(self.pose.heading) * other.speed
-        self.speed = other.speed
+        self._initialize_speed(other.speed)
+
+    def _initialize_speed(self, speed: float):
+        self.speed = speed
+        velocity = radians_to_vec(self.pose.heading) * speed
         self._client.resetBaseVelocity(self._bullet_id, [*velocity, 0])
```
iman512003, line 517: This is nice. We can also enhance the performance by reseting the rotational velocities of the wheels if it has some lags.

iman512003 (approved):

PR #155, opened 2020-11-11
WIP: Add desired trajectory sensor

liamchzh:
**This is a draft.** 

What we want to accomplish here is to provide a desired trajectory in the obs according to the task (u-turn, cut-in) that the agent is doing.

To be discussed:
1. Ways to pass a specified task to an agent? (
2. Is it allowed for agents to choose whether they need the desired trajectory in observation during runtime?

For question 1, I can see 3 ways to do it:
1. At the mission level
2. By `SocialAgentActor` config
3. By agent interface


liamchzh:
Finalized the work in #176, close this one.

Issue #75, opened 2020-10-26
Heterogeneous Agent Computing


Gamenot:
We want 5 social agents all on different machines contributing towards one simulation.

## core
- [x] Seamless hijack of vehicles going through bubbles.
- [x]  A single agent in the simulation from a remote machine
- [x] 5 agents from remote machines

#### prep work
- [x] #88 Merge `Agent` and `Policy`
- [x] #89 `RemoteAgent` to use `async def act(obs)` instead of `send_obs()`, `recv_act()`
- [x] #53   Pool social agents by `agent_specs`.

### networking
- [ ]

davidrusu:
A first implementation completed in #236

PR #162, opened 2020-11-16
Merge agent and agent policy

### LINKED ISSUE #88: Merge Agent and Policy
Currently Agent simply wraps AgentPolicy, if we can merge them we can simplify our API a little.

* remove special case for social agents applying adapters in the agent class
* if we can avoid it, try not to break user code, add a deprecation warning if possible.

davidrusu:
This PR merges Agent + AgentPolicy. The motivation for this change is to simplify our API surface area before we take on heterogeneous agent computing directly in #75.

AgentPolicy and Agent were nearly identical classes, doesn't make sense to have both.

I've made sure to keep these changes backwards compatible. We print noisy messages to our users to switch to the `Agent` class if we detect old api useage, but from a functional point of view, things should just work with the old API as well.

fixes #88

### DIFF
--- a/benchmark/default_model.py
+++ b/benchmark/default_model.py
@@ -1,6 +1,6 @@
 """
 This file contain default network for rllib training,
-and can be used for policy evaluation
+and can be used for agent evaluation
 """
 import pickle
 import tensorflow as tf
@@ -21,7 +21,7 @@ tf1, tf, tfv = try_import_tf()
 BASE_DIR = Path(__file__).expanduser().absolute().parent.parent
 
 
-class RLLibTFCheckpointPolicy(Agent):
+class RLLibTFCheckpointAgent(Agent):
     def __init__(self, load_path, algorithm, policy_name, yaml_path):
         load_path = str(load_path)
         if algorithm == "ppo":
@@ -141,7 +141,7 @@ class RLLibTFCheckpointPolicy(Agent):
         return action
 
 
-class RLLibTFSavedModelPolicy(Agent):
+class RLLibTFSavedModelAgent(Agent):
     def __init__(self, load_path, algorithm, policy_name, observation_space):
         load_path = str(load_path)
         self._prep = ModelCatalog.get_preprocessor_for_space(observation_space)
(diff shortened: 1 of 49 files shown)


liamchzh (approved):

benchmark/default_model.py
```diff
@@ -21,7 +21,7 @@
 BASE_DIR = Path(__file__).expanduser().absolute().parent.parent
 
 
-class RLLibTFCheckpointPolicy(AgentPolicy):
+class RLLibTFCheckpointPolicy(Agent):
```
davidrusu, line 24: Rename this to `RLLibTFCheckpointAgent`

smarts/core/agent.py
```diff
@@ -19,73 +19,57 @@
...
-    def act_with_adaptation(self, env_observation):
-        observation = self.observation_adapter(env_observation)
-        policy_action = self.act(observation)
-        action = self.action_adapter(policy_action)
-        return action
+# Remain backwards compatible with existing Agent's
+class AgentPolicy(Agent):
+    # we cannot use debtcollector here to signal deprecation because the wrapper object is
+    # not pickleable, instead we simply print.
+    print("[DEPRECATED] AgentPolicy has been replaced with `smarts.core.agent.Agent`")
```
JVillella, line 72: log a warning (via global logger) vs a print?

smarts/core/agent.py
```diff
@@ -131,33 +117,32 @@ class AgentSpec:
...
-            #       them proactively.
-            #
-            #       perform a self-test here to ensure things are rigged
-            #       up properly. ie.  generate some fake obs based on the
-            #       interface, pass it through the obs adapter, run that
-            #       through the policy, pass the policy action through
-            #       the action_adapter and ensure the controller accepts
-            #       the produced action.
-            del policy
+            print(f"[DEPRECATED] `perform_self_test` has been deprecated: {self}")
```
JVillella, line 128: logs instead of prints?

JVillella (approved):

JVillella:
Nice changes :)

Issue #63, opened 2020-10-22
Core functionality tests for trap management


Gamenot:
* [ ] Test trap successfully grabs vehicle
* [ ] Test trap emits vehicle with speed when it fails to capture a social vehicle
* [ ] Test trap does not capture pre-existing Agent vehicles
* [x] Reactivation functionality removed

PR #163, opened 2020-11-16
WIP: Speed up imitation learning training by allowing users to jump to a timestamp

liamchzh:
During the training, we will replace one of the vehicles in history data with an agent. However, this agent might appear in the middle of an episode but SMARTS still step through the replay data from the beginning until the agent shows up. 

In this PR, I exposed a param `start_time` to allow us to set the frame start time.

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -268,7 +268,7 @@ class SMARTS(ShowBase):
 
         self._teardown_agent_vehicles(agents_to_cleanup)
 
-    def reset(self, scenario: Scenario):
+    def reset(self, scenario: Scenario, start_time=0):
         if scenario == self._scenario and self._reset_agents_only:
             agent_ids = self._agent_manager.teardown_ego_agents()
             self._teardown_agent_vehicles(agent_ids)
@@ -283,7 +283,8 @@ class SMARTS(ShowBase):
         self._traffic_history_provider.set_replaced_ids(scenario.missions.keys())
 
         self.taskMgr.clock.reset()
-        self._elapsed_sim_time = 0
+        self.taskMgr.clock.set_frame_time(start_time)
+        self._elapsed_sim_time = start_time
 
         self._vehicle_states = [v.state for v in self._vehicle_index.vehicles]
         observations, _, _, _ = self._agent_manager.observe(self)
--- a/smarts/core/trap_manager.py
+++ b/smarts/core/trap_manager.py
@@ -42,24 +42,19 @@ class Trap:
     mission: Mission
     exclusion_prefixes: Sequence[str]
     reactivation_time: float
-    remaining_time_to_reactivation: float
+    emit_agent_at: float
     patience: float
     default_entry_speed: float
 
-    def step_trigger(self, dt: float):
-        self.remaining_time_to_reactivation -= dt
+    def is_ready(self, elapsed_sim_time: float):
+        return elapsed_sim_time >= self.emit_agent_at
 
-    @property
-    def ready(self):
-        return self.remaining_time_to_reactivation < 0
-
-    @property
-    def patience_expired(self):
+    def patience_expired(self, elapsed_sim_time: float):
         """Patience recommendation to wait for better capture circumstances"""
-        return self.remaining_time_to_reactivation < -self.patience
+        return elapsed_sim_time - self.emit_agent_at > self.patience
 
     def reset_trigger(self):
-        self.remaining_time_to_reactivation = self.reactivation_time
+        pass
 
     @property
     def reactivates(self):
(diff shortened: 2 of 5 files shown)


smarts/core/trap_manager.py
```diff
@@ -45,18 +45,21 @@ class Trap:
...
     def ready(self):
-        return self.remaining_time_to_reactivation < 0
+        return self.remaining_time_to_reactivation - self.elapsed_sim_time < 0
 
     @property
     def patience_expired(self):
         """Patience recommendation to wait for better capture circumstances"""
-        return self.remaining_time_to_reactivation < -self.patience
+        return (
+            self.remaining_time_to_reactivation - self.elapsed_sim_time < -self.patience
```
Gamenot, line 61: This bypasses the patience now I think if the starting `elapsed_sim_time` is greater than `patience`.

smarts/core/trap_manager.py
```diff
@@ -45,18 +45,21 @@ class Trap:
     reactivation_time: float
     remaining_time_to_reactivation: float
     patience: float
+    elapsed_sim_time: float = 0
 
-    def step_trigger(self, dt: float):
-        self.remaining_time_to_reactivation -= dt
+    def update_sim_time(self, elapsed_sim_time: float):
+        self.elapsed_sim_time = elapsed_sim_time
```
Gamenot, line 51: I think with this `remaining_time_to_reactivation` no longer maintains its original meaning. I think this also breaks the ability for the trap to reset and hijack another vehicle. Although I am not sure the functionality was ever used. This is fine for now but put up an issue for this leftover to either remove the functionality or fix it.

smarts/core/smarts.py
```diff
@@ -258,7 +258,7 @@ def _step(self, agent_actions):
         extras = dict(scores=scores)
         return observations, rewards, dones, extras
 
-    def reset(self, scenario: Scenario):
+    def reset(self, scenario: Scenario, start_time=0):
```
Gamenot, line 261: Adding this option to the `Scenario` reset has implications on all of the providers and co-simulators. Have you checked the implications on the SUMO simulation?

Gamenot (changes requested):

smarts/core/trap_manager.py
liamchzh: I'll take another look

smarts/core/trap_manager.py
liamchzh: `reactivation_time` is assigned to -1, which means for now it would not be reused, right?

smarts/core/smarts.py
liamchzh: Hmm, I need to double check, now this `start_time` will not affect SUMO.

liamchzh: https://github.com/huawei-noah/SMARTS/pull/163/commits/9ac9db0541bf52d35d39cefb4e9488317b74867e makes SUMO's cumulative sim time aligned with elapsed_sim_time

smarts/core/trap_manager.py
Gamenot: Yes sorry, it is fine, I do not think it is used at current. Reactivation should be stripped out as part of https://github.com/huawei-noah/SMARTS/issues/63.

Gamenot (approved):

aurorahcx (approved):

JVillella (changes requested):
I don't think this is the right design. The `reset(..., start_time=...)` isn't actually doing what the API makes it seem. One would think it's offsetting the sim by the given start_time, but really just changing the internal clock.

This looks to be a one-off solution for the history data replay work and not useful to anything else (or applicable to anything else). Can we just offset the timestamps of the data replay data directly / modify the data replay input?

liamchzh:
> I don't think this is the right design. The `reset(..., start_time=...)` isn't actually doing what the API makes it seem. One would think it's offsetting the sim by the given start_time, but really just changing the internal clock.
> 
> This looks to be a one-off solution for the history data replay work and not useful to anything else (or applicable to anything else). Can we just offset the timestamps of the data replay data directly / modify the data replay input?

Yes, the side effect of changing the clock is hidden, I am not sure about the change of `reset` too. I can try your suggested approach by maintaining an `offset` that just applied to the history provider.

liamchzh:
> > I don't think this is the right design. The `reset(..., start_time=...)` isn't actually doing what the API makes it seem. One would think it's offsetting the sim by the given start_time, but really just changing the internal clock.
> > This looks to be a one-off solution for the history data replay work and not useful to anything else (or applicable to anything else). Can we just offset the timestamps of the data replay data directly / modify the data replay input?
> 
> Yes, the side effect of changing the clock is hidden, I am not sure about the change of `reset` too. I can try your suggested approach by maintaining an `offset` that just applied to the history provider.

Underestimated the difficulty. Not only the history provider has the clock issue, but also the trap manager. The time that an agent is emitted, is specified in the mission's start time.

So two things:
1. the history provider needs to know the offset
2. mission starts at time 0 sec (Supporting multiple agent missions makes it harder :thinking: )

I will try to work towards the approach above.

PR #169, opened 2020-11-17
Issue 99: sensor camera position not being updated during airlock

aurorahcx:
For scenario straight:

Before stepping agent sensors for shadowed actors in bubble
Screenshot from 2020-11-17 07-06-20

After stepping agent sensors for shadowed actors in bubble
Screenshot from 2020-11-17 07-08-52

### DIFF
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -206,13 +206,18 @@ class AgentManager:
         return sim.vehicle_index.vehicle_by_id(vehicle_id).trip_meter_sensor()
 
     def step_agent_sensors(self, sim):
-        for vehicle_id in sim.vehicle_index.agent_vehicle_ids:
-            sensor_state = sim.vehicle_index.sensor_state_for_vehicle_id(vehicle_id)
-            Sensors.step(self, sensor_state)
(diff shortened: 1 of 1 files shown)


Gamenot (approved):
My thought with this is that there might be reason to instead mark and filter by vehicles with sensors rather than agent vehicles and shadowed vehicles as a future improvemnet

smarts/core/agent_manager.py
```diff
@@ -206,13 +206,18 @@ def _vehicle_score(self, vehicle_id, sim):
         return sim.vehicle_index.vehicle_by_id(vehicle_id).trip_meter_sensor()
 
     def step_agent_sensors(self, sim):
-        for vehicle_id in sim.vehicle_index.agent_vehicle_ids:
```
JVillella, line 209: That's right, the agents shadowing vehicles need to be sent the observations.

JVillella: I believe we can make this PR a one line change by just doing,

```python
for vehicle_id in sim.vehicle_index.vehicle_ids_by_actor_id(include_shadowers=True):
```

JVillella (changes requested):

JVillella (approved):

PR #170, opened 2020-11-17
Tensorflow upgrade

Gamenot:


Gamenot:
Intermediary merge to get the merges in from the original #166.

PR #173, opened 2020-11-17
Fix memory growth tests in master

Gamenot:
Install `pympler` in action but not add to requirements since it is not a main dependency.


davidrusu (approved):

JVillella (approved):
Pympler is great

liamchzh (approved):

PR #174, opened 2020-11-17
Only count waypoint if we have travelled sufficiently from prev wp

davidrusu:
With the equally spaced waypoint change, we can no longer rely on the fact that waypoints are fixed on the map, so we have to track distance from last waypoint instead.

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -862,7 +862,7 @@ class TripMeterSensor(Sensor):
             or wp_edge in self._mission_planner.route.edges
         )
 
-        if most_recent_wp != new_wp and should_count_wp:
+        if np.linalg.norm(new_wp.pos - most_recent_wp.pos) > 0.5 and should_count_wp:
             self._dist_travelled += TripMeterSensor._compute_additional_dist_travelled(
                 most_recent_wp, new_wp
             )
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -862,8 +862,11 @@ class TripMeterSensor(Sensor):
             or wp_edge in self._mission_planner.route.edges
         )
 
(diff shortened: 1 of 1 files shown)


smarts/core/sensors.py
```diff
@@ -862,7 +862,7 @@ def append_waypoint_if_new(self, new_wp):
             or wp_edge in self._mission_planner.route.edges
         )
 
-        if most_recent_wp != new_wp and should_count_wp:
+        if np.linalg.norm(new_wp.pos - most_recent_wp.pos) > 0.5 and should_count_wp:
```
Gamenot, line 865: Do we have a constant or variable we can use to describe what is going on here instead of 0.5?

Gamenot (changes requested):

smarts/core/sensors.py
JVillella, line 865: .5 assumes distance is 1 unit? but it's actually parameterizable right?

davidrusu: good point, I'll move it to a variable.

davidrusu: With the changes to the equally spaced waypoints, the waypoint spacing parameter does not have as much of an effect since the waypoints move around with the vehicle.

0.5 is a reasonable approximation to the old behaviour and should behave consistently across different spacings as well

JVillella: Is this not saying: "if we're passed the mid point..."?

Gamenot (approved):

aurorahcx (approved):

smarts/core/sensors.py
davidrusu: This change is saying "is the current waypoint we are holding more than 0.5 meters from the last waypoint that I recorded, if so, record this waypoint".

PR #176, opened 2020-11-17
Support UTurn task

liamchzh:
* A task can be specified along with the `Mission`
* The mission planner will calculate an u-turn trajectory instead of waypoints if the mission has an u-turn task.
* Add an u-turn scenario for demonstration (May delete it before merging)

ezgif com-gif-maker (7)
ezgif com-gif-maker (6)


smarts/core/vehicle.py
```diff
@@ -374,13 +375,23 @@ def attach_sensors_to_vehicle(sim, vehicle, agent_interface, mission_planner):
             )
 
         if agent_interface.waypoints:
-            vehicle.attach_waypoints_sensor(
-                WaypointsSensor(
-                    vehicle=vehicle,
-                    mission_planner=mission_planner,
-                    lookahead=agent_interface.waypoints.lookahead,
+            # TODO
```
liamchzh, line 378: @davidrusu @JVillella @Gamenot This is the design where I couldn't make the call. Currently, I replace the original waypoints with a u-turn trajectory, which makes it easier for agents to do the task, but the downsides are: 1. mix the waypoints and desired trajectory conceptually 2. The changes might not be expected to users.

smarts/core/sensors.py
```diff
@@ -909,6 +911,115 @@ def teardown(self):
...
 
+class UTurnTrajectorySensor(Sensor):
+    def __init__(self, vehicle, sim, mission_planner):
+        self._vehicle = vehicle
+        self._sim = sim
+        self._mission_planner = mission_planner
+        self._horizon = 50
+
+    def __call__(self):
+        # TODO: Reuse codes from RoadWaypointsSensor,
```
liamchzh, line 922: @iman512003 I believe we need to tune the trajectory. The flaws I can see now: 1. the agent won't land in the middle of the target lane 2. the trajectories of incoming edge and oncoming edge are not consistent.

liamchzh:
ezgif com-gif-maker (7)
ezgif com-gif-maker (6)

smarts/core/vehicle.py
JVillella: Let's talk about it in our team sync this evening?

scenarios/uturn/scenario.py
```diff
@@ -0,0 +1,48 @@
...
+    "all": (
+        [
+            t.SocialAgentActor(
+                name="open-agent", agent_locator="open_agent:open_agent-v0"
+            ),
+        ],
+        [
+            t.Mission(
+                t.Route(begin=("edge-west-WE", 0, 50), end=("edge-west-EW", 0, "max")),
+                task=t.UTurn(target_lane_index=0),
```
liamchzh, line 28: * The map should have a connection between the 2 edges so that the mission planner still can plan the route.
* If a uturn task is specified, then the route should cut early. Otherwise, there should a default uturn if the route requires it. (low priority)

smarts/core/mission_planner.py
```diff
@@ -182,3 +192,88 @@ def _edge_ids(self, pose: Pose, lane_id: str = None):
...
+        wp = self._waypoints.closest_waypoint(pose.position)
+        current_edge = self._road_network.edge_by_lane_id(wp.lane_id)
+        if not start_edge.oncoming_edges:
+            return []
+
+        oncoming_edge = start_edge.oncoming_edges[0]
+        oncoming_lanes = oncoming_edge.getLanes()
+        target_lane_index = self._mission.task.target_lane_index
+        if len(oncoming_lanes) <= target_lane_index:
+            return []
```
davidrusu, line 213:
```suggestion
        target_lane_index = min(target_lane_index, len(oncoming_lanes) - 1)
```

davidrusu (approved):
It's a good first pass, I think we can iterate over multiple PR's, I'd vote to merge and improvements can be done iteratively.

iman512003 (approved):

smarts/core/mission_planner.py
```diff
@@ -182,3 +192,88 @@ def _edge_ids(self, pose: Pose, lane_id: str = None):
...
+
+        heading = pose.heading
+        target_heading = target.heading
+        lane_width = target_lane.getWidth()
+        lanes = (len(current_edge.getLanes())) + (
+            len(oncoming_lanes) - target_lane_index
+        )
+
+        if current_edge.getID() != oncoming_edge.getID():
+            # agent at the start edge
```
iman512003, line 229: I think one reason that the vehicle has some offset during the second part of the uturn is related to the following figure(4 is the end target position and 1 is the vehicle position):
Screenshot from 2020-11-23 23-15-35
I think if we force the 3 point to be located on the center of the lane aligned with 4 then we can ensure smoother transition to the end, as @davidrusu mentioned we can work on this in the next PRs.

iman512003 (approved):

smarts/core/mission_planner.py
iman512003: Very nice PR, LGTM.

JVillella:
@JVillella

smarts/core/mission_planner.py
```diff
@@ -182,3 +192,88 @@ def _edge_ids(self, pose: Pose, lane_id: str = None):
             edge_ids.append(next_edges[0].getID())
 
         return edge_ids
+
+    def uturn_waypoints(self, pose: Pose):
```
JVillella, line 196: This method is a little too low-level to be in the mission planner. Perhaps wrap in a helper we call out to.

JVillella:
👍 I'm with David, let's merge as is. Some clean-up tuning can be done to the uturn method later.

JVillella:
(also, I would support deleting the scenario before merging in, or moving the uturn code to another scenario - my preference)

smarts/core/mission_planner.py
liamchzh: Thanks, @iman512003. You're right, I was fighting with the position of point 3 yesterday but couldn't get it on the right position.

smarts/core/mission_planner.py
liamchzh: OK, I will add a TODO since this method is relying on the mission, road_network, waypoints.

Issue #157, opened 2020-11-13
/tmp/smarts directory owner cause other user to run smarts error


charleschen6:
there are multi user in one machine, they all want to run smarts with python agent.py scenario_path
but after one user success run smarts, the other one run will meet below problem:
```
Traceback (most recent call last):
  File "/usr/share/sumo/tools/traci/main.py", line 115, in connect
    conn = Connection(host, port, proc)
  File "/usr/share/sumo/tools/traci/connection.py", line 51, in __init__
    self._socket.connect((host, port))
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/zhengang/828_release_smarts_virtual_map/share/object_model/config/smarts_config/agents/rl_single_agent_client0.py", line 134, in <module>
    seed=args.seed,
  File "agents/rl_single_agent_client0.py", line 113, in main
    observations = env.reset()
  File "SMARTS/smarts/env/hiway_env.py", line 179, in reset
    env_observations = self._smarts.reset(scenario)
  File "SMARTS/smarts/core/smarts.py", line 270, in reset
    self.setup(scenario)
  File "SMARTS/smarts/core/smarts.py", line 318, in setup
    provider_state = self._setup_providers(self._scenario)
  File "SMARTS/smarts/core/smarts.py", line 600, in _setup_providers
    provider_state.merge(provider.setup(scenario))
  File "SMARTS/smarts/core/sumo_traffic_simulation.py", line 231, in setup
    self._initialize_traci_conn()
  File "SMARTS/smarts/core/sumo_traffic_simulation.py", line 142, in _initialize_traci_conn
    waitBetweenRetries=0.5,
  File "/usr/share/sumo/tools/traci/main.py", line 121, in connect
    raise TraCIException("TraCI server already finished")
traci.exceptions.TraCIException: TraCI server already finished
```

then I find that  
```shell
/tmp | grep smarts
```
the path /tmp/smarts belong to the first smarts run user, when `chown -R   {other_user} /tmp/smarts` and `chgrp -R   {other_user} /tmp/smarts` , the specify other user could run smarts success, the first user will meet same probllem.

charleschen6:
change the all log_dir path to the current user home directory

liamchzh:
@charleschen6 Thanks for reporting this.

@Gamenot This could be a problem if researchers are sharing a machine.

davidrusu:
[...]

PR #180, opened 2020-11-18
Give a per user smarts log file

Gamenot:
This to address #157

### DIFF
--- a/smarts/core/scenario.py
+++ b/smarts/core/scenario.py
@@ -162,7 +162,7 @@ class Scenario:
         route: str = None,
         missions: Dict[str, Mission] = None,
         social_agents: Dict[str, SocialAgent] = None,
-        log_dir: str = "/tmp/smarts/_sumo_run_logs",
+        log_dir: str = None,
         surface_patches: list = None,
     ):
 
@@ -173,7 +173,7 @@ class Scenario:
         self._bubbles = Scenario._discover_bubbles(scenario_root)
         self._social_agents = social_agents or {}
         self._surface_patches = surface_patches
-        self._log_dir = os.path.abspath(log_dir)
+        self._log_dir = self._resolve_log_dir(log_dir)
 
         self._validate_assets_exist()
         self._road_network = SumoRoadNetwork.from_file(self.net_filepath)
@@ -714,6 +714,16 @@ class Scenario:
 
     def mission(self, agent_id):
         return self._missions.get(agent_id, None)
+
+    def _resolve_log_dir(self, log_dir):
+        home_dir = os.path.expanduser("~")
+        _, home_user = os.path.split(home_dir)
+        if log_dir is None:
+            log_dir = os.path.abspath(f"/tmp/smarts_{home_user}/_sumo_run_logs")
+        else:
+            log_dir = os.path.abspath(log_dir)
+
+        return log_dir
 
     def _validate_assets_exist(self):
         assert Scenario.is_valid_scenario(self._root)
--- a/smarts/core/scenario.py
+++ b/smarts/core/scenario.py
@@ -22,9 +22,10 @@ import json
 import logging
 import math
 import os
-import sys
 import pickle
 import random
+import stat
+import tempfile
 import uuid
 
 import numpy as np
@@ -40,7 +41,7 @@ from .sumo_road_network import SumoRoadNetwork
 from .waypoints import Waypoints
 from .coordinates import Heading
 from .utils.math import vec_to_radians
-from .utils.file import path2hash, file_md5_hash
+from .utils.file import make_dir_in_smarts_log_dir, path2hash, file_md5_hash
 from .utils.id import SocialAgentId
 from .route import ShortestRoute
 from smarts.sstudio import types as sstudio_types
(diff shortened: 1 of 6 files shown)


smarts/core/scenario.py
```diff
@@ -715,6 +715,16 @@ def bubbles(self):
     def mission(self, agent_id):
         return self._missions.get(agent_id, None)
 
+    def _resolve_log_dir(self, log_dir):
+        home_dir = os.path.expanduser("~")
+        _, home_user = os.path.split(home_dir)
+        if log_dir is None:
+            log_dir = os.path.abspath(f"/tmp/smarts_{home_user}/_sumo_run_logs")
```
davidrusu, line 722: `/tmp` is not something we can rely on to be cross platform.

Use `mkdtemp` with a `SMARTS_` prefix instead. https://docs.python.org/3.9/library/tempfile.html#tempfile.mkdtemp

davidrusu (changes requested):
/tmp is not cross platform

smarts/core/scenario.py
liamchzh: +1

smarts/core/utils/file.py
```diff
@@ -82,3 +85,26 @@ def file_md5_hash(file_path: str) -> str:
...
+
+    # if not os.path.exists(temp_dir):
+    #     original_mask = os.umask(0)
+    #     try:
+    #         os.makedirs(temp_dir, mode=0o777)
+    #         os.chmod(temp_dir, mode=0o777)
+    #     finally:
+    #         os.umask(original_mask)
+
+    # return os.path.join(temp_dir, f"{home_user}")
```
Gamenot, line 105: This above would work on linux and macos. The question would be windows.

smarts/core/scenario.py
```diff
@@ -22,9 +22,10 @@
 import logging
 import math
 import os
-import sys
 import pickle
 import random
+import stat
+import tempfile
```
liamchzh, line 28: Are these 2 imports being used?

smarts/sstudio/generators.py
```diff
@@ -20,6 +20,7 @@
 import os
 import random
 import logging
+from smarts.core.utils.file import make_dir_in_smarts_log_dir
```
liamchzh, line 23: This line should be in the 3rd part of imports.

smarts/core/utils/file.py
```diff
@@ -17,10 +17,13 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
+from functools import lru_cache
 import hashlib
 import os
 import shutil
 import dataclasses
+import stat
```
liamchzh, line 25: I believe `stat` is not used.

smarts/sstudio/generators.py
Gamenot: Sorry, I have automatic imports that tend to throw the imports in random places. :facepalm:

smarts/core/utils/file.py
liamchzh: Can `tempfile` take care of the permission?

smarts/sstudio/generators.py
liamchzh: Seems you have a new tool to help re-order the imports. I like that re-ordering.

smarts/core/scenario.py
```diff
@@ -22,27 +22,32 @@
...
 from functools import lru_cache
-from itertools import cycle, product
+from itertools import cycle
+from itertools import product
 from pathlib import Path
-from typing import Any, Dict, Sequence, Tuple
+from typing import Any
+from typing import Dict
+from typing import Sequence
+from typing import Tuple
```
davidrusu, line 37: why are we spreading out these imports across multiple lines?

smarts/core/utils/file.py
davidrusu: If the intention is to have a one directory that is consistent across runs, then we shouldn't rely on /tmp/

Let's create a `.smarts` directory in the user's home directory and put sumo logs under there.

smarts/core/scenario.py
Gamenot: I used a formatter and the justification for it is that it allows for easier navigation of merges.

davidrusu: Wait, you used a formatter other than black! Heathen!!

In all seriousness, I do prefer the old format of grouping all modules from the same parent on one line.

smarts/core/utils/file.py
Gamenot: > Can `tempfile` take care of the permission?
> 
> Sorry, I am not sure what's the problem here. I mean if `tempfile.mkdtemp` works well, why would we need to handle the permission?

There only consideration is that we are changing the file structure from the original to be completely flat. If that is already fine then there are no other issues.

Gamenot: > If the intention is to have a one directory that is consistent across runs, then we shouldn't rely on /tmp/
> 
> Let's create a `.smarts` directory in the user's home directory and put sumo logs under there.

That can be done. The commented out solution uses a combination of the home directory name and `tempfile.gettempdir()`(which finds the same temp directory used by `tempfile.mkdtemp`.) The original reason for using temp over the home directory was to allow for automatic cleanup of the logs, which we could suspect can become quite large.

smarts/core/scenario.py
Gamenot: Fair enough.

smarts/core/utils/file.py
```diff
@@ -82,3 +85,15 @@ def file_md5_hash(file_path: str) -> str:
         hasher.update(f.read().encode())
 
     return str(hasher.hexdigest())
+
+
+@lru_cache(maxsize=1)
+def smarts_log_dir() -> str:
+    ## Following should work for linux and macos
+    smarts_dir = os.path.join(os.path.expanduser("~"), "smarts")
```
liamchzh, line 93: `.smarts`?

JVillella (approved):
Assuming we use local dir over temp as the conversation was going, rest looks good to me

smarts/core/utils/file.py
```diff
@@ -82,3 +85,15 @@ def file_md5_hash(file_path: str) -> str:
         hasher.update(f.read().encode())
 
     return str(hasher.hexdigest())
+
+
+@lru_cache(maxsize=1)
```
davidrusu, line 90: No need for a cache here.

davidrusu (approved):
Just get rid of the lru cache, besides that LGTM

PR #182, opened 2020-11-18
Bump tensorflow version for security reasons

Gamenot:
This patch should fix cases using `requirements.txt` (i.e. docker and the CI) and set `Setup.py` to use the newest revision of `2.2.*`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
         #       and ray[rllib] under [train] it  prevents rllib from getting installed.
         #       For simplicity we just install both here. In the future we may want to
         #       address this bug head on to keep our SMARTS base install more lean.
-        "ray[rllib]==0.8.*",  # We use Ray for our multiprocessing needs
+        "ray[rllib]==0.8.3",  # We use Ray for our multiprocessing needs
         # The following are for Scenario Studio
         "yattag",
         # The following are for testing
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -66,7 +66,7 @@
         "PyYAML",
     ],
     extras_require={
-        "train": ["tensorflow==2.2", "torch==1.3.0", "torchvision==0.4.1"],
+        "train": ["tensorflow==2.2.*", "torch==1.3.*", "torchvision==0.4.*"],
```
liamchzh, line 69: While can we use a fixed version?

Gamenot: I would prefer to use the newest revision of a version because it fixes issues with the version. If it is better to just lock it, that is fine too.

liamchzh: I think locking the dependencies is necessary and important. Using a newer version brings us benefits but might also cause unexpected issues.

liamchzh: BTW, speaking of dependency management, I have been thinking use [poetry](https://python-poetry.org/) as the management tool. Reasons:
1. Adopt PEP 518 (looks like a better dependency resolver? :smile: )
2.  Can still install poetry-managed project directly: `pip install -e .`

Gamenot: OK, my thought was that `requirements.txt` is where we have the locked version. 

`ray[rllib]`, `tensorflow`, `torch` all use revision numbers for fixes to and since we have a test of `Setup.py` against `requirements.txt` we know when the version has changed and problems have occurred.

liamchzh (approved):

davidrusu (approved):

Issue #181, opened 2020-11-18
[ 93%] FAILED smarts/env/tests/test_rllib_hiway_env.py::test_rllib_hiway_env  smarts/env/tests/test_social_agent.py::test_social_agents


JianmingTONG:
I have already passed all Setup instructions except make test. ` pip install -e .[dev]` is also passed.
When I run make test, 1 env failed and no further reason reported.

Might I request help from anyone on how to solve it?

My environment is python3.7 on ubuntu 18.04 with the following package.
```
Package                           Version   Location
--------------------------------- --------- ------------------------------
absl-py                           0.10.0
actionlib                         1.12.1
aiohttp                           3.6.3
angles                            1.9.12
apipkg                            1.5
astor                             0.8.1
astunparse                        1.6.3
async-timeout                     3.0.1
atari-py                          0.2.6
attrs                             20.2.0
Automat                           20.2.0
beautifulsoup4                    4.9.3
bondpy                            1.8.5
cachetools                        4.1.1
camera-calibration                1.15.0
camera-calibration-parsers        1.11.13
casadi                            3.5.5
catkin                            0.7.29
certifi                           2020.6.20
chardet                           3.0.4
click                             7.1.2
cloudpickle                       1.3.0
colorama                          0.4.4
commonmark                        0.9.1
constantly                        15.1.0
controller-manager                0.18.2
controller-manager-msgs           0.18.2
coverage                          5.3
cv-bridge                         1.13.0
cycler                            0.10.0
Cython                            0.29.21
decorator                         4.4.2
diagnostic-analysis               1.9.7
diagnostic-common-diagnostics     1.9.7
diagnostic-updater                1.9.7
dm-tree                           0.1.5
dynamic-reconfigure               1.6.3
evdev                             1.3.0
execnet                           1.7.1
filelock                          3.0.12
future                            0.18.2
gast                              0.3.3
gazebo-plugins                    2.8.7
gazebo-ros                        2.8.7
gencpp                            0.6.5
geneus                            2.2.6
genlisp                           0.4.16
genmsg                            0.5.16
gennodejs                         2.0.1
genpy                             0.6.14
google                            3.0.0
google-auth                       1.23.0
google-auth-oauthlib              0.4.2
google-pasta                      0.2.0
grpcio                            1.32.0
gym                               0.17.3
h5py                              2.10.0
hyperlink                         20.0.1
idna                              2.10
image-geometry                    1.13.0
imageio                           2.9.0
[...]

PR #188, opened 2020-11-19
Mitigate failures in `test_rllib_hiway_env`

Gamenot:
The goal here is to lower the number of CPUs used by `ray.tune` and further lower the memory used to support those CPUs since there only needs to be a small number of workers to test out `rllib_hiway_env` we do not need all of the CPUs to be active. 

The issue that brought this up: https://github.com/huawei-noah/SMARTS/issues/181

### DIFF
--- a/smarts/core/remote_agent_buffer.py
+++ b/smarts/core/remote_agent_buffer.py
@@ -40,7 +40,8 @@ class RemoteAgentBuffer:
         self.destroy()
 
     def destroy(self):
-        atexit.unregister(self.destroy)
+        if atexit.unregister is not None:
+            atexit.unregister(self.destroy)
         self._quiescing = True
         if self._replenish_thread_is_running():
             self._replenish_thread.join()  # wait for the replenisher to finish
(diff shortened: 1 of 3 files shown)


smarts/core/remote_agent_buffer.py
```diff
@@ -40,7 +40,8 @@ def __del__(self):
         self.destroy()
 
     def destroy(self):
-        atexit.unregister(self.destroy)
+        if atexit.unregister is not None:
```
davidrusu, line 43: This is odd, why do we need it? Isn't this is always true?

Gamenot: It does not seem to always be true. I am not sure why.

liamchzh (approved):

smarts/core/remote_agent_buffer.py
JVillella: Pretty strange, docs don't indicate unregister could be none. Do you have an example snippet in the logs?

JVillella (approved):
Code looks good, just a comment regarding the unregister guard code.

smarts/core/remote_agent_buffer.py
Gamenot: I was seeing this in the logs:

image

PR #190, opened 2020-11-20
Point Via 

Gamenot:
Implementation of point vias just as observations.

### DIFF
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -104,7 +104,7 @@ class Accelerometer:
 
 @dataclass
 class Checkpoints:
-    """ Observation of nearby checkpoints """
+    """Observation of nearby checkpoints"""
 
     sensor_range_radius: float = 50
 
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -104,7 +104,7 @@ class Accelerometer:
 
 @dataclass
 class Goalpoints:
-    """ Observation of nearby checkpoints """
+    """ Observation of nearby goalpoints """
 
     near_acquisition_radius: float = 50
 
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -103,8 +103,8 @@ class Accelerometer:
 
 
 @dataclass
-class Checkpoints:
-    """Observation of nearby checkpoints"""
+class ViaPoints:
+    """Observation of nearby via_points"""
 
     sensor_range_radius: float = 50
 
@@ -219,7 +219,7 @@ class AgentInterface:
     Enable acceleration and jerk observations.
     """
 
-    checkpoint: Union[Checkpoints, bool] = True
+    via_point: Union[ViaPoints, bool] = True
     """
     Points that must be hit in order to successfully end an episode at the end goal.
     """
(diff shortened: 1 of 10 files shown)


smarts/core/agent_interface.py
```diff
@@ -102,6 +102,13 @@ class Accelerometer:
     pass
 
 
+@dataclass
+class Goalpoints:
+    """ Observation of nearby goalpoints """
```
JVillella, line 107: Can we remove the extra spaces?

smarts/core/agent_interface.py
```diff
@@ -102,6 +102,13 @@ class Accelerometer:
...
+
+    near_acquisition_radius: float = 50
```
JVillella, line 109: Can you just call it radius? Goal radius seems pretty clear what it is.

smarts/core/agent_interface.py
```diff
@@ -102,6 +102,13 @@ class Accelerometer:
     pass
 
 
+@dataclass
+class Goalpoints:
```
JVillella, line 106: GoalPoint?

smarts/core/route.py
```diff
@@ -17,12 +17,22 @@
...
+from smarts.sstudio.types import EdgePointVia
+
+from .utils.sumo import sumolib  # isort:skip
+from sumolib.net.edge import Edge  # isort:skip
 
-from .utils.sumo import sumolib
-from sumolib.net.edge import Edge
+
+@dataclass(frozen=True)
+class Goalpoint:
```
JVillella, line 33: GoalPoint?

smarts/core/route.py
```diff
@@ -33,6 +43,10 @@ def edges(self) -> List[Edge]:
         """An (unordered) list of edges that this route covers"""
         raise NotImplementedError()
 
+    @property
+    def goalpoints(self) -> List[Any]:
```
JVillella: What is the rational of putting goal points as part of the route interface, why don't we just implement them as a sep. entity all together? Furthermore, I don't like that the goalpoints concept has worked its way all the way up to the top level route interface.

smarts/core/sensors.py
```diff
@@ -111,6 +113,13 @@ class DrivableAreaGridMap(NamedTuple):
     data: np.ndarray
 
 
+@dataclass
+class GoalpointData:
+    nearest_remaining_goalpoint: Any
+    hit_goalpoint: Any
+    near_goalpoints: Any
```
JVillella: Any?

smarts/core/sensors.py
```diff
@@ -111,6 +113,13 @@ class DrivableAreaGridMap(NamedTuple):
     data: np.ndarray
 
 
+@dataclass
+class GoalpointData:
```
JVillella: GoalPointData

smarts/core/route.py
Gamenot: There is no way to separate the points from the route interface since the checkpoints have to be on the route. If they are not on the route then you need to specify them separately from the route which forces changes on either the route or the checkpoints(which should be along the route.)

smarts/core/agent_interface.py
Gamenot: The intention here is sensor radius, I have clarified it.

smarts/core/agent_interface.py
Gamenot: Certainly, for some reason I cannot stop the habit of padding these doc-strings.

PR #191, opened 2020-11-20
Support individual vehicles being done (i.e. for Boids)

### LINKED ISSUE #23: Handle boid vehicle going off-road
Split from #17. Presently,

> When a boid agent vehicle goes off road, a new sumo vehicle replace it and is stuck at the road side

We want the behavior to hand over control to SUMO and have SUMO attempt bringing the vehicle back on course. If the vehicle goes back into the bubble - not airlocking along the way - it should not get hijacked.

JVillella:
closes #23

We can now do this,
boid-offroad-demo

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -255,7 +255,10 @@ class Sensors:
         rgb = vehicle.rgb_sensor() if vehicle.subscribed_to_rgb_sensor else None
         lidar = vehicle.lidar_sensor() if vehicle.subscribed_to_lidar_sensor else None
 
-        done, events = Sensors._is_done_with_events(sim, agent_id, sensor_state)
+        done, events = Sensors._is_done_with_events(
+            sim, agent_id, vehicle, sensor_state
+        )
+
         return (
             Observation(
                 events=events,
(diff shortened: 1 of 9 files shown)


smarts/core/sumo_traffic_simulation.py
```diff
@@ -426,7 +426,7 @@ def _move_vehicle(self, vehicle_id, position, heading, speed):
             x=x,
             y=y,
             angle=heading,  # only used for visualizing in sumo-gui
-            keepRoute=0b010,  # gives vehicle freedom to move off road
+            keepRoute=0,
```
JVillella, line 429: This should be fine for now since we end the episode when vehicles go off route. It could however be a problem now if it's too sensitive.

Gamenot: Is there a particular reason for changing this flag? I am not confident in this since we currently have an option to be able to turn off our off-route and I am not sure it plays into our definition of off-route which does not count the oncoming lanes as off-route.

Gamenot (changes requested):

smarts/core/sumo_traffic_simulation.py
JVillella: If we don't do this, SUMO is unable to bring the vehicle back on the road. We'll have to do some manual vehicle movement to get it back to a region SUMO can pick up where we left off.

Gamenot: OK, this may be fine but you will need to check if it works with oncoming lanes or it will disrupt training.

smarts/core/smarts.py
```diff
@@ -254,24 +255,48 @@ def _step(self, agent_actions):
...
+
+            return done
+
+        def done_vehicle_ids(dones, vehicle_index):
+            vehicle_ids = set()
+            for agent_id, done in dones.items():
+                if self._agent_manager.is_boid_agent(agent_id):
+                    for vehicle_id in done:
+                        if done[vehicle_id]:
+                            vehicle_ids.add(vehicle_id)
```
davidrusu, line 271:
```suggestion
                    vehicle_ids |= {vehicle_id for vehicle_id in done if done[vehicle_id]}
```

davidrusu (approved):
Nice cleanups

smarts/core/sumo_traffic_simulation.py
JVillella: Docs seem to suggest =0 should be fine https://sumo.dlr.de/docs/TraCI/Change_Vehicle_State.html#move_to_xy_0xb4

smarts/core/smarts.py
JVillella: Tweaked it in the PR ahead.

smarts/core/sumo_traffic_simulation.py
JVillella: According to docs (https://sumo.dlr.de/docs/TraCI/Change_Vehicle_State.html#move_to_xy_0xb4) it seems as though it should be fine.

Gamenot (approved):

smarts/core/sumo_traffic_simulation.py
Gamenot: OK, I was confused for a few minutes, but I think the `--lanechange.duration` option is what is allowing this now. Nothing further blocking.

smarts/core/sensors.py
```diff
@@ -323,7 +329,46 @@ def _is_done_with_events(cls, sim, agent_id, sensor_state):
         return done, events
 
     @classmethod
-    def _agent_vehicle_is_off_route_and_wrong_way(cls, sim, agent_id):
+    def _agent_reached_goal(cls, sim, vehicle):
```
Gamenot, line 332: This seems oddly named to me considering it deals with vehicle and mission. Perhaps `_vehicle_reached_mission_goal()`?

smarts/core/smarts.py
```diff
@@ -424,11 +449,9 @@ def destroy(self):
 
         super().destroy()
 
-    def _teardown_agent_vehicles(self, agent_ids):
-        torndown_vehicles = self._vehicle_index.teardown_vehicles_by_actor_ids(
-            agent_ids
-        )
-        self._clear_collisions(agent_ids)
+    def _teardown_agent_vehicles(self, vehicle_ids):
```
Gamenot, line 452: I was confused about the use of this method when reading above. It does not seem to have anything to do with agents and just with vehicles perhaps something like `_teardown_vehicles_by_ids()` or `_teardown_vehicles()`?

JVillella: Good point, I updated it in #203

Issue #99, opened 2020-10-30
Camera images are blank in the first few steps when a new agent is created inside bubble


Nickniyihan:
Around 10 steps upon  the creation of boid agent, that is, 10 steps after a vehicle is hijacked by bubble and policy.act(obs) has been requested 10 times, with obs.ogm and obs.drivable_area fed to act() being blank!

aurorahcx:
The social agent is already created in _airlock_social_vehicle_with_social_agent() in the bubble manager, causing observe() to ask for grid maps while step() hasn't been called to update camera position, resulting in blank images. 
step() is not triggered because hijacked vehicle is not yet added to agent_vehicle_ids in vehicle index during airlock, until later _hijack_social_vehicle_wtih_social_agent() is called in bubble manager to swtich the control to agent. 

Waiting for feedback from @JVillella to check if this is the intended behavior of the bubble manager design.

Nickniyihan:
@aurorahcx Thanks for the thorough investigation! My question is that will this cause the agent to miss several (more than one) steps of camera images?

aurorahcx:
There is no frame mismatch between grid maps and other observation information. Currently the hijacked vehicle is providing blank images during airlock when it first enters the bubble, as it is in the process of being controlled by an agent. 

I guess what we would like to confirm is, whether it is ok for the hijacked vehicle to have blank grid maps during airlock.

PR #202, opened 2020-11-22
Perform sensor stepping for all vehicles with sensors

aurorahcx:
Currently, we manually match the vehicle list to be the same for step_sensors() and observe() in agent manager to avoid step mismatch. However, this is error-prone. 

Instead, it would be more robust to perform sensor stepping for all vehicles with sensors.

This is to address improvement suggestion by @Gamenot in https://github.com/huawei-noah/SMARTS/pull/169.

### DIFF
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -208,8 +208,7 @@ class AgentManager:
         return sim.vehicle_index.vehicle_by_id(vehicle_id).trip_meter_sensor()
 
     def step_sensors(self, sim):
-        for vehicle_id in sim.vehicle_index.sensor_states().keys():
-            sensor_state = sim.vehicle_index.sensor_states()[vehicle_id]
+        for vehicle_id, sensor_state in sim.vehicle_index.sensor_states().items():
             Sensors.step(self, sensor_state)
 
             vehicle = sim.vehicle_index.vehicle_by_id(vehicle_id)
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -120,7 +120,7 @@ class AgentManager:
         for v_id in vehicle_ids:
             vehicle = sim.vehicle_index.vehicle_by_id(v_id)
             agent_id = self._vehicle_with_sensors[v_id]
-            sensor_state = sim.vehicle_index.sensor_states()[vehicle.id]
+            sensor_state = sim.vehicle_index.sensor_state_for_vehicle_id(vehicle.id)
             observations[agent_id], dones[agent_id] = Sensors.observe(
                 sim, agent_id, sensor_state, vehicle
             )
@@ -152,7 +152,9 @@ class AgentManager:
                 ]
                 # returns format of {<agent_id>: {<vehicle_id>: {...}}}
                 sensor_states = {
-                    vehicle.id: sim.vehicle_index.sensor_states()[vehicle.id]
+                    vehicle.id: sim.vehicle_index.sensor_state_for_vehicle_id(
+                        vehicle.id
+                    )
                     for vehicle in vehicles
                 }
                 observations[agent_id], dones[agent_id] = Sensors.observe_batch(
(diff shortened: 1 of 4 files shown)


smarts/core/agent_manager.py
```diff
@@ -209,17 +207,14 @@ def _vehicle_reward(self, vehicle_id, sim):
...
 
-    def step_agent_sensors(self, sim):
-        for agent_id in self.active_agents:
-            for vehicle_id in sim.vehicle_index.vehicle_ids_by_actor_id(
-                agent_id, include_shadowers=True
-            ):
-                sensor_state = sim.vehicle_index.sensor_state_for_vehicle_id(vehicle_id)
-                Sensors.step(self, sensor_state)
+    def step_sensors(self, sim):
+        for vehicle_id in sim.vehicle_index.sensor_states().keys():
```
aurorahcx, line 211: Perform sensor stepping for all vehicles with sensors, not only for the ones controlled by active agents

Gamenot, line 211: Although `keys()` is explicit it will work just as:

```suggestion
        for vehicle_id in sim.vehicle_index.sensor_states():
```

smarts/core/agent_manager.py
```diff
@@ -209,17 +207,14 @@ def _vehicle_reward(self, vehicle_id, sim):
...
+            sensor_state = sim.vehicle_index.sensor_states()[vehicle_id]
```
Gamenot, line 212: Although `keys()` is explicit it might be better to go with:

```suggestion
        for vehicle_id, sensor_state in sim.vehicle_index.sensor_states().items():
```

smarts/core/vehicle_index.py
```diff
@@ -526,8 +526,8 @@ def build_social_vehicle(
 
         return vehicle
 
-    def sensor_state_for_vehicle_id(self, vehicle_id):
-        return self._sensor_states[vehicle_id]
+    def sensor_states(self):
+        return self._sensor_states
```
Gamenot, line 530: I think this is fine. A dictionary-like object is a reasonable interface here.

Gamenot (approved):

smarts/core/vehicle_index.py
JVillella: Can we keep the old? It's consistent with the other method signatures (*_for_vehicle_id).

smarts/core/agent_manager.py
```diff
@@ -209,17 +207,13 @@ def _vehicle_reward(self, vehicle_id, sim):
...
 
-    def step_agent_sensors(self, sim):
-        for agent_id in self.active_agents:
-            for vehicle_id in sim.vehicle_index.vehicle_ids_by_actor_id(
-                agent_id, include_shadowers=True
-            ):
-                sensor_state = sim.vehicle_index.sensor_state_for_vehicle_id(vehicle_id)
-                Sensors.step(self, sensor_state)
+    def step_sensors(self, sim):
+        for vehicle_id, sensor_state in sim.vehicle_index.sensor_states().items():
```
JVillella, line 211: Can you use `agent_vehicle_ids` instead that way we avoid returning the entire sensor state dictionary?

JVillella (changes requested):
Primary issue is returning a dictionary for sensor states.

smarts/core/agent_manager.py
aurorahcx: agent_vehicle_ids do not include shadowed agents, which was the original problem that causes https://github.com/huawei-noah/SMARTS/issues/99

smarts/core/vehicle_index.py
aurorahcx: All I need is a list of vehicles with sensors, is it not possible to maintain such a list outside of vehicle index?

aurorahcx: Ok, I have kept sensor_state_for_vehicle_id().

Instead, I have created sensor_states_items() to return a new view of the sensor states dictionary's items, similar to how it is done for vehicleitems().

Gamenot: > Can we keep the old? It's consistent with the other method signatures (*_for_vehicle_id). Returning the dictionary allows for manipulation of the reference data in an uncontrolled manner.

I had missed the reference manipulation. I would see it as fine if it were exposed as a copy. @aurorahcx 's `items()` seems to work fine now.

JVillella (approved):
`items()` is a nice idea :)

PR #203, opened 2020-11-22
Make SMARTS logs prettier

### LINKED ISSUE #213: Clean-up SMARTS logs
They've gotten out of hand recently: lots of runtime warnings, pybullet build time, panda3d pipe types, socket disconnect, malloc errors, etc. Looks very unpleasant.

JVillella:
closes #213 

Lots of noise in the logs including deprecation warnings, errors, etc. Looks quite bad. I took a first pass at it, but there's more todo before this PR can be merged.

- [x] Warning: mesh geometry_0 has a primitive with no material, using an empty RenderState
- [x] pybullet build time: Oct 18 2020 20:17:57
- [x] /Users/julian/.pyenv/versions/3.7.9/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject
  return f(*args, **kwds)


smarts/core/agent.py
```diff
@@ -63,9 +63,10 @@ def act(self, obs):
 class AgentPolicy(Agent):
     # we cannot use debtcollector here to signal deprecation because the wrapper object is
     # not pickleable, instead we simply print.
-    logger.warning(
-        "[DEPRECATED] AgentPolicy has been replaced with `smarts.core.agent.Agent`"
-    )
+    def __init__(self):
```
liamchzh, line 66: :+1: The `DEPRECATED` log is no longer dupplicated?

liamchzh: One more suggestion, remove the `{self}` of line 126, 131, 138 to make the log less annoying.

smarts/core/utils/pybullet.py
```diff
@@ -0,0 +1,28 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from .logging import disable_logging
+
+with disable_logging():
```
liamchzh, line 23: genius! :100:

smarts/core/agent.py
Gamenot: Alternate suggestion is to put the `{self}`s in `debug` logging.

JVillella:
Ok, all clean now. That turned out to be incredibly painful...

smarts/core/agent.py
JVillella: It is deprecated still, but the code was in the wrong spot.

smarts/core/remote_agent.py
```diff
@@ -106,9 +106,8 @@ def terminate(self):
             if self._conn:
                 self._conn.close()
 
-            if self._agent_proc.poll() is not None:
-                self._agent_proc.kill()
-                self._agent_proc.wait()
+            self._agent_proc.kill()
+            self._agent_proc.wait()
```
Gamenot, line 110: Were we leaking processes?

JVillella: Yes that's what the logs were saying.

smarts/core/utils/logging.py
```diff
@@ -29,3 +34,50 @@ def timeit(name: str):
...
+    # macOS
+    c_stdout = ctypes.c_void_p.in_dll(libc, "__stdoutp")
+
+
+def try_fsync(fd):
+    try:
+        os.fsync(fd)
+    except OSError:
+        # On GH actions was returning an OSError: [Errno 22] Invalid argument
+        pass
```
Gamenot, line 52: Not completely comfortable with skipping all `OSError`s here can we filter just for `[Errno 22]`?

smarts/core/utils/logging.py
```diff
@@ -29,3 +34,50 @@ def timeit(name: str):
...
+        os.close(devnull_fno)
+
+        os.dup2(dup_stdout_fno, original_stdout_fno)
+        try_fsync(dup_stdout_fno)
+        sys.stdout = os.fdopen(dup_stdout_fno, "w")
+
+        def close():
+            sys.stdout.close()
+
+        atexit.register(close)
```
Gamenot, line 83: It looks like this is registered every time `supress_stdout()` is used? I have seen this as a leak through global objects before.

smarts/core/utils/logging.py
JVillella: Why not? if the fsync fails, it's not the end of the world.

Gamenot: That is fair.

Gamenot (changes requested):

smarts/core/utils/logging.py
JVillella: That's a good point, let me think of a solution.

JVillella: This is intentional, notice it's closing the dup stream which is unique per process and changes per run.

JVillella: More ideally, we could reuse the old stdout (1), but I was unable to. Can tinker w/ it in the future. That would allow removing the need for the atexit.register.

liamchzh (approved):

Gamenot (approved):

liamchzh:
However, I am still getting logs from using `sh`, maybe this is an opportunity for us to make a contribution to `sh` :smile: 

```
/home/liamchzh/.virtualenvs/smarts-2/lib/python3.7/site-packages/sh.py:28: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
```

JVillella:
@liamchzh run pip install -r requirements.txt - you're running an old version of sh. Fixed it already :)

smarts/core/utils/logging.py
JVillella: Ok, I just improved the implementation. Now we don't need the close.

Gamenot: OK, please mark it somehow as being a problem. It will leak not so nicely since it is being called to wrap the `traci.connect()` code which can get called an uncapped number times when running SMARTS.

Gamenot (approved):
Please see my final comment.

PR #205, opened 2020-11-22
Simplify remote_agent_buffer logic with concurrent.futures

### LINKED ISSUE #53: Large delays when social vehicle gets hijacked in bubble

davidrusu:
With this change, I no longer experience any delays when vehicles enter/leave bubbles. Smooth as butter :)

fixes #53

### DIFF
--- a/smarts/core/remote_agent_buffer.py
+++ b/smarts/core/remote_agent_buffer.py
@@ -44,8 +44,13 @@ class RemoteAgentBuffer:
         if atexit.unregister is not None:
             atexit.unregister(self.destroy)
 
-        for remote_agent in self._agent_buffer:
-            remote_agent.terminate()
+        for remote_agent_future in self._agent_buffer:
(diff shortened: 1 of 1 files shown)


Adaickalavan (approved):
New code looks simpler than the old code.

smarts/core/remote_agent_buffer.py
```diff
@@ -19,82 +19,62 @@
...
+from concurrent import futures
 
 from threading import Lock, Thread
 
 from .remote_agent import RemoteAgent, RemoteAgentException
 
 
 class RemoteAgentBuffer:
-    def __init__(self, buffer_size=1):
+    def __init__(self, buffer_size=3):
```
Gamenot, line 30: I think this deserves a doc string mentioning that `buffer_size` should be greater than 0.

Gamenot (approved):

Issue #198, opened 2020-11-20
Large vehicle model and interface


iman512003:
1. Create a urdf file for large vehicle model.
2. Expose vehicle type at the agent interface level.

PR #214, opened 2020-11-24
Large vehicle urdf interface

iman512003:
This is to address #198 . This PR is an attempt to create the large vehicle urdf file and initial interface for exposing the vehicle type.


smarts/core/agent_interface.py
```diff
@@ -226,6 +231,7 @@ def __post_init__(self):
         self.accelerometer = AgentInterface._resolve_config(
             self.accelerometer, Accelerometer
         )
+        assert self.vehicle_type == "bus" or self.vehicle_type == "car"
```
davidrusu, line 234:
```suggestion
        assert self.vehicle_type in {"car", "bus"}
```

smarts/core/agent_interface.py
```diff
@@ -237,7 +243,7 @@ def from_type(requested_type: AgentType, **kwargs):
             max_episode_steps:
                 The total number of steps this interface will observe before expiring
         """
-
+        vehicle_type = kwargs.get("vehicle_type", "car")
```
davidrusu, line 246: I just realized we don't need to set the vehicle type here at all, this line will take care of overwriting any parameters for us:

https://github.com/huawei-noah/SMARTS/pull/214/files#diff-09e41aaa060eed1eb0690b068f085ab0de6a9d7aae708a50c2cef760936d49f3R326

smarts/core/models/controller_parameters.yaml
```diff
@@ -1,14 +1,51 @@
...
+            windup_gain: 0.01
+    chassis:
+            chassis_aero_force_gain : 0.63
+            max_brake_gain : 10000
+            max_turn_radius : 2.2
+            wheel_radius : 0.31265
+            max_torque : 1000
+            max_btorque : 800
+            max_steering : 12.56
+            steering_gear_ratio : 17.4
```
davidrusu, line 51: add a newline to the end of the file

smarts/core/vehicle.py
```diff
@@ -314,17 +316,38 @@ def build_agent_vehicle(
...
             showbase=sim,
             chassis=AckermannChassis(
                 pose=start_pose,
                 bullet_client=sim.bc,
-                vehicle_filepath=filepath,
+                vehicle_filepath=vehicle_filepath,
                 tire_parameters_filepath=tire_filepath,
                 friction_map=surface_patches,
-                controller_parameters_filepath=controller_filepath,
+                controller_parameters_filepath=controller_parameters,
```
davidrusu, line 350: rename this argument to be just `controller_paremeters=controller_parameters`

smarts/core/chassis.py
```diff
@@ -229,35 +229,50 @@ def __init__(
...
+            with pkg_resources.path(models, "vehicle.urdf") as path:
+                vehicle_filepath = str(path.absolute())
+
+        if controller_parameters_filepath is None:
+            with pkg_resources.path(
+                models, "controller_parameters.yaml"
+            ) as controller_path:
+                controller_filepath = str(controller_path.absolute())
+            with open(controller_filepath, "r") as controller_file:
+                controller_parameters_filepath = yaml.safe_load(controller_file)["car"]
```
davidrusu, line 242: We resolve these paths when we build the AckermanChassis in `vehicle.py`, I think we can remove these.

smarts/core/agent_interface.py
```diff
@@ -207,6 +207,11 @@ class AgentInterface:
     The choice of action space, this action space also decides the controller that will be enabled.
     """
 
+    vehicle_type: str = "car"
```
davidrusu, line 210: How did you choose the word `"car"`, is this the first time we've used this word to refer to the normal "car" vehicle? I know SUMO uses "passenger". I want to make sure we are being consistent about our vehicle types.

davidrusu (changes requested):

smarts/core/agent_interface.py
liamchzh: Is `car` from the [VEHICLE_CONFIGS](https://github.com/huawei-noah/SMARTS/blob/master/smarts/core/vehicle.py#L74)?  I think `car` is too general and should be replaced with `sedan` or something.

smarts/core/scenario.py
```diff
@@ -662,7 +662,10 @@ def plane_filepath(self):
 
     @property
     def vehicle_filepath(self):
-        return os.path.join(self._root, "vehicle.urdf")
+        for fname in os.listdir(self._root):
```
liamchzh, line 665: We are breaking some conventions. Is it possible to return a dict like `{"bus": "bus.urdf", "car": "vehicle.urdf"}`, then in `smarts/core/vehicle.py` no need to do the vehicle_filepath discovery.

smarts/core/vehicle.py
```diff
@@ -314,17 +316,38 @@ def build_agent_vehicle(
...
         )
 
+        if agent_interface.vehicle_type == "car":
+            urdf_name = "vehicle"
+        elif agent_interface.vehicle_type == "bus":
+            urdf_name = "bus"
+        else:
+            raise Exception("Vehicle type does not exist!!!")
+
+        if (vehicle_filepath is None) or not os.path.exists(vehicle_filepath):
```
liamchzh, line 326: As I mentioned above, if the vehicle_filepath is a dict like `{"bus": "bus.urdf", "car": "vehicle.urdf"}`, then you can locate the urdf based on the agent_interface.vehicle_type: `urdf_filepath = vehicle_filepath[agent_interface.vehicle_type]`. Likely need to rename `vehicle_filepath` to something else.

smarts/core/chassis.py
```diff
@@ -51,6 +51,14 @@
     RoadSurfacePatch,
 )
 
+vehicle_filepath, controller_parameters = None, None
+with pkg_resources.path(models, "vehicle.urdf") as path:
+    vehicle_filepath = str(path.absolute())
```
liamchzh, line 56: This `vehicle_filepath` is used for? Couldn't find any place using this variable.

smarts/core/scenario.py
liamchzh: Yes, now I see why.

smarts/core/chassis.py
iman512003: Thanks, this is the default value for the `chassis` class for the tests as in many of them instances of chassis are created without explicitly defining the paths, these are default values for them

davidrusu (approved):
looks good :ship:

liamchzh (approved):

PR #217, opened 2020-11-24
Use hijack state instead of prev_pos for transitions

### LINKED ISSUE #216: Travelling bubbles are missing in loop scenario
As shown in the screenshot, BUBBLE-AGENT-* appear in the score board, but no bubbles to display:
image

This is because `bubble_geometry` emitted to Envision in `_try_emit_envision_state` is empty.

JVillella:
closes #216 

Turns out the `in_*(prev_pos)` mechanics don't work for travelling bubbles because the previous and current position could be nearly identical while the travelling bubble forces a state change.

**(some other tests are failing now, needs further investigation)**

### DIFF
--- a/smarts/core/bubble_manager.py
+++ b/smarts/core/bubble_manager.py
@@ -233,28 +233,21 @@ class Cursor:
         )
 
         transition = None
+        # XXX: When a travelling bubble disappears and an agent is airlocked or
+        #      hijacked. It remains in that state.
         # TODO: Depending on step size, we could potentially skip transitions (e.g.
         #       go straight to relinquish w/o hijacking first). This may be solved by
         #       time-based airlocking. For robust code we'll want to handle these
         #       scenarios (e.g. hijacking if didn't airlock first)
-        if (
-            is_social
-            and not is_shadowed
-            and is_admissible
-            and in_airlock(pos)
-            and not (in_airlock(prev_pos) or in_bubble(prev_pos))
-        ):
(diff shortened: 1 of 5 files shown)


smarts/core/bubble_manager.py
```diff
@@ -233,28 +233,21 @@ def in_airlock(pos_):
         )
 
         transition = None
+        # XXX: When a travelling bubble disappears and an agent is airlocked or
+        #      hijacked. It remains in that state.
         # TODO: Depending on step size, we could potentially skip transitions (e.g.
         #       go straight to relinquish w/o hijacking first). This may be solved by
         #       time-based airlocking. For robust code we'll want to handle these
         #       scenarios (e.g. hijacking if didn't airlock first)
-        if (
```
liamchzh, line 240: Can you give an explanation of why deleting `in_*(prev_pos)` would work? Sorry, I didn't understand the problem when looking at the code

aurorahcx (commented):
I tested the change, the travelling bubble works now. For the two tests that are failing, should we disable them for now?

smarts/core/bubble_manager.py
JVillella: This delta only works if the vehicle is moving. If the vehicle is stationary and the bubble is travelling the delta is 0.

liamchzh (approved):

aurorahcx (approved):

PR #218, opened 2020-11-25
Ignore folders to reduce docker image size.

Adaickalavan:
Currently, the `huaweinoah/smarts:v0.4.4` docker image in Dockerhub expands to an on disk size of 9.45GB. 

Hence, added a `.dockerignore` file to ignore folders while building the docker image, in order to reduce its size.


Gamenot (approved):
Looks great. Perhaps you may want to add more filters similar to the .gitignore.

We would have to see what @JVillella says about it since he had an idea of how to go about using it when making the docker image. It would best to settle on how it should be used rather than piecing together some kind of chimera.

JVillella:
Great! Thanks for adding this Adai

My sense is the `.dockerignore` can basically be our `.gitignore`... we could symlink for simplicity :) there's lots more beyond the ignored directories you have here we should ignore.

JVillella (approved):

PR #224, opened 2020-11-25
Use docker container directly in CI

Gamenot:
This does not add much to the size but makes the image more useful.

### DIFF
--- a/.github/workflows/ci-base-tests.yml
+++ b/.github/workflows/ci-base-tests.yml
@@ -2,8 +2,6 @@ name: SMARTS CI
 
 on: [push]
 
-env:
-  venv_dir: .venv
 jobs:
   test:
     runs-on: ubuntu-18.04
@@ -21,8 +19,6 @@ jobs:
             -logfile ./xdummy.log \
             -config /etc/X11/xorg.conf :1 &
           cd $GITHUB_WORKSPACE
-          python3.7 -m venv "${venv_dir}"
-          . "${venv_dir}/bin/activate"
           pip install --upgrade pip
           pip install wheel
           pip install -r requirements.txt
@@ -30,6 +26,5 @@ jobs:
       - name: Run smoke tests
         run: |
           cd $GITHUB_WORKSPACE
-          . "${venv_dir}/bin/activate"
           make test
 
--- a/.github/workflows/ci-base-tests.yml
+++ b/.github/workflows/ci-base-tests.yml
@@ -7,7 +7,7 @@ env:
 jobs:
   test:
     runs-on: ubuntu-18.04
-    container: huaweinoah/smarts:v0.4.3-pre
+    container: huaweinoah/smarts:v0.4.6
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-base-tests.yml
```diff
@@ -2,12 +2,10 @@ name: SMARTS CI
 
 on: [push]
 
-env:
-  venv_dir: .venv
```
JVillella, line 6: What about PRs that add a new package? Looks like a docker build will need to be deployed first...but a new package needs to go out before we'd deploy a new build...

Sounds like a 🐔 and 🥚

Dockerfile
```diff
@@ -57,6 +57,10 @@ RUN echo "Cleaning-up"
 RUN apt-get autoremove && \
     rm -rf /var/lib/apt/lists/*
 
+RUN add-apt-repository ppa:git-core/ppa
+RUN apt-get update
+RUN apt-get install -y git
```
JVillella, line 62: For convenience? What's the use case?

JVillella (changes requested):
See inline comments

Dockerfile
Gamenot: This is mainly for convenience for development. Perhaps removing the `apt-get install` would be far enough, the docker instance does not have git and `apt` does not know where it is located by default.

.github/workflows/ci-base-tests.yml
Gamenot: That would be the case but we run an update to the packages already in in the scripts

https://github.com/huawei-noah/SMARTS/blob/master/.github/workflows/ci-base-tests.yml#L26-L29

liamchzh (approved):

.github/workflows/ci-base-tests.yml
```diff
@@ -21,15 +19,12 @@ jobs:
             -logfile ./xdummy.log \
             -config /etc/X11/xorg.conf :1 &
           cd $GITHUB_WORKSPACE
-          python3.7 -m venv "${venv_dir}"
-          . "${venv_dir}/bin/activate"
           pip install --upgrade pip
           pip install wheel
-          pip install -r requirements.txt
+          pip install --no-deps -r requirements.txt
```
JVillella, line 24: why?

if they're cached it'll just go over em quickly

JVillella (approved):
Small comment, otherwise 👍

.github/workflows/ci-base-tests.yml
Gamenot: I was getting dependency collisions.

Dockerfile
Gamenot: This was removed because it was not needed.

Gamenot:
This has suddenly stopped working. Not suitable to go in as is.

.github/workflows/ci-base-tests.yml
JVillella: This is likely _not_ the solution to that. If these are dep collisions of our own dependencies than we need to ease up our requirements.txt / setup.py to let them resolve.

JVillella (changes requested):
Have an issue w/ the --no-deps usage.

Gamenot:
https://github.com/huawei-noah/SMARTS/pull/587 covers this in a better way.

PR #227, opened 2020-11-26
Fix collision detection

### LINKED ISSUE #221: Collision check failure
Collision is failing to trigger in commit https://github.com/huawei-noah/SMARTS/commit/7104d4b974303f484af41995647c605a6929c2a1 in some cases.

`python examples/single_agent.py scenarios/intersections/4lane_t`

collision_failure

AC:

- [ ] Figure out why collision tests do not find issue
- [ ] Update collision tests to check for this issue
- [ ] Fix collision

liamchzh:
The issue was caused by:
1. The sim maintains the wrong keys of the agent collisions.
2. `sensor.py` passes a wrong vehicle ID to query the collision.

close #221 

* [x] Add a sim-level test to ensure collision detection works (Previously we test the detection at chassis level)
* [x] Fix the bug.

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -285,7 +285,7 @@ class Sensors:
         done_criteria = interface.done_criteria
 
         collided = (
-            sim.vehicle_did_collide(vehicle) if done_criteria.collision else False
+            sim.vehicle_did_collide(vehicle.id) if done_criteria.collision else False
         )
         is_off_road = (
             cls._vehicle_is_off_road(sim, vehicle) if done_criteria.off_road else False
--- a/smarts/core/tests/test_collision.py
+++ b/smarts/core/tests/test_collision.py
@@ -1,17 +1,26 @@
 import math
 import importlib.resources as pkg_resources
 import time
-
 from pathlib import Path
 
-from smarts.core.utils import pybullet
-from smarts.core.utils.pybullet import bullet_client as bc
 import pytest
 
+from helpers.scenario import temp_scenario
 from smarts.core import models
+from smarts.core.agent_interface import (
+    AgentInterface,
+    ActionSpaceType,
+)
 from smarts.core.chassis import AckermannChassis, BoxChassis
 from smarts.core.coordinates import Heading, Pose
+from smarts.core.scenario import Scenario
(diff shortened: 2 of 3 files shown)


aurorahcx (approved):

aurorahcx:
Maybe in another PR, is it possible to update the collision test to catch this issue?

liamchzh:
> Maybe in another PR, is it possible to update the collision test to catch this issue?

Yes, should have a goalkeeper to ensure that.

aurorahcx:
Nice test! I see that the updated test can catch the mistake in smarts.py, but doesn't catch the mistake in sensors.py. Perhaps it is more robust to check both events in observations and done?

smarts/core/tests/test_collision.py
```diff
@@ -197,3 +206,51 @@ def test_ackerman_chassis_size_unchanged(bullet_client: bc.BulletClient):
...
+
+
+def test_sim_level_collision(smarts, scenarios):
+    scenario = next(scenarios)
+    smarts.reset(scenario)
+
+    collisions = []
+
+    for _ in range(30):
+        observations, _, _, _ = smarts.step({AGENT_1: "keep_lane"})
```
aurorahcx, line 252: Nice test! I see that the updated test can catch the mistake in smarts.py, but doesn't catch the mistake in sensors.py. Perhaps it is more robust to check both events in observations and dones?

liamchzh: Yes, good point. I updated the test.

smarts/core/sensors.py
```diff
@@ -285,7 +285,7 @@ def _is_done_with_events(cls, sim, agent_id, vehicle, sensor_state):
         done_criteria = interface.done_criteria
 
         collided = (
-            sim.vehicle_did_collide(vehicle) if done_criteria.collision else False
+            sim.vehicle_did_collide(vehicle.id) if done_criteria.collision else False
```
Gamenot, line 288: Since this became a problem we may want to assert inside `vehicle_did_collide`

Gamenot (approved):

liamchzh:
@Gamenot We need to release a new version after this one goes in since the smarts service is using SMARTS as a library.

Issue #145, opened 2020-11-09
Interpolate along shape waypoints instead of along waypoints.


davidrusu:
Instead of interpolating along the fully generated waypoints, we can interpolate instead along the shape waypoints when generating waypoint paths.

This could give us a nice little performance improvement.

@iman512003

PR #232, opened 2020-11-27
Add shape to LinkledWaypoints

iman512003:
This is to address #145 .
comparison


smarts/core/waypoints.py
```diff
@@ -507,6 +514,7 @@ def _process_interp_for_lane_wp(
                     lane_index=shape_wp.wp.lane_index,
                 ),
                 nexts=[],
+                is_shape_wp=0,
```
Gamenot, line 517: If this is going to be binary I would prefer `True|False`, otherwise if it has more options it should be an enum.

smarts/core/waypoints.py
```diff
@@ -290,22 +291,27 @@ def _equally_spaced_path(self, path, point):
...
-                waypoint.wp.lane_index
-            )
-            ref_waypoints_coordinates["ref_wp_lane_width"].append(
-                waypoint.wp.lane_width
-            )
-            ref_waypoints_coordinates["ref_wp_speed_limit"].append(
-                waypoint.wp.speed_limit
-            )
+        for idx, waypoint in enumerate(path):
+            if idx == 0 or idx == len(path) - 1 or waypoint.is_shape_wp == 1:
```
Gamenot, line 295:
```suggestion
            if not waypoint.is_shape_wp or idx > 0 and idx < len(path) - 1:
                continue
```

smarts/core/waypoints.py
```diff
@@ -112,6 +112,7 @@ def __eq__(self, other):
         "wp",  # Waypoint: current waypoint
         "nexts",  # list of LinkedWaypoint: list of next immediate waypoints
         # it's a list of waypoints because a path may branch at junctions
+        "is_shape_wp",
```
liamchzh, line 115: Is it possible to assign a default value?

smarts/core/waypoints.py
liamchzh: Agree.

smarts/core/waypoints.py
iman512003: Thanks, sure, but the code which you have attached has issues, according to the `De Morgan's laws` in the following link:

https://en.wikipedia.org/wiki/De_Morgan%27s_laws
The correct syntax would be something like this
```
if not waypoint.is_shape_wp and 0<idx<len(path) - 1:
                continue
```

smarts/core/waypoints.py
iman512003: Thanks, sure.

smarts/core/waypoints.py
iman512003: Thanks, sure.

davidrusu (approved):
I see why you went the way you did with this implementation, but as it stands this complicates the Waypoints module just a little bit more, we should justify it with a little bit of performance testing before merging in this experiment.

A simple flamegraph of an ego agent with a large waypoint lookahead before and after this change should show us what we need.

Besides that, looks reasonable :+1:

iman512003:
> I see why you went the way you did with this implementation, but as it stands this complicates the Waypoints module just a little bit more, we should justify it with a little bit of performance testing before merging in this experiment.
> 
> A simple flamegraph of an ego agent with a large waypoint lookahead before and after this change should show us what we need.
> 
> Besides that, looks reasonable +1



> I see why you went the way you did with this implementation, but as it stands this complicates the Waypoints module just a little bit more, we should justify it with a little bit of performance testing before merging in this experiment.
> 
> A simple flamegraph of an ego agent with a large waypoint lookahead before and after this change should show us what we need.
> 
> Besides that, looks reasonable +1
Thanks,
comparison

smarts/core/waypoints.py
```diff
@@ -112,7 +112,9 @@ def __eq__(self, other):
...
     ],
+    defaults=[False],
```
JVillella, line 117: That is not right; you need to put a default for every field. E.g. `[None, [], False]`.

JVillella (approved):

smarts/core/waypoints.py
iman512003: Thanks, sure, I changed it to [`None,[], False]`, I thought we only needed the default value for the `is_shape_wp`.

liamchzh (approved):

PR #237, opened 2020-11-30
Mention environment reward in the observations.rst

Gamenot:


docs/sim/observations.rst
```diff
@@ -71,6 +71,10 @@ See implemention in :class:`smarts.core.sensors`
 
 Then, you can choose the observations needed through :class:`smarts.core.agent_interface.AgentInterface` and process these raw observations through :class:`smarts.core.observation_adapter`.
 
+=======
+Rewards
+=======
+The reward from smarts is given by a calculation in the `TripMeterSensor`. This is 0 or values `< -0.5 | 0.5 <` relating to distance traveled in meters when a vehicle has gone at least 0.5 meters since the last reward. `env_reward` from the smarts environment directly use this.
```
liamchzh, line 77: How does this `< -0.5 | 0.5 <` mean here?

JVillella: Minor: Do we need to mention TripMeterSensor? It's not extensible to the user unless they modify the source. I saw these docs as public API docs.

This is hard to understand as Liam said,

>This is 0 or values `< -0.5 | 0.5 <`

JVillella (changes requested):
Few minor comments

docs/sim/observations.rst
Gamenot: I have clarified it.

liamchzh: I am not sure whether we need to give the exact range in this `Rewards` explanation. 

But I am fine with both.

liamchzh (approved):

docs/sim/observations.rst
Gamenot: I had a question about what the reward was from someone doing research so I would say it is better to.

aurorahcx (approved):

Adaickalavan (approved):

Issue #219, opened 2020-11-25
Sort out the use case for the SMARTS Docker container


Gamenot:
## Use cases:
- Production
- Local development (creating an image on the spot with the current repository state *could be useful)
- CI

## Core:
- [x] Decide on how the docker container should be used
- [x] Update documentation for the use cases
- [x] Add a `.dockerignore`
   - Filter out any files that are not part of the SMARTS repository (.venv, .vscode, etc.)

@Adaickalavan Please change this issue as you find solutions to address the use cases.

Adaickalavan:
We have two options while building a Docker image for SMARTS. Namely, (i) `pip install smarts` or (ii) `COPY . /src`. 

1. `pip install smarts`
+ Here, we pip install the smarts package from PyPI index into the Docker image.
+ smarts source code frozen at PyPI release version used.
+ Further codes within the Docker can use smarts via `import smarts` command. 
+ Used primarily in production.
+ Less likely to be used as a base layer for other Docker images.
+ For example, if one needs NVIDIA support, he would use the NVIDIA Docker image as base layer and build smarts on top of it by copying the contents of smarts Dockerfile into his own new Dockerfile.  
  
2. `COPY . /src`
+ Here, we copy smarts source code from local host into Docker image.
+ smarts source code in Docker image mirrors state in local host at the time the Docker image was built.
+ We have the option of executing `pip install -e .` inside the Docker image, which installs smarts in an editable mode from the `/src` project path inside the Docker image.
+ Further codes within the Docker can use smarts by directly referring to the source files or by `import smarts` command.
+ Used primarily in development.

Maintaining only one Docker image makes the project more streamlined. The cucrrent Docker image with `COPY . /src` command may be more preferable, as Docker image with `pip install smarts` is less likely to be utilised. If agreeable, then no change is needed to the project now. Hence, the current Docker image caters more towards development.

If you intend to build a smarts Docker image as a base layer for production, you merely need to replace the lines 
```
COPY . /src
RUN cd /usr/local/lib/python3.7 && \
    mkdir site-packages && \
    cd site-packages && \
    python /src/setup.py develop 
``` 
to 
```
RUN pip install smarts
```
in the Dockerfile.

A dockerignore file is being added to the project through merge request #218.

Adaickalavan:
Summary of two approaches to using Dockers:

1. Have two Dockerfiles, namely, Dev.Dockerfile and Prod.Dockerfile
[...]

PR #242, opened 2020-12-01
Separate Dockerfiles for development and production.

Adaickalavan:
As discussed in issue #219, we may maintain only one official Docker image in DockerHub which is primarily catered towards development needs.

However, to make it easier for end users variable needs, we may provide two Dockerfiles. Namely, `Prod.Dockerfile` to build production Docker image and  `Dev.Dockerfile` to build development Docker image. The `Readme.md` file has been updated to reflect this idea.

The official hosted Docker image shall only be built from `Dev.Dockerfile`. 

Actually, the only difference between `Prod.Dockerfile` and `Dev.Dockerfile` is whether we use `RUN pip install smarts` or `COPY . /src`.


JVillella (approved):

JVillella (changes requested):
Sorry meant to request changes :)

Adaickalavan:
This idea was in response to the "Update documentation for the use cases" checkbox in issue #219. Instead of simply explaining how to modify the Dockerfile to suit production, I thought of providing a sample production Dockerfile for reference.

What changes do you suggest? 

We can drop this idea, if we think that multiple Dockerfiles bloat the codebase.

PR #263, opened 2020-12-03
Improvments on cut in and u-turn interaction between agents and social vehicles

liamchzh:
1. Add trigger radius to task mission
2. Introduce `max_speed` to `TrafficActor`. This will allow setting the max speed lower than the road speed limit, which makes it easier to do a cut in or u turn.

The u-turn and cut in task become meaningful only when agents have interaction with other vehicles. Before we can pin a bubble to a social vehicle, we introduce this `trigger radius` to allow users to specify a radius, the task will be triggered if any vehicles within this radius.

ezgif com-gif-maker (9)

 The agent keeps going straight until the social vehicle shows up.

### DIFF
--- a/smarts/core/mission_planner.py
+++ b/smarts/core/mission_planner.py
@@ -48,7 +48,7 @@ class MissionPlanner:
         self._route = None
         self._road_network = road_network
         self._did_plan = False
-        self._trigger_task = False
+        self._task_is_triggered = False
 
     def random_endless_mission(
         self, min_range_along_lane=0.3, max_range_along_lane=0.9
@@ -272,7 +272,7 @@ class MissionPlanner:
             vehicle=vehicle, radius=radius
         )
 
-        if not neighborhood_vehicles and not self._trigger_task:
+        if not neighborhood_vehicles and not self._task_is_triggered:
             return []
 
         start_lane = self._road_network.nearest_lane(
@@ -286,7 +286,7 @@ class MissionPlanner:
         if not start_edge.oncoming_edges:
             return []
 
-        self._trigger_task = True
+        self._task_is_triggered = True
 
         oncoming_edge = start_edge.oncoming_edges[0]
         oncoming_lanes = oncoming_edge.getLanes()
--- a/smarts/core/mission_planner.py
+++ b/smarts/core/mission_planner.py
@@ -29,7 +29,7 @@ from .sumo_road_network import SumoRoadNetwork
 from .scenario import EndlessGoal, LapMission, Mission, Start, default_entry_tactic
 from .waypoints import Waypoint, Waypoints
 from .route import ShortestRoute, EmptyRoute
-from .coordinates import Pose
+from .coordinates import Pose, Heading
 from .route import ShortestRoute, EmptyRoute
 from .scenario import EndlessGoal, LapMission, Mission, Start
 from .sumo_road_network import SumoRoadNetwork
@@ -245,7 +245,7 @@ class MissionPlanner:
         prev = position[:2]
         for i in range(len(p_x)):
             pos = np.array([p_x[i], p_y[i]])
-            heading = vec_to_radians(pos - prev)
+            heading = Heading(vec_to_radians(pos - prev))
             prev = pos
             lane = self._road_network.nearest_lane(pos)
             lane_id = lane.getID()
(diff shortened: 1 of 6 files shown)


zoo/policies/open-agent/open_agent/agent.py
```diff
@@ -405,7 +405,8 @@ def update_impatience(self, ego_state):
 
     def act(self, obs):
         ego = obs.ego_vehicle_state
-        wps = obs.waypoint_paths[len(obs.waypoint_paths) // 2]
+        lane_index = min(len(obs.waypoint_paths) - 1, ego.lane_index)
+        wps = obs.waypoint_paths[lane_index]
```
liamchzh, line 409: The intention of this change is to keep the OpEn stay in the current lane.

smarts/core/mission_planner.py
```diff
@@ -263,9 +264,17 @@ def cut_in_waypoints(self, sim, pose: Pose, vehicle):
...
-    def uturn_waypoints(self, pose: Pose):
+    def uturn_waypoints(self, sim, pose: Pose, vehicle):
         # TODO: 1. Need to revisit the approach to calculate the U-Turn trajectory.
         #       2. Wrap this method in a helper.
+        radius = self._mission.task.trigger_radius
+        neighborhood_vehicles = sim.neighborhood_vehicles_around_vehicle(
+            vehicle=vehicle, radius=radius
+        )
+
+        if not neighborhood_vehicles and not self._trigger_task:
```
davidrusu, line 275: I do think we should break out a separate `Task.triggered()` method

liamchzh: Yes, it's more intuitive to let `Task` maintain the `triggered` status. I tried but turns out there are some drawbacks and limitation:

1. The `Task` becomes mutable. 
2. `Task` object will be used between episodes, which means we have to introduce a mechanism to reset the `Task` when an episode ends.

No.1 is fine, it did raise an `unhashable` error but can be addressed by adding a custom `__hash__` method.
No.2 is more tricky and hard to workaround.

I reverted my changes, just renamed it to `_task_is_triggered`.

smarts/core/mission_planner.py
```diff
@@ -228,7 +229,7 @@ def cut_in_waypoints(self, sim, pose: Pose, vehicle):
         target_offset = self._road_network.offset_into_lane(
             target_lane, target_position
         )
-        if offset < target_offset:
+        if offset < target_offset + 5:
```
Gamenot, line 232: In theory this padding should also be parameterizable. It might be useful calculate the padding using some value related to the road speed or vehicle size, or parameterization from the task.

smarts/core/mission_planner.py
```diff
@@ -299,7 +310,7 @@ def uturn_waypoints(self, pose: Pose):
             # agent at the start edge
             p0 = pose.position[:2]
             distance = (
-                10 * abs(abs(target_heading - heading) - math.pi / 2) / (math.pi / 2)
+                15 * abs(abs(target_heading - heading) - math.pi / 2) / (math.pi / 2)
```
Gamenot, line 313: This curve value perhaps deserves to be part of the task attributes.

Gamenot (commented):
Some comments but looks good.

zoo/policies/open-agent/open_agent/agent.py
liamchzh: @davidrusu this change needs a close look

smarts/core/mission_planner.py
liamchzh: This one shouldn't be exposed, because it's leaking implementation details rather than a configurable param to users.

smarts/core/mission_planner.py
liamchzh: Hmm, the only reason not doing it now is because if it's parameterizable then we should make sure the input from users is valid, reasonable, and practical.

smarts/core/mission_planner.py
Gamenot: OK, if I am reading this correctly, this equation models an intent for a larger offset based on the difference between vehicle headings. Perhaps you are right here.

smarts/core/mission_planner.py
Gamenot: :ok_hand:

Gamenot (approved):

zoo/policies/open-agent/open_agent/agent.py
davidrusu: does lane index map directly to the waypoint path index? You may need to inspect the waypoints themselves to and look for a path that starts with a waypoint that has the correct edge id and lane index.

liamchzh: I see. Need to check the lane_index of the first waypoint. Also need to ensure works at intersections.

liamchzh: I loop through the `waypoint_paths` to see if I can match one, choose the first path that has the same lane_index.

davidrusu (approved):

PR #264, opened 2020-12-03
247 Authentication of listeners and clients

### LINKED ISSUE #247: Secure Zoo Worker Communication
We make use of `multiprocessing.Client` and `multiprocessing.Listener` to communicate between SMARTS, Zoo Workers, and Agent Runners.

Currently we do not perform any authentication on these servers so anyone who knows the IP/PORT of these workers can communicate with them.

`Client/Listener` have the option to take an `authkey` to authenticate requests, we should use this to protect these workers since the `run_agent.py` scripts execute the code they are given.

https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Listener
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Client

AC:
  * `scl zoo worker` Authenticates requests with a secret key
  * `run_agent.py` Authenticates connections with the same secret key
  * `SMARTS` takes the secret key as an argument and uses it to authenticate with these zoo workers.

Adaickalavan:
Authentication added to 
+ multiprocessing.connection.Client, and
+ multiprocessing.connection.Listener.

Authentication is an optional feature. If no authentication key is provided, then a byte string key of b'secret' is used by default.

Closes #247

### DIFF
--- a/examples/single_agent.py
+++ b/examples/single_agent.py
@@ -35,7 +35,7 @@ class ChaseViaPointsAgent(Agent):
         )
 
 
-def main(scenarios, headless, num_episodes, seed, max_episode_steps=None):
+def main(scenarios, headless, num_episodes, seed, authkey=None, max_episode_steps=None):
     agent_spec = AgentSpec(
         interface=AgentInterface.from_type(
             AgentType.Laner, max_episode_steps=max_episode_steps
@@ -53,6 +53,7 @@ def main(scenarios, headless, num_episodes, seed, max_episode_steps=None):
         sumo_headless=True,
         seed=seed,
         # zoo_workers=[("143.110.210.157", 7432)], # Distribute social agents across these workers
+        authkey=authkey,
         # envision_record_data_replay_path="./data_replay",
     )
 
@@ -73,6 +74,11 @@ def main(scenarios, headless, num_episodes, seed, max_episode_steps=None):
 
 if __name__ == "__main__":
     parser = default_argument_parser("single-agent-example")
+    parser.add_argument(
+        "--authkey",
+        help="Authentication key for connection to run agent",
+        default=None,
+    )
     args = parser.parse_args()
 
     main(
@@ -80,4 +86,5 @@ if __name__ == "__main__":
         headless=args.headless,
         num_episodes=args.episodes,
         seed=args.seed,
+        authkey=args.authkey,
     )
--- a/examples/single_agent.py
+++ b/examples/single_agent.py
@@ -35,7 +35,9 @@ class ChaseViaPointsAgent(Agent):
         )
 
 
-def main(scenarios, headless, num_episodes, seed, authkey=b'secret', max_episode_steps=None):
+def main(
+    scenarios, headless, num_episodes, seed, authkey=b"secret", max_episode_steps=None
+):
     agent_spec = AgentSpec(
         interface=AgentInterface.from_type(
             AgentType.Laner, max_episode_steps=max_episode_steps
@@ -73,11 +75,11 @@ def main(scenarios, headless, num_episodes, seed, authkey=b'secret', max_episode
 
 
 if __name__ == "__main__":
-    parser = default_argument_parser("single-agent-example") 
+    parser = default_argument_parser("single-agent-example")
     parser.add_argument(
         "--authkey",
         help="Authentication key for connection to run agent",
-        default=b'secret',
+        default=b"secret",
         type=bytes,
     )
     args = parser.parse_args()
--- a/examples/single_agent.py
+++ b/examples/single_agent.py
@@ -35,7 +35,9 @@ class ChaseViaPointsAgent(Agent):
         )
 
 
-def main(scenarios, headless, num_episodes, seed, auth_key, max_episode_steps=None):
+def main(
+    scenarios, headless, num_episodes, seed, auth_key=b"None", max_episode_steps=None
+):
     agent_spec = AgentSpec(
         interface=AgentInterface.from_type(
             AgentType.Laner, max_episode_steps=max_episode_steps
--- a/examples/single_agent.py
+++ b/examples/single_agent.py
@@ -36,7 +36,7 @@ class ChaseViaPointsAgent(Agent):
 
 
 def main(
-    scenarios, headless, num_episodes, seed, auth_key=b"secret", max_episode_steps=None
+    scenarios, headless, num_episodes, seed, auth_key, max_episode_steps=None
 ):
     agent_spec = AgentSpec(
         interface=AgentInterface.from_type(
@@ -77,18 +77,18 @@ def main(
 if __name__ == "__main__":
     parser = default_argument_parser("single-agent-example")
     parser.add_argument(
-        "--authkey",
+        "--auth_key",
         type=str,
-        default="secret",
+        default=None,
         help="Authentication key for connection to run agent",
     )
     args = parser.parse_args()
-    args.authkey = str.encode(args.authkey)
+    args.auth_key = str.encode(str(args.auth_key))
 
     main(
         scenarios=args.scenarios,
         headless=args.headless,
         num_episodes=args.episodes,
         seed=args.seed,
-        auth_key=args.authkey,
+        auth_key=args.auth_key,
     )
--- a/examples/single_agent.py
+++ b/examples/single_agent.py
@@ -78,11 +78,12 @@ if __name__ == "__main__":
     parser = default_argument_parser("single-agent-example")
     parser.add_argument(
         "--authkey",
+        type=str,
+        default="secret",
         help="Authentication key for connection to run agent",
-        default=b"secret",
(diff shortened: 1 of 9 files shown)


examples/single_agent.py
```diff
@@ -35,7 +35,9 @@ def act(self, obs: Observation):
         )
 
 
-def main(scenarios, headless, num_episodes, seed, max_episode_steps=None):
+def main(
+    scenarios, headless, num_episodes, seed, authkey=b"secret", max_episode_steps=None
+):
```
Gamenot, line 40: Please change the default for `authkey` to be `None`. A value like `"secret"` is a bit arbitrary.

Also please rename to `auth_key` since our convention splits by "_" between words.

smarts/core/agent_manager.py
```diff
@@ -41,9 +41,11 @@ class AgentManager:
          time.
     """
 
-    def __init__(self, interfaces, zoo_workers=None):
+    def __init__(self, interfaces, zoo_workers=None, authkey=b"secret"):
```
Gamenot, line 44:
```suggestion
    def __init__(self, interfaces, zoo_workers=None, auth_key=None):
```

smarts/core/remote_agent_buffer.py
```diff
@@ -32,17 +32,23 @@
 
 
 class RemoteAgentBuffer:
-    def __init__(self, zoo_worker_addrs=None, buffer_size=3):
+    def __init__(self, zoo_worker_addrs=None, authkey=b"secret", buffer_size=3):
```
Gamenot, line 35:
```suggestion
    def __init__(self, zoo_worker_addrs=None, auth_key=None, buffer_size=3):
```

smarts/core/remote_agent_buffer.py
```diff
@@ -83,15 +89,17 @@ def destroy(self):
...
+        )
         self._local_zoo_worker.start()
 
         local_address = ("0.0.0.0", local_port)
 
         # Block until the local zoo server is accepting connections.
         for i in range(retries):
             try:
-                conn = Client(local_address, family="AF_INET")
+                conn = Client(local_address, family="AF_INET", authkey=self._authkey)
```
Gamenot, line 102: The naming convention should not leak outside.

```suggestion
                conn = Client(local_address, family="AF_INET", authkey=self._auth_key)
```

smarts/core/smarts.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         timestep_sec=0.1,
         reset_agents_only=False,
         zoo_workers=None,
+        authkey=b"secret",
```
Gamenot, line 91:
```suggestion
        auth_key=None,
```

smarts/env/hiway_env.py
```diff
@@ -81,6 +83,7 @@ def __init__(
         envision_endpoint=None,
         envision_record_data_replay_path=None,
         zoo_workers=None,
+        authkey=b"secret",
```
Gamenot, line 86:
```suggestion
        auth_key=None,
```

smarts/core/remote_agent_buffer.py
```diff
@@ -32,17 +32,23 @@
...
+            buffer_size: 
+                Number of RemoteAgents to pre-initialize and keep running in the background, must be non-zero (default: 3)
         """
         assert buffer_size > 0
 
         self._log = logging.getLogger(self.__class__.__name__)
 
         self._local_zoo_worker = None
         self._local_zoo_worker_addr = None
+        self._authkey = authkey
```
Gamenot, line 51: It might be useful to add a validation for the `auth_key` here.

smarts/zoo/worker.py
```diff
@@ -85,6 +89,15 @@ def listen(port):
...
         "Listens for requests to allocate agents and executes them on-demand"
     )
-    parser.add_argument("--port", help="Port to listen on", default=7432, type=int)
+    parser.add_argument(
+        "--port", type=int, default=7432, help="Port to listen on",
+    )
+    parser.add_argument(
+        "--authkey",
+        type=str,
+        default="secret",
```
Gamenot, line 98: A default at this level is OK.

smarts/core/remote_agent.py
```diff
@@ -32,7 +32,7 @@ class RemoteAgentException(Exception):
 
 
 class RemoteAgent:
-    def __init__(self, address, socket_family, connection_retries=100):
+    def __init__(self, address, socket_family, authkey, connection_retries=100):
```
Adaickalavan, line 35: def __init__(self, address, socket_family, auth_key, connection_retries=100):

examples/single_agent.py
```diff
@@ -53,6 +55,7 @@ def main(scenarios, headless, num_episodes, seed, max_episode_steps=None):
         sumo_headless=True,
         seed=seed,
         # zoo_workers=[("143.110.210.157", 7432)], # Distribute social agents across these workers
+        authkey=authkey,
```
Adaickalavan, line 58:
```suggestion
        auth_key=auth_key,
```

examples/single_agent.py
Adaickalavan: + I have renamed all `authkey` occurrences with `auth_key`.
+ `auth_key` should be of type byte string. `None` is of type `NoneType`. Hence, I have changed default `auth_key` to be `b"None"`.

smarts/core/agent_manager.py
Adaickalavan: Changed default value of `auth_key` of type byte string to be `b"None"`.

smarts/core/remote_agent_buffer.py
Adaickalavan: Changed default value of auth_key of type byte string to be b"None".

smarts/core/remote_agent_buffer.py
Adaickalavan: Changed internal `self.authkey` values to `self._auth_key`.

smarts/core/smarts.py
Adaickalavan: Changed default value of auth_key of type byte string to be b"None".

smarts/env/hiway_env.py
Adaickalavan: Changed default value of auth_key of type byte string to be b"None".

smarts/zoo/worker.py
Adaickalavan: Changed default authentication settings.
1. Running agents in remote server
    + Authentication key is mandatory. 
    + Python command example:
        ```bash
        # Launch agent in remote servers
        $ python3.7 ./smarts/zoo/worker.py --auth_key secret
        # Start Smarts in local host
        $ python3.7 ./examples/single_agent.py scenarios/loop --auth_key secret
        ```
    + scl command example:
        ```bash
        # Launch agent in remote servers
        $ scl zoo worker secret
        # Start Smarts in local host
        $ python3.7 ./examples/single_agent.py scenarios/loop --auth_key secret
        ```
1. Running agents in local host
    + Authentication key is optional.
    + Python command example:
        ```bash
        # No user specified authentication key
        $ python3.7 ./examples/single_agent.py scenarios/loop  
        # User specificed optional authentication key
        $ python3.7 ./examples/single_agent.py scenarios/loop --auth_key secret
        ```
1. Although I have changed the default authentication settings as suggested, I somewhat believe setting default authentication on the remote worker side will be helpful for new users to get started easily without worrying about the various arguments in launching the code.

smarts/core/remote_agent_buffer.py
Adaickalavan: Type assertion has been added.

```python
assert isinstance(auth_key, bytes), f"Received auth_key of type {type(auth_key)}, but need auth_key of type <class 'bytes'>."
```

examples/single_agent.py
Gamenot: My opinion here is that the internal requirements should not show up in the API. I cannot see any particular reason why the `auth_key` argument should require a byte string at the interface level. Perhaps `None` could be replaced with an empty string but unless there is some compelling reason I do not think we should be using a byte string at this level.

Gamenot (changes requested):

examples/single_agent.py
davidrusu: We should really use the `NoneType` `None` here. 

From the docs: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Listener

> If authkey is given and not None, it should be a byte string and will be used as the secret key for an HMAC-based authentication challenge. No authentication is done if authkey is None. AuthenticationError is raised if authentication fails. 

So we do really want "no authentication" here on the client side, and that is enabled with `None` as the authkey

Adaickalavan: Changed default `auth_key` to `None`.

Adaickalavan:
This pull request is taken over by the pull request #284. 
Hence, this pull request is closed.

Issue #274, opened 2020-12-04
ActionSpace LaneWithContinuousSpeed specified speed mismatches the real vehicle speed


mg2015started:
example code ( https://github.com/huawei-noah/SMARTS/tree/274-lanespeed-example/lane_speed )

command
```bash
cd lane_speed
scl scenario build scenario/lane_speed 
python run_lanespeed.py  scenario/lane_speed 
```

results
![image](https://user-images.githubusercontent.com/20150115/101145191-68e62c80-3654-11eb-91ce-8f730513b096.png)

Issue #262, opened 2020-12-02
Investigation of Lane Change Speed and Curve


Gamenot:
Lane changing using the `Laner` may see vehicles slow down during lane change and makes the change sharply.

AC:
- [ ] Find the root cause of the issue.
- [ ] Fix problem or put up a new issue and link here.

Issue #258, opened 2020-12-02
Physics check of vehicles controlled by laner controller 


liamchzh:
This is a `laner with speed` agent. When it changes to the right lane, seems unrealistic compared to lane change in the real world. The action input is `(target_speed, -1)`.

The action of laner is dynamic, but just wondering if we can tune it to make it looks like a human driver.

A few things to check:
- What's the max angular velocity
- Is the speed slow down when it changes lane

![ezgif com-gif-maker](https://user-images.githubusercontent.com/2217445/100900857-4657e600-3491-11eb-9f98-472d724b2608.gif)

liamchzh:
@mg2015started @iman512003

mg2015started:
test branch: https://github.com/huawei-noah/SMARTS/tree/258-changelane-example/lane_change

```bash
python run_changelane.py scenario/lane_change
```

output: 
![image](https://user-images.githubusercontent.com/20150115/101046699-3688f080-35bc-11eb-9d1d-3ea13e83226a.png)

PR #277, opened 2020-12-04
Enhance speed tracking performance

iman512003:
This is to address #258, #262 , #274 #278. This is particularly for addressing lane change scenarios and speed reduction during such maneuvers. This is for **cut-in** scenario where the desired speed for the following figures is **50 km/hr=13.8**, the left figure is for 0.05 time step and the right one is for 0.1.
Screenshot from 2020-12-04 15-48-21
Also, currently with this pr we can make cut-in task with high speed **(100 km/hr)** feasible
before this pr, we had the following curve for the same cut-in scenario and desired speed.
speed_reduction1
From, above figure we can see that we had some steady state error together with transient speed drop during the **cut-in** maneuver.
The following changes are added to enhance the speed tracking scenario:
1) Integral and derivative controllers are added to speed tracking for laner(previously we only implemented them for trajectory tracking controller)
2) An additional feedforward term is added for speed tracking during lane change, physically, when the steering wheel changes the front lateral force has a component in the negative direction which reduces the speed, this feedforward term is added to cancel such speed drops.
3) The controller gains are also changing with vehicle speed to make higher speed (near 100 km/hr) feasible.


davidrusu (approved):
:ship:

smarts/core/controllers/lane_following_controller.py
```diff
@@ -211,9 +233,16 @@ def perform_lane_following(
             # gain is set to 4.5, the lower the value, the vehicle becomes
```
liamchzh, line 233: Does the comment here need to be updated? Your call~

liamchzh (approved):

smarts/core/controllers/lane_following_controller.py
iman512003: Thanks, sure.

PR #280, opened 2020-12-06
Envision Playback

JVillella:
envision-playback

Envision now stores the "frames" (the SMARTS step state we send over WS) for a given simulation in a "buffer" (really a linked list). When the accumulated frame data reaches the max capacity we began randomly deleting underlaying frame data. The linked list lets us quickly go to the next frame that has data.

Then per web client there is a "run loop" that can be "seeked" so users can go and forth in time. All run loops point to the same underlaying frames buffer.

### DIFF
--- a/envision/server.py
+++ b/envision/server.py
@@ -139,13 +139,16 @@ class Frames:
         """Sample random frames and clear their data to ensure we're under the max
         capacity size.
         """
+        bytes_to_mb = 1e-6
         sizes = [frame.size for frame in self._frames]
-        while sum(sizes) * 1e-6 > self._max_capacity:
+        while sum(sizes) * bytes_to_mb > self._max_capacity:
             # XXX: randint(1, ...), we skip the start frame because that is a "safe"
             #      frame we can always rely on being available.
             idx_to_delete = random.randint(1, len(self._frames) - 1)
             sizes[idx_to_delete] = 0
-            self._frames[idx_to_delete].data = None
+            self._frames[idx_to_delete - 1].next_ = self._frames[idx_to_delete].next_
+            del self._frames[idx_to_delete]
+            del self._timestamps[idx_to_delete]
 
 
 class WebClientRunLoop:
@@ -155,6 +158,7 @@ class WebClientRunLoop:
     """
 
     def __init__(self, frames, web_client_handler, timestep_sec, seek=None):
+        self._log = logging.getLogger(__class__.__name__)
         self._frames = frames
         self._client = web_client_handler
         self._timestep_sec = timestep_sec
@@ -171,7 +175,7 @@ class WebClientRunLoop:
 
     def run_forever(self):
         async def run_loop():
-            # Wait till the first frame is present
+            # If no frame, wait till one is present
             frame = self._frames.start_frame
             while frame is None:
                 await asyncio.sleep(0.1)
@@ -181,16 +185,15 @@ class WebClientRunLoop:
                 # Handle seek
                 if self._seek is not None and self._frames.start_time is not None:
                     frame = self._frames(self._frames.start_time + self._seek)
-                    frame = self._frame_with_data(frame)
                     self._seek = None
 
                 if frame is None:
-                    # TODO: This shouldn't happen, log a warning
+                    self._log.warning("Seek frame missing, reverting to start frame")
                     frame = self._frames.start_frame
 
                 closed = self._push_frame_to_web_client(frame)
                 if closed:
-                    # XXX: Log warning that the socket closed and we're exiting
+                    self._log.debug("Socket closed, exiting")
                     return
 
                 frame = await self._wait_for_next_frame(frame)
@@ -223,24 +226,10 @@ class WebClientRunLoop:
 
     async def _wait_for_next_frame(self, frame):
         while True:
+            # TODO: Consider using an asyncio queue instead
             await asyncio.sleep(self._timestep_sec)
-            next_frame = self._frame_with_data(frame.next_)
-            if next_frame is not None:
-                return next_frame
-
-    def _frame_with_data(self, frame):
-        # May loop through until it finds a frame with data
-        if frame is None:
-            return None
-
-        while frame.data is None:
-            frame = frame.next_
-
-            # No more frames left
-            if frame is None:
-                return None
-
-        return frame
+            if frame.next_ is not None:
+                return frame.next_
 
 
 class BroadcastWebSocket(tornado.websocket.WebSocketHandler):
@@ -256,7 +245,6 @@ class BroadcastWebSocket(tornado.websocket.WebSocketHandler):
     async def open(self, simulation_id):
         self._logger.debug(f"Broadcast websocket opened for simulation={simulation_id}")
         self._simulation_id = simulation_id
-        # TODO: Make capacity a command line parameter
         self._frames = Frames(max_capacity_mb=self._max_capacity_mb)
         FRAMES[simulation_id] = self._frames
         WEB_CLIENT_RUN_LOOPS[simulation_id] = set()
@@ -285,7 +273,7 @@ class StateWebSocket(tornado.websocket.WebSocketHandler):
         if simulation_id not in WEB_CLIENT_RUN_LOOPS:
             raise tornado.web.HTTPError(404)
 
-        # TODO: Set timestamp correctly...
+        # TODO: Set this appropriately
         timestep_sec = 0.1
         self._run_loop = WebClientRunLoop(
             frames=FRAMES[simulation_id],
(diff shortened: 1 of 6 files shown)


envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+            frame_idx = -1
+
+        return self._frames[frame_idx]
+
+    def _enforce_max_capacity(self):
+        """Sample random frames and clear their data to ensure we're under the max
+        capacity size.
+        """
+        sizes = [frame.size for frame in self._frames]
+        while sum(sizes) * 1e-6 > self._max_capacity:
```
Gamenot, line 143: We may want to use a named constant for the conversion to megabytes. 

I also think that if the number of frames become very large the random iteration over the whole collection is both inefficient and highly variable. It might be best to exclude the frames that already do not have size.

envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+        self._timestep_sec = timestep_sec
+        self._seek = seek
+        self._thread = None
+
+    def seek(self, offset_seconds):
+        self._seek = offset_seconds
+
+    def stop(self):
+        if self._thread:
+            self._thread.join(timeout=3)
```
Gamenot, line 169: I cannot tell where this is called. Might want to do something if this fails so the thread does not hang around at the end.

envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+
+    def _enforce_max_capacity(self):
+        """Sample random frames and clear their data to ensure we're under the max
+        capacity size.
+        """
+        sizes = [frame.size for frame in self._frames]
+        while sum(sizes) * 1e-6 > self._max_capacity:
+            # XXX: randint(1, ...), we skip the start frame because that is a "safe"
+            #      frame we can always rely on being available.
+            idx_to_delete = random.randint(1, len(self._frames) - 1)
```
Gamenot, line 146: I would also suggest not deleting the last frame.

envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+    def __call__(self, timestamp):
+        """Finds the nearest frame according to the given timestamp."""
+        frame_idx = bisect.bisect_left(self._timestamps, timestamp)
+        if frame_idx >= len(self._frames):
+            frame_idx = -1
+
+        return self._frames[frame_idx]
+
+    def _enforce_max_capacity(self):
+        """Sample random frames and clear their data to ensure we're under the max
```
Gamenot, line 139: I understand how this might work but I am not sure if I like it, random depopulation creates randomly sized holes.

Gamenot (changes requested):

envision/server.py
davidrusu: If we'd really like to use randomness here, how about we use the exponential distribution instead? with a median of `len(self._frames) / 4`

This will bias us to remove frames from the start of the list.

`idx_to_delete = min(len(self._frames) - 1, -1 * len(self._frames) / 4 * math.log(random.random())`

envision/web/src/index.js
```diff
@@ -26,7 +26,7 @@ import App from "./components/app.js";
 import "antd/dist/antd.dark.css";
 
 let client = new Client({
-  endpoint: window.location.origin,
+  endpoint: "http://localhost:8081",
```
davidrusu, line 29: why are you switching this back? did the `location.origin` not work for you?

envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+        capacity size.
+        """
+        sizes = [frame.size for frame in self._frames]
+        while sum(sizes) * 1e-6 > self._max_capacity:
+            # XXX: randint(1, ...), we skip the start frame because that is a "safe"
+            #      frame we can always rely on being available.
+            idx_to_delete = random.randint(1, len(self._frames) - 1)
+            sizes[idx_to_delete] = 0
+            self._frames[idx_to_delete].data = None
```
davidrusu, line 148: This frame list will still grow unbounded, how about we just remove the frame entirely. 
```suggestion
            if idx_to_delete > 0:
                self._frames[idx_to_delete - 1].next_ = self._frames[idx_to_delete].next_
            del self._frames[idx_to_delete]
```

envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+            return True
+
+    async def _wait_for_next_frame(self, frame):
+        while True:
+            await asyncio.sleep(self._timestep_sec)
+            next_frame = self._frame_with_data(frame.next_)
+            if next_frame is not None:
+                return next_frame
+
+    def _frame_with_data(self, frame):
```
davidrusu, line 231: After making the change above to completely remove a frame, you won't need this function any more

envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+        return self._timestamp
+
+    @property
+    def data(self):
+        return self._data
+
+    @data.setter
+    def data(self, data: str):
+        self._data = data
+        self._size = sys.getsizeof(data)
```
davidrusu, line 86: very cool

davidrusu (changes requested):
Looks great, just some questions and suggestions

envision/server.py
```diff
@@ -54,30 +64,214 @@ def options(self):
...
+
+    def __call__(self, timestamp):
+        """Finds the nearest frame according to the given timestamp."""
+        frame_idx = bisect.bisect_left(self._timestamps, timestamp)
+        if frame_idx >= len(self._frames):
+            frame_idx = -1
+
+        return self._frames[frame_idx]
+
+    def _enforce_max_capacity(self):
```
liamchzh, line 138: Should be cautious and conservative about memory management. The max capacity is applied to every simulation, and running a loop scenario could generate 500M data in less than 10 mins.

liamchzh:
And was having this error on my machine using your branch.

Screenshot from 2020-12-07 15-46-11

envision/server.py
JVillella: But fairly uniform over long time periods. There are other approaches as we discussed, each with their pros and cons. How about we revisit this later once we have a better sense of the pitfalls.

envision/web/src/index.js
JVillella: It didn't. `npm run` puts it on `8080` while the server always remains on `8081`.

envision/server.py
JVillella: Yes, unbounded; we're talking about 2 ints and a reference: 28+28+8 = 64 bytes. At 50 FPS we're looking at 3200 bytes/sec ~275MB every 24 hours (may be a bit larger in practice).

Not great.

The reason for this design was to simplify seeking via the timestamps; but let me take another look.

JVillella:
@liamchzh can you hard refresh your browser? I don't have that issue. Maybe your browser is caching.

envision/server.py
```diff
@@ -54,30 +64,202 @@ def options(self):
...
+                )
+            )
+            return False
+        except WebSocketClosedError:
+            return True
+
+    async def _wait_for_next_frame(self, frame):
+        while True:
+            # TODO: Consider using an asyncio queue instead
+            await asyncio.sleep(self._timestep_sec)
```
davidrusu, line 230: it'd be cool to let the user configure this timestep in the future!

davidrusu (approved):
Just a few suggestions :+1: once those are resolved

envision/server.py
```diff
@@ -54,30 +64,202 @@ def options(self):
...
+
+    def _enforce_max_capacity(self):
+        """Sample random frames and clear their data to ensure we're under the max
+        capacity size.
+        """
+        bytes_to_mb = 1e-6
+        sizes = [frame.size for frame in self._frames]
+        while sum(sizes) * bytes_to_mb > self._max_capacity:
+            # XXX: randint(1, ...), we skip the start frame because that is a "safe"
+            #      frame we can always rely on being available.
```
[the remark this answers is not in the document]
JVillella: We need the first frame to grab the elapsed time. I also like having a frame that I know is always safe / available (used for seeking edge case).

envision/server.py
JVillella: It needs to be derived from the run script / SMARTS.

envision/server.py
JVillella: That could be a nice change, there are pros and cons. Let's see how things are from actual use. We can come back to this.

envision/server.py
```diff
@@ -54,30 +64,201 @@ def options(self):
...
+    def _enforce_max_capacity(self):
+        """Sample random frames and clear their data to ensure we're under the max
+        capacity size.
+        """
+        bytes_to_mb = 1e-6
+        sizes = [frame.size for frame in self._frames]
+        while len(self._frames) >= 2 and sum(sizes) * bytes_to_mb > self._max_capacity:
+            # XXX: randint(1, ...), we skip the start frame because that is a "safe"
+            #      frame we can always rely on being available.
+            idx_to_delete = random.randint(1, len(self._frames) - 1)
```
Gamenot, line 146: I still have some concern that it might be best to exclude a number of the newest frames so that the most recent frames will be guaranteed to be viewable.

JVillella: Sure, doesn't hurt. I'll do that now.

Gamenot (approved):

PR #284, opened 2020-12-08
247 Authenticate

### LINKED ISSUE #247: Secure Zoo Worker Communication

Adaickalavan:
Authentication is an optional feature.  Authentication added to
+ multiprocessing.connection.Client, and
+ multiprocessing.connection.Listener.

This pull request addresses issue #247, and supersedes an earlier pull request #264. Pull request #264 is closed in favour of this pull request.
 
Closes #247

### DIFF
--- a/smarts/core/remote_agent_buffer.py
+++ b/smarts/core/remote_agent_buffer.py
@@ -32,7 +32,7 @@ from .utils.networking import find_free_port
 
 
 class RemoteAgentBuffer:
-    def __init__(self, zoo_worker_addrs=None, auth_key=b"secret", buffer_size=3):
+    def __init__(self, zoo_worker_addrs=None, auth_key=b"None", buffer_size=3):
         """
         Args:
             zoo_worker_addrs:
@@ -99,7 +99,7 @@ class RemoteAgentBuffer:
         # Block until the local zoo server is accepting connections.
         for i in range(retries):
             try:
-                conn = Client(local_address, family="AF_INET", auth_key=self._auth_key)
+                conn = Client(local_address, family="AF_INET", authkey=self._auth_key)
                 break
             except Exception as e:
                 self._log.error(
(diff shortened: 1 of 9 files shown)


smarts/core/remote_agent_buffer.py
```diff
@@ -44,6 +49,12 @@ def __init__(self, zoo_worker_addrs=None, buffer_size=3):
         self._local_zoo_worker = None
         self._local_zoo_worker_addr = None
 
+        assert isinstance(
+            auth_key, (str, type(None))
+        ), f"Received auth_key of type {type(auth_key)}, but need auth_key of type <class 'string'> or <class 'NoneType'>."
+        self._auth_key = auth_key if auth_key else ""
+        self._auth_key_conn = str.encode(auth_key) if auth_key else None
```
Gamenot, line 56: nit: This is fine but I am not convinced we need to store this variable since it is mostly a duplication of `self._auth_key` and `str.encode()` is not costly.

Gamenot (approved):

PR #294, opened 2020-12-10
Improved error message

### LINKED ISSUE #186: Error in `Smarts.__init__()` is misleading
The current error swallows all causes when there are a few other known reasons for the issue. (e.g. another known issue here is the hard requirement for a singular `Showbase` instance.)

https://github.com/huawei-noah/SMARTS/blob/d9f609b634f50346a61fe38c3c4a11e826eb2bdd/smarts/core/smarts.py#L96-L104

Adaickalavan:
This attempts to improve the error message thrown by `SMARTS` class when initializing its parent class `ShowBase`.

Closes #186 .

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -95,8 +95,8 @@ class SMARTS(ShowBase):
         except Exception as e:
             # Known reasons for this failing:
             raise Exception(
-                "SMARTS: Error in initializing parent class ShowBase."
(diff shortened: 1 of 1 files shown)


davidrusu (approved):

JVillella (approved):
👍 Looks good. Debatable if we should expose "ShowBase" as a parent class to users.

PR #298, opened 2020-12-10
Bump OpEn version 0.1.188

liamchzh:
Forgot to bump the version since I made a small change in #263


Gamenot (approved):

liamchzh:
@davidrusu The way I built the wheel is using `python setup.py bdist_wheel`. But I noticed python_bindings_open_agent_solver also has been modified, should I commit the solver?

davidrusu (approved):

davidrusu:
@liamchzh if you didn't modify the equations in the build_problem() function, then you don't need to commit the solver

PR #300, opened 2020-12-11
Internal lane snapping to fill gaps

### LINKED ISSUE #273: Fix road geometry caps and junctions
image

This is on the minicity scenario.

* We should either make the caps flat or fix the holes if we keep them round.
* Some tight junctions have holes in them (this could be because they're sep. lanes and SUMO has a gap there - i.e. not a problem for us to fix at this time)

JVillella:
closes #273 

This PR fixes these (with a hammer),

Screen Shot 2020-12-11 at 12 52 23 PM
Screen Shot 2020-12-11 at 12 52 16 PM

Now we have,

Screen Shot 2020-12-11 at 12 52 16 PM

### DIFF
--- a/smarts/core/sumo_road_network.py
+++ b/smarts/core/sumo_road_network.py
@@ -30,7 +30,7 @@ import trimesh.scene
 from shapely import ops
 from shapely.geometry import LineString, MultiPolygon, Polygon
 from shapely.geometry.base import CAP_STYLE, JOIN_STYLE
-from shapely.ops import triangulate
+from shapely.ops import triangulate, snap
 from trimesh.exchange import gltf
 
 from .utils.math import rotate_around_point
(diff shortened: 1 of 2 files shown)


aurorahcx (approved):

Gamenot (approved):

smarts/core/sumo_road_network.py
```diff
@@ -121,6 +125,30 @@ def _compute_road_polygons(self, scale):
 
         return polys
 
+    def _snap_internal_edges(self, lane_to_poly, snap_threshold=2):
+        # HACK: Internal edges that have tight curves, when buffered their ends do not
```
Gamenot, line 129: I am not sure this is a hack since the underlying data is not conducive to fitted segments.

JVillella: It in fact is, I can explain more this evening.

PR #302, opened 2020-12-11
What would life be like without supervisors?

JVillella:
```bash
scl experiments run examples/egoless.py scenarios/loop
```

Do you like this life?

🤷 

EDIT: Folks may be able to get behind,

```bash
scl run examples/egoless.py scenarios/loop
```


JVillella:
(Note, this is not actually the experiments workshop, I initially thought the experiments workshop being the single interface into SMARTS could handle spinning up Envision. However, thinking about this more,

* Envision still needs to be runnable outside SMARTS
* It would be convenient to have one command to run both SMARTS and Envision
* Perhaps we have an scl smarts run <script> <args>, experiment workshop will still be involved at the script and args level most likely
)

Issue #272, opened 2020-12-03
Fix the oscillation of OpEn agent 


liamchzh:
We can see from the driven path, the OpEn trajectory sometimes becomes a little bit curly even on a straight lane. 

here are 2 major problems:
1. lane keeping: OpEn is not as smooth as other agents like `laner`
2. lane change: oscillation of the cut-in trajectory
3. the speed is way too lower than the speed limit of the road (should be able to drive at around 50 km/h).

Potential improvements:

1. ~~use a smaller timestep~~ (@liamchzh need to clarify. Reply: not an ultimate solution, should be working well with different timestep)
2. @iman512003 please fill this out.
3. Interpolate the cut-in waypoints. 

In the following recording, the OpEn is using default gain config.

![ezgif com-gif-maker (10)](https://user-images.githubusercontent.com/2217445/101188301-851da580-3623-11eb-9727-019438eeb5d2.gif)

PR #309, opened 2020-12-15
Fix oscillation open

iman512003:
This is to address #272 .


liamchzh (approved):

Gamenot (approved):
I did see a case where the vehicle got stuck trying to follow a turn on the `4lane_t` (it is hard to reproduce because that map is not deterministic) but overall seems better in terms of oscillation.

PR #315, opened 2020-12-15
Mix Proximity and Heading in Waypoint Selection

Gamenot:


smarts/core/waypoints.py
```diff
@@ -307,7 +307,8 @@ def _closest_linked_wp_in_kd_tree_with_pose_batched(
         return [
             sorted(
                 l_wps,
-                key=lambda _lwp: abs(poses[idx].heading.relative_to(_lwp.wp.heading)),
```
Gamenot, line 310: Sorting by heading was not enough here since when you have multiple lanes on an edge the correct waypoint becomes ambiguous.

The posed solution here mixes a quadratic distance growth and the difference in heading. Meaning distant points will get sorted to the back and heading differences will have strong bearing on close points.

liamchzh (approved):

iman512003 (approved):

PR #326, opened 2020-12-17
Fix `replace()` use in AgentSpec

Gamenot:


smarts/core/agent.py
```diff
@@ -146,7 +146,27 @@ def __post_init__(self):
     def replace(self, **kwargs) -> "AgentSpec":
         """Return a copy of this AgentSpec with the given fields updated."""
 
-        return replace(self, **kwargs)
+        replacements = [
+            ("policy_builder", "agent_builder"),
+            ("policy_params", "agent_params"),
+            ("perform_self_test", None),
```
aurorahcx, line 152: Would `None` be informative enough for the warning message below?

Gamenot: I used `None` because it should not be in the `kwargs`. I can fix this I think by asserting that `None` is not in the kwargs.

aurorahcx: Maybe I wasn't being clear due to lack of context, what should the users do if they use "perform_self_test"? Should they get a deprecated warning that instructs them not to use this, because there's no replacement (`None`)?

Gamenot: I see what you mean now, I have addressed it.

aurorahcx (approved):

smarts/core/agent.py
```diff
@@ -146,7 +146,34 @@ def __post_init__(self):
...
+        ]
+
+        assert None not in kwargs, f"Error: replace contains invalid key: ({None})"
```
Adaickalavan, line 155: I guess `f" "` is not needed as there is no string formatting. Alternatively, we can improve the error message as:

```suggestion
        assert None not in kwargs, f"Error: kwargs input to replace() function contains invalid key `None`: {kwargs}"

```

Gamenot: This seems like a reasonable change.

Adaickalavan (approved):

PR #329, opened 2020-12-17
328 fix `Heading.relative_to()` assert on greater than 3 pi difference

Gamenot:


smarts/core/coordinates.py
```diff
@@ -46,13 +48,28 @@ def as_lwh(self):
...
+
+    def __new__(self, x: Union[SupportsFloat, SupportsIndex] = ...) -> "Heading":
+        """A override to constrain heading to -pi to pi"""
+        value = x
+        if isinstance(value, (int, float)):
+            value = value % (2 * math.pi)
+            if value > math.pi:
+                value -= 2 * math.pi
+            if value < -math.pi:
+                value += 2 * math.pi
```
aurorahcx, line 62: I think `%` only returns non-negative value, so perhaps the check for `value < -math.pi` is not needed.

Gamenot: You are right, I intended `math.fmod()` here.

aurorahcx (approved):

Adaickalavan (approved):

PR #236, opened 2020-11-30
Heterogenous Agent Computing

davidrusu:
This PR contains a first pass at Heterogeneous Agent Computing. 

The main idea is that we have the Zoo Worker installed on a server and listens for SMARTS sessions requesting remote social agents.

```bash
# on a server
scl zoo worker

# locally

gym.make(..., zoo_workers=[("143.125.34.1", 1200)])
```

Sequence diagram for SMARTS allocating a remote social agent
zoo-compute-sequence-diagram

PR #342, opened 2020-12-20
Add user-defined simulation names and display them in Envision

aurorahcx:
image


envision/web/src/index.js
```diff
@@ -29,7 +29,7 @@ import { ToastProvider } from "react-toast-notifications";
 import Snack from "./components/snack";
 
 let client = new Client({
-  endpoint: window.location.origin,
+  endpoint: "http://localhost:8081",
```
aurorahcx, line 32: This was changed in https://github.com/huawei-noah/SMARTS/commit/eeea1d6c6ba4a93e70e110385a5cfb12bead4bea

But I am getting the following error when I run `npm start`: 
`Uncaught (in promise) SyntaxError: Unexpected token < in JSON at position 0` with `window.location.origin=http://localhost:8080`

So I reverted back the change.

envision/client.py
```diff
@@ -73,9 +75,18 @@ def __init__(
         num_retries: int = 5,
         wait_between_retries: float = 0.5,
         output_dir: str = None,
+        client_id: str = None,
```
liamchzh, line 78: How about renaming the `client_id`, because now it has a different meaning.

envision/client.py
```diff
@@ -73,9 +75,18 @@ def __init__(
...
     ):
         self._log = logging.getLogger(self.__class__.__name__)
-        client_id = str(uuid.uuid4())[:8]
+
+        current_time = datetime.now().strftime("%Y%m%d%H%M%S%f")[:-4]
```
liamchzh, line 82: The readability of a timestamp like `2020122110132406` is really bad. Perhaps just `%H:%M:%S` is enough.

envision/client.py
```diff
@@ -73,9 +75,18 @@ def __init__(
...
+        if client_id:
+            # String length limit for display is 20 characters
+            # Replace all special (non-alphanumeric) characters to "_" to avoid invalid key values
+            client_id = re.sub("\W+", "_", client_id[:20]) + "_"
+            client_id += current_time
```
liamchzh, line 87: Can we use a string formating here? `client_id = f"{client_id}_{current_time}"`. Then you get rid of the `_` suffix and string concatenation. Need to replace the `client_id` if you rename the var.

envision/web/src/components/header.js
```diff
@@ -74,13 +74,13 @@ export default function Header_({
...
           value={selectValue}
-          style={{ width: 200 }}
+          style={{ width: 400 }}
           onChange={(value) => onSelectSimulation(value)}
         >
           <Option value="all">All Simulations</Option>
           <OptGroup label="Simulations">
             {simulationIds.map((id) => (
-              <Option key={id} value={id}>{`Simulation ${id}`}</Option>
+              <Option key={id} value={id}>{`Sim ${id}`}</Option>
```
liamchzh, line 83: Probably need to also rename the `id` here?

envision/web/src/index.js
liamchzh: This is where `window.location.origin` was introduced https://github.com/huawei-noah/SMARTS/pull/236/files#r533447521

aurorahcx: When you made some changes and re-compiled the distribution using `npm run build`, does the new build include your changes? For me it still does not with `window.location.origin`, even when I use the `scl` command

liamchzh: The new build did reflect my changes. Maybe something went wrong during `npm run build`.

aurorahcx: Ok I cleared all browser cache and it works now, sorry for the confusion.

liamchzh (approved):

examples/ego_open_agent.py
```diff
@@ -20,12 +20,13 @@
 AGENT_ID = "Agent-007"
 
 
-def main(scenarios, headless, num_episodes, seed):
+def main(scenarios, sim_name, headless, num_episodes, seed):
```
Gamenot, line 23: nit:`experiment_name` might be more correct here.

envision/web/src/components/header.js
```diff
@@ -60,12 +60,27 @@ export default function Header_({
     setEgoView(egoView_);
   }
 
+  function displaySimName(id) {
```
Gamenot, line 63: If you are worried about user supplied string, since the input comes from a user, it might be best to sanitise the input here.

Gamenot (approved):

envision/web/src/components/header.js
aurorahcx: I am not sure a super long string can still be a valid key value on the client side?

examples/ego_open_agent.py
aurorahcx: I saw a lot of places that use the word "simulation" in our code, is it supposed to be different from "experiment"?

envision/web/src/components/header.js
Gamenot: Alright!

PR #355, opened 2020-12-24
WIP: Env score

adreena:
Adding a module to highway environment to calculate environment score. Ultra can override this module to assign the reward_adapter as some evaluation score to Observations objects.

Main files to look at:
- smarts/core/sensors.py
- smarts/env/hiway_env.py
- smarts/ultra/env/ultra_env.py

@junluo-huawei @Gamenot we don't need to merge the changes, it's just a suggestion


adreena:
@Gamenot we can do the same for info_adapter, adding a tag for logs and populating it based on overriding a module, I'm adding that suggestion to this branch

smarts/env/hiway_env.py
```diff
@@ -170,10 +175,14 @@ def step(self, agent_actions):
             reward = rewards[agent_id]
             info = infos[agent_id]
 
+            raw_reward = reward
+            raw_observation = copy.deepcopy(observation)
             rewards[agent_id] = agent_spec.reward_adapter(observation, reward)
             observations[agent_id] = agent_spec.observation_adapter(observation)
+            observations[agent_id]['env_score'] = self.assign_env_score(raw_observation, raw_reward)
```
Gamenot, line 182: This seems reasonable.

PR #361, opened 2020-12-29
Upgrade ray version

Gamenot:


setup.py
```diff
@@ -46,7 +46,7 @@
         #       and ray[rllib] under [train] it  prevents rllib from getting installed.
         #       For simplicity we just install both here. In the future we may want to
         #       address this bug head on to keep our SMARTS base install more lean.
-        "ray[rllib]==0.8.6",  # We use Ray for our multiprocessing needs
+        "ray[rllib]==1.0.1.post1",  # We use Ray for our multiprocessing needs
```
liamchzh, line 49: 1.1.0 is published. Shall we use this instead?

Gamenot: 1.1.0 seems to be a bit new, (Dec 24th). It might be reasonable given another week.

liamchzh: Sure. `1.1.0` also might have breaking changes.

liamchzh (approved):

setup.py
Gamenot: While that is true, I am more worried about how well it has been tested.

tools/regression_rllib.py
```diff
@@ -109,6 +109,7 @@ def run_experiment(log_path, experiment_name, training_iteration=100):
     analysis = tune.run(
         "PPO",
         name=experiment_name,
+        mode="max",
```
Adaickalavan, line 112: Given no `metric` attribute in `tune.run()`, what is `mode=max` maximizing here?

Gamenot: This was being slightly aggressive with adding default `mode` in case of analysis being used since I was seeing related assertions.

Adaickalavan (approved):

PR #363, opened 2020-12-30
Fix Envision to automatically load the simulation being selected

aurorahcx:
Fix Envision to automatically load the simulation being selected:
Add a dictionary to map a web socket for each simulation

Before:
dxFM0tPqKE

After:
X6QFOh911w

### DIFF
--- a/envision/client.py
+++ b/envision/client.py
@@ -84,7 +84,7 @@ class Client:
         if sim_name:
             # String length limit for display is 20 characters
             # Replace all special (non-alphanumeric) characters to "_" to avoid invalid key values
-            sim_name = re.sub("\W+", "_", sim_name[:20])
+            sim_name = re.sub(r"\W+", "_", sim_name[:20])
             client_id = f"{sim_name}_{client_id}"
 
         if endpoint is None:
--- a/envision/client.py
+++ b/envision/client.py
@@ -82,7 +82,7 @@ class Client:
         current_time = datetime.now().strftime("%Y%m%d%H%M%S%f")[:-4]
         client_id = current_time
         if sim_name:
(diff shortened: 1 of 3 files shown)


envision/client.py
```diff
@@ -84,7 +84,7 @@ def __init__(
         if sim_name:
             # String length limit for display is 20 characters
             # Replace all special (non-alphanumeric) characters to "_" to avoid invalid key values
-            sim_name = re.sub("\W+", "_", sim_name[:20])
+            sim_name = re.sub(r"\W+", "_", sim_name[:20])
```
aurorahcx, line 87: "Note that any invalid escape sequences in Python’s usage of the backslash in string literals now generate a `DeprecationWarning` and in the future this will become a `SyntaxError`. The solution is to use Python’s raw string notation for regular expression patterns."

https://docs.python.org/3/library/re.html

envision/web/src/client.js
```diff
@@ -31,8 +31,10 @@ export default class Client {
     this._maxRetries = retries;
     this._glb_cache = {};
 
-    this._socket = null;
-    this._flushStream = false;
+    this._socket = {};
```
Gamenot, line 34: Since this is now a set, perhaps `_sockets`?

envision/web/src/client.js
```diff
@@ -31,8 +31,10 @@ export default class Client {
...
+    this._flushStream = {};
```
Gamenot, line 35: Plural same as above.

Gamenot (changes requested):
Some small comments but looks good.

liamchzh (approved):

Gamenot (approved):
LGTM

PR #364, opened 2020-12-30
Fix Envision multi-instance display

aurorahcx:
Fix Envision multi-instance display:
Register gltf loader locally under different names for different scenarios

Before:
3S9A02NfjD

After:
k0QNmjP0dI


Gamenot (approved):
Looks good to me.

liamchzh (approved):

PR #248, opened 2020-12-01
Minicity Scenario

JVillella:
closes #267 
(makes progress on #268)

minicity-1

This will take some iterations to make it nice. But this is the first big chunk of bringing in the Toronto Younge/Bloor map, creating the scenario, and some fixes along the way.

PR #374, opened 2021-01-01
Fix bubble geometry built from MapZone

aurorahcx:
- Changed the logic to select splited shapes in MapZone
- Changed the definition of max cut in MapZone

I am not sure if this is a good solution for this problem because I did it in a rush. It mostly serves as a starting point. It is worth of discussion and double checking. Feel free to make adjustment on top of this.

I am inspired by this PR: https://github.com/huawei-noah/SMARTS/pull/248/files

The dark-red color of the bubbles shown below is only for illustration purpose.

**4lane_t** (The bubble on the left is built with MapZone)
Screenshot from 2021-01-01 10-55-46

**minicity** (note that some bubbles have longer length with changes to `max_cut` in this PR)
Screenshot from 2021-01-01 10-23-10
Screenshot from 2021-01-01 10-34-48
Screenshot from 2021-01-01 10-47-42

### DIFF
--- a/smarts/sstudio/types.py
+++ b/smarts/sstudio/types.py
@@ -527,8 +527,8 @@ class MapZone(Zone):
             if len(lane_shape) == 1:
                 return lane_shape[0]
 
+            # We assume that there are only two splited shapes to choose from
             keep_index = 0
-
             if lane_shape[1].minimum_rotated_rectangle.contains(
                 Point(expected_midpoint)
             ):
--- a/smarts/sstudio/types.py
+++ b/smarts/sstudio/types.py
@@ -513,9 +513,7 @@ class MapZone(Zone):
             else:
                 return float(offset)
 
-        def pick_remaining_shape_after_split(
-            geometry_collection, expected_midpoint, lane
-        ):
+        def pick_remaining_shape_after_split(geometry_collection, expected_point, lane):
             lane_shape = geometry_collection
             if not isinstance(lane_shape, GeometryCollection):
                 return lane_shape
@@ -529,9 +527,7 @@ class MapZone(Zone):
 
             # We assume that there are only two splited shapes to choose from
             keep_index = 0
-            if lane_shape[1].minimum_rotated_rectangle.contains(
-                Point(expected_midpoint)
-            ):
+            if lane_shape[1].minimum_rotated_rectangle.contains(expected_point):
                 # 0 is the discard piece, keep the other
                 keep_index = 1
 
@@ -570,8 +566,10 @@ class MapZone(Zone):
             # Second cut takes into account shortening of geometry by `min_cut`.
             max_cut = min(min_cut + geom_length, lane_length)
 
-            midpoint = road_network.world_coord_from_offset(
-                lane, lane_offset + geom_length * 0.5
+            midpoint = Point(
+                road_network.world_coord_from_offset(
+                    lane, lane_offset + geom_length * 0.5
+                )
             )
 
             lane_shape = road_network.split_lane_shape_at_offset(
--- a/smarts/sstudio/types.py
+++ b/smarts/sstudio/types.py
@@ -23,7 +23,7 @@ import collections.abc as collections_abc
 from dataclasses import dataclass, field
 from typing import Any, Dict, Optional, Sequence, Tuple, Union
 
-from shapely.geometry import GeometryCollection, MultiPolygon, Polygon
+from shapely.geometry import GeometryCollection, MultiPolygon, Polygon, Point
 from shapely.ops import unary_union
 
 from smarts.core import gen_id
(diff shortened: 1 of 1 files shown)


smarts/sstudio/types.py
```diff
@@ -513,29 +513,6 @@ def resolve_offset(offset, geometry_length, lane_length):
...
-            if len(lane_shape) not in {1, 2}:
-                return None
-
-            if len(lane_shape) == 1:
-                return lane_shape[0]
-
-            expected_bbox_area = lane.getWidth() * length
-            keep_index = 0
-            if (lane_shape[0].minimum_rotated_rectangle.area - expected_bbox_area) < (
-                lane_shape[1].minimum_rotated_rectangle.area - expected_bbox_area
```
aurorahcx, line 531: I didn't really understand how this area comparison works...but it is not selecting the correct index in 4lane_t.

smarts/sstudio/types.py
```diff
@@ -565,21 +542,27 @@ def pick_remaining_shape_after_split(geometry_collection, length, lane):
 
             min_cut = min(lane_offset, lane_length)
             # Second cut takes into account shortening of geometry by `min_cut`.
-            max_cut = min(min_cut + geom_length, lane_length - min_cut)
+            max_cut = min(min_cut + geom_length, lane_length)
```
aurorahcx, line 545: I assume that max_cut should always be greater than min_cut, `lane_length - min_cu`t would make max_cut less than min_cut.

smarts/sstudio/types.py
```diff
@@ -565,21 +542,27 @@ def pick_remaining_shape_after_split(geometry_collection, length, lane):
...
                 lane_shape, lane, min_cut
             )
-            lane_shape = pick_remaining_shape_after_split(lane_shape, min_cut, lane)
-            if lane_shape is None:
-                continue
+            if isinstance(lane_shape, GeometryCollection):
+                if len(lane_shape) not in {1, 2}:
+                    continue
+                else:
+                    lane_shape = lane_shape[0]
```
aurorahcx, line 554: lane_shape[0] contains the right half of the splitted shape, lane_shape[1] contains the left half of the splitted.
Here is selecting the right half splitted by min_cut.

smarts/sstudio/types.py
```diff
@@ -565,21 +542,27 @@ def pick_remaining_shape_after_split(geometry_collection, length, lane):
...
-            lane_shape = pick_remaining_shape_after_split(lane_shape, max_cut, lane)
-            if lane_shape is None:
-                continue
+            if isinstance(lane_shape, GeometryCollection):
+                if len(lane_shape) not in {1, 2}:
+                    continue
+                elif len(lane_shape) == 1:
+                    lane_shape = lane_shape[0]
+                else:
+                    lane_shape = lane_shape[1]
```
aurorahcx, line 565: Here is selecting the left half splitted by max_cut.

smarts/sstudio/types.py
Gamenot: Yes this was guessing that the larger of the two geometries was most correct and seems to have resulted in the wrong behaviour. The change might work but it will need to be tested with a box cut in another direction.

Gamenot:
Thanks @aurorahcx, thank you for bringing up discussion of this issue.

smarts/sstudio/types.py
```diff
@@ -525,13 +525,12 @@ def pick_remaining_shape_after_split(geometry_collection, length, lane):
             if len(lane_shape) == 1:
                 return lane_shape[0]
 
-            expected_bbox_area = lane.getWidth() * length
+            # We assume that there are only two splited shapes to choose from
             keep_index = 0
-            if (lane_shape[0].minimum_rotated_rectangle.area - expected_bbox_area) < (
-                lane_shape[1].minimum_rotated_rectangle.area - expected_bbox_area
-            ):
+            if lane_shape[1].minimum_rotated_rectangle.contains(expected_point):
```
Gamenot, line 530: The core of this change is to use the midpoint inside the cut geometry to check if the second cut is viable.

smarts/sstudio/types.py
Gamenot: I think this means we need to have a test to make sure that extremely short lanes and extremely long lane geometry do not cause errors.

It is correct to limit the `max_cut` to the `lane_length`, if we want to push back off the end of the lane it should be calculated in the `min_cut`.

Gamenot (approved):
In my testing it works as intended.

smarts/sstudio/types.py
liamchzh: Now it makes more sense. I think Aurora is right, the max_cut should less than the length.

liamchzh (approved):

PR #380, opened 2021-01-05
Ultra

adreena:
I'm adding ultra changes under smarts/ultra


smarts/ultra/worker.py
```diff
@@ -0,0 +1,109 @@
+import logging
```
Gamenot, line 1: What is the purpose of this file?

adreena: @Gamenot removed them

AlexLewandowski (commented):
there are still references to smarts.ultra.registry in train.py and evaluate.py

Gamenot:
@AlexLewandowski , thank you for spotting that.

Issue #347, opened 2020-12-23
Aggressiveness adjustments


liamchzh:
Cut In: 
The failing reason is that the agent is not able to catch up within a short period when it starts 50m behind the social vehicle.
- [x] Increase the speed limit (The further, the faster of the speed)
- [x] Trigger distance should consider the social vehicle's speed.

U Turn:
- [x] Increase the safety distance
- [x] Wait a bit longer to let the social vehicle pass (Agent will collide into social vehicle if they are at the adjacent lanes)

PR #383, opened 2021-01-06
Fix uturn aggressiveness

iman512003:
This is to address #347. The sensor test for uturn is failing.

### DIFF
--- a/smarts/core/mission_planner.py
+++ b/smarts/core/mission_planner.py
@@ -329,6 +329,7 @@ class MissionPlanner:
         lane_id_list = []
         for idx in oncoming_lanes:
             lane_id_list.append(idx.getID())
+
         if n_lane.getID() not in lane_id_list:
             return []
         # The aggressiveness is mapped from [0,10] to [0,0.8] domain which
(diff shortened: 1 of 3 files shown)


liamchzh (approved):

smarts/core/mission_planner.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         self, waypoints: Waypoints, road_network: SumoRoadNetwork, agent_behavior=None
     ):
         self._waypoints = waypoints
-        self._agent_behavior = agent_behavior or AgentBehavior(aggressiveness=5)
+        self._agent_behavior = agent_behavior
```
Gamenot, line 48: This was a default in the case that someone applied a task in a scenario but did not match it with a specified aggressiveness. If there is not going to be a default here we may need an assertion when translating the agent spec and mission into an agent.

Gamenot (approved):

smarts/core/mission_planner.py
iman512003: Thanks, I added it back.

Gamenot:
Looks ready to merge.

Issue #226, opened 2020-11-26
Support multiple traffic history data in one scenario


zbzhu99:
When generating scenarios with history replay vehicles, we specify the traffic history files to be used via the `traffic_histories` parameter. This parameter currently supports passing in a list containing file names, but actually, only the first file in that list will be used.

The use of multiple historical data files (replayed one by one) in one scenario should be supported.

Example scenario and related traffic history files:

[interaction_dataset_merging.zip](https://github.com/huawei-noah/SMARTS/files/5603745/interaction_dataset_merging.zip)

liamchzh:
* https://github.com/huawei-noah/SMARTS/blob/master/smarts/core/scenario.py#L529
* https://gitlab.smartsai.xyz/smarts/SMARTS/-/merge_requests/758/diffs
* https://gitlab.smartsai.xyz/smarts/imitation-learning/-/tree/master/scenarios

PR #390, opened 2021-01-07
Add support for multiple traffic history files (#226)

JingfeiPeng:
* Scenarios now will use all traffic history files in traffic_histories

* Added one test case for checking traffic histories

### DIFF
--- a/smarts/core/scenario.py
+++ b/smarts/core/scenario.py
@@ -181,6 +181,7 @@ class Scenario:
         social_agents: Dict[str, SocialAgent] = None,
         log_dir: str = None,
         surface_patches: list = None,
+        traffic_history: dict = None,
     ):
 
         self._logger = logging.getLogger(self.__class__.__name__)
@@ -197,6 +198,7 @@ class Scenario:
         self._net_file_hash = file_md5_hash(self.net_filepath)
         self._waypoints = Waypoints(self._road_network, spacing=1.0)
         self._scenario_hash = path2hash(str(Path(self.root_filepath).resolve()))
+        self._traffic_history = traffic_history or {}
 
     def __repr__(self):
         return f"""Scenario(
@@ -263,24 +265,31 @@ class Scenario:
             routes = Scenario.discover_routes(scenario_root) or [None]
             agent_missions = agent_missions or [None]
             social_agents = social_agents or [None]
+            traffic_histories = Scenario.discover_traffic_histories(scenario_root) or [
+                None
+            ]
 
             roll_routes = 0
             roll_agent_missions = 0
             roll_social_agents = 0
+            roll_traffic_histories = 0
 
             if shuffle_scenarios:
                 roll_routes = random.randint(0, len(routes))
                 roll_agent_missions = random.randint(0, len(agent_missions))
                 roll_social_agents = random.randint(0, len(social_agents))
+                roll_traffic_histories = random.randint(0, len(traffic_histories))
 
             for (
                 concrete_route,
                 concrete_agent_missions,
                 concrete_social_agents,
+                conrete_traffic_history,
             ) in product(
                 np.roll(routes, roll_routes, 0),
                 np.roll(agent_missions, roll_agent_missions, 0),
                 np.roll(social_agents, roll_social_agents, 0),
+                np.roll(traffic_histories, roll_traffic_histories, 0),
             ):
                 concrete_social_agent_missions = {
                     agent_id: mission
@@ -306,6 +315,7 @@ class Scenario:
                     },
                     social_agents=concrete_social_agents,
                     surface_patches=surface_patches,
+                    traffic_history=conrete_traffic_history,
                 )
 
     @staticmethod
@@ -505,13 +515,9 @@ class Scenario:
         self._missions.update(ego_mission)
 
     def discover_missions_of_traffic_histories(self):
-        current_traffic_history = self.discover_traffic_histories()[0] or {}
-
         vehicle_missions = {}
         # sort by timestamp
-        sorted_history = sorted(
-            current_traffic_history.items(), key=lambda d: float(d[0])
-        )
+        sorted_history = sorted(self.traffic_history.items(), key=lambda d: float(d[0]))
         for t, vehicle_states in sorted_history:
             for vehicle_id in vehicle_states:
                 if vehicle_id not in vehicle_missions:
@@ -526,8 +532,9 @@ class Scenario:
 
         return vehicle_missions
 
-    def discover_traffic_histories(self):
-        path = os.path.join(self._root, "traffic_histories.pkl")
+    @staticmethod
+    def discover_traffic_histories(scenario_root):
+        path = os.path.join(scenario_root, "traffic_histories.pkl")
         if not os.path.exists(path):
             return []
 
@@ -535,7 +542,7 @@ class Scenario:
         with open(path, "rb") as f:
             files = pickle.load(f)
             for file_name in files:
-                with open(os.path.join(self._root, file_name), "r") as history_file:
+                with open(os.path.join(scenario_root, file_name), "r") as history_file:
                     traffic_histories.append(json.loads(history_file.read()))
 
         return traffic_histories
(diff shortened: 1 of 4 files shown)


smarts/core/scenario.py
```diff
@@ -181,6 +181,7 @@ def __init__(
         social_agents: Dict[str, SocialAgent] = None,
         log_dir: str = None,
         surface_patches: list = None,
+        traffic_history: dict = {},
```
liamchzh, line 184: In Python, probably not good practice to use `{}` or `[]` as default value. Using `None` is better in this case.

smarts/core/tests/test_mutiple_traffic_data.py
```diff
@@ -0,0 +1,75 @@
...
+        [str(create_scenario)], [AGENT_ID]
+    )
+    scenarios = list(iterator)
+
+    assert len(scenarios) == 8  # 2 social agents x 2 missions x 2 histories
+
+    traffic_history_provider = TrafficHistoryProvider()
+
+    for scenario in scenarios:
+        assert scenario.traffic_history  # all scenarios have history
```
liamchzh, line 72: Maybe also check traffic_history is exactly what we need to put in the scenario instance? Just to void assigning the same history data to the scenario.

smarts/core/traffic_history_provider.py
```diff
@@ -21,7 +21,7 @@
 from typing import Set
 
 from .controllers import ActionSpaceType
-from .coordinates import Pose, Heading
+from .coordinates import Pose, Heading, BoundingBox
```
liamchzh, line 24: Thanks for putting this back. I mistakenly removed it a long time ago.

smarts/core/tests/test_mutiple_traffic_data.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+import shutil
+import tempfile
+
+from pathlib import Path
+from unittest.mock import MagicMock
```
liamchzh, line 6: Add a new blank line below line 6 to separate the imports here.

smarts/core/tests/test_mutiple_traffic_data.py
```diff
@@ -0,0 +1,75 @@
...
+        gen_social_agent_missions(
+            scenario, social_agent_actor=actors, name=name, missions=missions,
+        )
+
+    return scenario
+
+
+def test_mutiple_traffic_data(create_scenario):
+    Scenario.discover_traffic_histories = MagicMock(
+        return_value=[{"1": "history"}, {"2": "history2"}]
```
liamchzh, line 60: Although this is a mock, try using real mock data if possible, like `{"0.1": {"1": {"vehicle_id": "1", "vehicle_type": "car", ...}`.

liamchzh (commented):
Mostly Looks good, just a few comments. :+1:

smarts/core/scenario.py
```diff
@@ -789,6 +798,13 @@ def _validate_assets_exist(self):
 
         os.makedirs(self._log_dir, exist_ok=True)
 
+    def set_traffic_history(self, traffic_history):
+        self._traffic_history = traffic_history
+
+    @property
+    def traffic_history(self):
+        return self._traffic_history
```
Gamenot, line 806: Another option

```suggestion
    @property
    def traffic_history(self):
        return self._traffic_history
        
     @traffic_history.setter
     def traffic_history(self, traffic_history):
        self._traffic_history = traffic_history
```

smarts/core/tests/test_mutiple_traffic_data.py
```diff
@@ -0,0 +1,75 @@
...
+def create_scenario(scenario_parent_path):
+    scenario = Path(scenario_parent_path) / "cycles"
+    scenario.mkdir()
+
+    shutil.copyfile(
+        Path(__file__).parent / "maps/6lane.net.xml", scenario / "map.net.xml"
+    )
+    generate_glb_from_sumo_network(
+        str(scenario / "map.net.xml"), str(scenario / "map.glb")
+    )
```
Gamenot, line 32: There is a utility in `smarts.core.tests.helpers` that does this. You can look at: https://github.com/huawei-noah/SMARTS/blob/99b689a7e01d891242b13b32de7f871753c7b8df/smarts/core/tests/test_boids.py#L32-L52 

`yield` in pytest fixtures means the remaining after the yield will be run when the test ends for purposes of cleanup.

Gamenot (changes requested):
Some comments

smarts/core/tests/test_mutiple_traffic_data.py
JingfeiPeng: Thanks! I initially copied the above code from `smarts/core/tests/test_scenario.py` . Should I apply the same there?

Also, I noticed `make test` is not picking up `test_scenario.py` as one of the test files, and `pytest smarts/core/tests/test_scenario.py` also doesn't run the test, which seems strange

Gamenot: > Thanks! I initially copied the above code from `smarts/core/tests/test_scenario.py` . Should I apply the same there?  

Yes, if that is not a problem.
> Also, I noticed `make test` is not picking up `test_scenario.py` as one of the test files, and `pytest smarts/core/tests/test_scenario.py` also doesn't run the test, which seems strange  

Thanks for noticing could you change the following line to `def test_scenario_variations_of_social_agents(scenario_root):`

https://github.com/huawei-noah/SMARTS/blob/99b689a7e01d891242b13b32de7f871753c7b8df/smarts/core/tests/test_scenario.py#L70

smarts/core/scenario.py
```diff
@@ -505,7 +515,9 @@ def set_ego_missions(self, ego_mission):
         self._missions.update(ego_mission)
 
     def discover_missions_of_traffic_histories(self):
-        current_traffic_history = self.discover_traffic_histories()[0] or {}
+        current_traffic_history = (
+            Scenario.discover_traffic_histories(self.root_filepath)[0] or {}
```
JingfeiPeng, line 519: Note: still need to adjust this to use all traffic_histories

smarts/core/scenario.py
```diff
@@ -263,24 +265,31 @@ def variations_for_all_scenario_roots(
             routes = Scenario.discover_routes(scenario_root) or [None]
             agent_missions = agent_missions or [None]
             social_agents = social_agents or [None]
+            traffic_histories = Scenario.discover_traffic_histories(scenario_root) or [
```
liamchzh, line 268: I am considering getting rid of the `[{}]` in the `discover_traffic_histories() or [{}]`, i.e., we don't handle the edge case(`discover_traffic_histories` returns empty list `[]` ) here, maybe `traffic_histories = Scenario.discover_traffic_histories(scenario_root) or [None]` is better since it's aligned with the code above, and let the history provider to handle the case when history is `None`.

smarts/core/scenario.py
```diff
@@ -505,13 +515,9 @@ def set_ego_missions(self, ego_mission):
...
 
     def discover_missions_of_traffic_histories(self):
-        current_traffic_history = self.discover_traffic_histories()[0] or {}
-
         vehicle_missions = {}
         # sort by timestamp
-        sorted_history = sorted(
-            current_traffic_history.items(), key=lambda d: float(d[0])
-        )
+        sorted_history = sorted(self.traffic_history.items(), key=lambda d: float(d[0]))
```
liamchzh, line 520: `traffic_history` might become a `None` if apply the above comment.

JingfeiPeng: Technically I had set `self._traffic_history = traffic_history or {}` in the initialization of Scenario. So applying the comment above would not make traffic_history become `None`. `xxx = xxx or {}` was also the convention in the initialization of Scenario

Gamenot:
I will review this after the meeting

smarts/core/scenario.py
```diff
@@ -526,16 +532,17 @@ def discover_missions_of_traffic_histories(self):
 
         return vehicle_missions
 
-    def discover_traffic_histories(self):
-        path = os.path.join(self._root, "traffic_histories.pkl")
+    @staticmethod
+    def discover_traffic_histories(scenario_root):
```
Gamenot, line 536: My only last thought around this is that the traffic histories could be very large but that might be an issue for another time.

Gamenot (approved):

liamchzh (approved):
:ship:

Issue #54, opened 2020-10-22
Crash from `done` during `HiWayEnv.reset()`


Gamenot:
Done occuring during an environment reset causes AGENT_ID to not be in the observations after the next step.

```
import logging

import gym

from smarts.core.agent_interface import AgentInterface, AgentType
from smarts.core.agent import AgentSpec, Agent
from smarts.core.sensors import Observation

from examples import default_argument_parser


logging.basicConfig(level=logging.INFO)

AGENT_ID = "Agent-007"


class KeepLaneAgent(Agent):
    def act(self, obs: Observation):
        return "keep_lane"


def main(scenarios, max_episode_steps=None):
    agent_spec = AgentSpec(
        interface=AgentInterface.from_type(
            AgentType.Laner, max_episode_steps=max_episode_steps
        ),
        agent_builder=KeepLaneAgent,
    )

    env = gym.make(
        "smarts.env:hiway-v0",
        scenarios=scenarios,
        agent_specs={AGENT_ID: agent_spec},
        headless=False,
        visdom=False,
    )

    agent = agent_spec.build_agent()

    #done occurs here under the hood
    observations = env.reset()
    agent_obs = observations[AGENT_ID]
    agent_action = agent.act(agent_obs)

    # This works but should probably warn or assert because the agent no longer exists
    observations, rewards, dones, infos = env.step({AGENT_ID: agent_action})

    # This passes because done happened in `reset()` and the `env.step()` is no longer concerned with the agent.
    assert not dones["__all__"]

    # This crashes because the agent is done
    agent_obs = observations[AGENT_ID]

    env.close()


if __name__ == "__main__":
    parser = default_argument_parser("single-agent-example")
    args = parser.parse_args()

    main(
        scenarios=args.scenarios,
        max_episode_steps=0,
    )

```

The easiest way to reproduce this bug is to set the `AgentInterface.max_episode_steps` to 0. This will cause the agent to be done on the first step. However, there may be cases where the agent does terminate on the first step. We should find that out intelligently rather than crashing.

AC:
- [ ] `SMARTS.reset()` and `env.reset()` communicates agent `dones`
- [ ] `SMARTS` gives a warning when an agent is done on its first step.
- [ ] `SMARTS` provides a warning when trying to give an action for a `done` agent.

EDIT: example updated.

JingfeiPeng:
kkk

PR #397, opened 2021-01-11
Prevent crash from `done` during `HiWayEnv.reset()` (#54)

JingfeiPeng:
Added a warning when an agent is done on its first step.
Added a warning when trying to give an action for a done agent.

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -48,6 +48,8 @@ from .masks import RenderMasks
 from .scenario import Mission, Via
 from .waypoints import Waypoint
 
+logger = logging.getLogger(__name__)
+
 
 class VehicleObservation(NamedTuple):
     id: str
@@ -291,6 +293,13 @@ class Sensors:
             sim, agent_id, vehicle, sensor_state
         )
 
+        if (
+            done
+            and sensor_state.steps_completed == 1
+            and agent_id in sim.agent_manager.ego_agent_ids
+        ):
+            logger.warning(f"{agent_id} is done on the first step")
+
         return (
             Observation(
                 events=events,
@@ -359,6 +368,7 @@ class Sensors:
             wrong_way=is_wrong_way,
             not_moving=is_not_moving,
         )
+
         return done, events
 
     @classmethod
@@ -564,6 +574,10 @@ class SensorState:
     @property
     def mission_planner(self):
         return self._mission_planner
+
+    @property
+    def steps_completed(self):
+        return self._step
 
 
 class CameraSensor(Sensor):
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -181,6 +181,13 @@ class SMARTS(ShowBase):
             self.destroy()
             raise  # re-raise
 
+    def _check_if_acting_on_active_agents(self, agent_actions):
+        for agent_id in agent_actions.keys():
+            if agent_id not in self._agent_manager.ego_agent_ids:
+                self._log.warning(
+                    f"Attempted to perform actions on non-existing agent, {agent_id} "
+                )
+
     def _step(self, agent_actions):
         """Steps through the simulation while applying the given agent actions.
         Returns the observations, rewards, and done signals.
@@ -210,6 +217,7 @@ class SMARTS(ShowBase):
 
         # 2. Step all providers and harmonize state
         provider_state = self._step_providers(all_agent_actions, dt)
+        self._check_if_acting_on_active_agents(agent_actions)
 
         # 3. Step bubble manager and trap manager
         self._vehicle_index.sync()
--- a/smarts/env/hiway_env.py
+++ b/smarts/env/hiway_env.py
@@ -157,7 +157,6 @@ class HiWayEnv(gym.Env):
             agent_id: self._agent_specs[agent_id].action_adapter(action)
             for agent_id, action in agent_actions.items()
         }
-
         observations, rewards, agent_dones, extras = self._smarts.step(agent_actions)
 
         infos = {


examples/test_example.py
```diff
@@ -0,0 +1,67 @@
+import logging
+
+import gym
```
JingfeiPeng, line 3: will remove this file before merge

smarts/core/sensors.py
```diff
@@ -291,6 +292,11 @@ def observe(sim, agent_id, sensor_state, vehicle):
             sim, agent_id, vehicle, sensor_state
         )
 
+        print(f'observe steps: {sensor_state.steps_completed}')
+        print(f'observe done: {done}')
+        if done and sensor_state.steps_completed == 1:
+            logger.warning('Agent is done at the first step')
```
JingfeiPeng, line 298: The print statements here show this for test_example:
`observe steps: 0
observe done: True
observe steps: 0
observe done: True
observe steps: 0
observe done: True
observe steps: 0
observe done: True
observe steps: 1
observe done: False
observe steps: 1
observe done: False
observe steps: 1
observe done: False
observe steps: 1
observe done: False
observe steps: 2
observe done: False
observe steps: 2
observe done: False
observe steps: 2
observe done: False
observe steps: 1
observe done: True
WARNING:smarts.core.sensors:Agent is done at the first step
observe steps: 2
observe done: False
observe steps: 3
observe done: False
observe steps: 3
observe done: False
observe steps: 3
observe done: False
observe steps: 3`

It shows this for using a normal example like single agent:
`observe steps: 0
observe done: True
observe steps: 0
observe done: True
observe steps: 0
observe done: True
observe steps: 0
observe done: True
observe steps: 1
observe done: False
observe steps: 1
observe done: False
observe steps: 1
observe done: False
observe steps: 1
observe done: False
observe steps: 2
observe done: False
observe steps: 2
observe done: False
observe steps: 2
observe done: False
observe steps: 2
observe done: False
observe steps: 1
observe done: False
observe steps: 3
observe done: False
observe steps: 3
observe done: False
observe steps: 3
observe done: False
observe steps: 3
observe done: False
observe steps: 2
observe done: False
observe steps: 4
observe done: False
observe steps: 4
observe done: False
observe steps: 4
observe done: False
observe steps: 4
observe done: False
observe steps: 3
observe done: False
observe steps: 5
observe done: False
observe steps: 5
observe done: False
observe steps: 5
observe done: False
observe steps: 5
observe done: False
observe steps: 4
observe done: True`

smarts/core/sensors.py
```diff
@@ -291,6 +293,11 @@ def observe(sim, agent_id, sensor_state, vehicle):
             sim, agent_id, vehicle, sensor_state
         )
 
+        # print(f'observe steps: {sensor_state.steps_completed} agentId: {agent_id}')
+        # print(f'observe done: {done}')
+        if done and sensor_state.steps_completed == 1 and agent_id == "Agent-007":
+            logger.warning(f"{agent_id} is done at the first step")
```
Gamenot, line 299:
```suggestion
            logger.warning(f"{agent_id} is done on the first step")
```

smarts/core/sensors.py
```diff
@@ -291,6 +293,11 @@ def observe(sim, agent_id, sensor_state, vehicle):
             sim, agent_id, vehicle, sensor_state
         )
 
+        # print(f'observe steps: {sensor_state.steps_completed} agentId: {agent_id}')
+        # print(f'observe done: {done}')
+        if done and sensor_state.steps_completed == 1 and agent_id == "Agent-007":
```
Gamenot, line 298: OK, one main thing here.

We cannot rely on the agent id to be `Agent-007` since that is a user field. 


It might be useful to bring in the `AgentIndex` here to query for agents to separate them from the social agents. Then again it may still be worth warning for social agents. I have no hard preference. Someone else may be more opinionated on this issue.

smarts/core/smarts.py
```diff
@@ -234,6 +241,8 @@ def _step(self, agent_actions):
 
         observations, rewards, scores, dones = self._agent_manager.observe(self)
 
+        self.check_if_acting_on_non_existing_agent(agent_actions, dones)
```
Gamenot, line 244: I would suggest trying this out using more than one agent to see if this works properly.

smarts/core/smarts.py
```diff
@@ -318,13 +327,14 @@ def reset(self, scenario: Scenario):
...
 
+        default_dones = {"Agent-007": False}
         if len(self._agent_manager.ego_agent_ids):
             while len(observations_for_ego) < 1:
-                observations_for_ego, _, _, _ = self.step({})
+                observations_for_ego, _, default_dones, _ = self.step({})
 
         self._reset_providers()
 
-        return observations_for_ego
```
Gamenot, line 327: This is what we may need conversation about tonight.

Gamenot (changes requested):

smarts/core/tests/test_smarts_reset.py
```diff
@@ -63,7 +63,7 @@ def neighborhood_vehicle_poses(agent_obs):
     scenario = next(scenarios)
 
     seed(42)
-    obs = smarts.reset(scenario)
+    obs, dones = smarts.reset(scenario)
```
JingfeiPeng, line 66: keep the reset interface the same and leave changing reset interface to another issue (another milestone)

smarts/core/smarts.py
JingfeiPeng: True, this wouldn't work for multi-agent case
when running:
``` python examples/multi_agent.py scenarios/intersections/6lane```

```
WARNING:SMARTS:Attempted giving action to an non-existing agent, Agent 0 
WARNING:SMARTS:Attempted giving action to an non-existing agent, Agent 1 
WARNING:SMARTS:Attempted giving action to an non-existing agent, Agent 3 
│               0/10 │               0.79 │                 59 │               7.92 │              6lane │                    │ -65129523927046146 │   162.84 - Agent 2 │
pybullet build time: Dec 23 2020 01:47:29
pybullet build time: Dec 23 2020 01:47:29
pybullet build time: Dec 23 2020 01:47:29
pybullet build time: Dec 23 2020 01:47:29
WARNING:SMARTS:Attempted giving action to an non-existing agent, Agent 3 
WARNING:SMARTS:Attempted giving action to an non-existing agent, Agent 0 
WARNING:SMARTS:Attempted giving action to an non-existing agent, Agent 1
```

Comment (liamchzh) [reply smarts/core/smarts.py:244]: @JingfeiPeng The example you gave above, actually is called "multi scenarios". "Multi agents" means we have multiple agents in the example.

File: smarts/core/smarts.py [hunk 8]
```diff
@@ -181,6 +181,13 @@ def step(self, agent_actions):
             self.destroy()
             raise  # re-raise
 
+    def check_if_acting_on_non_existing_agent(self, agent_actions, dones):
+        for agent_id in agent_actions:
+            if agent_id not in dones:
```
liamchzh, line 186: `if agent_id not in dones` or `if agent_id in dones`? 

Anyway, please see the comments below, I think this checking needs `dones` from the previous step.

smarts/core/smarts.py
liamchzh: I think this `check_if_acting_on_non_existing_agent` is not working as we expected, because we are calling it after `observe`, i.e., the `dones` are calculated based on the current `agent_actions`, and seems to be against our intention.

smarts/core/smarts.py
```diff
@@ -672,6 +672,13 @@ def matches_provider_action_spaces(agent_id, action_spaces):
             interface = self._agent_manager.agent_interface_for_agent_id(agent_id)
             return interface.action_space in action_spaces
 
+        # Check if actions are on active ego agents
+        for agent_id, action in ego_agent_actions.items():
```
Gamenot, line 676: If only the keys are needed you can do:
```suggestion
        for agent_id in ego_agent_actions:
```
or 
```suggestion
        for agent_id in ego_agent_actions.keys():
```

smarts/core/smarts.py
```diff
@@ -661,7 +661,7 @@ def _reset_providers(self):
         for provider in self.providers:
             provider.reset()
 
-    def _step_providers(self, actions, dt) -> List[VehicleState]:
+    def _step_providers(self, actions, dt, ego_agent_actions) -> List[VehicleState]:
```
Gamenot, line 664: This is a bit unwieldy, we are asking for both all the actions and then, further, some of the actions. Perhaps send in only the keys of the active agents?

The method may be the wrong place to perform validation, the agents can be validated a method a level up as the last part of step 2. Only because the warnings are useful rather than critical.

JingfeiPeng: Very true indeed

smarts/core/smarts.py
```diff
@@ -181,6 +181,13 @@ def step(self, agent_actions):
             self.destroy()
             raise  # re-raise
 
+    def _check_if_acting_on_active_agents(self, agent_actions):
+        for agent_id in agent_actions.keys():
+            if agent_id not in self._agent_manager.ego_agent_ids:
+                self._log.warning(
+                    f"Attempted giving action to an non-existing agent, {agent_id} "
```
liamchzh, line 188: Maybe use `perform` instead of `give`? But I am not sure whether to use to do or doing here :smile:

liamchzh (approved):
Looks good!

Just a reminder:
1. Delete the example before you merge
2. Add more info in the PR description to provide more context (In case people revisit this PR)

Gamenot (approved):
LGTM

Issue #358, opened 2020-12-28
What about a pure trajectory interpolation controller?


valaxkong:
I consider that it is necessary to offering a pure trajectory interpolation controller to move vehicle without considering vehicle dynamic details. It can enhance input compatibility and is helpful to make developer more focus on designing interaction algorithm.

valaxkong:
@Gamenot 
Hi, I have posted a patch to solve this issue. Waiting for review.

liamchzh:
@valaxkong Thanks for contributing to SMARTS!

@iman512003 Please also take a look when you have a chance.

Gamenot:
@valaxkong are you still interested in adding the test for this controller?

valaxkong:
@Gamenot Yes.Sorry for the delay. I'll finish the work by February 28th.

PR #400, opened 2021-01-12
Add trajectory interpolation controller #358

valaxkong:
For issue #358 .
This controller move(set) vehicle to a specific pose which is interpolated from input trajectory.


smarts/core/controllers/trajectory_interpolation_controller.py
```diff
@@ -0,0 +1,122 @@
...
+            speed = 0.0
+        else:
+            ms = cls.interpolate(ms0, ms1, sim.timestep_sec)
+
+            center_position = ms[X_INDEX : Y_INDEX + 1]
+            center_heading = Heading(ms[THETA_INDEX])
+            pose = Pose.from_center(center_position, center_heading)
+            speed = ms[VEL_INDEX]
+
+        vehicle.set_pose(pose)
```
Gamenot, line 121: This seems like a reasonable call but it will not work unless the vehicle chassis is a box model at current, except the Ackermann chassis is the standard for agents. This is a problem since the call in the Ackermann chassis goes down to `bullet_client.resetBasePositionAndOrientation()` which disrupts the bullet simulation and causes collision filtering artefacts.

@liamchzh we have some options here, we can force the Ackermann chassis `set_pose()` to work as expected or swap the vehicle chassis to box model when using a kinematic controller.

Gamenot: Let's support a fix within this sprint.

smarts/core/controllers/trajectory_interpolation_controller.py
```diff
@@ -0,0 +1,122 @@
...
+        return trajectory[:, end_index - 1], trajectory[:, end_index]
+
+    @classmethod
+    def perform_trajectory_interpolation(
+        cls, sim, agent_id, vehicle, trajectory: np.ndarray, controller_state,
+    ):
+        """Move vehicle with trajectory interpolation
+
+        If you want vehicle stop at a specific pose,
+        trajectory[TIME_INDEX][0] should be setted as infi
```
liamchzh, line 85: Sorry, I am confused. Do you mean "should be set as infinite"?

valaxkong: Yes. Sorry for the typo. I'll fix it.

valaxkong: @liamchzh I fixed it up with a force push.

smarts/core/controllers/trajectory_interpolation_controller.py
```diff
@@ -0,0 +1,122 @@
...
+
+        speed = 0.0
+        pose = []
+        if math.isinf(ms0[TIME_INDEX]) or math.isinf(ms1[TIME_INDEX]):
+            center_position = ms0[X_INDEX : Y_INDEX + 1]
+            center_heading = Heading(ms0[THETA_INDEX])
+            pose = Pose.from_center(center_position, center_heading)
+            speed = 0.0
+        else:
+            ms = cls.interpolate(ms0, ms1, sim.timestep_sec)
```
Gamenot, line 114: Just wondering how you got this to work since `sim.timestep_sec` does not change?

```suggestion
            ms = cls.interpolate(ms0, ms1, sim._elapsed_sim_time)
```

smarts/core/controllers/trajectory_interpolation_controller.py
```diff
@@ -0,0 +1,122 @@
...
+            ValueError: Error when input trajectory has nan field.
+            RuntimeError: Error when input trajectory is empty
+        """
+        assert isinstance(vehicle.chassis, AckermannChassis)
+        if len(trajectory[TIME_INDEX]) <= 0:
+            raise RuntimeError("Input trajectory is empty!")
+        if cls.ndarray_has_nan(trajectory):
+            raise ValueError("Has nan in trajectory")
+
+        ms0, ms1 = cls.locate_motion_state(trajectory, sim.timestep_sec)
```
Gamenot, line 104:
```suggestion
        ms0, ms1 = cls.locate_motion_state(trajectory, sim._elapsed_sim_time)
```

Gamenot:
Would you happen to have an example of how this is used?

Gamenot:
I branched off of this #443 since I found I needed to test the `set_pose()` fix with the controller already available.

smarts/core/controllers/trajectory_interpolation_controller.py
valaxkong: Yes, time of trajectory point should be relative. In my experience, trajectory should be updated at every planning cycle. Therefore, it is common to give trajectory point a relative time.

Gamenot: OK, I was not sure how it was used but this should be fine then.

iman512003: So from my understanding, the main loop in the `locate_motion_state` finds the trajectory point near the next time step considering the case of mismatch between the temporal resolution of the trajectory and the time step of SMARTS, and using the obtained trajectory point and the point before it  from `locate_motion_state` we perform the interpolation for the amount of `sim.timestep_sec`. If my understanding is true, then this is ok for the case that the trajectory resolution is bigger than smarts time step,  however, for the case that the temporal resolution of the trajectory is smaller than the `sim.timestep_sec` this may result in significant error, in this case we might use an assert to ensure the temporal resolution relation or modify the return values of the `locate_motion_state`

iman512003:
It would also be great if we can have both examples and a test for this controller.

smarts/core/controllers/trajectory_interpolation_controller.py
valaxkong: @iman512003 Your understanding is right. But I think it still be fine when the temporal resolution of the trajectory is smaller than the time step. I will submit some examples and tests to ensure the validity.

valaxkong:
@Gamenot @iman512003 @liamchzh  
Sorry, I think I messed up this PR...I am going to create a new PR for the same topic.

PR #311, opened 2020-12-15
129 lane change in intersection

Gamenot:
It draws somewhat badly but it shows the lane change working.

BadVisualLaneChangeInIntersection

PR #409, opened 2021-01-13
Roll back the changes for cutin

iman512003:
This is to fix(roll back to Liam's pr) the new issue regarding the cutin task failure for the behavioral model.

The broken logic was introduced by #311 .


liamchzh (approved):

Gamenot (approved):
OK, my bad if this caused issues.

PR #410, opened 2021-01-13
Fixes to improve mini-city

Gamenot:
In trying to use the minicity map there were a few issues that came up.

I found that, perhaps due to an artefact of the map cutter, the streets often had u-turns built in which were causing traffic congestion so I mostly fixed the map.

Also in fixing the map I damaged the traffic light schedule. And running it for a long time I found that an vehicle with sensors somehow had their sensors replaced.


smarts/core/vehicle.py
```diff
@@ -531,9 +531,9 @@ def _meta_create_sensor_functions(self):
         for sensor_name in sensor_names:
 
             def attach_sensor(self, sensor, sensor_name=sensor_name):
-                assert (
-                    getattr(self, f"_{sensor_name}", None) is None
-                ), f"{sensor_name} already added"
+                # assert (
```
Gamenot, line 534: This comment out will be reverted but I found that somehow vehicles with sensors were having their sensor replaced.

Run `python examples/egoless.py scenarios/minicity`

scenarios/minicity/scenario.py
```diff
@@ -4,18 +4,28 @@
...
 traffic = t.Traffic(
     flows=[
         t.Flow(
             route=t.RandomRoute(),
-            rate=60 * 60,
+            rate=60 * 2,
             actors={
                 t.TrafficActor(
                     name="car",
                     vehicle_type=random.choice(
```
liamchzh, line 17: Nit: `numpy.random.choice` allows you to provide probabilities.

smarts/core/vehicle.py
liamchzh: This should be reverted.

liamchzh (approved):

JingfeiPeng (approved):

PR #431, opened 2021-01-14
Hasten CI

### LINKED ISSUE #152: Decrease CI time to run tests
It can be up to an hour (or worse: cancelled) to run the main tests. It should be less than 10 minutes, ideally only a few.

Move to Circle CI or pay GitHub.

Adaickalavan:
+ Original CI duration: about 35 minutes
+ Current CI duration: about 15 minutes
+ Changes: CI config file now uses strategy matrix which helps run test cases in parallel.
+ Longest individual test case `/smarts/env/tests/test_rllib_hiway_env.py` takes a minimum raw time of 10 minutes to run (excluding setup time)

Closes #152 .


liamchzh:
Very impressive and worth doing such an enhancement before we migrate to self-host workers or any paid CI.

liamchzh (approved):

Gamenot (approved):
Thanks for being thorough with this.

Gamenot:
@Adaickalavan I think this is enough time spent on this problem. This should go in once the test-formatting is addressed.

smarts/core/remote_agent_buffer.py
```diff
@@ -197,8 +197,6 @@ def get_manager_channel_stub(addr):
         # Wait until the grpc server is ready or timeout after 30 seconds
         grpc.channel_ready_future(channel).result(timeout=30)
     except grpc.FutureTimeoutError:
-        raise RemoteAgentException(
```
liamchzh, line 200: :+1:

liamchzh (commented):
Is the caching buying us more time? :smile:

Issue #405, opened 2021-01-12
OpEn "teleports" after collision with MAD vehicle


liamchzh:
By teleport, I mean the agent vehicle appears again after the crash with the same vehicle ID.

liamchzh:
Seems that this PR https://github.com/huawei-noah/SMARTS/pull/369 introduced the teleport.

The difference after this change is we will wait for the SUMO process exit gracefully.

PR #435, opened 2021-01-17
Remove all vehicle in SUMO when reset

liamchzh:
This is to fix the agent teleport issue. #405 

Every time when we swap/reset a scenario, we will reload SUMO, this works well if we can successfully terminate SUMO. However, if the SUMO server continues running, it will take over the controller of our agents. To ensure the cleanup is nicely done, we manually remove the vehicles in the sumo traffic sim's `teardown`.

### DIFF
--- a/smarts/core/sumo_traffic_simulation.py
+++ b/smarts/core/sumo_traffic_simulation.py
@@ -273,6 +273,10 @@ class SumoTrafficSimulation:
             self._traci_conn.close()
             self._traci_conn = None
 
+    def _remove_all_vehicles(self):
+        for vehicle_id in self._non_sumo_vehicle_ids.union(self._sumo_vehicle_ids):
+            self._traci_conn.vehicle.remove(vehicle_id)
+
     def teardown(self):
         self._log.debug("Tearing down SUMO traffic sim %s" % self)
         if not self._is_setup:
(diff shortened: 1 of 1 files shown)


iman512003 (approved):

smarts/core/sumo_traffic_simulation.py
```diff
@@ -273,6 +273,10 @@ def _close_traci_and_pipes(self):
             self._traci_conn.close()
             self._traci_conn = None
 
+    def _remove_all_vehicles(self):
```
Gamenot, line 276: This method name is slightly misleading since it is specifically remove vehicles that are not SUMO vehicles.

Gamenot (changes requested):

smarts/core/sumo_traffic_simulation.py
liamchzh: Probably the `union` is misleading or not explicitly enough?

Gamenot: Oh sorry, when I originally looked at this I was reviewing late at night and submitted it when I got up.

Is the MAD vehicle somehow included in this?

Gamenot (approved):

smarts/core/sumo_traffic_simulation.py
liamchzh: Yes, including MAD vehicles.

PR #439, opened 2021-01-18
Fix history vehicles key error + not recreate SMARTS instance 

JingfeiPeng:
Originally it takes from 1s to 0.06 second to initialize smarts instance for each loop of imitation learning. 
Time reduced to exponentially small by not re-instantiating SMARTS for each loop.

### DIFF
--- a/examples/history_vehicles_replacement_for_imitation_learning.py
+++ b/examples/history_vehicles_replacement_for_imitation_learning.py
@@ -20,8 +20,11 @@ class KeepLaneAgent(Agent):
 
 def main(scenarios, headless, seed):
     scenarios_iterator = Scenario.scenario_variations(scenarios, [])
-    first_time = True
-    smarts = None
+    smarts = SMARTS(
+        agent_interfaces={},
+        traffic_sim=SumoTrafficSimulation(headless=True, auto_start=True),
+        envision=Envision(),
+    )
 
     for _ in scenarios:
         scenario = next(scenarios_iterator)
@@ -38,15 +41,7 @@ def main(scenarios, headless, seed):
             )
             agent = agent_spec.build_agent()
 
-            if first_time:
-                smarts = SMARTS(
-                    agent_interfaces={agent_id: agent_spec.interface},
-                    traffic_sim=SumoTrafficSimulation(headless=True, auto_start=True),
-                    envision=Envision(),
-                )
-                first_time = False
-            else:
-                smarts.switch_ego_agent({agent_id: agent_spec.interface})
+            smarts.switch_ego_agent({agent_id: agent_spec.interface})
 
             observations = smarts.reset(scenario)
 
--- a/examples/history_vehicles_replacement_for_imitation_learning.py
+++ b/examples/history_vehicles_replacement_for_imitation_learning.py
@@ -29,7 +29,6 @@ def main(scenarios, headless, seed):
         agent_missions = scenario.discover_missions_of_traffic_histories()
 
         for agent_id, mission in agent_missions.items():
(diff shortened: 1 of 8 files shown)


smarts/core/remote_agent_buffer.py
```diff
@@ -196,7 +196,9 @@ def get_manager_channel_stub(addr):
     try:
         # Wait until the grpc server is ready or timeout after 30 seconds
         grpc.channel_ready_future(channel).result(timeout=30)
-    except grpc.FutureTimeoutError:
-        raise RemoteAgentException("Timeout in connecting to remote zoo manager.")
+    except grpc.FutureTimeoutError as e:
+        raise RemoteAgentException(
+            "Timeout in connecting to remote zoo manager."
+        ) from e
```
liamchzh, line 202: @Adaickalavan already fixed this.

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -20,6 +20,9 @@ def act(self, obs):
 
 def main(scenarios, headless, seed):
     scenarios_iterator = Scenario.scenario_variations(scenarios, [])
+    first_time = True
```
liamchzh, line 23: If we initialize at the beginning with an empty agent_interfaces - `SMARTS(agent_interfaces={})`, the code would be more concise.

smarts/core/trap_manager.py
```diff
@@ -145,6 +145,10 @@ def largest_vehicle_plane_dimension(vehicle):
                 ),
             )
             for v_id in sorted_vehicle_ids:
+                # Skip the capturing process if history traffic is used
+                if sim.scenario.traffic_history != {}:
```
liamchzh, line 149:
```suggestion
                if not sim.scenario.traffic_history:
```

smarts/core/smarts.py
```diff
@@ -369,6 +369,10 @@ def setup(self, scenario: Scenario):
     def add_provider(self, provider):
         self._providers.append(provider)
 
+    def switch_ego_agent(self, agent_interface):
```
liamchzh, line 372: Maybe replace `switch` with a more accurate verb?

smarts/core/remote_agent_buffer.py
Adaickalavan: Here, the `grpc.FutureTimeoutError` may only mean one thing, which is that we are unable to connect to the server due to a timeout. 

Since there is no other information to be gained from this error, and to keep the error message concise and helpful to the user, I believe there is no need to print out the actual timeout error using `from e` which does not look very beautiful.

The originally fixed code is fine:
```
except grpc.FutureTimeoutError:
    raise RemoteAgentException("Timeout in connecting to remote zoo manager.")
```

smarts/core/trap_manager.py
JingfeiPeng: Thanks, though to be precise, the correction should be `if sim.scenario.traffic_history:`

liamchzh (approved):

smarts/core/trap_manager.py
```diff
@@ -145,6 +145,10 @@ def largest_vehicle_plane_dimension(vehicle):
                 ),
             )
             for v_id in sorted_vehicle_ids:
+                # Skip the capturing process if history traffic is used
+                if sim.scenario.traffic_history:
```
liamchzh, line 149: One assumption here is when doing imitation learning, there will be no SUMO traffic, therefore we wanted to disable the trapping.

Adaickalavan (approved):

PR #441, opened 2021-01-18
Option for resetting a scenario without restart or reload SUMO

liamchzh:


smarts/core/sumo_traffic_simulation.py
```diff
@@ -60,11 +60,11 @@ def __init__(
         self,
         headless=True,
         time_resolution=0.1,
-        num_clients=1,
```
liamchzh, line 63: This param is not being used.

smarts/core/sumo_traffic_simulation.py
```diff
@@ -273,6 +274,10 @@ def _close_traci_and_pipes(self):
             self._traci_conn.close()
             self._traci_conn = None
 
+    def _remove_all_vehicles(self):
```
liamchzh, line 277: This change is duplicated with #435

JingfeiPeng (approved):

Adaickalavan (approved):

PR #442, opened 2021-01-18
Fix the teleport issue for BM

iman512003:
This is to address #405 .

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -449,7 +449,6 @@ class SMARTS(ShowBase):
 
     def destroy(self):
         self.teardown()
-        self._traffic_sim.destroy()
 
         if self._envision:
             self._envision.teardown()
@@ -458,6 +457,7 @@ class SMARTS(ShowBase):
             self._visdom.teardown()
 
         self._agent_manager.destroy()
+        self._traffic_sim.destroy()
         self._bullet_client.disconnect()
 
         super().destroy()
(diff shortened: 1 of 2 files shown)


smarts/core/smarts.py
```diff
@@ -449,6 +449,7 @@ def teardown(self):
 
     def destroy(self):
         self.teardown()
+        self._traffic_sim.destroy()
```
iman512003, line 452: I am not sure whether this is the best place to add this function.

smarts/core/sumo_traffic_simulation.py
```diff
@@ -108,11 +106,10 @@ def __repr__(self):
     def __str__(self):
         return repr(self)
 
-    def _destroy(self):
-        if atexit:
-            atexit.unregister(self._destroy)
-
+    def destroy(self):
         self._close_traci_and_pipes()
+        if not self._is_setup:
```
iman512003, line 111: If I do not add this then the traffic test will fail when it calls for destroying smarts and throw an error which mentioned that no sumo process exists.

smarts/core/smarts.py
Gamenot: This should be fine, although, it might be better to put it after the `self._agent_manager.destroy()`

Gamenot (approved):
LGTM

Adaickalavan (approved):

smarts/core/smarts.py
liamchzh: Should be fine.

liamchzh (approved):

PR #450, opened 2021-01-19
Release v0.4.12

Gamenot:
This branch is to be tested and then merged.


liamchzh (approved):

Adaickalavan:
Although the tests passed, the [CI](https://github.com/huawei-noah/SMARTS/runs/1729705788) test for merge #420 appears to slow down the tests in the `./tests` folder. It would be comforting to understand why this happens.

liamchzh:
> Although the tests passed, the [CI](https://github.com/huawei-noah/SMARTS/runs/1729705788) test for merge #420 appears to slow down the tests in the `./tests` folder. It would be comforting to understand why this happens.

Thanks for reminding us.

@Gamenot Do you think we should revert #420 ?

Gamenot:
> > Although the tests passed, the [CI](https://github.com/huawei-noah/SMARTS/runs/1729705788) test for merge #420 appears to slow down the tests in the `./tests` folder. It would be comforting to understand why this happens.
> 
> Thanks for reminding us.
> 
> @Gamenot Do you think we should revert #420 ?

I investigated it deeply, it looks like the problem with it had nothing to do with the upgrade of `ray` and `rllib` and only did not occur because of how the code operated, it could happen anyway but was not caught in the test because of setting `PYTHONHASHSEED` just happened to make it short. I have fixed the problem so we do not need to revert.

examples/rllib.py
```diff
@@ -62,17 +97,22 @@ def main(
...
         perturbation_interval=300,
         resample_probability=0.25,
         # Specifies the mutations of these hyperparams
+        # See: `ray.rllib.agents.trainer.COMMON_CONFIG` for common hyperparams
         hyperparam_mutations={
             "lr": [1e-3, 5e-4, 1e-4, 5e-5, 1e-5],
-            "rollout_fragment_length": lambda: random.randint(128, 16384),
-            "train_batch_size": lambda: random.randint(2000, 160000),
+            "rollout_fragment_length": lambda: rollout_fragment_length,
+            "train_batch_size": lambda: train_batch_size,
```
Gamenot, line 115: These two lines were the problem since they used random values. The core of it was that the `ray.tune()` `stop` parameter was only being checked per training batch so with extremely large training batches it would get checked very infrequently.

Gamenot:
Pipeline passed relatively quickly: https://github.com/huawei-noah/SMARTS/actions/runs/503282058
image

JingfeiPeng (approved):

examples/rllib.py
```diff
@@ -19,27 +27,52 @@
...
 
+        episode.user_data["ego_speed"] = []
 
-def on_episode_step(info):
-    episode = info["episode"]
-    single_agent_id = list(episode._agent_to_last_obs)[0]
-    obs = episode.last_raw_obs_for(single_agent_id)
-    episode.user_data["ego_speed"].append(obs["speed"])
+    def on_episode_step(
+        self,
```
liamchzh, line 45: `self` and `*` seem not to be used?

Adaickalavan (approved):

examples/rllib.py
Gamenot: Sure, yes, the * is not needed here it was auto-generated from the base method.

Issue #433, opened 2021-01-15
Decouple logging envision data from headless mode


liamchzh:
`envision_record_data_replay_path` ([hiway_env.py](https://github.com/huawei-noah/SMARTS/blob/master/smarts/env/hiway_env.py#L61)) enables data logging, but it requires headless also set to `False`.

The reason is that the Envision client also takes care of the logging. Here is what happened in pseudo code:

```
if not headless:
    client send envision data
        if enable data logging:
            write to files
        ...
```

The data logging and headless mode should be decoupled.

PR #451, opened 2021-01-20
Decouple logging keep headless envision (#433)

JingfeiPeng:
Added a headless mode in envision client. When enabled, envision will only run the logging but not send state data to server

The other option is to decouple the logging code from envision and have logging be controlled in smarts.py, which requires more changes.

And feel free to suggest other options.

### DIFF
--- a/envision/client.py
+++ b/envision/client.py
@@ -76,8 +76,10 @@ class Client:
         wait_between_retries: float = 0.5,
         output_dir: str = None,
         sim_name: str = None,
+        headless: bool = False,
     ):
         self._log = logging.getLogger(self.__class__.__name__)
+        self.headless = headless
 
         current_time = datetime.now().strftime("%Y%m%d%H%M%S%f")[:-4]
         client_id = current_time
@@ -96,14 +98,15 @@ class Client:
             self._logging_thread = self._spawn_logging_thread(output_dir, client_id)
             self._logging_thread.start()
 
-        self._state_queue = Queue()
-        self._thread = self._connect(
-            endpoint=f"{endpoint}/simulations/{client_id}/broadcast",
-            queue=self._state_queue,
-            num_retries=num_retries,
-            wait_between_retries=wait_between_retries,
-        )
-        self._thread.start()
+        if not self.headless:
+            self._state_queue = Queue()
+            self._thread = self._connect(
+                endpoint=f"{endpoint}/simulations/{client_id}/broadcast",
+                queue=self._state_queue,
+                num_retries=num_retries,
+                wait_between_retries=wait_between_retries,
+            )
+            self._thread.start()
 
     def _spawn_logging_thread(self, output_dir, client_id):
         output_dir = Path(f"{output_dir}/{int(time.time())}")
@@ -222,7 +225,7 @@ class Client:
         )
 
     def send(self, state: types.State):
-        if self._thread.is_alive():
+        if not self.headless and self._thread.is_alive():
             self._state_queue.put(state)
         if self._logging_thread:
             self._logging_queue.put(state)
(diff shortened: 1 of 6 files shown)


JingfeiPeng:
@Gamenot @liamchzh any thought on this decoupling approach by having an option in envision client? 

@Gamenot If I understood correctly, the approached you proposed to do decoupling is to have an logging interface that is controlled by env and have env pass the logging service to smarts,py so smarts.py can send data to this interface, then this service sends data to envision client if not headless. Is this correct?

liamchzh:
This is also the first implementation that I can think of, simple and intuitive.

I also suggest doing profiling on `try_emit_envision_state`, just wondering how much time it spends on constructing the state (`state = envision_types.State()`) and sending it to Envision server (`self._envision.send(state)`). I think with and without Envision server up will have a huge impact on those times, so we can measure the time with the following mode:

| Envision server | headless | replay_path |
| ------------- | ------------- | ------------- |
| Y | Y | Y |
| N | N | Y |
| N | Y | Y |

JingfeiPeng:
`try_emit_envision_state` average timing for each scenario: 
**intersections/6lane**: 0.14ms to 0.16ms
**intersections/4lane_t**: 0.15ms-0.35ms
**loop:**: 0.15ms-0.35ms
**History_vehicles_replacement**: 0.15ms - 0.35ms





> This is also the first implementation that I can think of, simple and intuitive.
> 
> I also suggest doing profiling on `try_emit_envision_state`, just wondering how much time it spends on constructing the state (`state = envision_types.State()`) and sending it to Envision server (`self._envision.send(state)`). I think with and without Envision server up will have a huge impact on those times, so we can measure the time with the following mode:
> Envision server 	headless 	replay_path
> Y 	N 	Y
> N 	N 	Y
> N 	Y 	Y

`try_emit_envision_state` average timing for each scenario: 
**intersections/6lane**: 0.14ms to 0.16ms
**intersections/4lane_t**: 0.15ms-0.35ms
**loop:**: 0.15ms-0.35ms
**History_vehicles_replacement**: 0.15ms - 0.35ms

Tested time spent from smarts.py `self._envision.send(state)` to client.py `ws.send(state)`
| Envision server | headless | replay_path | result |
| ------------- | ------------- | ------------- | ------------- |
| Y | N | Y | 5ms to 11ms for history_vehicles, 40ms to 70ms for 6lane, 15m to 45ms for 4lane_t (most are 20~ ms
| N | N | Y | _connect() in client fails as no server available to connect, prints errors on screen and wouldn't reach sending step |
| N | Y | Y | Since headless is disabled, will not enter send data to server thread, logging data on file takes on average 15ms from self._envision.send(state) to f.write(f"{state}\n") in client |

liamchzh:
`try_emit_envision_state` takes time even without envision server, doesn't it?

JingfeiPeng:
> `try_emit_envision_state` takes time even without envision server, doesn't it?

right, it will take time as long as `self._envision` exists which could be true during headless mode, I will add another condition in the check  for return `if not self._envision or self._envision.headless`

envision/client.py
```diff
@@ -76,8 +76,10 @@ def __init__(
         wait_between_retries: float = 0.5,
         output_dir: str = None,
         sim_name: str = None,
+        headless: bool = False,
     ):
         self._log = logging.getLogger(self.__class__.__name__)
+        self.headless = headless
```
Gamenot, line 82: I think this should be private at least for now since it looks like state change afterwards can break the object.
```suggestion
        self._headless = headless
```

Gamenot (approved):

envision/client.py
Gamenot, line 82: I think this should be private at least for now since it looks like state change afterwards can break the object.
```suggestion
        self._headless = headless
```

Gamenot (changes requested):
A concern but otherwise looks good.

liamchzh (approved):

PR #477, opened 2021-01-25
Remove ijson dependency in ULTRA

JenishPatel99:
ULTRA now uses a stable version of SMARTS, slight tweak in the installation process is needed. README needs to updated as well to accommodate the installation changes. As of now, ULTRA uses SMARTS v0.4.6


ultra/install.sh
```diff
@@ -0,0 +1,11 @@
...
+
+pip install --upgrade pip
+pip install -e .
+
+dir="$PWD"
+smartsdir="$(dirname "$dir")"
+
+export PYTHONPATH=$smartsdir
+
+
```
liamchzh, line 11: Remove the extra blank line?

ultra/install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+pip install --upgrade pip
+pip install -e .
+
+dir="$PWD"
+smartsdir="$(dirname "$dir")"
+
+export PYTHONPATH=$smartsdir
```
liamchzh, line 9: My feeling is that we should install smarts as a lib instead of adding the SMARTS path to `PYTHONPATH`.

cc @Gamenot

Gamenot: @JenishPatel99 is this how you tried to solve the issue with the repository not having the path to the ultra package from within the directory?.

JenishPatel99: At first the problem was that the PYTHONPATH did not include the absolute path to SMARTS, causing ultra module to be not found. So I decided to export the path and the ModuleNotFoundError was solved. However to do, I noticed that the export the python path is temporary and every time we close python and start it back, the path is gone. So, to solve that I echoed back the export command into /.bashrc script. It seems that it permanently stores the path in PYTHONPATH

ultra/setup.py
```diff
@@ -0,0 +1,44 @@
...
+setup(
+    name="ultra",
+    description="Unprotected Left Turn using Reinforcement-learning Agents",
+    version="0.1",
+    packages=find_packages(),
+    include_package_data=True,
+    zip_safe=True,
+    python_requires=">=3.7",
+    install_requires=[
+        "smarts[train]==0.4.6",  # Stable version
```
liamchzh, line 39: The latest is 0.4.12 https://pypi.org/project/smarts/

ultra/install.sh
liamchzh: Probably not a good idea to solve it by having a script and source it every time. 

Having a `setup.py` is a good start, you can pin down the dependencies. Then you should be able to import smarts from the installed site-packages after install SMARTS (`pip install smarts`).

JenishPatel99: @liamchzh As you mentioned above that it is better to have smarts as a library, that was my first approach by editing the setup.py file in ultra/ . However, the issue was that people who are using ultra will create an virtual environment inside of the ultra directory and not in smarts. 

Whenever, I tried installing the dependencies using the "pip install -e ." in ultra, it was able to get the mentioned version of smarts in the setup.py and install it into site-packages in the venv inside of ultra (not in SMARTS) 

Coming back a directory up (SMARTS/) to run the ultra commands, there was a module not found error, saying that ultra is not a module. So, my guess was that the path for the SMARTS/ was not initialized in PYTHONPATH and caused python to not locate the ultra package properly. 

So, my approach now is to create a one-time executable shell script which downloads the dependencies, adds an absolute path to the bashrc script, which then solves the issue of the Module not found and also gives a permanent solution to not running the script everytime.

ultra/setup.py
JenishPatel99: The ultra team wants to have a stable version of smarts, and I was suggested that v0.4.6 was very stable. In the future we will update it to imply the latest version release of smarts

ultra/install.sh
liamchzh: > Coming back a directory up (SMARTS/) to run the ultra commands, there was a module not found error, saying that ultra is not a module.

@JenishPatel99  May I know the reason why you need to go to SMARTS directory?
- - - - 
@Gamenot @adreena  I am wondering what's your thought regarding this. I do have a few concerns, e.g., this is a hack t the search path.

JenishPatel99: @liamchzh I discussed with @Gamenot and I was under the impression that the ultra will also rely SMARTS repo in addition with smarts package. The confusion seems to be settled. Please ignore this pull request

AlexLewandowski (approved):

christianjans (approved):

PR #485, opened 2021-01-27
fix collision between agent and bubble agent vehicle

JingfeiPeng:


smarts/core/vehicle_index.py
```diff
@@ -382,7 +388,8 @@ def switch_control_to_agent(
                 is_hijacked=hijacking,
             )
         )
-
+        interface = sim.agent_manager.agent_interface_for_agent_id(agent_id)
+        print(f"Before: {agent_id} action: {interface}") # interfce none here
```
JingfeiPeng, line 392: interface is None here

smarts/core/bubble_manager.py
```diff
@@ -451,9 +451,11 @@ def _hijack_social_vehicle_with_social_agent(
             hijacking=True,
             recreate=False,
         )
-
+        interface = sim.agent_manager.agent_interface_for_agent_id(agent_id)
+        print(f"After: {agent_id} action: {interface.action}")
```
JingfeiPeng, line 455: Interface is not none and can access the action

Gamenot:
OK, I took a look at this and from what I can see it is a problem with using the TargetPose action space here.

JingfeiPeng:
Smarts will only call controller's perform_action if the agent's action is in _dynamic_action_spaces.

We want the vehicle to have boxChassis so vehicle index.py, switch_control_to_agent check if agent's action is not in _dynamic_action_spaces and if so, create BoxChassis

To let the vehicle to reach the controller's perform_action method, we need to put TargetPose in dynnamic_action_spaces....

This causes vehicle index switch_control_to_agent to not create box model for the agents, and have the TargetPose agents to use AckermanChassis. Though this can be solved by making vehicle Index check if agent action is TargetPose or Multi-targetpose.

@Gamenot  But it seems the controllers are for vehicles with Ackerman chassis, and using MotionPlannerProvider for bubble agents with BoxChassis is that not bad of a choice?

JingfeiPeng:
Finished ensuring collision issue is fixed and proper use of boxChassis for non dynamic agents.
Will create another issue for refactoring MotionPlannerProvider into a controller similar to Trajectory_Tracking controller

smarts/core/vehicle.py
```diff
@@ -483,6 +498,14 @@ def sync_to_panda3d(self):
         pos, heading = self._chassis.pose.as_panda3d()
         self._np.setPosHpr(*pos, heading, 0, 0)
 
+    @lru_cache(maxsize=1)
+    def _warn_non_kinematic_vehicle_set_pose(self):
```
Gamenot, line 502: This was a bad naming scheme on my part. Also not called.

smarts/core/vehicle.py
```diff
@@ -483,6 +498,14 @@ def sync_to_panda3d(self):
...
+        if self._has_stepped and isinstance(self._chassis, AckermannChassis):
+            logging.warning(
+                f"Agent `{self._id}` has called set pose after step."
+                "This may cause collision problems"
+            )
+
     # TODO: Merge this w/ speed setter as a set GCD call?
     def set_pose(self, pose: Pose):
```
Gamenot, line 510: I think here ^.

Gamenot (approved):

smarts/core/chassis.py
```diff
@@ -164,9 +165,7 @@ def dimensions(self) -> BoundingBox:
 
     @property
     def contact_points(self) -> Sequence:
-        contact_points = _query_bullet_contact_points(
-            self._client, self._bullet_body, -1
-        )
+        contact_points = _query_bullet_contact_points(self._client, self.bullet_id, -1)
```
Gamenot, line 168:
```suggestion
        contact_points = _query_bullet_contact_points(self._client, self._bullet_body.bullet_id, -1)
```

Gamenot (changes requested):

Gamenot (approved):

iman512003 (approved):

PR #487, opened 2021-01-27
Clean up args and help strings in evaluate.py and train.py

AlexLewandowski:


ultra/evaluate.py
```diff
@@ -0,0 +1,218 @@
...
+    parser = argparse.ArgumentParser("intersection-single-agent-evaluation")
+    parser.add_argument(
+        "--task", help="Tasks available : [1, 2]", type=str, default="1"
+    )
+    parser.add_argument(
+        "--level",
+        help="Levels available : [easy, medium, hard, no-traffic]",
+        type=str,
+        default="easy",
+    )
```
christianjans, line 157: Just thinking, should these be mandatory arguments?
```suggestion
    parser.add_argument(
        "--task", help="Tasks available : [1, 2]", type=str, required=True
    )
    parser.add_argument(
        "--level",
        help="Levels available : [easy, medium, hard, no-traffic]",
        type=str,
        required=True,
    )
```

ultra/train.py
```diff
@@ -0,0 +1,182 @@
...
+    parser = argparse.ArgumentParser("intersection-single-agent")
+    parser.add_argument(
+        "--task", help="Tasks available : [1, 2]", type=str, default="1"
+    )
+    parser.add_argument(
+        "--level",
+        help="Levels available : [easy, medium, hard, no-traffic]",
+        type=str,
+        default="easy",
+    )
```
christianjans, line 128: Same thing here obviously, should these be mandatory?

christianjans (commented):
Looks great! Just a comment about the task and level arguments.

ultra/train.py
AlexLewandowski: They aren't mandatory since there are defaults. For training however, we might be able to think of a better api for users to select the env/task/scenario.

ultra/evaluate.py
AlexLewandowski: They aren't mandatory since there are defaults. They currently match the defaults in train.py, but maybe we don't want to do that.

ultra/train.py
christianjans: Hmm.. yeah okay. We can maybe change this in a future ULTRA version.

ultra/evaluate.py
christianjans: Actually sorry, for now I think it's probably okay to have default values here; just for the initial ULTRA version.

ultra/train.py
```diff
@@ -0,0 +1,182 @@
...
+                timestep_sec=float(args.timestep),
+                headless=args.headless,
+                policy_class=policy_class,
+                seed=args.seed,
+            )
+        ]
+    )
+    # finally:
+    #     time.sleep(1)
+    #     ray.shutdown()
```
JenishPatel99, line 182: Sorry for the late review. Everything seems good, train.py will be updated from my side

JenishPatel99 (commented):
Seems good

christianjans (approved):
Looks good!

JenishPatel99 (approved):

Issue #493, opened 2021-01-28
Do not remove vehicles emitted by MAD Co-simulator


liamchzh:
To integrate SMARTS service with H, the key requirements are:
- [x] #441 Do NOT restart or reload SUMO
- [ ] Do NOT remove vehicles emitted by the co-simulator

In #435 we introduced a clean up to remove all vehicles when resetting a scenario. To avoid removing vehicles emitted by the co-simulator, I built a "hack" version of SMARTS (4.12a0) and put it in the SMARTS service repo.

original master:
```
    def _remove_all_vehicles(self):
        for vehicle_id in self._non_sumo_vehicle_ids.union(self._sumo_vehicle_ids):  # remove all vehicles
            self._traci_conn.vehicle.remove(vehicle_id)
```

becomes:
```
    def _remove_all_vehicles(self):
        for vehicle_id in self._non_sumo_vehicle_ids:  # only remove agents
            self._traci_conn.vehicle.remove(vehicle_id)
```

To address, a few approaches we can take:
1. Revert #435, which turns out is not necessary since every time SUMO will be reloaded.
2. Introduce a MAD mode, which excludes the vehicles emitted by the co-simulator when removing the vehicles in the scene.
3. Maintain a "hack" version in the SMARTS service repository.

PR #495, opened 2021-01-28
Add MAD mode (#493)

JingfeiPeng:
Used approach 2 in #493

### DIFF
--- a/smarts/core/sumo_traffic_simulation.py
+++ b/smarts/core/sumo_traffic_simulation.py
@@ -66,7 +66,7 @@ class SumoTrafficSimulation:
         endless_traffic=True,
         allow_reload=True,
         debug=True,
-        kybersim_mode=False
+        kybersim_mode=False,
     ):
         self._kybersim_mode = kybersim_mode
         self._log = logging.getLogger(self.__class__.__name__)
@@ -278,7 +278,9 @@ class SumoTrafficSimulation:
         if self._kybersim_mode:
             vehicles_to_remove = self._non_sumo_vehicle_ids
         else:
-            vehicles_to_remove = self._non_sumo_vehicle_ids.union(self._sumo_vehicle_ids)
+            vehicles_to_remove = self._non_sumo_vehicle_ids.union(
+                self._sumo_vehicle_ids
+            )
 
         for vehicle_id in vehicles_to_remove:
             self._traci_conn.vehicle.remove(vehicle_id)
--- a/smarts/core/sumo_traffic_simulation.py
+++ b/smarts/core/sumo_traffic_simulation.py
@@ -66,9 +66,9 @@ class SumoTrafficSimulation:
         endless_traffic=True,
         allow_reload=True,
         debug=True,
-        kybersim_mode=False,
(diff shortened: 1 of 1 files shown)


smarts/core/sumo_traffic_simulation.py
```diff
@@ -272,7 +274,15 @@ def _close_traci_and_pipes(self):
...
-        for vehicle_id in self._non_sumo_vehicle_ids.union(self._sumo_vehicle_ids):
+        vehicles_to_remove = None
+        if self._kybersim_mode:
+            vehicles_to_remove = self._non_sumo_vehicle_ids
+        else:
+            vehicles_to_remove = self._non_sumo_vehicle_ids.union(
+                self._sumo_vehicle_ids
+            )
+
+        for vehicle_id in vehicles_to_remove:
```
JingfeiPeng, line 285: Could also change this and the above to the following if needed
```
(
            self._non_sumo_vehicle_ids
            if self._kybersim_mode
            else self._non_sumo_vehicle_ids.union(self._sumo_vehicle_ids)
        )
```

liamchzh:
Nice! Actually, I was going to discuss the approaches in the parking log since we are introducing more toggles/params to meet the requirements, which I think should be fully considered before we get our hands dirty.

A few things:
* `allow_reload` and `kybersim_mode` are params customized for the KyberSim, should we keep the `allow_reload` or not
* If we decide to use separate params, then `kybersim_mode` is a bit too ambiguous
* We had talked about only exclude the vehicles with a specific prefix, how do we extend the API to support this

smarts/core/sumo_traffic_simulation.py
```diff
@@ -66,7 +66,9 @@ def __init__(
         endless_traffic=True,
         allow_reload=True,
         debug=True,
+        kybersim_mode=False,
```
Gamenot, line 69: Use a different name.

smarts/core/sumo_traffic_simulation.py
```diff
@@ -66,7 +66,9 @@ def __init__(
         endless_traffic=True,
         allow_reload=True,
         debug=True,
+        integration_mode=False,
     ):
```
Gamenot, line 70: We need a docstring here describing these parameters.

Gamenot (approved):

Adaickalavan (approved):

smarts/core/sumo_traffic_simulation.py
```diff
@@ -54,6 +54,20 @@ class SumoTrafficSimulation:
...
+        auto_start:
+            False to pause simulation when SMARTS run, and wait for user to click 
+            start on sumo-gui
+        endless_traffic:
+            Not remove vehicles by re-adding back vehicles that exit 
+        allow_reload:
+            True to reload existing SUMO
+        remove_agents_only_mode: 
+            Remove only agent vehicles used by SMARTS and not delete other sumo
+            vehicles when teardown
```
JingfeiPeng, line 70: Feel free to point out anything inaccurate about the description

PR #498, opened 2021-01-28
Add policy and log directory arguments to train.py

JenishPatel99:


AlexLewandowski:
Are the files in test/ supposed to be part of the commit?

Issue #488, opened 2021-01-27
OpEn U-Turn & Cut-In Documentation


Gamenot:
A document describing the ideas behind tuning the OpEn agent, U-Turn, and Cut-In and how to change the parameters and hyperparameters.

liamchzh:
This document should cover:
1. What's the approach that we chose to accomplish the tasks (U-Turn & Cut-in)
2. What are the responsibilities of the mission planner and OpEn
3. The ideas behind the trajectory calculation of u-turn and cut-in (What are the params and how do they affect the trigger timing)
4. Other topics that Chendong might be interested in (Jeffer can help clarify in the regular meetings)

JingfeiPeng:
Done this is added in `docs/cutin-uturn-docs.rst`

PR #503, opened 2021-01-29
Add docs for uturn and cut in tasks

iman512003:
This is to address #488 .


Gamenot (approved):
Thanks @iman512003 this is great.

JingfeiPeng (approved):

Issue #346, opened 2020-12-22
Refactor the u-turn and cut-in scenarios


iman512003:
No tuning should occur.

Simplify the code.

This ticket will close when the client is satisfied with the u-turn and cut-in tasks.

Gamenot:
Closed in favour of #617.

PR #505, opened 2021-01-29
Refactor uturn and cutin

iman512003:
This is to address #346 .


Gamenot (approved):
This looks like a good cleanup pass

PR #507, opened 2021-01-29
Updated train.py and evaluate.py to include new flags

JenishPatel99:


ultra/train.py
```diff
@@ -146,13 +152,28 @@ def train(task, num_episodes, policy_class, eval_info, timestep_sec, headless, s
     parser.add_argument(
         "--seed", help="environment seed", default=2, type=int,
     )
+    parser.add_argument(
+        "--log-dir", help="log directory location", default="logs", type=str,
```
christianjans, line 156: Maybe make the description start with an uppercase? Something like:
```suggestion
        "--log-dir", help="Log directory location", default="logs", type=str,
```
to match what Alex has in #487.

ultra/train.py
```diff
@@ -146,13 +152,28 @@ def train(task, num_episodes, policy_class, eval_info, timestep_sec, headless, s
...
+    )
     args = parser.parse_args()
 
     num_cpus = max(
         1, psutil.cpu_count(logical=False) - 1
     )  # remove `logical=False` to use all cpus
 
-    policy_class = "ultra.baselines.sac:sac-v0"
+    with open("/home/jenish/Desktop/temp/SMARTS/ultra/agent_pool.json", "r") as f:
```
christianjans, line 164: Oh maybe this hardcoded path to the agent_pool.json file has to be changed?

ultra/train.py
```diff
@@ -146,13 +152,28 @@ def train(task, num_episodes, policy_class, eval_info, timestep_sec, headless, s
...
+        objects = ijson.items(f, "agents")
+        for o in objects:
+            policy_pool = o
```
christianjans, line 167: Oh wait sorry, what does this part do here?

ultra/train.py
```diff
@@ -19,7 +19,8 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
-import os
+import os, sys
+import ijson
```
christianjans, line 23: Maybe reading the JSON for the agents can be done with Python's built-in JSON package? Just to avoid another dependency? I'm not sure, what are your thoughts @JenishPatel99 and @AlexLewandowski ?

ultra/train.py
```diff
@@ -124,6 +124,12 @@ def train(task, num_episodes, policy_class, eval_info, timestep_sec, headless, s
         type=str,
         default="easy",
     )
+    parser.add_argument(
+        "--policy",
+        help="Policies available : [PPO, SAC, TD3, DQN]",
```
christianjans, line 129: Oh perhaps the help argument needs to be changed so that the stuff inside the brackets match available policies? Something like the following maybe?
```suggestion
        help="Policies available: [ppo, sac, ddpg, dqn, bdqn]",
```

christianjans (commented):
Looks great! Just a few comments and questions.

ultra/train.py
JenishPatel99: That sounds good, i thought when we merge it, Alex's changes will overwrite mine, I guess

ultra/train.py
JenishPatel99: Smh, I actually forgot about that lol

ultra/train.py
JenishPatel99: The way ijson extracts the objects, I for some reason wasn't able to do just policy = objects[0], It demanded me to loop once

ultra/train.py
JenishPatel99: I was familiar the functions that were needed to extract those keys from the file using ijson, I was not sure about the main JSON package. I will take a look into that

AlexLewandowski (approved):

christianjans (approved):
Looks good!

Issue #464, opened 2021-01-22
Allow user to specify a initial speed for u-turn task


liamchzh:
https://github.com/huawei-noah/SMARTS/blob/c4e17f403ac82b6e420a18b80b088f32cd233d7c/smarts/sstudio/types.py#L371-L380

Suggestion: add an attribute in `UTurn`, then let the mission planner set the speed limit for OpEn.

PR #516, opened 2021-02-01
Add initial speed uturn

iman512003:
This is to address #464 .

### DIFF
--- a/smarts/core/mission_planner.py
+++ b/smarts/core/mission_planner.py
@@ -212,7 +212,7 @@ class MissionPlanner:
 
         return edge_ids
 
-    def cut_in_waypoints(self, sim, pose: Pose, vehicle, base_waypoint_generator):
+    def cut_in_waypoints(self, sim, pose: Pose, vehicle):
         aggressiveness = self._agent_behavior.aggressiveness or 0
 
         neighborhood_vehicles = sim.neighborhood_vehicles_around_vehicle(
@@ -296,7 +296,7 @@ class MissionPlanner:
             trajectory.append(wp)
         return [trajectory]
 
-    def uturn_waypoints(self, sim, pose: Pose, vehicle, initial_speed):
+    def uturn_waypoints(self, sim, pose: Pose, vehicle):
         # TODO: 1. Need to revisit the approach to calculate the U-Turn trajectory.
         #       2. Wrap this method in a helper.
 
@@ -305,6 +305,9 @@ class MissionPlanner:
         ego_wps = self._waypoints.waypoint_paths_on_lane_at(
             ego_position, ego_lane.getID(), 60
         )
+        if self._mission.task.initial_speed is None:
+            default_speed = ego_wps[0][0].lane_speed
+        default_speed = self._mission.task.initial_speed
         ego_wps_des_speed = []
         for px in range(len(ego_wps[0])):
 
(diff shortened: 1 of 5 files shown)


Gamenot (approved):

smarts/sstudio/types.py
```diff
@@ -372,7 +372,8 @@ class TrapEntryTactic(EntryTactic):
 class UTurn:
     trigger_radius: int = 100
     """This task will be triggered if any vehicles within this radius"""
-
+    initial_speed: float = None
+    """This is the initial speed for the vehilce before it starts u-turn maneuver"""
```
Adaickalavan, line 376:
```suggestion
    """This is the initial speed for the vehicle before it starts u-turn maneuver"""
```

iman512003: Thanks @Adaickalavan ,

smarts/core/mission_planner.py
```diff
@@ -299,12 +300,28 @@ def cut_in_waypoints(self, sim, pose: Pose, vehicle, base_waypoint_generator):
...
+        if self._mission.task.initial_speed is None:
+            default_speed = ego_wps[0][0].speed_limit
+        else:
+            default_speed = self._mission.task.initial_speed
+        ego_wps_des_speed = []
+        for px in range(len(ego_wps[0])):
+            new_wp = replace(ego_wps[0][px], speed_limit=default_speed)
+            ego_wps_des_speed.append(new_wp)
+
+        ego_wps_des_speed = [ego_wps_des_speed]
```
Adaickalavan, line 318: Is this `ego_wps_des_speed` variable meant to have two nested lists `[[ ... ]]` ? Is the outer list necessary?

iman512003: Thanks, Yes, exactly, it should be a list of lists, otherwise it throws an error.

Adaickalavan (approved):

smarts/core/mission_planner.py
```diff
@@ -299,12 +300,28 @@ def cut_in_waypoints(self, sim, pose: Pose, vehicle, base_waypoint_generator):
...
         # TODO: 1. Need to revisit the approach to calculate the U-Turn trajectory.
         #       2. Wrap this method in a helper.
+
+        ego_position = pose.position[:2]
+        ego_lane = self._road_network.nearest_lane(ego_position)
+        ego_wps = self._waypoints.waypoint_paths_on_lane_at(
+            ego_position, ego_lane.getID(), 60
+        )
+        if self._mission.task.initial_speed is None:
+            default_speed = ego_wps[0][0].speed_limit
```
JingfeiPeng, line 310: Is `ego_wps[0][0]` guaranteed to exist? Maybe it is helpful to set a variable here like 
`{description_of_ego_wps[0][0]} = ego_wps[0][0]`

JingfeiPeng (approved):

smarts/core/mission_planner.py
iman512003: Thanks, ego_wps are coming from `waypoints`, where it calls for 
```
closest_linked_wp = self._closest_linked_wp_in_kd_tree_batched(
            [point], self._waypoints_by_lane_id[lane_id], lane_kd_tree, k=1
        )[0][0] 
```
which is the closest linked waypoint, at worst case scenario it will return the beginning or the end of the lane waypoint unless it is off the lanes where in that case the off road will be triggered.

PR #521, opened 2021-02-02
Run CI on pull request from forked repo

Adaickalavan:
When a pull request is made from a forked repo, the CI does not run as it is currently configured to be triggered only on a `push`. Hence, added `pull_request` to the trigger conditions of the CI workflows.

Note: Having both `push` and `pull_request` triggers may cause some duplicated GitHub Actions, but it is a simple solution to prevent erroneous merge commits.


Adaickalavan:
Closed as superseded by #524 .

PR #522, opened 2021-02-02
Update train.py and evaluate.py to include new flags (format fix)

JenishPatel99:


AlexLewandowski:
I think because this pull request was pointed at master, there are some additional files being changed. It might be better to merge #523 instead.

PR #523, opened 2021-02-02
fix format in train.py

JingfeiPeng:
seems like the ci test is failing on develop because of format on train.py.


JenishPatel99:
I think because we had the incorrect version of black which caused the fail test

AlexLewandowski (approved):
Was using the wrong version of black, sorry about that!

Adaickalavan (approved):

JenishPatel99 (approved):

PR #524, opened 2021-02-02
Run CI on pull request from forked repo

Adaickalavan:
When a pull request is made from a forked repo, the CI does not run as it is currently configured to be triggered only on a `push`. Hence, added `pull_request` to the trigger conditions of the CI workflows.


.github/workflows/ci-ultra-tests.yml
```diff
@@ -1,11 +1,10 @@
-name: ULTRA CI
+name: ULTRA CI Base Tests
```
Gamenot, line 1: I think we should keep changes specific to ULTRA separate from those changing SMARTS as a general case.

Gamenot (approved):

iman512003 (approved):

.github/workflows/ci-base-tests.yml
```diff
@@ -1,10 +1,11 @@
 name: SMARTS CI Base Tests
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   test:
     runs-on: ubuntu-18.04
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```
Gamenot, line 8: This is a funny looking solution for the doubled up tests.

Gamenot (approved):

JingfeiPeng (approved):

.github/workflows/ci-base-tests.yml
Adaickalavan: Unfortunately, GitHub Actions does not provide a method to stop duplicated actions. Hence, the next best method would be to include a conditional statement. 

This conditional statement ensures that this test will only be triggered by
+ a `pull_request` that originates from a fork, or
+ a `push`.

Issue #519, opened 2021-02-02
Remove policy flag and policy_class from ultra/evaluate.py


AlexLewandowski:
Currently you need to specify the policy so that the constructed agent matches the model loaded. Ideally, we would only specify the directory to the model being loaded and read policy_class from there.

PR #544, opened 2021-02-03
Update evaluate.py to remove policy flag and add log directory flag

JenishPatel99:
Addressing issue : #519 

Changes made
- Removed the policy flag from evaluate.py to prevent unnecessary policy input
- Added log-dir flag option to evaluate.py to request designated directory for new experiments
- Updated evaluate_test.py to conform to new log-dir flag

### DIFF
--- a/ultra/evaluate.py
+++ b/ultra/evaluate.py
@@ -161,7 +161,7 @@ if __name__ == "__main__":
     )
     parser.add_argument("--models", default="models/", help="Directory to saved models")
     parser.add_argument(
-        "--episodes", help="Number of training episodes", type=int, default=5
+        "--episodes", help="Number of training episodes", type=int, default=200
     )
     parser.add_argument(
         "--timestep", help="Environment timestep (sec)", type=float, default=0.1
--- a/ultra/evaluate.py
+++ b/ultra/evaluate.py
@@ -183,7 +183,8 @@ if __name__ == "__main__":
         os.makedirs(args.log_dir)
 
     m = re.search(
-        "ultra\.[a-zA-Z0-9]+\.[a-zA-Z0-9]+\:[a-zA-Z0-9]+\-[a-zA-Z0-9]+", args.models
+        "ultra(\.)*([a-zA-Z0-9_]*\.)+([a-zA-Z0-9_])+\:[a-zA-Z0-9_]+((\-)*[a-zA-Z0-9_]*)*",
+        args.models,
     )
 
     try:
--- a/ultra/evaluate.py
+++ b/ultra/evaluate.py
@@ -116,7 +116,7 @@ def evaluate(
     summary_log = LogInfo()
     logs = []
 
-    for episode in episodes(num_episodes, etag=policy_class, dir=log_dir):
+    for episode in episodes(num_episodes, etag=policy_class, log_dir=log_dir):
         observations = env.reset()
         state = observations[agent_id]
         dones, infos = {"__all__": False}, None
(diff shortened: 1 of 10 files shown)


Gamenot (approved):

ultra/tests/test_train.py
```diff
@@ -54,37 +54,38 @@ def test_locate_log_directory(self):
...
 
-    # def test_train_single_agent(self):
-    #     if os.path.exists("ultra/tests/logs"):
-    #         shutil.rmtree("ultra/tests/logs")
+    def test_train_single_agent(self):
+        if os.path.exists("ultra/tests/logs"):
+            shutil.rmtree("ultra/tests/logs")
 
-    #     os.system("pkill -9 ray")
+        os.system("pkill -9 ray")
```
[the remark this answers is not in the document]
JenishPatel99: I was wondering if we are going to replace ray with another distributed solution? #549

Apart from killing the ray processes, I tried to ray.shutdown() prior to new ray executions (ray.init()). This eliminated the ray leakage. I felt like it was worth sharing

ultra/evaluate.py
```diff
@@ -156,15 +159,9 @@ def evaluate(
...
-    parser.add_argument(
-        "--policy",
-        help="Policies available : [ppo, sac, ddpg, dqn, bdqn]",
-        type=str,
-        default="sac",
-    )
     parser.add_argument("--models", default="models/", help="Directory to saved models")
     parser.add_argument(
-        "--episodes", help="Number of training episodes", type=int, default=200
+        "--episodes", help="Number of training episodes", type=int, default=5
```
AlexLewandowski, line 164: I think we can leave this the same for now. If not, we should update the README as well.

AlexLewandowski (approved):

ultra/tests/scenarios/task00/config.yaml
```diff
@@ -28,4 +28,42 @@ levels:
...
+          specs: [[50kmh,low-density,0.21],[70kmh,low-density,0.20],[100kmh,low-density,0.20], #61%
+                  [50kmh,mid-density,0.11],[70kmh,mid-density,0.11],[100kmh,mid-density,0.11], #33%
+                  [50kmh,no-traffic,0.01],[70kmh,no-traffic,0.01],[100kmh,no-traffic,0.01], #3%
+                  [50kmh,high-density,0.01],[70kmh,high-density,0.01],[100kmh,high-density,0.01]] # 3%
+        3lane_c:
+          percent: 0.5
+          specs: [[50kmh,low-density,0.21],[70kmh,low-density,0.20],[100kmh,low-density,0.20], #61%
+                  [50kmh,mid-density,0.11],[70kmh,mid-density,0.11],[100kmh,mid-density,0.11], #33%
+                  [50kmh,no-traffic,0.01],[70kmh,no-traffic,0.01],[100kmh,no-traffic,0.01], #3%
+                  [50kmh,high-density,0.01],[70kmh,high-density,0.01],[100kmh,high-density,0.01]] # 3%
```
AlexLewandowski, line 69: Is eval_test needed for the additional specs or something else?

JenishPatel99: So yeah, what I was encountering when using simple scenarios (--task 00 --level easy) was that the evaluation process was really slow. I decided to add additional diverse scenarios, which then allowed to decreases the number of steps per episode. The eval test is merely used for test_evaluation.py

ultra/evaluate.py
JenishPatel99: oops, I forgot to revert it back to its original. value I will do it right now

Issue #341, opened 2020-12-19
SMARTS Will Not Try to Reconnect to Envision


Gamenot:
After running a SMARTS instance with envision playback, SMARTS will not reconnect to Envision if Envision goes down.

Steps to reproduce:  
- `console_1$ scl envision start`  
- `console_2$ python examples/single_agent.py scenarios/loop`  
- Confirm example is displayed in envision by navigating to `localhost:8081` in a browser.  
- `console_1$ ^C`
- `console_1$ scl envision start`
- Refreshing the `localhost:8081` will no longer update from the `console_2` simulation.

AC:
- [ ] SMARTS will attempt to connect to an available Envision at a certain interval even if disconnected from Envision.

PR #545, opened 2021-02-03
SMARTS_reconnect_envision_when_disconnected (#341)

JingfeiPeng:
After envision is disconnected, SMARTS will forever try to reconnect envision every 5 seconds. 
Envision will check if there are running simulation every 3 seconds.


envision/web/src/components/app.js
```diff
@@ -55,16 +55,19 @@ function App({ client }) {
...
         }
       }
-
       setSimulationIds(ids);
-    })();
+    }
+
+    // checks if there is new simulation running every 3 seconds.
+    const interval = setInterval(fetchRunningSim, 3000);
+    return () => clearInterval(interval)
```
JingfeiPeng, line 70: Suppose we run envision and navigate to localhost:8081, and then run scenario. If we only change in client.py, envision needs to be refreshed to see the running simulation. This makes envision if there are any running simulation every 3 seconds.

envision/web/src/components/app.js
```diff
@@ -55,16 +55,19 @@ function App({ client }) {
   const matchedSimulationId = routeMatch ? routeMatch.params.simulation : null;
 
   useEffect(() => {
-    (async () => {
+    const fetchRunningSim = async () => {
       let ids = await client.fetchSimulationIds();
       if (ids.length > 0) {
         if (!matchedSimulationId || !ids.includes(matchedSimulationId)) {
-          history.push(`/${ids[0]}`);
+          history.push(`/${ids[ids.length-1]}`);
```
JingfeiPeng, line 62: a small improvement: if there are mutiple scenario running. envision displays the first one which is oldest one. choosing the last one displays the latest one.

envision/web/src/components/app.js
```diff
@@ -55,16 +55,19 @@ function App({ client }) {
...
         }
       }
-
       setSimulationIds(ids);
-    })();
+    };
+
+    // checks if there is new simulation running every 3 seconds.
+    const interval = setInterval(fetchRunningSim, 3000);
+    return () => clearInterval(interval);
```
Gamenot, line 70: OK, this should work.

envision/client.py
```diff
@@ -211,16 +201,18 @@ def run_socket(endpoint, num_retries, wait_between_retries, threadlocal):
...
 
                 if not connection_established:
-                    self._log.info(
-                        f"Attempting to connect to Envision tries={tries}/{num_retries}"
-                    )
+                    self._log.info(f"Attempting to connect to Envision tries={tries}")
+                else:
+                    # when connection closed, retry again every 5 seconds
+                    wait_between_retries = 5
+                    tries = 0
```
Gamenot, line 208: I do not think resetting `tries` is needed. Or if you are resetting `tries` you will need to flip `connection_established` to false again.

```suggestion
                    self._log.info(f"Connection to Envision lost. Attempting to reconnect.")
```
```suggestion
                    tries = 0
                    self._log.info(f"Connection to Envision lost. Attempting to reconnect.")
                    connection_established = false
                    setattr(threadlocal, "connection_established", connection_established)
```

Gamenot (approved):

envision/client.py
Adaickalavan, line 208: Here, I think don't need to reset `tries=0` since number of `tries` is not being checked in the `while` loop.

There might be conflict between #567 and this pull request. I suggest to merge #567 and #545 together first, before merging into the `develop` branch.

Adaickalavan:
There might be conflict between #567 and this pull request. I suggest to merge #567 and #545 together first, before merging into the `develop` branch.

envision/client.py
JingfeiPeng: The `wait_between_retries` is from `_connect`, the default there is 0.5 seconds, which is pretty fast. This is good for trying to establish the initial connection. But I guess when connection is lost, we don't have to try reconnect that frequently. So I set it to 5s.

Adaickalavan (approved):

Gamenot:
This can be merged.

PR #560, opened 2021-02-09
Reset initial vehicle speed in UTurn

iman512003:
This is to reset the vehicle initial speed to the desired value which is specified for uturn scenarios.

### DIFF
--- a/smarts/core/vehicle.py
+++ b/smarts/core/vehicle.py
@@ -45,6 +45,7 @@ from .sensors import (
     WaypointsSensor,
 )
 from .utils.math import rotate_around_point
+from smarts.sstudio.types import UTurn
 from functools import lru_cache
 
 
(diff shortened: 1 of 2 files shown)


JingfeiPeng (approved):

smarts/core/vehicle.py
```diff
@@ -312,6 +313,10 @@ def build_agent_vehicle(
         vehicle_type = "passenger"
         chassis_dims = VEHICLE_CONFIGS[vehicle_type].dimensions
 
+        if isinstance(mission_planner.mission.task, UTurn):
+            if mission_planner.mission.task.initial_speed:
+                initial_speed = mission_planner.mission.task.initial_speed
```
Gamenot, line 318: OK, this is fine. A new issue needs to be brought up around the consideration of generalising this initial speed parameter by acquiring it from the mission itself rather than from the task.

Gamenot (approved):

smarts/core/vehicle.py
iman512003: Thanks @Gamenot , I created #563 .

PR #565, opened 2021-02-10
Bugtest SUMO Crashing on Turns

Gamenot:
This one seems to occur in just the second episode of loop. Replicate with just a call to `supervisord` and Eclipse SUMO sumo Version 1.8.0.

### DIFF
--- a/examples/single_agent.py
+++ b/examples/single_agent.py
@@ -21,7 +21,7 @@ class ChaseViaPointsAgent(Agent):
             len(obs.via_data.near_via_points) < 1
             or obs.ego_vehicle_state.edge_id != obs.via_data.near_via_points[0].edge_id
         ):
-            return (obs.waypoint_paths[0][0].speed_limit, 0)
+            return (obs.waypoint_paths[0][0].speed_limit -5, 0)
 
         nearest = obs.via_data.near_via_points[0]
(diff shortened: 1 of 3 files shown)


Gamenot:
@Adaickalavan I found out one of the reasons why we could not reproduce the sumo crash on docker.

For some reason Docker is installing a different version of smarts in the image.

Gamenot:
I have put up an image that should be able to reproduce the problem https://hub.docker.com/layers/huaweinoah/smarts/bugtest_0.4.13-sumo_1_8_0/images/sha256-51025124ec25309e923893c59c569939d788983fd76d7e5b3bac116d9c3dddcb?context=repo .

Adaickalavan:
I ran the given commands on the docker image.

Test outcomes are as follows. Each test was started afresh by restarting the docker and re-running all commands.

| Test # | Outcome |
| :----: | ---- |
| 1 | Tracii connection error in 3rd episode |
| 2 | Finished all 10 episodes without any errors |
| 3 | Finished all 10 episodes without any errors |
| 4 | Finished all 10 episodes without any errors |

Perhaps, the docker image should be tested multiple number of times by more people @iman512003 @JingfeiPeng.

Gamenot:
Going to try using #612 to get this to become fully reproducible.

Gamenot:
#619 is looking at this issue more in depth.

PR #567, opened 2021-02-10
Parallel process to send state to Envision server

### LINKED ISSUE #283: Envision performance tuning
Envision cost seems to be about 40 fps on a regularly 100 fps run. That should turn out to be about 7 milliseconds.

We want to ideally cut that down to 3 or less.

Adaickalavan:
A parallel process in `./envision/client.py`, instead of a thread, is used to send the state to the Envision server.

Commands used
```python
# In terminal A
scl envision start -s ./scenarios -p 8081

# In terminal B - headless 
python3.7 ./examples/single_agent.py ./scenarios/loop/  --episodes 10 --headless

# In terminal B - with Envision
python3.7 ./examples/single_agent.py ./scenarios/loop/  --episodes 10

# In terminal B - with Envision and with logging
# set envision_record_data_replay_path="./data_replay" in gym.make() function in single_agent.py
python3.7 ./examples/single_agent.py ./scenarios/loop/  --episodes 10

# In terminal B - test data replay
scl scenario replay -d ./data_replay
```

Python profiler was used to time the main function.

| Case | Multithreaded | Multiprocessing |
| ----- | ----- | ----- |
| Headless | 174 s | 174 s |
| Envision | 219 s | 184 s |
| Envision and Logging | 270 s | 193 s |

Saved ~3.5s per episode for Envision and ~3.5s per episode for Logging.

Closes #283.


examples/single_agent.py
```diff
@@ -52,7 +52,7 @@ def main(scenarios, sim_name, headless, num_episodes, seed, max_episode_steps=No
         sumo_headless=True,
         seed=seed,
         # zoo_addrs=[("10.193.241.236", 7432)], # Sample server address (ip, port), to distribute social agents in remote server.
-        # envision_record_data_replay_path="./data_replay",
+        # envision_record_data_replay_path="./.data_replay",
```
JingfeiPeng, line 55: is this needed?

JingfeiPeng (approved):

examples/single_agent.py
Adaickalavan: Renamed the dot folder.

envision/client.py
```diff
@@ -1,39 +1,43 @@
...
-import re
-import time
-import uuid
+
 import json
 import logging
+import multiprocessing
+import numpy as np
+import re
 import threading
```
Gamenot, line 28: I think `threading` is not used anymore.

envision/client.py
```diff
@@ -181,19 +187,19 @@ def on_error(ws, error):
             self._log.error(f"Connection to Envision terminated with: {error}")
 
         def on_open(ws):
-            setattr(threadlocal, "connection_established", True)
+            nonlocal connection_established
+            connection_established = True
```
Gamenot, line 191: OK, yes, this makes more sense than using meta-programming.

Gamenot (approved):

envision/client.py
Adaickalavan: Good catch. I have removed the unnecessary `import threading` statement.

PR #570, opened 2021-02-10
Add multi-agent functionality to ULTRA episodes function

christianjans:
Currently, the Episode object used to record agent data during training and evaluation only supports the recording of a single agent's data.

This change will allow the Episode object to record the training and evaluation data of multiple agents.


AlexLewandowski:
Would it be more convenient to have a default AGENT_ID arg in episodes, like 007, so that we won't have to change every instance of episodes?

christianjans:
Yeah, good point! I just had a thought that would get rid of agent_ids. These new commits allow whatever is using episodes to track whichever agent without having to specify an agent_ids parameter in episodes(...).

JenishPatel99 (approved):

ultra/ultra/utils/episode.py
```diff
@@ -60,11 +53,10 @@ def __init__(self):
             "off_route": 0,
             "reached_goal": 0,
             "timed_out": 0,
-            "episode_length": 0,
+            "episode_length": 1,
```
AlexLewandowski, line 56: Just wondering why this needed to be changed from 0 to 1?

AlexLewandowski (approved):

ultra/ultra/utils/episode.py
christianjans: Oh yeah for sure, originally this episode length data was updated with the total number of steps taken in the episode once the episode ended. This total number of steps started at 1 and incremented each time a step was taken when the agent was present in the scenario:
https://github.com/huawei-noah/SMARTS/blob/feee8fd8a1f0c10ab2aaf6f12acc8c9cc0f861af/ultra/utils/episode.py#L137
https://github.com/huawei-noah/SMARTS/blob/feee8fd8a1f0c10ab2aaf6f12acc8c9cc0f861af/ultra/utils/episode.py#L202
This worked when there was just one agent, but now that there could potentially be multiple agents in the same episode (each agent potentially taking a different number of steps), this episode length data needs to be updated for each agent. This is what the step() method does: https://github.com/huawei-noah/SMARTS/blob/57dafec07612a18bce6ad3aeb497a6e087d32467/ultra/ultra/utils/episode.py#L93-L94
But in order to achieve the same data as the original episode.py recorded, we have to start the initial step count at 1 so that when we normalize by the step count...
https://github.com/huawei-noah/SMARTS/blob/57dafec07612a18bce6ad3aeb497a6e087d32467/ultra/ultra/utils/episode.py#L96-L105
we get the same numbers in the Tensorboard as we did before this change.

PR #571, opened 2021-02-10
Create Uturn cutin example scenarios

JingfeiPeng:
U turn scenario video
https://user-images.githubusercontent.com/36685292/107585848-eb4b1980-6bcc-11eb-86e9-59d88a71d64e.mp4

Cut in scenario video (Updated with new lane length of 250

https://user-images.githubusercontent.com/36685292/107602734-92917600-6bf8-11eb-9b08-8e9aebc963b4.mp4

Used part of the map in 4lane


scenarios/cutin/map.net.xml
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Gamenot, line 1: I think it might be useful to make the map a little bit longer.

smarts/core/mission_planner.py
```diff
@@ -423,6 +424,10 @@ def uturn_waypoints(self, sim, pose: Pose, vehicle):
                 speed_limit = neighborhood_vehicles[0].speed
 
         self._task_is_triggered = True
+        # uturn_activated_distance = math.sqrt((neighborhood_vehicles[0].pose.position[0]-ego_position[0])**2
+        #     +(neighborhood_vehicles[0].pose.position[1]-ego_position[1])**2
+        # )
+        # print(f"uturn activated at distance: {uturn_activated_distance}")
```
Gamenot, line 430: Is this needed or just for debug?

scenarios/cutin/scenario.py
```diff
@@ -0,0 +1,41 @@
+import os
```
Gamenot, line 1: I could suggest putting all the BM scenarios in a subfolder like "scenarios/behavior_models/<task>".

smarts/core/mission_planner.py
JingfeiPeng: Actually both, it would be beneficial to log the distance between the u-turn vehicle and neighbor vehicle when u-turn is activated. I will use a logger for this. However

```
        uturn_activated_distance = math.sqrt((neighborhood_vehicles[0].pose.position[0]-ego_position[0])**2
            +(neighborhood_vehicles[0].pose.position[1]-ego_position[1])**2
        )
        print(f"uturn activated at distance: {uturn_activated_distance}")

JingfeiPeng:
I tried also changing the uturn lane length to 250 then a strange behavior occurred... @iman512003 

https://user-images.githubusercontent.com/36685292/107599791-bc926a80-6bef-11eb-9201-6911af44ca82.mp4

Gamenot (approved):

iman512003 (approved):

Issue #572, opened 2021-02-11
Termination conditions for ULTRA are too conservative


AlexLewandowski:
Evaluation can take a long time because termination conditions are too conservative. This is particular to evaluation, because the agent picks the greedy action and does not explore. Often the car will stop and not move, or drive past the intersection and stay in the other lane blocking traffic. This prevents collisions and the episode ends due to a time out.

Possible termination conditions:

- [x] reduce max_num_steps to 200
- [ ] on blocking traffic for too long (i.e. traffic violations?)
- [ ] when the intersection is passed, or the turn is no longer possible.
- [ ] when the score is sufficiently negative.

AlexLewandowski:
For now, try lowering max_episode_steps to ~200

AlexLewandowski:
max_steps seems to be sufficient, closing for now.

PR #584, opened 2021-02-16
Add max-episode-steps flag and use in tests

AlexLewandowski:
The ULTRA CI tests can be slow due to the high number of steps. Control over the max number of steps is also good to have for both training and evaluation in general. Somewhat related to termination conditions issue #572.

Also includes some path fixes for config.yaml.


Gamenot (approved):
LGTM

JenishPatel99 (approved):

christianjans (approved):

PR #591, opened 2021-02-17
Add RLlib and multi-agent functionality to ULTRA

christianjans:
### Desecription
Allows for multi-agent training in ULTRA. Tasks can now be defined with multiple ego missions in their config.yaml. The scenarios of the task are then built with these missions, allowing for multiple ego agents to entertain the scenario in the same episode and potentially at the same time.

Note that the number of missions defined in the task should be equal to the number of agents performing the task.

### Considerations
Some things to consider/complete before merging are:
- [x] Use multi-agent version of ULTRA episodes
- [x] Take care of offset error on maps
  - Potentially requires adding more options for the config.yaml mission specifications (e.g. start time, route lane, etc.)
  ```sh
  2021-02-17 15:19:57,888 ERROR worker.py:1012 -- Possible unhandled error from worker: ray::__main__.train() (pid=20701, ip=10.208.237.111)
  File "python/ray/_raylet.pyx", line 445, in ray._raylet.execute_task
  File "ultra/train.py", line 79, in train
    observations = env.reset()
  File "/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/env/hiway_env.py", line 176, in reset
    scenario = next(self._scenarios_iterator)
  File "/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/core/scenario.py", line 223, in variations_for_all_scenario_roots
    scenario_root, agents_to_be_briefed
  File "/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/core/scenario.py", line 314, in discover_agent_missions
    for actor_and_mission in missions
  File "/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/core/scenario.py", line 314, in <listcomp>
    for actor_and_mission in missions
  File "/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/core/scenario.py", line 551, in _extract_mission
    position, _ = to_position_and_heading(*mission.route.end, road_network,)
  File "/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/core/scenario.py", line 539, in to_position_and_heading
    lane_vector = road_network.lane_vector_at_offset(lane, offset)
  File "/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/core/sumo_road_network.py", line 227, in lane_vector_at_offset
    f"Offset={end_offset} goes out further than the end of "
  ValueError: Offset=143.0 goes out further than the end of lane=(edge-north-SN_0, length=136.25)
  ```
- [x] Figure out why mission start times are so far apart when missions are different
  - Potentially requires adding more options for the config.yaml mission specifications (e.g. start time, route lane, etc.)
- [x] Change the config.yaml of other ULTRA tasks and in the ULTRA tests
- [x] Decide what to do about evaluation... Perhaps the evaluation check condition needs to be changed to support multiple agents.
  - Should each evaluate separately and only against social agents? Or
  - Should other ego agents participate in the evaluation?
- [x] Change train to accept a variable amount of agent policies (probably just through a command-line flag)
- [x] Add multiagent tests to ULTRA CI tests
- [x] Option to specify custom agent ids (ensure to assert the agent ids and policy classes have the same length)
- [x] Change the etag in the train episodes
- [x] Refactor the agent_classes, agent_specs, and agents dictionaries in train.py

### Concerns
- Is there any reason that the `wait_to_hijack_limit_s` is set to 1000? https://github.com/huawei-noah/SMARTS/blob/feee8fd8a1f0c10ab2aaf6f12acc8c9cc0f861af/ultra/scenarios/common/distributions.py#L160 Can this be changed?
- Is there any reason that the beginning and end of the ego route have different ranges for their random offset? https://github.com/huawei-noah/SMARTS/blob/feee8fd8a1f0c10ab2aaf6f12acc8c9cc0f861af/ultra/scenarios/generate_scenarios.py#L149-L152 Can these be changed?


AlexLewandowski:
> * Is there any reason that the beginning and end of the ego route have different ranges for their random offset? https://github.com/huawei-noah/SMARTS/blob/feee8fd8a1f0c10ab2aaf6f12acc8c9cc0f861af/ultra/scenarios/generate_scenarios.py#L149-L152
>    Can these be changed?

I do think these can be changed. We didn't have time to discuss this during the monday meeting, but I think starting the vehicles at the intersection might be more interesting for the multiagent setting. Multiagent interaction in ULTRA happens mostly at the intersection, so we may as well fast-track the multi-agent interaction.

AlexLewandowski:
You might need to merge develop to fix the tests?

christianjans:
> You might need to merge develop to fix the tests?

Oh good call, I will try this.

ultra/ultra/baselines/agent_spec.py
```diff
@@ -60,11 +60,16 @@ def __new__(
...
         if experiment_dir:
-            print(f"LOADING SPEC from {experiment_dir}/spec.pkl")
-            with open(f"{experiment_dir}/spec.pkl", "rb") as input:
-                spec = dill.load(input)
+            print(
+                f"Loading spec for {agent_id} from {experiment_dir}/agent_metadata.pkl"
+            )
+            with open(f"{experiment_dir}/agent_metadata.pkl", "rb") as metadata_file:
+                agent_metadata = dill.load(metadata_file)
+                spec = agent_metadata["agent_specs"][agent_id]
```
JenishPatel99, line 71: I think we discussed this in the ultra meeting, but just affirming that the older experiment files would not be compatible with the new multi-agent structure and how spec is pickled

christianjans: Yes exactly. If one had an experiment that they had run with ULTRA < 0.2 and wanted to perform evaluation on it (python ultra/evaluate.py ...), then it would not work with this new version. This new version's evaluate.py looks for an agent_metadata.pkl file in the experiment directory that contains everything the old spec.pkl did, but also more information (such as the policy class string associated with the agent(s) in the experiment) that was not included, but is helpful.

JenishPatel99: Do we think we should mention this in the documentation or comment it in the evaluate file? I do not know how many people are using ultra at this moment but I dont think it is alot, but nevertheless a warning of some type would be great

christianjans: Oh yah good call. I will look into updating the documentation and then probably add it there.

ultra/tests/task/test_multiagent_mock_2lane_t_70kmh_p-test-flow-1/map.net.xml
```diff
@@ -0,0 +1,135 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
AlexLewandowski: (applies to all xml files, glb, pkl?)

ultra/ultra/evaluate.py
```diff
@@ -58,24 +54,37 @@ def evaluation_check(
     headless,
     log_dir,
 ):
-    agent_itr = episode.get_itr(agent_id)
+    agent_iterations = {agent_id: episode.get_itr(agent_id) for agent_id in agent_ids}
+    lowest_agent_iteration = min(agent_iterations.values())
```
AlexLewandowski, line 58: I think this can be simplified when we change to episode based evaluation. However, why is lowest agent iteration used?

ultra/ultra/scenarios/common/distributions.py
```diff
@@ -669,7 +669,7 @@
         },
         "ego_hijacking_params": {
             "zone_range": [5, 10],
-            "wait_to_hijack_limit_s": 1000,
+            "wait_to_hijack_limit_s": 10,
```
AlexLewandowski, line 672: How was the value of 10 chosen here?

ultra/tests/task/test_multiagent_mock_2lane_t_70kmh_p-test-flow-1/map.net.xml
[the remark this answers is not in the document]
christianjans: I think for now it would be best to keep them. They are not generated before certain CI tests and thus must be shipped with ULTRA (see the current scenarios used for single agent CI tests: https://github.com/huawei-noah/SMARTS/tree/master/ultra/tests/task). Perhaps we can change the tests in the future to generate these on the fly so that they do not have to be shipped with ULTRA.

.github/workflows/ci-ultra-tests.yml
```diff
@@ -32,7 +32,8 @@ jobs:
           cd ultra
           . .venv/bin/activate
           scl scenario build-all ultra/scenarios/pool
-          pytest -v ./tests/
+          pytest -v ./tests/ \
+          --ignore=./tests/test_ultra_package.py \
```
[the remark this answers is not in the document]
christianjans: Good call, it is probably no longer necessary. I will remove this change and see if the test passes.

ultra/docs/multiagent.md
```diff
@@ -0,0 +1,76 @@
...
+   levels:
+     no-traffic:  # This is the level name.
+       train:
+         total: 10  # There will be 10 training scenarios.
+         intersection_types:
+           2lane_c:
+             percent: 1.0  # 100% of these scenarios will be 2 lane, c-intersections.
+             specs: [[50kmh,no-traffic,0.34],[70kmh,no-traffic,0.33],[100kmh,no-traffic,0.33]]
+       test:
+         total: 2  # There will be 10 testing scenarios.
```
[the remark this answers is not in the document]
christianjans: Oh awesome good catch, thanks!

ultra/ultra/evaluate.py
christianjans: Good question, yeah this can definitely be simplified when we move to episode-based evaluation. But just in the interim, I thought that it would be best to evaluate once we know all agents have reached the desired eval_rate threshold before we evaluate all the agents together in a test scenario (rather than having just one, or some, agent reach the threshold and then evaluating all the agents in the test scenario). Let me know your thoughts on this or if you have a better idea.

ultra/ultra/scenarios/common/distributions.py
christianjans: Yeah good question. So we still want to allow for the opportunity to hijack a social vehicle if the opportunity becomes available. Therefore, setting this value to something like 0 would not be good. However, if the value is too large and the scenario has little social vehicle traffic, it will wait too long looking for a social vehicle to hijack, delaying the spawn of the ego vehicle. By lowering the value to 10, it allows for the ability of an ego vehicle to still hijack a social vehicle if one is available, but will also not wait more than 10 time steps before just spawning a new vehicle for the ego vehicle rather than hijack an existing vehicle. Let me know if I can explain this better or if there are any more questions.

christianjans:
Hi @AlexLewandowski and @JenishPatel99 I just wanted to draw your attention to the config.yaml files of the tasks. Let me know what you think of the ego_missions changes. This allows for multi-agent training with single agent evaluation and may also be good for future extensibility in ULTRA task definitions, but please let me know your honest thoughts or if you recognize any other ways to accomplish this.

JenishPatel99 (approved):

AlexLewandowski (approved):

JenishPatel99 (approved):

PR #596, opened 2021-02-17
Speed up imitation learning jump to timestamp

JingfeiPeng:
Original issue was that during imitation learning, smarts had to step through all the time histories before the ego car appears. This is obvious when envision has large amount of time of no activity. 

Fixed this issue by adding a `start_time_offset` in traffic_history_provider, and exposed an api in smarts for `history_vehicles_replacement_for_imitation_learning.py` to set this start_time_offset. Traffic_history_provider will only fetch histories from time `timestamp+start_time_offset`. 

Related previous PR: #163

### DIFF
--- a/examples/history_vehicles_replacement_for_imitation_learning.py
+++ b/examples/history_vehicles_replacement_for_imitation_learning.py
@@ -1,5 +1,11 @@
 import logging
 
+from smarts.core.smarts import SMARTS
+from smarts.core.agent_interface import AgentInterface, AgentType
+from smarts.core.agent import AgentSpec, Agent
+from smarts.core.sumo_traffic_simulation import SumoTrafficSimulation
+from smarts.core.scenario import Scenario
+from smarts.core.scenario import Mission
 from envision.client import Client as Envision
 from examples import default_argument_parser
 from smarts.core.agent import Agent, AgentSpec
@@ -23,14 +29,11 @@ def main(scenarios, headless, seed):
         traffic_sim=SumoTrafficSimulation(headless=True, auto_start=True),
         envision=Envision(),
     )
-
     for _ in scenarios:
         scenario = next(scenarios_iterator)
         agent_missions = scenario.discover_missions_of_traffic_histories()
 
         for agent_id, mission in agent_missions.items():
-            scenario.set_ego_missions({agent_id: mission})
-
             agent_spec = AgentSpec(
                 interface=AgentInterface.from_type(
                     AgentType.Laner, max_episode_steps=None
@@ -40,7 +43,13 @@ def main(scenarios, headless, seed):
             agent = agent_spec.build_agent()
 
             smarts.switch_ego_agent({agent_id: agent_spec.interface})
-
+            smarts.history_set_start_elapsed_time(mission.start_time)
+            modified_mission = Mission(
+                goal=agent_missions[agent_id].goal,
+                start_time=0.0,
+                start=agent_missions[agent_id].start,
+            )
+            scenario.set_ego_missions({agent_id: modified_mission})
             observations = smarts.reset(scenario)
 
             dones = {agent_id: False}
--- a/examples/history_vehicles_replacement_for_imitation_learning.py
+++ b/examples/history_vehicles_replacement_for_imitation_learning.py
@@ -1,4 +1,5 @@
 import logging
+from dataclasses import replace
 
 from smarts.core.smarts import SMARTS
 from smarts.core.agent_interface import AgentInterface, AgentType
@@ -44,11 +45,7 @@ def main(scenarios, headless, seed):
 
             smarts.switch_ego_agent({agent_id: agent_spec.interface})
             smarts.history_set_start_elapsed_time(mission.start_time)
-            modified_mission = Mission(
-                goal=agent_missions[agent_id].goal,
-                start_time=0.0,
-                start=agent_missions[agent_id].start,
-            )
+            modified_mission = replace(mission, start_time=0.0)
             scenario.set_ego_missions({agent_id: modified_mission})
             observations = smarts.reset(scenario)
 
--- a/examples/history_vehicles_replacement_for_imitation_learning.py
+++ b/examples/history_vehicles_replacement_for_imitation_learning.py
@@ -40,7 +40,9 @@ def main(scenarios, headless, seed):
 
             smarts.switch_ego_agent({agent_id: agent_spec.interface})
             # required: get traffic_history_provider and set time offset
-            traffic_history_provider = smarts.get_provider_by_type(TrafficHistoryProvider)
+            traffic_history_provider = smarts.get_provider_by_type(
+                TrafficHistoryProvider
+            )
             assert traffic_history_provider
             traffic_history_provider.set_start_time(mission.start_time)
 
--- a/examples/history_vehicles_replacement_for_imitation_learning.py
+++ b/examples/history_vehicles_replacement_for_imitation_learning.py
@@ -8,6 +8,7 @@ from smarts.core.agent_interface import AgentInterface, AgentType
 from smarts.core.scenario import Mission, Scenario
 from smarts.core.smarts import SMARTS
 from smarts.core.sumo_traffic_simulation import SumoTrafficSimulation
+from smarts.core.traffic_history_provider import TrafficHistoryProvider
 
 logging.basicConfig(level=logging.INFO)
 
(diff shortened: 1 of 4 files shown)


smarts/core/scenario.py
```diff
@@ -116,7 +116,7 @@ class Via:
     required_speed: float
 
 
-@dataclass(frozen=True)
```
JingfeiPeng, line 119: Had to set the `mission.start_time` to 0 otherwise smarts.reset will still step through time_steps until original mission.start_time

smarts/core/sumo_traffic_simulation.py
```diff
@@ -330,9 +330,6 @@ def step(self, provider_actions, dt, elapsed_sim_time) -> ProviderState:
         """
         Args:
             dt: time (in seconds) to simulate during this simulation step
-            managed_vehicles: dict of {vehicle_id: (x, y, heading)}
```
JingfeiPeng, line 333: noticed this arg actually doesn't exist here so removed the doc

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -42,7 +40,13 @@ def main(scenarios, headless, seed):
             agent = agent_spec.build_agent()
 
             smarts.switch_ego_agent({agent_id: agent_spec.interface})
-
+            smarts.history_set_start_elapsed_time(mission.start_time)
+            modified_mission = Mission(
+                goal=agent_missions[agent_id].goal,
+                start_time=.0,
```
JingfeiPeng, line 46: @Gamenot The default start_time specified in Mission class is 0.1, I tried using 0 or 0.1 here and both works. Is there a reason why the default start_time in Mission class is 0.1?

```
@dataclass(frozen=True)
class Mission:
    start: Start
    goal: Goal
    # An optional list of edge IDs between the start and end goal that we want to
    # ensure the mission includes
    route_vias: Tuple[str] = field(default_factory=tuple)
    start_time: float = 0.1
    entry_tactic: EntryTactic = None
    task: Tuple[CutIn, UTurn] = None
    via: Tuple[Via, ...] = ()

    @property
    def has_fixed_route(self):
        return not self.goal.is_endless()

    def is_complete(self, vehicle, distance_travelled):
        return self.goal.is_reached(vehicle)
```

JingfeiPeng:
Speed boost:
Before this change, the time for finishing `python examples/history_vehicles_replacement_for_imitation_learning.py ~/datasets/imitation-learning/scenarios/interaction_dataset_merging` is around 1.5 hours long. 
After this change, the time spent is around 20~30minutes.

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -42,7 +40,13 @@ def main(scenarios, headless, seed):
             agent = agent_spec.build_agent()
 
             smarts.switch_ego_agent({agent_id: agent_spec.interface})
-
+            smarts.history_set_start_elapsed_time(mission.start_time)
+            modified_mission = Mission(
+                goal=agent_missions[agent_id].goal,
+                start_time=0.0,
+                start=agent_missions[agent_id].start,
+            )
```
Gamenot, line 48: I would suggest using `dataclasses.replace()`.

```suggestion
            modified_mission = replace(mission, start_time=0.0)
```

smarts/core/traffic_history_provider.py
```diff
@@ -31,6 +31,10 @@ def __init__(self):
         self._is_setup = False
         self._current_traffic_history = None
         self.replaced_vehicle_ids = set()
+        self.start_time_offset = 0
+
+    def set_start_time(self, start_time: float):
+        self.start_time_offset = start_time
```
Gamenot, line 37: A validation to make sure this is positive would be useful.

smarts/core/smarts.py
```diff
@@ -478,6 +478,11 @@ def attach_sensors_to_vehicles(self, agent_spec, vehicle_ids):
     def observe_from(self, vehicle_ids):
         return self._agent_manager.observe_from(self, vehicle_ids)
 
+    def history_set_start_elapsed_time(self, start_time: float):
```
Gamenot, line 481: I am hesitant about this change since it specifically applies to the traffic history provider. Ideally I would want this to be a general `move_to_time()`.

smarts/core/traffic_history_provider.py
```diff
@@ -69,10 +73,12 @@ def step(self, provider_actions, dt, elapsed_sim_time) -> ProviderState:
         if (
             not self._current_traffic_history
             or timestamp is None
-            or str(timestamp) not in self._current_traffic_history
+            or str(round(timestamp + self.start_time_offset, 1))
```
Gamenot, line 76: I would suggest moving `time_with_offset` before this if statement and using it in place of the rounded value here.

Gamenot (changes requested):
Some comments.

smarts/core/traffic_history_provider.py
JingfeiPeng: the timestamp `or timestamp is None` check here might be `None` which causes an error when adding None with self.start_time_offset

JingfeiPeng: I guess I could move this check `str(round(timestamp + self.start_time_offset, 1)) not in self._current_traffic_history` to be after `time_with_offset` definition?

smarts/core/smarts.py
JingfeiPeng: Liam's original PR on this issue tried to do that actually [on_this_line](https://github.com/huawei-noah/SMARTS/pull/163/files#diff-52e8b6282acc13c9a94300f00440b8a71ccf60c8c810becb246659aa297aa40aR287)
But JVillella mentioned: 
```
I don't think this is the right design. The reset(..., start_time=...) isn't actually doing what the API makes it seem. One would think it's offsetting the sim by the given start_time, but really just changing the internal clock.

This looks to be a one-off solution for the history data replay work and not useful to anything else (or applicable to anything else). Can we just offset the timestamps of the data replay data directly / modify the data replay input?
```
This seems to saying that move_to_time isn't useful to anything else besides history replay. The conclusion from the original PR was to:
1. the history provider needs to know the offset
2. mission starts at time 0 sec
which is done on this PR

examples/history_vehicles_replacement_for_imitation_learning.py
Gamenot: The reason was a thought to give a step before adding the vehicle but with that case the default should probably be a lower value like `1e-5`.

smarts/core/smarts.py
JingfeiPeng: Another difficulty with setting the `self._elapsed_sim_time` is that it has to be done in `def reset` like the original PR. 

Reasons:
The goal is to reset self._elapsed_sim_time to be the time when the agent occurs. There are two hypothetical options to do this and avoid doing it in `def reset`:
1: we call `move_to_time()` after `reset()`, but this doesn't work since `reset` will keep stepping until the agent appears. Calling `move_to_time()` will just set elapsed_sim_time to the same value. There is no performance gain from doing this since reset still stepped through all timestamps before agent occurs
2: we call `move_to_time()` before `reset()`. This doesn't work either because `reset` will set `_elapsed_sim_time` to 0.

smarts/core/traffic_history_provider.py
Gamenot: Oh sure, I missed that.

smarts/core/smarts.py
Gamenot: I see that, I guess my concern here is that it is ballooning the API. It seems to be the wrong level to affect a specific provider from SMARTS instance. For instance I can say that the `SumoTrafficSimulation` also could use this time skip.

JingfeiPeng: I agree that this isn't a good level to affect a specific provider. I noticed we have the following apis in smarts.py
```
    def add_provider(self, provider):
    @property
    def providers(self):
        # TODO: Add check to ensure that action spaces are disjoint between providers
        # TODO: It's inconsistent that pybullet is not here
        return self._providers
```
Perhaps `history_vehicles_replacement_for_imitation_learning.py` should get smarts's providers and loop through the list to find traffic_history_provider and set the time_offset that way?

Gamenot: Yes, that was what I was thinking.

Gamenot: Yes, that was what I also arrived at. Ideally, though, we can have a better interface for getting the providers like `get_provider(<type>)`. This would be somewhat in line with something like the ECS architecture.

JingfeiPeng: Sounds good! i will do that

smarts/core/smarts.py
```diff
@@ -651,6 +651,11 @@ def providers(self):
         # TODO: It's inconsistent that pybullet is not here
         return self._providers
 
+    def get_provider_by_type(self, requested_type):
+        for provider in self._providers:
+            if isinstance(provider, requested_type):
+                return provider
```
Gamenot, line 657: This is good for now, we can improve it can be optimised later.

Gamenot (approved):

smarts/core/smarts.py
JingfeiPeng: If it is a small optimization we could do it now. I found this interesting way to do it. would this make it more optimized? 
```
     def get_provider_by_type(self, requested_type): 
         return next((p for p in self._providers if isinstance(p, requested_type) None)
```

Adaickalavan (approved):

PR #598, opened 2021-02-18
Update open gains with curvature

iman512003:
This is to Fix BM on curved road and cut-in scenarios.


JingfeiPeng (approved):

smarts/core/mission_planner.py
```diff
@@ -227,6 +228,12 @@ def cut_in_waypoints(self, sim, pose: Pose, vehicle):
 
         target_vehicle = neighborhood_vehicles[0]
         target_position = target_vehicle.pose.position[:2]
+
+        target_velocity = (
```
iman512003, line 232: We can calculate the velocity vector more generally, Do you think this is sufficient for this deadline? @Gamenot @JingfeiPeng

Gamenot: This works assuming the vehicle is moving only on x. I know this delays things slightly but could we have the more general form? Just so it does not fail on vehicles heading on y.

Gamenot (approved):

PR #604, opened 2021-02-19
Update OpEn 0.1.193

Gamenot:


JingfeiPeng:
should we remove some of the old wheel files?

Gamenot:
> should we remove some of the old wheel files? (189 and 190)

Yes, we should remove the old ones, especially since they are up on PyPI.

Gamenot:
Going back over this PR, I see it is better to just not have the open binaries in repository in the first place.

JingfeiPeng:
Might be useful to provide the commands to build and install open-agent and solver under the readme here: https://github.com/huawei-noah/SMARTS/tree/master/zoo/policies/open-agent

Issue #462, opened 2021-01-21
Refactor ULTRA State Preprocessor


christianjans:
make baseline state_preprocessor accessible to any agents
  - a new policy can simply plug it into their code and use it
  - State preprocessor depends on baseline agent_spec [baseline spec has some key features that state_preprocessor looks for them] 
     - one way to tackle this is to get rid of observation_adapter and move all of the information extraction into state_preprocessor so we break the dependency of state_preprocessor to agent_spec.observation_adapter, and we wouldn't pass any observation_adapter to agent_spec and do all of that inside our state_preprocessor. 
  - I think we need to keep the rest of low_dimensional states in the state_preprocessor intact [once they're stabilized]
  - main usages of state_preprocessor are:
    - 1. act() to process state before passing it to policy network
    - 2. update networks to process data fetched from replay_memory

christianjans:
Closed with the merge of #591.

PR #616, opened 2021-02-23
ULTRA RLlib: State Preprocessor Refactoring

christianjans:
### Info
Refactoring of the state preprocessor. Most of this refactoring work was done before the changes of RLlib integration and assumed that the `BaselineAdapter` object was no longer going to be used.

Now it seems like the `StatePreprocessor` object is closer to being deprecated than the `BaselineAdapter`. However, if it is still deemed useful, feel free to look at the refactored state preprocessor.

Will potentially address #462.

### Changes
This refactoring allows for users of ULTRA to easily create their own state preprocessor by inheriting from the `StatePreprocessor` and use static methods from this base class to grab specific parts of a raw environment observation.

Additionally, the `StatePreprocessor` class comes with the abstract method `_preprocess_state(state: Observation, *args, **kwargs)` which must be overridden by any child class. This method will be called to preprocess the given environment state.

See the `BaselineStatePreprocessor` for more on inheriting from `StatePreprocessor`.

### Testing
These changes have been tested to show that they produce the same processed states as without these refactoring changes.


JenishPatel99 (approved):

ultra/ultra/baselines/common/baseline_state_preprocessor.py
```diff
@@ -0,0 +1,231 @@
...
+        self._state_description = self.get_state_description(
+            social_vehicle_config, observation_waypoints_lookahead, action_size
+        )
+
+    @staticmethod
+    def get_state_description(
+        social_vehicle_config, observation_waypoints_lookahead, action_size
+    ):
+        return {
+            "images": {},
```
adreena, line 60: I removed images from all of the RLlib implementations because we don't use them in baseline, can you test train_rllib with this?

adreena (commented):
the "images" in state_preprocessor is not included in the rllib observation_space and it might break rllib-train.py because rllib is very sensitive to the observation and observation_space.
please check rllib-train in DEBUG mode and make sure things work fine for rllib

ultra/ultra/baselines/common/baseline_state_preprocessor.py
christianjans: Okay yes, if we simply uncomment
https://github.com/huawei-noah/SMARTS/blob/17fa18134c658b595b50d79f0d112f50c2064e98/ultra/ultra/baselines/common/baseline_state_preprocessor.py#L89
then we get this error:
```sh
ValueError: ('Observation for a Box/MultiBinary/MultiDiscrete space should be an np.array, not a Python list.', {'images': {}, 'low_dim_states': array([ 0.4453365 ,  0.        , -0.        ,  0.        , -1.2080997 ,
        1.1596714 ,  0.        ,  0.        ,  0.        ,  0.09338313,
        0.        ,  0.18676627,  0.        ,  0.2801494 ,  0.        ,
        0.37353253,  0.        ,  0.46691567,  0.        ,  0.5602988 ,
        0.        ,  0.65368193,  0.        ,  0.74706507,  0.        ,
        0.8404482 ,  0.        ,  0.93383133,  0.        ,  1.0272144 ,
        0.        ,  1.1205976 ,  0.        ,  1.2139807 ,  0.        ,
        1.3073639 ,  0.        ,  1.400747  ,  0.        ,  1.4941301 ,
        0.        ,  1.5875132 ,  0.        ,  1.6808964 ,  0.        ,
        1.7742795 ,  0.463     ], dtype=float32), 'social_vehicles': array([[-3.2000002e-02,  5.2642679e-01, -4.7114328e-01,  5.7698415e-17],
       [ 7.7859737e-02,  1.1286021e+00,  1.2947618e-04, -9.5708752e-01],
       [ 3.1775606e-01,  1.1606046e+00,  4.2924758e-17,  7.0101452e-01],
       [ 6.6240573e-01,  1.0966046e+00, -2.2666438e-16, -1.2339033e+00],
       [ 9.4935733e-01,  1.1606046e+00,  5.5122443e-17,  9.0021777e-01],
       [-1.4803296e+00,  1.0954689e+00,  4.3548187e-04, -5.8546180e-01],
       [-1.4826225e+00,  1.1274673e+00,  4.2566488e-04, -5.7226384e-01],
       [ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00],
       [ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00],
       [ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00]],
      dtype=float32)})
```
If we don't have images to preprocess, then it works if we simply change
https://github.com/huawei-noah/SMARTS/blob/17fa18134c658b595b50d79f0d112f50c2064e98/ultra/ultra/baselines/common/baseline_state_preprocessor.py#L89
to something like `images = np.zeros(1)`. But then if we do this, it may just be better to get rid of the images in the baseline state description, baseline state preprocessor, and baseline adapter, because they aren't used.

I think Alex just wanted to have the options to include images through the Gym wrapper, not necessarily for the baselines. Is this correct @AlexLewandowski? Sorry I could be incorrectly interpreting what you have said.

But anyways, since the baselines don't even use images, I think it would be okay to just remove images from the baseline state description, baseline state preprocessor, and baseline adapter (as long as Alex didn't need them for anything).

adreena: it's up to you guys if you wanna keep it. I think the images would be more complex as it needs to include dimensions of the image and also consider the number of images to stack. I think it's best to remove images for now and let Alex experiment with them himself instead of including them in baseline

christianjans: Okay great, sounds good.

adreena (approved):
looks good

ultra/ultra/baselines/common/baseline_state_preprocessor.py
```diff
@@ -0,0 +1,200 @@
...
+import torch
+
+from ultra.baselines.common.state_preprocessor import StatePreprocessor
+from ultra.utils.common import rotate2d_vector
+
+
+class BaselineStatePreprocessor(StatePreprocessor):
+    """The State Preprocessor used by the baseline agents."""
+
+    _NUM_LOOKAHEAD = 100
```
AlexLewandowski, line 33: Why do we need this hard-coded value? I know it was there before, but I think it's used only in extract_closest_waypoint and it already defaults to 100. Can we always use observation_num_lookahead?

ultra/ultra/baselines/adapter.py
```diff
@@ -97,7 +95,6 @@ def observation_space(self):
             social_vehicle_shape = self.social_capacity * self.num_social_features
         return gym.spaces.Dict(
             {
-                # "images": gym.spaces.Box(low=0, high=1e10, shape=(1,)),
                 "low_dim_states": gym.spaces.Box(
                     low=-1e10,
                     high=1e10,
```
AlexLewandowski, line 100: Should we include images here?

AlexLewandowski: Also, do the values actually vary from -1e10 to 1e10? Or is this just for convenience.

ultra/ultra/baselines/common/baseline_state_preprocessor.py
christianjans: Ahh good call. I will get rid of it. And good question, @adreena can you comment on this? Can we just always use the `observation_num_lookahead` in `params.yaml` for the `num_lookahead` parameter when we call `extract_closest_waypoint(...)`? It is done here: https://github.com/huawei-noah/SMARTS/blob/c8473049b4b52ad3eb29f7c7eb93753c2d70df55/ultra/ultra/baselines/common/state_preprocessor.py#L102 But not here: https://github.com/huawei-noah/SMARTS/blob/c8473049b4b52ad3eb29f7c7eb93753c2d70df55/ultra/ultra/baselines/adapter.py#L58 was there any reason why it was set to 100?

ultra/ultra/baselines/adapter.py
christianjans: I'm not sure why they vary like this, I imagine it's for convenience, maybe @adreena can comment on this again? And we had a discussion about whether to include images for the baselines or or not (see the resolved conversation in Kimia's first review of this Pull Request). We decided against including them because the baselines do not use them whether they are in their observation or not. But did you have some reasons to keep the images in the baselines' observations @AlexLewandowski?

AlexLewandowski: Nope, I was just looking for clarification. Looks good then!

AlexLewandowski (approved):

Issue #618, opened 2021-02-24
Error occurs during implementation of different encoders (new adapter format)


JenishPatel99:
- Problem : When trying to execute baselines using pointnet_encoder or precog_encoder, the following error occurs
-- pointnet_encoder log (using DQN
`(pid=14444) pybullet build time: Dec 23 2020 01:47:29
(pid=14444) LOADING SPEC from logs/experiment-2021.2.24-19:23:31-ultra.baselines.dqn:dqn-v0/spec.pkl
(pid=14444) /SMARTS/SMARTS/ultra/ultra/utils/common.py:107: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
(pid=14444) Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
(pid=14444)   for k, v in zip(index_to_action, np.arange(len(index_to_action)).astype(np.int))
(pid=14444) pybullet build time: Dec 23 2020 01:47:29
(pid=14444) pybullet build time: Dec 23 2020 01:47:29
(pid=14444) pybullet build time: Dec 23 2020 01:47:29
(pid=14444) loading from : logs/experiment-2021.2.24-19:23:31-ultra.baselines.dqn:dqn-v0/models/89
(pid=14444) Model loaded
(pid=14444) ╭────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────╮
(pid=14444) │            Episode │           Sim/Wall │        Total Steps │          Steps/Sec │              Score │
(pid=14444) ├────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┤
(pid=14444) /SMARTS/SMARTS/ultra/.venv/lib/python3.7/site-packages/smarts/core/utils/logging.py:73: ResourceWarning: unclosed file <_io.TextIOWrapper name=132 mode='w' encoding='UTF-8'>
(pid=14444)   sys.stdout = original_stdout
(pid=14444) ResourceWarning: Enable tracemalloc to get the object allocation traceback
(pid=14447) ╰────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────╯
[...]

PR #620, opened 2021-02-25
Update agent spec to retrieve encoder info from policy parameters

JenishPatel99:
Addresses Issue : #618


ultra/ultra/baselines/agent_spec.py
```diff
@@ -72,8 +72,10 @@ def __new__(
...
                 spec = new_spec
         else:
+            policy_dir = "/".join(inspect.getfile(policy_class).split("/")[:-1])
+            policy_params = load_yaml(f"{policy_dir}/params.yaml")
             social_vehicle_params = dict(
-                encoder_key="no_encoder",
+                encoder_key=policy_params["social_vehicles"]["encoder_key"],
                 social_policy_hidden_units=128,
                 social_polciy_init_std=0.5,
                 num_social_features=4,
```
christianjans, line 81:
```suggestion
                social_policy_hidden_units=policy_params["social_vehicles"]["social_policy_hidden_units"],
                social_polciy_init_std=policy_params["social_vehicles"]["social_polciy_init_std"],
                num_social_features=policy_params["social_vehicles"]["num_social_features"],
```

ultra/ultra/baselines/agent_spec.py
```diff
@@ -82,8 +84,6 @@ def __new__(
                 social_capacity=10,
```
christianjans, line 84:
```suggestion
                social_capacity=policy_params["social_vehicles"]["social_capacity"],
```

christianjans: Also in https://github.com/huawei-noah/SMARTS/blob/46d74c415d610c565bd481489d80f93809b3aa31/ultra/ultra/baselines/agent_spec.py#L83 maybe have it reference the policy_params so it's not hard-coded? Let me know your thoughts.

christianjans (commented):
Looks good, just a few suggestions to make some of the numbers not hard-coded. Let me know your thoughts!

ultra/ultra/baselines/agent_spec.py
adreena: yes these were left for cleanup. as long as we have access to policy_params it should be ok to replace these

ultra/ultra/baselines/agent_spec.py
```diff
@@ -82,8 +84,6 @@ def __new__(
...
             )
             adapter = BaselineAdapter(social_vehicle_params=social_vehicle_params)
-            policy_dir = "/".join(inspect.getfile(policy_class).split("/")[:-1])
```
adreena, line 85: why did u remove these? policy_params should be read from the saved version 
if you don't read the policy_params from the saved file, it could lose its original values from when the experiment was started.  here is the problem scenario:
 1- start experiment 1 with param x = 100
 2- start experiment 2 with param x = 200
 3- experiment 1 starts evaluation and creates AgentSpec with x = 200 instead of x = 100

so for avoiding this situation I saved the params from the beginning of each experiment and load them once evaluation of that experiment need them

adreena (commented):
the changes look good, just make sure the policy params are not removed from the agent_spec else condition

ultra/ultra/baselines/agent_spec.py
JenishPatel99: I am too sure what it means by "policy params should be read from the saved version". Currently I have the read the config file and extract those policy params here:

```         
            else:
                 policy_dir = "/".join(inspect.getfile(policy_class).split("/")[:-1])
                 policy_params = load_yaml(f"{policy_dir}/params.yaml")
                 social_vehicle_params = dict(	           
```
Instead of extracting the policy params after line 86, I moved it up (still in the else clause) so I can extract the social vehicle params at the same time and get the rest of the policy params.

ultra/ultra/baselines/agent_spec.py
JenishPatel99: Thank you for letting me know, the changes are implemented, please check de21070

ultra/ultra/baselines/agent_spec.py
adreena: ok I checked your agent_spec and those lines were not removed.

christianjans (approved):

AlexLewandowski (approved):

adreena (commented):
looks good

PR #621, opened 2021-02-25
[WIP] Ultra Gym wrapper

AlexLewandowski:
TODO:

- [ ] Fill in low-level state option using gym.spaces.Dict
- [x] Option for image resolution
- [x] "frame stacking" will need 2-3 images for markov state (i.e. to calculate speed and acceleration)
- [ ] Documentation for research use


AlexLewandowski:
With the current commit, we can actually use out-of-the-box algorithms from rllib as so, 

```
import ray
from ray import tune
from ultra.env.ultra_gym import UltraGym

ray.init()
tune.run(
    "DQN",
    stop={"episode_reward_mean": 100},
    config={
        "env": UltraGym,
        "num_gpus": 0,
        "num_workers": 1,
        "lr": tune.grid_search([0.01, 0.001, 0.0001]),
        "monitor": False,
    },
)
```

JenishPatel99:
@AlexLewandowski I am closing the PR and branch because it seems to be outdated

PR #623, opened 2021-02-25
Remove ultra package from site-packages during ci test

JenishPatel99:
- When using smarts==0.4.13, an ultra folder is added to the site-packages in the virtual env. When running any features, the ultra package from the site-packages is referenced (in the path) rather than the developing ultra folder. Not too sure why this is happening, because when using smarts<0.4.11 there isn't ultra package folder created in the virtual env.


AlexLewandowski:
A more permanent fix should change line 14 in setup.py from

```
    packages=find_packages(exclude="tests"),
```
to:
```
    packages=find_packages(exclude=["tests", "ultra"]),
```

AlexLewandowski (approved):

christianjans (approved):

PR #634, opened 2021-03-01
Limit memory use of loaded traffic histories

JingfeiPeng:
Summary of changes:

- Added ijson to load partial json values instead of loading the whole file [ijson](https://pypi.org/project/ijson/)
- Added TrafficHistoryService which uses ijson to load json values in batches. Currently the batch size is 300.

**Loading process is as follows:**
The scenario has 3500 timesteps
During init:
Load first 0->300 in main process 
Ask child process  to prepare the next 300 -> 600

when simulation reachs timestamp range 300->600, send request to child process and gets the previously prepared 300->600
Child process prepares 600 ->900

when simulation reachs timestamp range 600->900, send request to child process and gets the previously prepared 600->900
Child process prepares 900 ->1200

and so on

Will fix test and add a new test


smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @property
+    def traffic_history(self):
+        return {**self._current_traffic_history, **self._prev_batch_history}
+
+    def _prepare_next_batch(self):
+        self._request_queue.put(
+            RequestHistoryRange(
+                start_index=self._range_start,
+                end_index=self._range_start + self._batch_size,
```
Gamenot, line 116: I might replace this end index with count. Otherwise it will look like the following to get the batch size intended:

```suggestion
                end_index=self._range_start + self._batch_size - 1,
```

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+                ):
+                    self._current_traffic_history[t] = vehicles_state
+        self._range_start += self._batch_size
+        # prepares the next batch
+        self._prepare_next_batch()
+        self._receive_data_conn.recv()
+
+    def _fetch_history(
+        self, send_data, request_queue, history_file_path, total_timestep_size
+    ):
```
Gamenot, line 83: Not used:

```suggestion
    def _fetch_history(
        self, send_data, request_queue, history_file_path,
    ):
```

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+        self._receive_data_conn = receive_data_conn
+        self._request_queue = Queue()
+        self._fetch_history_proc = Process(
+            target=self._fetch_history,
+            args=(
+                send_data_conn,
+                self._request_queue,
+                self._history_file_path,
+                len(self._all_timesteps),
+            ),
```
Gamenot, line 58: This argument is not used:
```suggestion
                self._history_file_path,
            ),
```

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+                    and index <= self._range_start + self._batch_size
+                ):
+                    self._current_traffic_history[t] = vehicles_state
+        self._range_start += self._batch_size
+        # prepares the next batch
+        self._prepare_next_batch()
+        self._receive_data_conn.recv()
+
+    def _fetch_history(
+        self, send_data, request_queue, history_file_path, total_timestep_size
```
Gamenot, line 82: I could suggest using `send_data_conn` similar to the argument.

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+
+import ijson
+
+import smarts.core.scenario as scenario
+
+
+@dataclass
+class RequestHistoryRange:
+    start_index: int
+    end_index: int
```
Gamenot, line 31: nit: Could suggest count especially considering python's `range()` also uses count rather than an end index:
```suggestion
    history_count: int
```

smarts/core/tests/test_mutiple_traffic_data.py
```diff
@@ -117,14 +138,22 @@ def test_mutiple_traffic_data(create_scenario):
...
 
         traffic_history_provider.setup(scenario)
         assert (
-            traffic_history_provider._current_traffic_history
-            == scenario.traffic_history
+            traffic_history_provider._traffic_history_service
+            == scenario.traffic_history_service
         )
 
         use_first_history = not use_first_history
```
Gamenot, line 159: Rather this weird switching logic I think it would be better to pair up the scenarios with their respective traffic like:

```python
for scenario, traffic_history in zip(scenarios, [traffic_history_1, traffic_history_2]):
```

Alternatively parameterise the fixtures.

smarts/core/tests/test_mutiple_traffic_data.py
```diff
@@ -69,6 +70,24 @@
...
+    return p
+
+
+@pytest.fixture
+def create_history_file_2(tmp_path):
+    d = tmp_path / "sub2"
+    d.mkdir()
+    p = d / "traffic_history_2.json"
+    p.write_text(json.dumps(traffic_history_2))
+    return p
```
Gamenot, line 88: I think this could work as configuration, it would split the test into two but would reduce the complexity.

```suggestion
@pytest.fixture(params=[(traffic_history_1, "traffic_history_1"), (traffic_history_2, "traffic_history_2")])
def create_history_file(tmp_path, request):
    traffic_history, name = request.param
    d = tmp_path
    d.mkdir()
    path = d / f"{name}.json"
    path.write_text(json.dumps(traffic_history))
    return (traffic_history, path)
```

Then you can turn the test into this:

```
def test_mutiple_traffic_data(create_scenario, create_history_file):
    traffic_history, path = create_history_file
```

Gamenot (changes requested):
I will come back to review again

smarts/core/tests/test_mutiple_traffic_data.py
JingfeiPeng: I think this will cause `test_mutiple_traffic_data` to run twice each time with different traffic history file, right? The behavior we want from `Scenario.discover_traffic_histories` is so that it returns two history files at once in order to test if multiple traffic data will create multiple scenarios. 

I agree that we can simplify the two similar `create_history_file` though and I will work on that

JingfeiPeng: I think this will cause `test_mutiple_traffic_data` to run twice each time with different traffic history file, right? The behavior we want from `Scenario.discover_traffic_histories` is so that it returns two history files at once in order to test if multiple traffic data will create multiple scenarios. 

I agree that we can simplify the two similar `create_history_file` though and I will work on that

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+                end_index=self._range_start + self._batch_size,
+            )
+        )
+        self._range_start += self._batch_size
+
+    def fetch_history_at_timestep(self, timestep):
+        if timestep not in self._all_timesteps:
+            return {}
+        elif timestep in self.traffic_history:
+            return self.traffic_history[timestep]
```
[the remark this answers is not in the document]
JingfeiPeng: Actually we checked this before calling `fetch_history_at_timestep` in traffic_history provider `step`.
```
    def step(self, provider_actions, dt, elapsed_sim_time) -> ProviderState:
        timestamp = min(
            (
                float(ts)
                for ts in self._traffic_history_service.all_timesteps
                if float(ts) >= elapsed_sim_time
            ),
            default=None,
        )
```
We checked to use the timestep in all_timesteps that is the earliest after or equal to  `elapsed_sim_time`

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+                    if vehicle_id in vehicle_missions:
+                        continue
+                    vehicle_missions[vehicle_id] = scenario.Mission(
+                        start=scenario.Start(
+                            vehicles_state[vehicle_id]["position"][:2],
+                            scenario.Heading(vehicles_state[vehicle_id]["heading"]),
+                        ),
+                        goal=scenario.EndlessGoal(),
+                        start_time=float(t),
+                    )
```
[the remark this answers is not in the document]
JingfeiPeng: Could you elaborate on dumping this to a blob? What are we dumping exactly? Are we dumping the content in traffic_history.json into a blob file and whenever we need to access the data we access the blob file?

I think this might be costly in terms of performance but should be okay in terms of memory since we are getting `t, vehicles_state` for each timestamp and then overwrites t and vehicles_state with data in next timestamp.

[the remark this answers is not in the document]
JingfeiPeng: The performance cost of this is roughly similar to opening the json file using json.load and iterating through all its contents

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,153 @@
...
+        self, send_data, request_queue, history_file_path, total_timestep_size
+    ):
+        """prepare 1 batch ahead, when received request, immediately return the previously
+        prepared batch and prepares the next batch.
+        """
+        return_batch = {}
+        while True:
+            historyRange = request_queue.get()
+            assert isinstance(historyRange, RequestHistoryRange)
+            send_data.send(return_batch)
```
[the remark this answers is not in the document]
JingfeiPeng: Yes, in `__init__` of traffic_history_service it sends request to prepare the next batch, and does nothing with the first batch it receives

smarts/core/utils/traffic_history_serivce.py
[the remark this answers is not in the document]
JingfeiPeng: Also if we dump all the contents in json file into a blob file, then when loading the blob file, the size could be very large and we are loading all the contents into memory, this would kinda defeats the purpose of using ijson here?

smarts/core/scenario.py
```diff
@@ -514,7 +514,9 @@ def set_ego_missions(self, ego_mission):
         self._missions.update(ego_mission)
 
     def discover_missions_of_traffic_histories(self, vehicle_missions={}):
-        return self._traffic_history_service.fetch_agent_missions()
+        return Traffic_history_service.fetch_agent_missions(
+            self._traffic_history_service.history_file_path
+        )
```
Gamenot, line 519: This can work I think.

smarts/core/utils/traffic_history_serivce.py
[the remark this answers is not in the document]
Gamenot: Sorry, to clarify: when we get the first timestep we have to process a potentially massive history file and will have to reprocess the same vehicle multiple times because we want to find the earliest instance of a vehicle. We can skip looping over the json by taking the resulting `vehicle_missions` here and turning it into a binary. Then we can load those missions back in if available since it is faster.

For small files this will not have much effect but it would see a lot of improvement on massive traffic history files.

I do not think we need to do it now but it should be put up as an issue.

Gamenot (approved):

smarts/core/utils/traffic_history_serivce.py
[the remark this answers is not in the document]
JingfeiPeng: If I understood your comment correctly, you were saying this method would be called for every vehicle when we reaches its first timestamp. 
Actually this method `discover_missions_of_traffic_histories` is only called once per run of a scenario. 
It is called in `history_vehicles_replacement_for_imitation_learning.py` 
```
agent_missions = scenario.discover_missions_of_traffic_histories()
for agent_id, mission in agent_missions.items():
   ...
```
I think saving the `vehicle_missions` here would only be a somewhat helpful if there are multiple scenarios with same traffic_history_file, but the difference in performance shouldn't be too big

Gamenot (changes requested):
Please make that follow-up issue so I can approve.

smarts/core/scenario.py
```diff
@@ -540,8 +528,7 @@ def discover_traffic_histories(scenario_root):
         with open(path, "rb") as f:
             files = pickle.load(f)
             for file_name in files:
-                with open(os.path.join(scenario_root, file_name), "r") as history_file:
-                    traffic_histories.append(json.loads(history_file.read()))
+                traffic_histories.append(os.path.join(scenario_root, file_name))
```
Adaickalavan, line 531: Are the iterations of the `for` loop independent, such that they can be replaced with `map` to improve performance?

smarts/core/utils/traffic_history_serivce.py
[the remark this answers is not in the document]
JingfeiPeng: @Gamenot

smarts/core/scenario.py
JingfeiPeng: yes, good point thanks

smarts/core/scenario.py
```diff
@@ -44,6 +44,7 @@
 from .utils.file import file_md5_hash, make_dir_in_smarts_log_dir, path2hash
 from .utils.id import SocialAgentId
 from .utils.math import vec_to_radians
+from .utils.traffic_history_serivce import Traffic_history_service
 from .waypoints import Waypoints
```
Adaickalavan, line 48: Always use full paths rather than relative paths.

smarts/core/scenario.py
Adaickalavan: Are the iterations of the `for` loop independent, or can be made independent, such that they can be replaced with `map` or other constructs to improve performance and speed?

smarts/core/scenario.py
Adaickalavan, line 48: Always use full paths rather than relative paths.

smarts/core/tests/test_multiple_traffic_data.py
```diff
@@ -114,17 +129,17 @@ def test_mutiple_traffic_data(create_scenario):
...
+                == traffic_history[key]
+            )
 
         traffic_history_provider.setup(scenario)
         assert (
-            traffic_history_provider._current_traffic_history
-            == scenario.traffic_history
+            traffic_history_provider._traffic_history_service
+            == scenario.traffic_history_service
         )
```
Adaickalavan, line 145: Are the iterations of the `for` loop independent, or can be made independent, such that they can be replaced with `map` or other constructs to improve performance and speed?

smarts/core/tests/test_traffic_history_loading.py
```diff
@@ -0,0 +1,122 @@
...
+@pytest.fixture
+def create_scenario():
+    with temp_scenario(name="cycles", map="maps/6lane.net.xml") as scenario_root:
+        yield scenario_root
+
+
+def check_history(range_start, range_end, data):
+    for timestamp in data:
+        assert range_start <= float(timestamp) and float(timestamp) <= range_end
+        assert data[timestamp] == traffic_history[timestamp]
```
Adaickalavan, line 73: Are the iterations of the `for` loop independent, or can be made independent, such that they can be replaced with `map` or other constructs to improve performance and speed?

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,159 @@
...
+        with open(self._history_file_path, "rb") as f:
+            for index, (t, vehicles_state) in enumerate(
+                ijson.kvitems(f, "", use_float=True)
+            ):
+                self._all_timesteps.add(t)
+                if (
+                    self._range_start <= index
+                    and index < self._range_start + self._batch_size
+                ):
+                    self._current_traffic_history[t] = vehicles_state
```
Adaickalavan, line 74: Are the iterations of the `for` loop independent, or can be made independent, such that they can be replaced with `map` or other constructs to improve performance and speed?

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,159 @@
...
+            return_batch = {}
+            with open(history_file_path, "rb") as f:
+                for index, (t, vehicles_state) in enumerate(
+                    ijson.kvitems(f, "", use_float=True)
+                ):
+                    if (
+                        historyRange.start_index <= index
+                        and index < historyRange.start_index + historyRange.batch_count
+                    ):
+                        return_batch[t] = vehicles_state
```
Adaickalavan, line 102: Are the iterations of the `for` loop independent, or can be made independent, such that they can be replaced with `map` or other constructs to improve performance and speed?

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -0,0 +1,159 @@
...
+                    if vehicle_id in vehicle_missions:
+                        continue
+                    vehicle_missions[vehicle_id] = scenario.Mission(
+                        start=scenario.Start(
+                            vehicles_state[vehicle_id]["position"][:2],
+                            scenario.Heading(vehicles_state[vehicle_id]["heading"]),
+                        ),
+                        goal=scenario.EndlessGoal(),
+                        start_time=float(t),
+                    )
```
Adaickalavan, line 157: Are the iterations of the `for` loop independent, or can be made independent, such that they can be replaced with `map` or other constructs to improve performance and speed?

Adaickalavan (approved):

smarts/core/utils/traffic_history_serivce.py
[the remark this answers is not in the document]
Gamenot: OK, there is a bit more to it. I am assuming the converted NGSIM file looks like this:

```
traffic_history_1 = {
    "0.1": {
        "1": {
            "vehicle_id": "1",
            "vehicle_type": "car",
            "position": [1068.124, 959.455, 0],
            "speed": 5.128410474991252,
            "heading": -4.675796326794897,
            "vehicle_length": 4.22,
            "vehicle_width": 1.72,
        },
        "2": {
            "vehicle_id": "2",
            "vehicle_type": "car",
            "position": [1041.407, 956.583, 0],
            "speed": 2.9067860258367832,
            "heading": 1.4512036732051032,
            "vehicle_length": 4.16,
            "vehicle_width": 1.78,
        },
    },
    "0.2": {
        "1": {
            "vehicle_id": "1",
            "vehicle_type": "car",
            ...
        },
        ...
    },
    ...
}
```

So then the iteration is like this:

```            
for t, vehicles_state in ijson.kvitems(f, "", use_float=True):
    for vehicle_id in vehicles_state:
        if vehicle_id in vehicle_missions:
            continue
```

Then that means that the cause of the `if` statement here is because we revisit vehicles that have already been seen before but they are in a different frame (e.g. `vehicle: 1` in frame 1 and 2... and probably something like frames 3->882.) Looking at the NGSIM files I am seeing files as large as 250MB. The cost of reading a json file near that size is going to be terrible when all we need is the first known instance of the vehicle. 

Ideally we would want this to be dynamic but we seem constrained on that point.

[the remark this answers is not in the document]
JingfeiPeng: @Gamenot  yes, we do just need the first known instance of each vehicle in order to construct the mission. But since it is possible that a new vehicle id appears at a later timestamp, we had to iterate through all the timestamps and the vehicles at each timestamps.  

I will create the new issue.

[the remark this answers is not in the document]
JingfeiPeng: Issue created #640

smarts/core/scenario.py
JingfeiPeng: It seems many places in the code base uses relative paths, for example `smarts.py`. Converting 1 relative path to full path doesn't affect much, we should either convert all of them or not convert.

smarts/core/utils/traffic_history_serivce.py
JingfeiPeng: for a long block of code like this, is it suitable to use `map`? Since lambda functions can only be 1 line, I would have to define a new function for the following code:
```
self._all_timesteps.add(t)
if (
    self._range_start <= index
    and index < self._range_start + self._batch_size
):
    self._current_traffic_history[t] = vehicles_state
```

Gamenot (approved):

smarts/core/scenario.py
Adaickalavan: Indeed, I was suggesting to change all import paths to use full paths whenever you encounter them. I would prefer to always use full import paths compared to relative import paths. I do recognise the previous authors used relative paths, but I believe this is not a good practise. It leads to confusion and prone to error when the files are moved around.

smarts/core/utils/traffic_history_serivce.py
Adaickalavan: To avoid repetitions, the following comment is applicable to all of my previous comments on trying to avoid `for` loops.
 
A good practise would be to avoid executing code serially in a `for` loop if the contents of the `for` loop is independent of the iterations. Unnecessary `for` loops could eventually build up to slow down the code. It would be difficult to refactor the code at a later time, more so by a different author. Hence, it is mostly best that the code is well parallelized the first time it is written. If needed, additional functions may be written for usage in `map` or other constructs.

However, usage of `for` loops may be acceptable in cases where the `for` loop only minimally increases the run time or if the alternative parallel execution consumes more run time due to increase setup time.

You would be in the best position to determine whether the code can be refactored to avoid `for` loops to gain speed, if it is possible. If it is not possible to gain any further speed ups, than I am happy to continue merging this pull request and I have approved it.

PR #639, opened 2021-03-03
Update evaluate and train to evaluate per episode

AlexLewandowski:
TODO:

- [x] Introduce training episodes in evaluation to compare performance with explore = false.
- [x] Rename episode_reward to episode_return


ultra/ultra/evaluate.py
```diff
@@ -137,7 +142,7 @@ def evaluate(
 
         episode.reset(mode="Evaluation")
         while not dones["__all__"]:
-            action = agent.act(state, explore=False)
+            action = agent.act(state, explore=True)
```
christianjans, line 145: Oh interesting, are we allowing for exploration during evaluation?

ultra/tests/test_evaluate.py
```diff
@@ -130,6 +118,21 @@ def run_experiment():
...
+                    eval_episodes=1,
+                    max_episode_steps=2,
+                    episode_count=episode_count,
+                    policy_class=policy_class,
+                    scenario_info=("00", "eval_test"),
+                    timestep_sec=0.1,
+                    headless=True,
+                    log_dir=log_dir,
+                )
+                episode_count += 1
```
JenishPatel99, line 135: Why do we conduct the evaluation check after env.step() and agent.act()

AlexLewandowski: If we want training performance and evaluation to be performed at the same time, then evaluation has to wait unit after some episodes are over. The downside is we dont get to see evaluation at initialization of the agent.

ultra/ultra/evaluate.py
AlexLewandowski: If generalization is the gap between training and testing performance, then we need to make sure we're comparing apples to apples. If we explore during training and not during testing, then exploration can conflate the difference in performance. The alternative is to run the agent on the train set without explorating during evaluation (which may be a better idea, but more time consuming)

christianjans: Oh okay yes that makes sense, thanks!

ultra/ultra/utils/episode.py
```diff
@@ -219,10 +230,18 @@ def record_step(self, agent_id, infos, rewards, total_step=0, loss_output=None):
...
             agent_info.normalize()
 
+        if (old_episode is not None) and (eval_rate is not None):
+            for agent_id, agent_info in self.info[self.active_tag].items():
+                for key in agent_info.data:
+                    agent_info.data[key] = (
+                        agent_info.data[key] * eval_rate
+                        + old_episode.info[self.active_tag][agent_id].data[key]
+                    ) / eval_rate
+
```
christianjans, line 244: Wait sorry could you explain what this is here?

AlexLewandowski: Yeah, it's a bit confusing. I want to average training performance over the last eval_rate (lets call it N) episodes. To do this incrementally (one episode at a time), there is a recursive formula: avg_1:(n+1) = (avg_1:n * n + new_data)/(n+1). Since I know the average is going to be over N episodes, I just divide by N = eval_rate instead of n and n+1.

ultra/tests/test_evaluate.py
JenishPatel99: I see, sorry I think you talked about this yesterday but I may need a bit more clarification on why we need to perform the training and evaluation at the same?

AlexLewandowski: To measure generalization, we need to compare training and test performance. If we don't train and test at the same time, we won't be comparing the same agent because after training the agent's parameters will be updated.

JenishPatel99: Okay, that does make sense. The agents who be on different states and comparison between the evaluation and training would contain some offset.

ultra/ultra/utils/episode.py
christianjans: Oh right okay gotcha, makes sense. And then these averages will be recorded to TensorBoard every eval_rate episodes: https://github.com/huawei-noah/SMARTS/blob/360e0ebdfa7c90967b4fe9f6f9e3c49a10a334c5/ultra/ultra/train.py#L122-L123 Am I understanding this correctly?

AlexLewandowski: Yes exactly

JenishPatel99 (approved):

ultra/ultra/train.py
```diff
@@ -108,6 +108,8 @@ def train(
     # policy_classes list, transform it to an etag of "dqn-v0:ppo-v0".
     etag = ":".join([policy_class.split(":")[-1] for policy_class in policy_classes])
 
+    episode_count = 0
```
christianjans, line 111: I think the episode objects have an `index` attribute that keeps track of the episode number. Is there any reason why a separate `episode_count` variable was used instead of `episode.index`? Both are fine, but just wondering.

christianjans (approved):

ultra/ultra/train.py
AlexLewandowski: Missed this, thanks.

PR #643, opened 2021-03-04
Fix envision error 15

Gamenot:
image

Cleaning up the Envision exit error that showed up due to a process being cleaned up through timeout.

### DIFF
--- a/envision/client.py
+++ b/envision/client.py
@@ -213,6 +213,9 @@ class Client:
                 if not connection_established:
                     self._log.info(f"Attempt {tries} to connect to Envision.")
                 else:
+                    # No information left to send, connection is likely done
+                    if state_queue.empty():
+                        break
                     # When connection lost, retry again every 3 seconds
                     wait_between_retries = 3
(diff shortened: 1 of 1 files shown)


JingfeiPeng (approved):
LGTM

Adaickalavan (approved):

envision/client.py
```diff
@@ -213,6 +213,9 @@ def run_socket(endpoint, wait_between_retries):
                 if not connection_established:
                     self._log.info(f"Attempt {tries} to connect to Envision.")
                 else:
+                    # No information left to send, connection is likely done
+                    if state_queue.empty():
+                        break
```
Gamenot, line 218: To be clear, this can only happen if the connection has already been established, then disconnected, and the `state_queue` has run out of objects such as in `teardown()` on https://github.com/huawei-noah/SMARTS/pull/643/files#diff-7b43287d33aa76542305d80067527c750e75691693d63d8107edc59f5e8b79acR243-R245.

PR #644, opened 2021-03-04
made a few fixes and did some more cleanup

adreena:
I noticed some methods were throwing exception in ppo and fixed them
also did some refactoring to rllib and baseline_adapter to clean things up
@JenishPatel99 please review these and apply changes to your tests, if we could merge these changes to ultra-rllib and then merge everything to develop that'd be great

also  I added more algorithms to rllib but still waiting for dqn and sac integration


JenishPatel99:
Hey @adreena I have made some changes to the tests and a clean up a few statements. The idea of moving the social vehicle information to a separate params is great, especially now that everything is orchestrated by passing the agent_type through the adapter.

adreena:
Yes it should be separated from any changes that researchers apply to our
baseline params, these are tricky to find and might raise issues later such
as comparing experiments with a modified score rather than a fixed score

PR #646, opened 2021-03-05
Fix add ijson to setup

Gamenot:


JingfeiPeng (approved):

JingfeiPeng:
I initially added ijson in requirements.txt. Are we adding it in setup.py because if people install smarts package, dependencies in requirements.txt wouldn't be installed, and to install ijson when installing smarts package, we have to add it in setup.py?

Gamenot:
> I initially added ijson in requirements.txt. Are we adding it in setup.py because if people install smarts package, dependencies in requirements.txt wouldn't be installed, and to install ijson when installing smarts package, we have to add it in setup.py?

Yes it needs to be in setup.py or it will not be added to the package when we build it.

Adaickalavan (approved):

PR #647, opened 2021-03-05
Update common.py

Teelfee:
While the ego_closest_wp and  closest_wp are totally the same in this function.

remove extra variable closest_wp


Gamenot:
@Teelfee Hello! Thank you for contributing. This appears to be a good change for the health of the code. If you can, please sign the contributor agreement and we can get this change merged in.

Gamenot (approved):

Gamenot:
This change has become out of date.

PR #648, opened 2021-03-05
Fix ci error

JingfeiPeng:
fixed ci tests on smarts side, still needs to figure out how to fix ci test on Ultra side

### DIFF
--- a/.github/workflows/ci-base-tests.yml
+++ b/.github/workflows/ci-base-tests.yml
@@ -1,9 +1,6 @@
 name: SMARTS CI Base Tests
 
-on:
-  push:
-    branches:
-      - disabled
+on: [push, pull_request]
 
 jobs:
   test:
--- a/.github/workflows/ci-base-tests.yml
+++ b/.github/workflows/ci-base-tests.yml
@@ -33,10 +33,10 @@ jobs:
         run: |
           python3.7 -m venv .venv
           . .venv/bin/activate
-          pip install --upgrade --upgrade-strategy eager pip
-          pip install --upgrade --upgrade-strategy eager wheel
-          pip install --upgrade --upgrade-strategy eager -r requirements.txt
-          pip install --upgrade --upgrade-strategy eager -e .[train,test]
+          pip install --upgrade pip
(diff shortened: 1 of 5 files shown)


JingfeiPeng:
To reproduce the failing CI test locally, create two venv, one based on smarts ci and one based on Ultra's CI:

```
python3.7 -m venv .first
. .first/bin/activate
pip install --upgrade pip
pip install --upgrade wheel
pip install --upgrade -r requirements.txt
pip install --upgrade -e .[train,test]
```
then `cd ultra` and run `pytest tests/test_train.py` the test passes

Create venv according to Ultra's CI:
```
cd ultra
python3.7 -m venv .second
. .second/bin/activate
pip install --upgrade --upgrade-strategy eager pip
pip install --upgrade -e .
```

then run `pytest tests/test_train.py` we can reproduce the numpy import error
```
.ultra_venv/lib/python3.7/site-packages/smarts/core/utils/pybullet.py:27: in <module>
    from pybullet import *
E   ImportError: numpy.core.multiarray failed to import
```

JenishPatel99:
@JingfeiPeng I have managed to suppress the ultra ci error by upgrading the numpy package. A stackexchange thread seemed to give a solution here is the link [https://stackoverflow.com/questions/33859531/runtimeerror-module-compiled-against-api-version-a-but-this-version-of-numpy-is](url)

.github/workflows/ci-base-tests.yml
```diff
@@ -1,6 +1,9 @@
 name: SMARTS CI Base Tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - disabled
```
JingfeiPeng, line 6: this would disable the tests right? We probably want to fix it instead of disabling it

JenishPatel99: Yes, sorry I forgot to enable those tests after just testing the ultra-ci by itself. I have restored the others tests as well

Gamenot (approved):

Gamenot:
We should merge this into both develop and master.

PR #663, opened 2021-03-09
Fix envision button skip to newest when resumed

JingfeiPeng:
Currently there is a problem with envision's pause button, pausing works, but when resumed, envision will skip directly to the most latest timestamp. This PR fixes this issue.

Summary of changes:
1: in client.js, frames are obtained from the end of the list, which has the most current frame. Changed this to obtain frames from beginning of list.
2: Added a sleep function to have a minimum amount of wait time before loading the next frame. Since after done step 1, when paused, there will be a long list of existing frames waiting to be fetched, if we simply fetch them from list and update worldstate, the update is too fast for react to cause re-render. Previously this wasn't a problem since the speed of giving out available frames is 1 frame per eposide on the smarts side. Envision will always get the frame when it is available and wait for the next available frame.

### DIFF
--- a/envision/web/src/client.js
+++ b/envision/web/src/client.js
@@ -161,7 +161,8 @@ export default class Client {
             continue;
           }
 
-          let item = this._stateQueues[simulationId].pop();
+          // Removes the oldest element
+          let item = this._stateQueues[simulationId].shift();
           let elapsed_times = [
             item.current_elapsed_time,
             item.total_elapsed_time,
--- a/envision/web/src/components/simulation.js
+++ b/envision/web/src/components/simulation.js
@@ -157,17 +157,31 @@ export default function Simulation({
   useEffect(() => {
     let stopPolling = false;
     (async () => {
+      const msInSec = 1000;
       const it = client.worldstate(simulationId);
+      let prevElapsedTime = null;
+      let waitStartTime = null;
       let wstate_and_time;
       if (playing) wstate_and_time = await it.next();
       while (!stopPolling && playing && !wstate_and_time.done) {
         let wstate, elapsed_times;
         [wstate, elapsed_times] = wstate_and_time.value;
+        const currentTime = elapsed_times[0]
+        if (prevElapsedTime == null) {
+          // default: wait 50ms before playing the next frame
+          await sleep(50);
+        } else {
+          // msInSec*(currentTime-prevElapsedTime) is the time difference between 
+          //   current frame and previous frame
+          // Since we could have waited (Date.now() - waitStartTime) to get the current frame,
+          //    we deduct this amount from the time we will be waiting
+          await sleep(msInSec*(currentTime-prevElapsedTime)- (Date.now() - waitStartTime))
+        }
+        prevElapsedTime = currentTime
+
         setWorldState(wstate);
         onElapsedTimesChanged(...elapsed_times);
-
-        // play the frames every 50ms
-        await sleep(50);
+        waitStartTime = Date.now();
         wstate_and_time = await it.next();
       }
     })();
--- a/envision/web/src/components/simulation.js
+++ b/envision/web/src/components/simulation.js
@@ -163,13 +163,11 @@ export default function Simulation({
       while (!stopPolling && playing && !wstate_and_time.done) {
         let wstate, elapsed_times;
         [wstate, elapsed_times] = wstate_and_time.value;
-        if (!stopPolling) {
-          setWorldState(wstate);
-          onElapsedTimesChanged(...elapsed_times);
+        setWorldState(wstate);
+        onElapsedTimesChanged(...elapsed_times);
 
-          // play the frames at 100/40 = 20fps
-          await sleep(50);
-        }
+        // play the frames every 50ms
+        await sleep(50);
         wstate_and_time = await it.next();
       }
     })();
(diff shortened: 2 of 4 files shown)


envision/web/src/components/simulation.js
```diff
@@ -149,18 +149,26 @@ export default function Simulation({
...
     (async () => {
       const it = client.worldstate(simulationId);
-      let wstate_and_time = await it.next();
-      while (!wstate_and_time.done && playing) {
+      let wstate_and_time;
+      if (playing) wstate_and_time = await it.next();
+      while (!stopPolling && playing && !wstate_and_time.done) {
         let wstate, elapsed_times;
         [wstate, elapsed_times] = wstate_and_time.value;
         if (!stopPolling) {
```
Gamenot, line 166: Is this line needed anymore considering the `!stopPolling` in the while loop?

envision/web/src/components/simulation.js
```diff
@@ -149,18 +149,26 @@ export default function Simulation({
...
           setWorldState(wstate);
           onElapsedTimesChanged(...elapsed_times);
+
+          // play the frames at 100/40 = 20fps
+          await sleep(50);
```
Gamenot, line 171: This seems suspicious to me, this calculation wants `100/50` but since this waits in real-time what stops this from playing at ~200fps?

envision/web/src/client.js
```diff
@@ -161,7 +161,8 @@ export default class Client {
             continue;
           }
 
-          let item = this._stateQueues[simulationId].pop();
+          // Removes the oldest element
+          let item = this._stateQueues[simulationId].shift();
```
Gamenot, line 165: This makes more sense.

Gamenot (commented):
A question.

envision/web/src/components/simulation.js
JingfeiPeng: this line causes the code execution to sleep for 50ms before fetching the next `wstate_and_time`.

My comment was a bit incorrect, it should be 1s = 1000ms/50ms = 20 frames per second. This ensures the simulation doesn't run faster than 20 frames per second, 20 fps also seems to be the normal rate for smarts to send out the frames.

I updated the comment and this 50ms here is also open to changes.

envision/web/src/components/simulation.js
```diff
@@ -149,19 +149,25 @@ export default function Simulation({
...
         [wstate, elapsed_times] = wstate_and_time.value;
-        if (!stopPolling) {
-          setWorldState(wstate);
-          onElapsedTimesChanged(...elapsed_times);
-        }
+        setWorldState(wstate);
+        onElapsedTimesChanged(...elapsed_times);
+
+        // play the frames every 50ms
+        await sleep(50);
```
sah-huawei, line 170: Could you please point me to where the rate that Smarts is sending is controlled?  (I saw your comment that it's also 20fps, but I'm just curious to look at it.)

JingfeiPeng: yes, SMARTS is sending frame to envision at the following line: https://github.com/huawei-noah/SMARTS/blob/master/smarts/core/smarts.py#L970   `self._envision.send(state)`

Here smarts is just sending the frame whenever an episode is done. 20fps is just an estimate. Since when I use 20fps, the visulation looks normal when transitioning from before pause -> pause -> after resume. If I use a higher fps ( sleep less than 50ms) then i can see the vehicle looks like it is driving way faster after resume for a period of time. Then slows down again when reached the latest frame

JingfeiPeng, line 170: Will try to use the difference between the elapsed time of previous frame and current frame to decide how long to sleep. Also needs to take into account the time spent on waiting to get the next frame

envision/web/src/components/simulation.js
```diff
@@ -149,19 +149,42 @@ export default function Simulation({
...
+          await sleep(50);
+        } else {
+          // msInSec*(currentTime-prevElapsedTime) is the time difference between
+          //   current frame and previous frame
+          // Since we could have waited (Date.now() - waitStartTime) to get the current frame,
+          //    we deduct this amount from the time we will be waiting
+          await sleep(
+            msInSec * (currentTime - prevElapsedTime) -
+              (Date.now() - waitStartTime)
+          );
```
Gamenot, line 181: I think this biases towards losing time although it means we will never skip frames.

Gamenot (approved):

Adaickalavan (approved):

envision/web/src/components/simulation.js
```diff
@@ -149,19 +149,42 @@ export default function Simulation({
...
+          // default: wait 50ms before playing the next frame
+          await sleep(50);
+        } else {
+          // msInSec*(currentTime-prevElapsedTime) is the time difference between
+          //   current frame and previous frame
+          // Since we could have waited (Date.now() - waitStartTime) to get the current frame,
+          //    we deduct this amount from the time we will be waiting
+          await sleep(
+            msInSec * (currentTime - prevElapsedTime) -
+              (Date.now() - waitStartTime)
```
sah-huawei, line 180: Is it possible for this to end up being negative?  If so, should we avoid that?

JingfeiPeng: in very rare situations where the waiting time to get the next frame exceeds the time difference between the frames the sleep time could be negative. However this is fine, calling sleep with a negative numebr `sleep(-1000)` would return instantly just like `sleep(0)`

sah-huawei (approved):

Issue #378, opened 2021-01-04
Appropriately dispose of SMARTS at exit


Gamenot:
Instances of the SMARTS class hold on to a lot of resources that do not get automatically cleaned up on exit.

AC:
- [ ] Investigate to see if SMARTS instance can automatically close on exit
- [ ] SMARTS instance is cleaned up when destroyed (e.x. `SMARTS.destroy()` is called from `delete` or `atexit`)
- [ ] Introduce clear documentation to explain how to dispose of SMARTS instances appropriately.
- [ ] Update all examples to follow the appropriate practice.

Gamenot:
@sah-huawei Help Jeffer with this ticket.

JingfeiPeng:
Closed with #665

PR #665, opened 2021-03-10
Appropriately dispose smarts (#378)

JingfeiPeng:
- Ensured SMARTS destroy can be called safely multiple times
- Ensured Traffic_history_service clean up sub process

**Ensured Investigate to see if SMARTS instance can automatically close on exit**
Seems SMARTS instance cannot bt automatically close on exit. When running `test.py`, if user does not call `smarts.destroy()`, smarts.__del__  is not called before the next iteration, which causes a "mutiple showbase error". There is no guarantee on __del__ being called immediately after references to smarts is lost. So I think there is a need for users to call `destroy()` explicitly in smarts.

**SMARTS instance is cleaned up when destroyed (e.x. SMARTS.destroy() is called from delete or atexit)**
Done

**Introduce clear documentation to explain how to dispose of SMARTS instances appropriately.**
https://github.com/huawei-noah/SMARTS/blob/master/docs/sim/env.rst seems to have already mentioned to call `env.close`

**Update all examples to follow the appropriate practice.**
If we still forces user to call destroy() explicitly, then there is no change needed in examples.


smarts/core/utils/traffic_history_serivce.py
```diff
@@ -77,6 +80,17 @@ def __init__(self, history_file_path):
         self._prepare_next_batch()
         self._receive_data_conn.recv()
 
+    def teardown(self):
+        if self.is_in_use:
```
sah-huawei, line 84: I don't see where `history_file_path` gets reset to `None`.  So would this `is_in_use` check always succeed?  If so, then this `teardown()` method isn't really idempotent.

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -77,6 +80,17 @@ def __init__(self, history_file_path):
...
+            self._request_queue.put(Traffic_history_service.QueueDone)
+            self._request_queue.close()
+            self._request_queue = None
+            self._fetch_history_proc.join(timeout=3)
```
sah-huawei, line 88: Here you might check `isAlive()` and, if the join timed out, print some sort of warning message telling the user (in case they want to kill the process manually).  (imo, not that big of a deal though, either way.)

sah-huawei:
Also, should a `smarts.destroy()` call be added to more of the examples?  (Looks like not all of them use it.)

JingfeiPeng:
> Also, should a `smarts.destroy()` call be added to more of the examples? (Looks like not all of them use it.)

We should if the examples don't call smarts.destroy. Do you have an example of an example that didn't call smarts destroy? The examples under `examples` directory do call smarts.destroy when they call `env.close()`

sah-huawei:
> > Also, should a `smarts.destroy()` call be added to more of the examples? (Looks like not all of them use it.)
> 
> We should if the examples don't call smarts.destroy. Though do you have an example of an example that didn't call smarts destroy? The examples under `examples` directory do call smarts.destroy when they call `env.close()`

I saw one in `observation_collection_for_imitation_learning.py` but maybe that only exists on my branch?

smarts/core/utils/traffic_history_serivce.py
JingfeiPeng: Right, nice catch. thanks

JingfeiPeng:
> > > Also, should a `smarts.destroy()` call be added to more of the examples? (Looks like not all of them use it.)
> > 
> > 
> > We should if the examples don't call smarts.destroy. Though do you have an example of an example that didn't call smarts destroy? The examples under `examples` directory do call smarts.destroy when they call `env.close()`
> 
> I saw one in `observation_collection_for_imitation_learning.py` but maybe that only exists on my branch?

Right, that example is missing smarts.destroy. I added it now. Also checked all other examples either calls `env.close` or `smarts.destroy()`. The exceptions are rllib.py and rllib_agent.py which don't seem to use smarts or env.

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -77,6 +82,20 @@ def __init__(self, history_file_path):
         self._prepare_next_batch()
         self._receive_data_conn.recv()
 
+    def teardown(self):
+        if self.is_in_use:
+            self._request_queue.put(Traffic_history_service.QueueDone())
```
Gamenot, line 87: nit: Not very important in this case but I have had this thought that it might be useful to drain the queue before adding the `QueueDone` signal.

Gamenot (approved):

sah-huawei (approved):

smarts/core/utils/traffic_history_serivce.py
JingfeiPeng: while I was trying to drain the queue before adding `QueueDone`, I noticed the check for QueueDone was not reached in the child process.
```
        while True:
            print("start")
            historyRange = request_queue.get()
            print("got: ",historyRange)
            if type(historyRange) is Traffic_history_service.QueueDone:
                print("Found done")
                break
```
In fact, `print("got: ",historyRange)` was not reached either, then I tried putting other values in the queue and `print("got: ",historyRange)` also didn't show up. Though `teardown` method was reached. Is this related to putting items in a queue in `___del___` method?

iman512003 (approved):

smarts/core/utils/traffic_history_serivce.py
JingfeiPeng: The reason this happens is `__del__` for traffic_history_service happens even after smarts calls destroy, when teardown is called, the child process `self._fetch_history_proc` is no longer alive.....

PR #670, opened 2021-03-12
Waypoint optimizations

sah-huawei:
These are some optimizations that came from Adai's profiling and looking at flamegraph.
There is much more optimizing to be done (and I'll update the bottlenecks issue accordingly)... this was limited to just dealing with waypoints (and even there, more could still be done, but trying not to disrupt too much at this point).

Note that the Makefile change came out of our discussion today and is unrelated.

### DIFF
--- a/smarts/core/waypoints.py
+++ b/smarts/core/waypoints.py
@@ -120,6 +120,9 @@ class LinkedWaypoint:
         self.is_shape_wp = is_shape_wp
 
     def __hash__(self):
+        ## distinguish between different continuations here too
+        ## so the lru_cache on _waypoints_starting_at_waypoint() below
+        ## doesn't return the wrong set of waypoints.
         return hash(self.wp) + sum(hash(nwp.wp) for nwp in self.nexts)
 
 
@@ -128,7 +131,9 @@ class Edges:
         self.edge_ids = edge_ids
 
     def __hash__(self):
-        return hash(str(self.edge_ids))  # be careful (don't mutate edge_ids) / only use for lru_cache below.
+        ## be careful: don't mutate edge_ids while using this.
+        ## (only use for lru_cache below)
+        return hash(str(self.edge_ids))
 
 
 class Waypoints:
@@ -218,14 +223,14 @@ class Waypoints:
         return linked_waypoint.wp
 
     def waypoint_paths_on_lane_at(
-        self, point, lane_id, lookahead, filter_edges: Sequence[str] = None
+        self, point, lane_id, lookahead, filter_edge_ids: Sequence[str] = None
     ):
         lane_kd_tree = self._waypoints_kd_tree_by_lane_id[lane_id]
         closest_linked_wp = self._closest_linked_wp_in_kd_tree_batched(
             [point], self._waypoints_by_lane_id[lane_id], lane_kd_tree, k=1
         )[0][0]
         return self._waypoints_starting_at_waypoint(
-            closest_linked_wp, lookahead, tuple(point), Edges(filter_edges)
+            closest_linked_wp, lookahead, tuple(point), Edges(filter_edge_ids)
         )
 
     def waypoint_paths_at(self, pose, lookahead, filter_from_count=3, within_radius=5):
@@ -366,9 +371,10 @@ class Waypoints:
 
     @staticmethod
     def _inplace_unwrap(wp_array):
-        ## optimization hack adapted from
+        ## minor optimization hack adapted from
         ##  https://github.com/numpy/numpy/blob/v1.20.0/numpy/lib/function_base.py#L1492-L1546
-        ## to avoid unnecessary (slow) np array copy.
+        ## to avoid unnecessary (slow) np array copy
+        ## (as seen in profiling).
         p = np.asarray(wp_array)
         dd = np.subtract(p[1:], p[:-1])
         ddmod = np.mod(dd + math.pi, 2 * math.pi) - math.pi
--- a/smarts/core/waypoints.py
+++ b/smarts/core/waypoints.py
@@ -25,7 +25,6 @@ from dataclasses import dataclass
 from typing import Sequence
 
 import numpy as np
-from scipy.interpolate import interp1d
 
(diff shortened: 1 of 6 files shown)


smarts/core/waypoints.py
```diff
@@ -111,16 +112,28 @@ def __eq__(self, other):
...
-LinkedWaypoint = namedtuple(
-    "LinkedWaypoint",
-    [
-        "wp",  # Waypoint: current waypoint
-        "nexts",  # list of LinkedWaypoint: list of next immediate waypoints
+class LinkedWaypoint:
+    def __init__(self, wp=None, nexts=[], is_shape_wp=False):
+        self.wp = wp  # Waypoint: current waypoint
+        self.is_shape_wp = is_shape_wp
+        self.nexts = nexts  # list of LinkedWaypoint: list of next immediate waypoints
```
Adaickalavan, line 119: The following is a suggestion for the entire `LinkedWaypoint` class. Unfortunately, I could not highlight the entire class while commenting. 

I believe it is better to maintain the immutable data type, namely `NamedTuple`, with type hints. It can be done as follows:

```suggestion
class LinkedWaypoint(NamedTuple):
    wp: Waypoint = None  # Waypoint: current waypoint
    nexts: List[Waypoint] = [] # list of LinkedWaypoint: list of next immediate waypoints
    is_shape_wp: bool = False

    def __hash__(self):
        ## distinguish between different continuations here too
        ## so the lru_cache on _waypoints_starting_at_waypoint() below
        ## doesn't return the wrong set of waypoints.
        return hash(self.wp) + sum(hash(nwp.wp) for nwp in self.nexts)
```

smarts/core/waypoints.py
```diff
@@ -356,6 +367,21 @@ def _waypoints_starting_at_waypoint(
...
+        ## to avoid unnecessary (slow) np array copy
+        ## (as seen in profiling).
+        p = np.asarray(wp_array)
+        dd = np.subtract(p[1:], p[:-1])
+        ddmod = np.mod(dd + math.pi, 2 * math.pi) - math.pi
+        np.copyto(ddmod, math.pi, where=(ddmod == -math.pi) & (dd > 0))
+        ph_correct = ddmod - dd
+        np.copyto(ph_correct, 0, where=abs(dd) < math.pi)
+        p[1:] += ph_correct.cumsum(axis=-1)
+        return p
```
Adaickalavan, line 383: I believe `_inplace_unwrap()` can be placed as a utility function outside the `Waypoints` class, as it is not necessarily tied to waypoints.

Adaickalavan:
How much time did we save per episode by 
+ changing numpy/scipy implementation (i.e., np.unwrap, interp1d),
+ caching waypoints, and
+ reducing lookaheads?

smarts/core/waypoints.py
sah-huawei: done.

smarts/core/waypoints.py
sah-huawei: done.

smarts/core/waypoints.py
sah-huawei: done.

smarts/core/waypoints.py
sah-huawei: done.

sah-huawei:
> How much time did we save per episode by
> 
>     * changing numpy/scipy implementation (i.e., np.unwrap, interp1d),
> 
>     * caching waypoints, and
> 
>     * reducing lookaheads?

The total amount of _time_ saved will depend on many things (for example, depending on how many times each of these things are used in a scenario), but I have been watching both the average steps per second and (better) percentage spent in the functions (via the flame-graphs), which improved with each of the above.  However, I don't know the best way to report this as data here.  

Of the above 3 things, the caching of the waypoints is currently the _least_ helpful, because the cache is only rarely hit.  The current `lru_cache` (while convenient to add) only helps in the circumstance of redundant picalls to retrieve waypoints starting from the same waypoint, which is rare.  A better cache would handle _overlapping_ calls too (for example, allowing for the `lookahead` parameter not to have to match, and even for different starting waypoints, among other things).  However, given the complexity of that change, I decided to leave it until a later sprint -- we can file it as a separate issue as Tucker requested -- and just push what we have now.

smarts/core/waypoints.py
```diff
@@ -216,13 +232,10 @@ def waypoint_paths_on_lane_at(
         closest_linked_wp = self._closest_linked_wp_in_kd_tree_batched(
             [point], self._waypoints_by_lane_id[lane_id], lane_kd_tree, k=1
         )[0][0]
-
-        unlinked_waypoint_paths = self._waypoints_starting_at_waypoint(
-            closest_linked_wp, lookahead, point, filter_edge_ids
+        return self._waypoints_starting_at_waypoint(
+            closest_linked_wp, lookahead, tuple(point), Edges(filter_edge_ids)
```
Gamenot, line 236: nit: I somewhat disagree with creating the `Edges` class if it is going to just be used for an internal method. It is also possible to just force the use of a tuple for hashing and then also avoid the mutability concern you mention in the class.

Gamenot (approved):

Adaickalavan (approved):

smarts/core/waypoints.py
sah-huawei: agreed.  Just pushed the change to switch to tuple from the `Edges` class.

sah-huawei:
> Of the above 3 things, the caching of the waypoints is currently the _least_ helpful, because the cache is only rarely hit. The current `lru_cache` (while convenient to add) only helps in the circumstance of redundant picalls to retrieve waypoints starting from the same waypoint, which is rare. A better cache would handle _overlapping_ calls too (for example, allowing for the `lookahead` parameter not to have to match, and even for different starting waypoints, among other things). However, given the complexity of that change, I decided to leave it until a later sprint -- we can file it as a separate issue as Tucker requested -- and just push what we have now.

This is now tracked as Issue #679.

JingfeiPeng (approved):

Issue #640, opened 2021-03-03
Store the processed mission object from history json file as a temporary binary file


JingfeiPeng:
**Is your feature request related to a problem? Please describe.**
The `fetch_agent_missions` method in `traffic_history_service` currently could potentially loop through a large large json file by `ijson.kvitems(...)` but only the first occurrence of each vehicle is needed to construct the mission. 

**Describe the solution you'd like**
If multiple scenarios use the same history file, it would be good to store the final mission object as a binary file and load if we need the missions again for a different scenario.

**Additional context**
_Originally posted by @Gamenot in https://github.com/huawei-noah/SMARTS/pull/634#discussion_r585156257_

Gamenot:
A potential here is that we could create this file as a part of the conversion of the NGSIM dataset.

JingfeiPeng:
yes and intersection_data_set_merging. 

Though `ijson.kvitems` is also used at two other places: init of `Traffic_history_service` to know all the timestamp and initial batch of history, and in child process to fetch next batch of history. It might be interesting to have the first looping over the json file to also store multiple partitions of original large history file as binary files. Then load these binary files in child process instead of looping original json file again

JingfeiPeng:
Addressed in PR #681

PR #681, opened 2021-03-15
Store mission objects

iman512003:
This is to address #640 .


smarts/core/utils/traffic_history_serivce.py
```diff
@@ -163,6 +166,18 @@ def fetch_history_at_timestep(self, timestep: str):
 
     @staticmethod
     def fetch_agent_missions(history_file_path):
+        history_mission_filepath = os.path.join(
+            os.path.dirname(utils.__file__), "temp_vehicle_mission.pkl"
+        )
```
JingfeiPeng, line 171: Perhaps this get name operation could be made into a function in __init__.py of utils
```suggestion
        )
        def get_path_name(file_name):
            return os.path.join(
               os.path.dirname(utils.__file__), file_name
            )
```

then here
`history_mission_filepath = get_path_name("temp_vehicle_mission.pkl")`

smarts/core/utils/traffic_history_serivce.py
```diff
@@ -177,5 +192,8 @@ def fetch_agent_missions(history_file_path):
                         goal=scenario.EndlessGoal(),
                         start_time=float(t),
                     )
+        history_mission[history_file_path] = vehicle_missions
+        with open(history_mission_filepath, "wb") as f:
+            pickle.dump(history_mission, f)
```
JingfeiPeng, line 197: since each `agent_missions` object could be big, it might be helpful to use history_mission as a set to store paths to `agent_missions` pickle files instead of storing all `agent_missions` in 1 file. 

So this would be the structure of util folder
 utils 
      - temp_vehicle_mission.pkl 
      - tmp_agent_mission_folder
        -  agent_mission_for_scenario_A 
        -  agent_mission_for_scenario_B
So temp_vehicle_mission.pkl would store `("tmp_agent_mission_folder/agent_mission_for_scenario_A", "tmp_agent_mission_folder/agent_mission_for_scenario_B")`. 

However, I am not sure if `agent_missions` would be that big for this to be important.

Gamenot:
@iman512003 , @JingfeiPeng , and I had a conversation about the direction this should go.

- Store the missions `.pkl` inside the used scenario directory this way `scl scenario clean` can get to it
- Fix the spelling error `traffic_history_serivce.py`

smarts/core/utils/traffic_history_serivce.py
iman512003: Thanks,

JingfeiPeng (approved):

smarts/core/utils/traffic_history_service.py
```diff
@@ -162,7 +164,21 @@ def fetch_history_at_timestep(self, timestep: str):
         return {}
 
     @staticmethod
-    def fetch_agent_missions(history_file_path):
+    def fetch_agent_missions(history_file_path: str, scenario_root_path: str):
+        assert os.path.isdir(scenario_root_path)
+        history_mission_filepath = os.path.join(
+            scenario_root_path, "temp_vehicle_mission.pkl"
```
Gamenot, line 170: nit: I think this file does not fully describe what it is for. Perhaps `history_missions.pkl`?

Gamenot (approved):
Please rebase on develop and add to the change log.

PR #683, opened 2021-03-16
Add custom lane offset ranges in the ULTRA task configs

christianjans:
Allows for the ability to set the vehicle's initial position in the start lane, and also its end position in the end lane by specifying a range [min, max] for the `start_offset` and `end_offset`.

When generating a scenario, the missions generate a random start offset integer and end offset integer in the range [min, max]. This determines how far along the start lane the vehicle will be spawned in (in the case of the start offset), and how far along the end lane the vehicle has to go to reach its goal (in the case of the end offset).

@AlexLewandowski and @JenishPatel99, should `start_offset` and `end_offset` be optional fields for `ego_missions` or should these be explicitly defined (like they are now)? Let me know your thoughts on this, and anything else regarding these changes.

### DIFF
--- a/ultra/tests/scenarios/task00-multiagent/config.yaml
+++ b/ultra/tests/scenarios/task00-multiagent/config.yaml
@@ -5,10 +5,16 @@ levels:
       ego_missions:
       - start: south-SN  # Turn left going from South to West.
         end:   west-EW
+        start_offset: [50, 120]
+        end_offset:   [50, 120]
       - start: west-WE   # Turn right going from West to South.
         end:   south-NS
+        start_offset: [50, 120]
+        end_offset:   [50, 120]
       - start: east-EW   # Turn left going from East to South.
         end:   south-NS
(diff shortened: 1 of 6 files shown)


JenishPatel99:
@christianjans I believe it is a nice fine adjustment to control the initial position of the vehicle. Although, I may feel that the config files may get a bit crowded, and for cases where people may choose not to change the offset (from default), we should provide a default setting; i.e if there is no offset values provided in the config files

christianjans:
> @christianjans I believe it is a nice fine adjustment to control the initial position of the vehicle. Although, I may feel that the config files may get a bit crowded, and for cases where people may choose not to change the offset (from default), we should provide a default setting; i.e if there is no offset values provided in the config files then you use the defaults

Oh good call, I will ensure there is a default and remove the changes in the configs.

AlexLewandowski (approved):

JenishPatel99 (approved):

PR #691, opened 2021-03-17
Observation dictionary

### LINKED ISSUE #548: Make observation gym friendly
TODO : add more description

Adaickalavan:
1. Smarts environment was ensured to organically produce nested observations which are type checked. The observations only consists of `List`, `Tuple`, and `NamedTuple` data. 

1. When a class object is pickled, both its data and accompanying methods are pickled. This causes pickling to slow down and to bloat.

1. Wherever possible, the following was adhered to in order to make the code scalable, flexible, and faster.
    + Data is decoupled from methods
    + Inheritance is avoided, except for implementation of interfaces

1. An `obs_config.yaml` file is introduced. If provided, the dynamic observation lengths will either be truncated or padded to produce the fixed sizes denoted in the `obs_config.yaml` file. If `obs_config.yaml` file is not provided, Smarts will return observations with varying lengths at different time steps. 

Closes #548 .


Gamenot:
Since this will break the observations, the merge will need to be deferred until `0.5`.

sah-huawei:
@Adaickalavan Instead of splitting the methods out into a different class, did you consider using a pattern like:
```python
@dataclass(frozen=True)
class Waypoint:
    pos: np.ndarray
    heading: Heading
    lane_width: float
    speed_limit: float
    lane_id: str
    lane_index: int

    def dist_to(self, p):
        """Calculates straight line distance to the given 2D point"""
        return np.linalg.norm(self.pos - vec_2d(p))

    def relative_heading(self, h: Heading):
        etc

    def serialize(self):
        return cloudpickle.dumps(self.__dict__)

    @classmethod
    def deserialize(cls, serialized_obj):
       return cls(**pickle.loads(serialized_obj))
```
(Note: I didn't test this -- it might require some tweaking to get it to work.  I'm just trying to show the basic idea here.)

And then in `remote_agent.py`, you would do:
```
    def act(self, obs):
        # Run task asynchronously and return a Future.
        self._act_future = self._worker_stub.act.future(
            worker_pb2.Observation(payload=obs.serialize())
        )
```

sah-huawei:
> @Adaickalavan Instead of splitting the methods out into a different class, did you consider using a pattern like:
> 
> ```python
> @dataclass(frozen=True)
> class Waypoint:
>     pos: np.ndarray
>     heading: Heading
>     lane_width: float
>     speed_limit: float
>     lane_id: str
>     lane_index: int
> 
>     def dist_to(self, p):
>         """Calculates straight line distance to the given 2D point"""
>         return np.linalg.norm(self.pos - vec_2d(p))
> 
>     def relative_heading(self, h: Heading):
>         etc
> 
>     def serialize(self):
>         return cloudpickle.dumps(self.__dict__)
> 
>     @classmethod
>     def deserialize(cls, serialized_obj):
>        return cls(**pickle.loads(serialized_obj))
> ```
> 
> (Note: I didn't test this -- it might require some tweaking to get it to work. I'm just trying to show the basic idea here.)
> 
> And then in `remote_agent.py`, you would do:
> 
> ```python
>     def act(self, obs):
>         # Run task asynchronously and return a Future.
>         self._act_future = self._worker_stub.act.future(
>             worker_pb2.Observation(payload=obs.serialize())
>         )
> ```
> 
> I haven't yet gone through this review entirely, so there may be good reasons to do it the other way. I just wanted to get the suggestion/question out now, because I might not be able to get through the review for a little bit.

If the above works, you might then create a "Serializable"  base class that contains the `serialize` and `deserialize` methods, and have any data object that needs to be sent to the remote agent inherit from that.

sah-huawei:
after coming back to this, I guess you will need to include the object type in the serialized dict as well such that you can determine what type of object to create when deserializing.  I'm guessing this was the reason you opted for the other pattern, as it does add a little ugliness to the pattern.  However, I wonder whether it might still be worth doing, just for the sake of avoiding changing all of the observation code.  (This PR would be much shorter at least!)  :)  Something to consider...

Adaickalavan:
1. On the issue of splitting methods and data:
I think to achieve good code scalability and flexibility (inspired by GOlang coding philosophy)
    + keep data (in struct like objects) and methods/functions separate
    + more functional programming and less OOP style code
    + OOP style is more difficult to scale and to parallelize as we have to be careful of statefulness/side-effects

2. On the issue of inheriting from a “Serializable” base class:
I prefer to avoid it here, as
    + multiple inheritance does not work well for type `NamedTuple`, e.g., `def Observation(Serializable, NamedTuple):` 
    + inheritance brings difficulty in hierarchy/dependency tracking (inspired by GOlang coding philosophy)

3. On the issue of including class type in the serialized data:
Conceptually, I think we should only send the observation data to the remote agent and we should not be bound to recreate the corresponding class methods inside the remote agent. 
    + we should make no representation of the corresponding methods when we send the data to the remote agent.

sah-huawei:
(inline)

>     1. On the issue of splitting methods and data:
>        I think to achieve good code scalability and flexibility (inspired by GOlang coding philosophy)
>        
>        * keep data (in struct like objects) and methods/functions separate
>        * more functional programming and less OOP style code
>        * OOP style is more difficult to scale and to parallelize as we have to be careful of statefulness/side-effects

Except this isn't the pattern we've been following so far, so it's a fairly big change and makes our codebase inconsistent to do it partially, and probably not necessary yet. (And while I disagree about Parallelization and Scaling in OOP, it's not worth debating here.)  For better and worse, SMARTS is squarely OOP.

>     2. On the issue of inheriting from a “Serializable” base class:
>        I prefer to avoid it here, as
>        
>        * multiple inheritance does not work well for type `NamedTuple`, e.g., `def Observation(Serializable, NamedTuple):`
>        * inheritance brings difficulty in hierarchy/dependency tracking (inspired by GOlang coding philosophy)

Yeah, I agree about avoiding multiple inheritance.  Of course, it's easy enough to achieve the same effect (while avoiding touching so much of the code) by adding something like a `Serializer` class (not a base class) with static `serialize()` and `deserialize()` methods.

>     3. On the issue of including class type in the serialized data:
>        Conceptually, I think we should only send the observation data to the remote agent and we should not be bound to recreate the corresponding class methods inside the remote agent.
>        * we should make no representation of the corresponding methods when we send the data to the remote agent.

cloudpickle is _already_ sending the object types to the remote agent.  That's how it's able to unpickle things there.  And there's no need to send any other methods.  And we're not doing any sort of monkey patching of our objects as far as I know. 

Anyway, I'd be happy to discuss further in any case, but I suspect the above questions may be obsoleted anyway by your other (proto) PR, so I'll check that out first.

PR #697, opened 2021-03-19
Add hyperparameter tuning to ULTRA

### LINKED ISSUE #656: Ultra GridSearch
Adding gridSearch ability to train.py

this requires some changes to params.yaml, here is an example of the config we need to pass to tune.run():
```son
config = {
    "encoder": tune.sample_from(["no_encoder", "precog_encoder", "pointnet_encoder"]),
    "l2": tune.sample_from(lambda _: 2**np.random.randint(2, 9)),
    "lr": tune.loguniform(1e-4, 1e-1),
    "batch_size": tune.choice([2, 4, 8, 16])
}
```

At each trial, Ray Tune will randomly sample a combination of parameters from these search spaces. It will then train a number of models in parallel and find the best performing one among these. 

- We need to use the a scheduler to terminate bad performing trials early.
- We also need a reporter to report metrics to tune.
- Evaluation design needs to be reconsidered [ we might need to treat it like RLlib]

Tasks:
- [x] change params.yaml to params.py and change them to tune objects for all baselines
  - Have a separate params.yaml and tune_params.py
- [x] decide on model termination criteria `get_best_trial("loss", "min", "last")`
  - You can specify this in the command line flags
- [x] check Tensorboard handles and make sure they're descriptive enough [indicating which experiment is related to which plots? if the files are saved correctly in directories of experiments]
- [x] report the best model at then end of search 
- [x] figure out if we need the `reporter = CLIReporter(metric_columns=["loss", "accuracy", "training_iteration"])` it might be necessary for early termination of training trials
  - `CLIReporter` is the default `progress_reporter`, its output can be suppressed by specifying the `log_to_file` parameter or the `verbose` parameter
  - The scheduler is what determines early termination of trials
- [x] handle evaluation in each training [ do we need to evaluate for all trials or we keep evaluation just for the best result?] @AlexLewandowski we need to discuss evaluation for grid search
  - Solution: Do evaluation with the best config after with the best config is found
  - The best config is saved as a .yaml so we could just use this file and run evaluate.py?
   

@christianjans @JenishPatel99 I am still in the middle of implementing things but feel free to take a look at `ultra-gs` ppo, we can discuss it in details

christianjans:
Will close #656. Work taken from https://github.com/huawei-noah/SMARTS/tree/ultra-gs.


ultra/tests/test_tune.py
```diff
@@ -0,0 +1,112 @@
...
+        POLICY = "ppo"
+        EPISODES = 2
+        MAX_EPISODE_STEPS = 50
+        TIMESTEP = 0.1
+        HEADLESS = True
+        EVAL_EPISODES = 2
+        SAVE_RATE = 1
+        SEED = 2
+        LOG_DIR = os.path.join(TuneTest.OUTPUT_DIRECTORY, "tune_logs/")
+        CONFIG_MODULE = "ultra.baselines.ppo.ppo.tune_params"
```
JenishPatel99, line 47: Do we want to create a separate config in the test directory, instead of using the baseline one? I know it will not change anything drastically because the result params will be stored in a different file, but it will ensure that the tests are a standalone thing

ultra/ultra/baselines/agent_spec.py
```diff
@@ -79,23 +81,18 @@ def __new__(
...
+            # If policy_params is None, then there must be a params.yaml file in the same
+            # directory as the policy_class module.
+            if not policy_params:
+                policy_class_module_file = inspect.getfile(policy_class)
+                policy_class_module_directory = os.path.dirname(
+                    policy_class_module_file
+                )
+                policy_params = load_yaml(
+                    os.path.join(policy_class_module_directory, "params.yaml")
+                )
```
JenishPatel99, line 93: Will this implementation work with the regular train script? For static config files do we want to add a separate config path to the agent_pool?

JenishPatel99 (approved):

ultra/tests/test_tune.py
christianjans: Oh yes, great idea!

ultra/docs/tune.md
```diff
@@ -0,0 +1,53 @@
...
+- `--episodes`: The number of training episodes to perform for each sampled config (default is 10000).
+- `--max-episode-steps`: The maximum number of steps the agent is allowed to take per episode (default is 10000).
+- `--timestep`: The environment timestep in seconds (default is 0.1).
+- `--headless`: Whether to run tuning without Envision (default is True).
+- `--eval-episodes`: The number of evaluation episodes to perform on the best-performing config after tuning is complete (default is 200).
+- `--save-rate`: How often to save the agent, measured in episodes (default is 1000).
+- `--seed`: The environment seed (default is 2).
+- `--log-dir`: The directory to put the tune experiment's data (default is tune_logs/).
+- `--config-module`: The module containing a dictionary variable, 'config', that defines the config to tune (default is ultra.baselines.ppo.ppo.tune_params).
+- `--metric`: The metric to optimize for; either episode_length, episode_reward, or env_score (default is episode_reward).
```
AlexLewandowski, line 47: I think with the recent PR, these should be episode_return instead of episode_reward

AlexLewandowski (approved):

PR #706, opened 2021-03-24
Showbase refactor

### LINKED ISSUE #597: Turn `Showbase` inheritance into composition via singleton
Context #184 

Convert from `SMARTS` inheriting `Showbase` to `SMARTS` just using a Showbase instance.

- [ ] Partition rendering instance so that SMARTS instances do not share information via camera observations
- [ ] Unique identifiers for SMARTS instances for accessing Showbase nodes.
- [ ] Close down `Showbase` instance when all SMARTS instances are closed

sah-huawei:
Refactor to not inherit from ShowBase.  Closes #597.

Also, factored out a "Provider" base class that all Providers inherit because Ioriginally hoped to make the new Renderer a Provider too, but changed my mind because it didn't quite fit the concept.  But I kept the base class anyway to help clarify what currently exists for future refactors.


smarts/core/renderer.py
```diff
@@ -0,0 +1,261 @@
...
+            # https://www.panda3d.org/manual/?title=Multithreaded_Render_Pipeline
+            # loadPrcFileData('', 'threading-model Cull/Draw')
+            # have makeTextureBuffer create a visible window
+            # loadPrcFileData('', 'show-buffers true')
+        it = cls.__dict__.get("__it__")
+        if it is None:
+            cls.__it__ = it = object.__new__(cls)
+            it.init(timestep_sec)
+        else:
+            assert timestep_sec == it._timestep_sec
```
JingfeiPeng, line 71: question: for the subsequent times after the first time we create `_ShowBaseInstance`, do we allow users to create _ShowBaseInstance (or create smarts, since creating smarts creates _ShowBaseInstance) again with a different initial `timestep_sec`?

JingfeiPeng: If we don't allow that, maybe it is good add a message to let the user know

```suggestion
            assert timestep_sec == it._timestep_sec, "Running multiple instances of SMARTS must use same timestep second"
```

smarts/core/provider.py
```diff
@@ -46,3 +48,29 @@ def filter(self, vehicle_ids):
                 del self.vehicles[index]
             except ValueError:
                 continue
+
+
+class Provider:
+    """A Provider tracks a (sub)set of vehicles that all share the same action space.
+    This is a base class (interface) from which all Providers should inherit."""
```
JingfeiPeng, line 55: Should we make the existing providers inherit from this interface? So from `SumoTrafficSimulation`, `MotionPlannerProvider` and `TrafficHistoryProvider`

JingfeiPeng (approved):

smarts/core/renderer.py
sah-huawei: Agree.  Will add the message.
(BTW, as far as I know, we don't have any circumstances where multiple SMARTS instances are created simultaneously, which is where this would be an issue.  Back-to-back, but not overlapping usage of `_ShowBaseInstance` with different values for `timestep_sec` should be fine.)

smarts/core/provider.py
sah-huawei: Yes.  I thought I had!  Hmm... I think I went too far when I backed out the code where `Renderer` inherited it.  I will re-add.

smarts/core/renderer.py
```diff
@@ -0,0 +1,263 @@
...
+        lens = OrthographicLens()
+        lens.setFilmSize(width * resolution, height * resolution)
+
+        camera_np = self._showbase_instance.makeCamera(
+            buffer, camName=name, scene=self._root_np, lens=lens
+        )
+        camera_np.reparentTo(self._root_np)
+
+        # mask is set to make undesireable objects invisible to this camera
+        camera_np.node().setCameraMask(camera_np.node().getCameraMask() & mask)
```
Gamenot, line 261: I think my concern here is if this camera will see all vehicles even if it is from other SMARTS instances. I think that is the case with the current configuration.

sah-huawei: I thought the `scene=self._root_np` part of the `makeCamera()` call would limit what the camera sees to just the vehicle's scene.  However, I'm not sure of this (the documentation is unclear) and couldn't test it.  If so, this is different than the current situation because there we only have one root node under `render` called "sim", whereas here, each simulation has its own uniquely-named root node.

smarts/core/vehicle.py
```diff
@@ -507,9 +491,9 @@ def step(self, current_simulation_time):
     def control(self, *args, **kwargs):
         self._chassis.control(*args, **kwargs)
 
-    def sync_to_panda3d(self):
+    def sync_to_renderer(self):
         pos, heading = self._chassis.pose.as_panda3d()
-        self._np.setPosHpr(*pos, heading, 0, 0)
+        self._renderer_path.setPosHpr(*pos, heading, 0, 0)
```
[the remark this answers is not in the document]
sah-huawei: ok, I think I agree (I was, overall, unhappy with leaving `_renderer_path` in the `Vehicle` class, although moving it would have caused more/bigger ripples. 

 I assume you are suggesting that we add a dictionary to track vehicle nodes to the `Renderer` class, right?  (This can be keyed on the vehicle-id (`vid`) and populated by the call to `create_vehicle()` there, which I'm thinking to rename `create_vehicle_node` btw,)  This would require forwarding the `sync_to_renderer()` call to the `Renderer` object from each `Vehicle` object.  All of this is doable and cleaner, in my opinion.  So assuming this is what you meant, we agree.

I had considered this path, but stopped short when I saw `_follow_vehicle()` in the camera sensors (`sensors.py`) which is uglier / more tangled.  That was enough to scare me away from trying to do the above because I realized to truly get it "clean" would be harder.  Of course I could still do what I wrote in the previous paragraph and leave `follow_camera()` alone, but it seems to undermine the goal . Do you have thoughts/suggestions about this?

smarts/core/renderer.py
sah-huawei: Regarding your concern, notwithstanding my comment which I think means the scene graphs can be kept separate for different SMARTS instances, I do think it's valid:  that is, I don't think we will ever be able to overlap multiple, simultaneous SMARTS instances until we (a) make our own simulation clock in SMARTS cf. the comment I added there in `step()` and (b) figure out a way to cause ShowBase to render only one branch of the scene graph at a time.  I just added a couple more comments here in the Renderer class to try to clarify that.

Gamenot: I think there are a few things I can address here. Cameras in game engines are attached to a node in order to inherit the transformations (and other attributes) of the parent node. This generally means if the parent note is transformed (e.g. translation) the camera will maintain relative distance. The camera will still see all other renderable nodes regardless of existing on a different branch of the scene graph. From what I can see all of the nodes are connected via the the common root of the showbase `render` node.

I think there may be a few possible workarounds. But using `hide()` on each of the other simulation roots before rendering and then calling `show()` after might work best: https://docs.panda3d.org/1.10/python/reference/panda3d.core.NodePath?highlight=nodepath#panda3d.core.NodePath.hide

smarts/core/vehicle.py
```diff
@@ -149,25 +147,15 @@ def __init__(
             self._color = config.color
 
         # TODO: Move this into the VehicleGeometry class
-        self._np = self._build_model(pose, config, showbase)
+        self._renderer_path = renderer.create_vehicle(
```
Gamenot, line 150: I might consider naming this `_renderable`, renderer has the meaning of being the driver of rendering rather than an item to be rendered.

smarts/core/sensors.py
```diff
@@ -588,68 +579,39 @@ def steps_completed(self):
...
 
-    def _follow_vehicle(self, camera_np):
+    def _follow_vehicle(self):
         center = self._vehicle.position
         largest_dim = max(self._vehicle._chassis.dimensions.as_lwh)
+        camera_np = self._camera.camera_np
         camera_np.setPos(center[0], center[1], 20 * largest_dim)
         camera_np.lookAt(*center)
-        camera_np.setH(self._vehicle._np.getH())
+        camera_np.setH(self._vehicle.renderer_path.getH())
```
Gamenot, line 614: It should be possible to replace the heading with heading directly from the vehicle rather than from the vehicle renderable node.

smarts/core/sensors.py
```diff
@@ -588,68 +579,39 @@ def steps_completed(self):
...
+    def step(self):
+        self._follow_vehicle()
 
-    def _follow_vehicle(self, camera_np):
+    def _follow_vehicle(self):
         center = self._vehicle.position
         largest_dim = max(self._vehicle._chassis.dimensions.as_lwh)
+        camera_np = self._camera.camera_np
         camera_np.setPos(center[0], center[1], 20 * largest_dim)
         camera_np.lookAt(*center)
```
Gamenot, line 613: I think in this case it might be OK to continue to use the `NodePath` here directly. Cameras have a tendency to have privileged access to the renderer.

smarts/core/vehicle.py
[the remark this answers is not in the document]
Gamenot: > I assume you are suggesting that we add a dictionary to track vehicle nodes to the Renderer class, right? (This can be keyed on the vehicle-id (vid) and populated by the call to create_vehicle() there, which I'm thinking to rename create_vehicle_node btw,) This would require forwarding the sync_to_renderer() call to the Renderer object from each Vehicle object. All of this is doable and cleaner, in my opinion. So assuming this is what you meant, we agree.
Yes, renderables should submit to the renderer. I might even go further and remove the Renderer reference from the vehicle and instead force a batched position update.

I think leaving the rest of the camera alone could be fine after removing the `vehicle.renderable_path` use, cameras tend to have privileged access to the renderer. It is usually fine to allow that privileged access because when you manipulate a camera it should generally have no other effect except for a change in the attributes used by the renderer. I might prefer calling the guts of the `_wait_for_ram_image()` inside the renderer as well rather than the camera but it is fairly contained for now.

smarts/core/renderer.py
sah-huawei: I think `Camera.setScene` is the same as passing the `scene` parameter to the `makeCamera()` call, as I'm already doing. So I thnk we are already good because there is currently a difffernt scene root (all children of `render`) for each simulation.  But I can add the `hide()` and `show()` calls to `Renderer.render()`, or try with masks, if you like that alternative better.  However, note that all of these only solve issue (b) of my previous comment, issue (a) will still exist until we fix the clock.  If you think it's not biting off too much, I can go ahead and try to fix the clock too with this PR.

smarts/core/vehicle.py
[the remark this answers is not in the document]
sah-huawei: OK.  Will do something along these lines.

smarts/core/vehicle.py
sah-huawei: Yes, I agree.  That's better.  Will do.

smarts/core/sensors.py
sah-huawei: Good!  will do.

smarts/core/vehicle.py
sah-huawei: Actually, there's no need since `renderer_path` has been removed completely.

Gamenot (approved):

PR #712, opened 2021-03-26
Proto schema for observation

### LINKED ISSUE #680: Streamline communication w/ remote agent
As part of the profiling done for Issue #641, it was seen that approximately 50% of our current step time comes from pickling observations to be sent to the remote agent's act method.   This can be improved.

As a first step, @Adaickalavan is ensuring that cloudpickle will only serialize the actual observation _data_.

Once this is finished, we should profile again and see if more optimization is needed.

Adaickalavan:
**Notes**
1. Implemented proto schema for observations sent to remote agents.
1. gRPC proto results in better serialization.
1. With proto schema, security concerns can be allayed, as only data fitting the schema can be sent to the remote agent.
1. The environment only sends observation data to agents. Any corresponding methods are not sent to agents. 
1. `None` is of `NoneType`, and this does not have simple counterpart in strongly typed platforms.

**Performance improvement**
1. Setup
    + compare with before at e9964a7fed838a80e3a5cc980da9405a8d865793
    + test case: `python3.7 ./examples/single_agent.py ./scenarios/intersections/6lane/ --headless`

1. cProfile timing
Result: Per episode runtime improved by ~1.8 seconds.

    |Before|After|
    |:----:|:----:|
    | 40 seconds | 22 seconds |

1. Flamegraph
    ```bash
    $ cd <path/to/>SMARTS
    $ pip install git+https://github.com/asokoloski/python-flamegraph.git
    $ export script= ./examples/single_agent.py
    $ export scenario=./scenarios/intersections/6lane/
    $ make flamegraph
    ```

    Result: Percentage of total time taken by `act@remote_agent.py` decreased by ~4.9 times from 43.51% to 8.94%.

    _Before:_
    image

    _After:_
    image

**Future work:**
1. If possible variables, such as `position`, should be standardized to use np.array() instead of tuples.
1. Proto schema needs to be set for remote agent build specification and remote agent actions.
1. Any remaining type incompatibility and variability is expected to be ironed out when RL frameworks are interfaced with the SMARTS observation output.
 
Closes #680.


Gamenot:
@Adaickalavan. @sah-huawei  and I will review this pull request with the intention to filter out the non-bugfix interface changes.

Adaickalavan:
Perhaps you can do this on #787, instead of here, as types/issues/conflicts/organization tend to be fixed on newer pull requests.

PR #713, opened 2021-03-26
ULTRA Integrate zoo agents

JenishPatel99:


christianjans:
This feature is highly dependent on the outcome of #843. Monitor the progress of this issue to determine if this feature should still be completed.

If SMARTS' Agent class is removed, then we should create an ULTRA Agent class that can be used by anyone who wants to use ULTRA. An Agent class for ULTRA makes more sense than an Agent class in SMARTS, since ULTRA's training and evaluation requires a specific interface (methods and attributes) from the agent. The implementation of this ULTRA agent however, will be dependent on how #843 is resolved.

PR #727, opened 2021-03-31
map offset fixes

### LINKED ISSUE #716: Vehicle locations in sumo-gui does not match vehicle positions on envision
## **BUG REPORT**
**High Level Description**
The vehicle positions in sumo_gui are different from vehicle postions in envision

**SMARTS version**
0.4.15

This might be related to previous map location offset PR #673 

checkout `multi_agent_demo` branch, commit id: 4aee20350b97899bf202b6f3963a9f0b401b53c6 and run 
"python examples/game_of_tag.py scenarios/demo/"
Though vehicle positions in other scenarios worked fine. This might be related to the map creation

vehicle locations on sumo-gui should match postions in Envision

**Error logs and screenshots**
The envision vehicle position looks correct
image
sumo_gui vehicle positions are incorrect
image

sah-huawei:
Use possibly-shifted road network file when starting Sumo.  Note that we now keep around the shifted net file as `shifted_map-AUTOGEN.net.xml` in the same folder where `map.net.xml` was.  
This PR assumes that `sumo` (or `sumo-gui`) are started by SMARTS (within `sumo_traffic_simulation.py`).  There may still be problems if an external simulator uses the original (unshifted) `map.net.xml` directly, either when it starts Sumo or otherwise.  (But this risk may be mitigated if/when we refactor the sumo bridge to be controlled by SMARTS.)
Closes #716.

Also some snake_case variable_name style fixes along the way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ the missions for all agents.
 we now initially shift road networks (maps) that are offset back to the origin
 using [netconvert](https://sumo.dlr.de/docs/netconvert.html).
 We adapt Sumo vehicle positions to take this into account to allow Sumo to continue
-using the original coordinate system.  See Issue #325.
+using the original coordinate system.  See Issue #325 and #716.
 - Cleanly close down the traffic history provider thread. See PR #665.
 - Improved the disposal of a SMARTS instance. See issue #378.
 - Envision now resumes from current frame after un-pausing.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ we now initially shift road networks (maps) that are offset back to the origin
 using [netconvert](https://sumo.dlr.de/docs/netconvert.html).
 We adapt Sumo vehicle positions to take this into account to allow Sumo to continue
 using the original coordinate system.  See Issue #325.
+    - This fix will require all Scenarios to be rebuilt (`scl scenario build-all --clean ./scenarios`).
 - Cleanly close down the traffic history provider thread. See PR #665.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -33,7 +33,7 @@ the missions for all agents.
 we now initially shift road networks (maps) that are offset back to the origin
 using [netconvert](https://sumo.dlr.de/docs/netconvert.html).
 We adapt Sumo vehicle positions to take this into account to allow Sumo to continue
-using the original coordinate system.  See Issue #325.
+using the original coordinate system.  See Issue #325 and #716.
```
Gamenot, line 36: I think this change might require a Fixed entry under `[Unreleased]`.

smarts/core/scenario.py
```diff
@@ -530,8 +527,8 @@ def discover_traffic_histories(scenario_root):
             return []
 
         traffic_histories = []
-        with open(path, "rb") as f:
-            files = pickle.load(f)
+        with open(path, "rb") as inf:
+            files = pickle.load(inf)
```
JingfeiPeng, line 531: just curious what is `inf` an abbreviation for? My first reaction is "infinity"....

Gamenot: My guess is in_file, although `inf` does have that special meaning for numbers.

JingfeiPeng (approved):

smarts/core/scenario.py
sah-huawei: Yeah, Tucker is right.  It was supposed to be "input_file".  The reason for the change was because `f` was being used in the list comprehension immediately following this code, so I wanted to avoid any possible confusion.

PR #729, opened 2021-03-31
Renderer Optional

### LINKED ISSUE #725: Start SMARTS without x11
SMARTS will crash due to not having x11 but many of our users do not use image rendering.

Default to no rendering if x11 is not available.

**Describe alternatives you've considered**
Allow a toggle of rendering.

**Additional context**
#696 #184 
Most context: #605

sah-huawei:
Changed SMARTS to only load the `Renderer` (`ShowBase`) when a Vehicle tries to attach a camera-based sensor.  At that point, it will either fail (raise) because `ShowBase` isn't supported, or begin rendering the scene with all other vehicles too.  This way, if someone does not have an external display attached, Smarts should run ok assuming they don't try to use camera-based sensors.

Closes #725.


smarts/core/smarts.py
```diff
@@ -711,16 +724,10 @@ def neighborhood_vehicles_around_vehicle(self, vehicle, radius=None):
...
-                    for c in self._vehicle_collisions[vehicle_id]
-                    if c.collidee_id != self._ground_bullet_id
-                ]
-            )
-            > 0
-        )
+        for c in self._vehicle_collisions[vehicle_id]:
+            if c.collidee_id != self._ground_bullet_id:
+                return True
+        return False
```
sah-huawei, line 731: oops.  This is an unrelated change (very minor optimization) that I accidentally left in.  I think it's harmless, but I'm happy to take it back out of this PR if you think I should.

smarts/core/smarts.py
```diff
@@ -73,12 +73,21 @@ def __init__(
         timestep_sec=0.1,
         reset_agents_only=False,
         zoo_addrs=None,
+        renderer_optional=False,
```
Gamenot, line 76: If the goal is to make sure that setup is easy even for people that will not use rendering, perhaps even with the option of not installing `panda3d`, we may want this optional to be `True`.

Gamenot: Even with disregard to the former point, this still causes issues with basic setup since there is no configuration of the rendering through either of the environments.

smarts/core/smarts.py
```diff
@@ -73,12 +73,21 @@ def __init__(
...
         self._scenario: Scenario = None
-        self._renderer = Renderer(self._sim_id)
+        try:
+            self._renderer = Renderer(self._sim_id)
+        except Exception as e:
+            if not renderer_optional:
+                raise e
+            self._log.info(
+                f"Unable to create Renderer; continuing without one.\nException was:  {e}"
+            )
```
Gamenot, line 89: I would be tempted to use a warning rather than info here although I see that the warning is instead given in the sensors. That may be OK.

smarts/core/vehicle.py
```diff
@@ -431,6 +432,10 @@ def attach_sensors_to_vehicle(sim, vehicle, agent_interface, mission_planner):
             )
 
         if agent_interface.drivable_area_grid_map:
+            if not sim.renderer:
+                raise Exception(
+                    "A renderer is required to add a drivable_area_grid_map.  Ensure that `renderer_optional=False` is set when creating SMARTS instance."
+                )
```
Gamenot, line 438: All of the sensors use the same exception. Maybe this is a case for a new exception type?

Gamenot (approved):

smarts/core/smarts.py
sah-huawei: I agree that we may eventually want to make this default to `True`, although I didn't do it yet because I didn't want to break/change any existing tests or behaviours.  However, I can go ahead and make this change all the way through (including ensuring that the sanity tests all pass without a renderer).  

I will add something for this to the environment interfaces too.

smarts/core/smarts.py
sah-huawei: Yeah, I was torn about this.  I chose `info` because if they explicitly specified `renderer_optional=True`, they might not consider the absence of Panda3D as something to be worried about.

smarts/core/vehicle.py
sah-huawei: Good call.  Will add.

.gitignore
```diff
@@ -97,6 +97,7 @@ venv
 *.rou.alt.xml
 *.rou.xml
 *.trips.xml
+shifted_map-AUTOGEN.net.xml
```
sah-huawei, line 100: This should have been added in the `map_offset_fixes` PR (#727).

smarts/core/smarts.py
sah-huawei: Actually, as we just discussed offline, I'm going to try to go either further, only adding a Renderer to Smarts on demand  if there's a sensor that requires it (and getting rid of the `renderer_optional` parameter altogether).  Then I'll try to make the Panda3d dependency optional.

JingfeiPeng (approved):

PR #737, opened 2021-04-05
Fix headless argument behaviour in ULTRA

### LINKED ISSUE #736: Headless argument in ULTRA is not parsed correctly
## **BUG REPORT**
The `--headless` flag in ULTRA's scripts will always be `True` due to a [known issue](https://bugs.python.org/issue14392) in Python's argparse.

christianjans:
The `--headless` argument would always be True due to a known issue in Python's argparse library.

**Before:** These changes ensure that if `--headless False` is specified, the ULTRA script runs in non-headless mode; and if `--headless True` is specified or the headless flag is not set at all, the ULTRA script runs in headless mode.

**Now:** These changes ensure that the behaviour of the `--headless` flag is the same as in SMARTS. That is, if a script is run without the `--headless` flag, the script will run in non-headless mode (it will use Envision => the experiment can be viewed in a browser). If `--headless` is included as a flag, the script will run in headless mode (Envision will not be used => the experiment cannot be viewed in a browser).

Will close #736.


AlexLewandowski (approved):

Gamenot (approved):
I think this might add some confusion between the different ways of using `--headless` so I might suggest also changing the name of the argument.

JenishPatel99 (approved):

PR #739, opened 2021-04-05
Control number of episodes in ULTRA tests (rllib and train)

AlexLewandowski:
This speeds up tests, and provides more customization for setting up rllib_train. The changes are needed for eval-per-episode in particular, because the eval_check is done at init. As a result, test_train triggered evaluation_check and produced 200 rollouts.


christianjans:
I think these changes include the parallel evaluation and eval-per-episode branches as well. Are both supposed to be included?

ultra/ultra/rllib_train.py
```diff
@@ -149,6 +154,7 @@ def train(
 
     # Iteration value in trainer.py (self._iterations) is the technically the number of episodes
     for i in range(num_episodes):
+        print("i ASDASD", i)
```
christianjans, line 157: Is this supposed to be here?

ultra/ultra/evaluate.py
```diff
@@ -53,21 +53,22 @@ def evaluation_check(
...
     log_dir,
+    evaluation_task_ids,
 ):
     # Evaluate agents that have reached the eval_rate.
     agent_ids_to_evaluate = [
         agent_id
         for agent_id in agent_ids
-        if (episode.get_itr(agent_id) + 1) % eval_rate == 0
-        and episode.last_eval_iterations[agent_id] != episode.get_itr(agent_id)
+        if episode.index % eval_rate == 0
```
christianjans, line 62: Just curious, should this be:
```suggestion
        if (episode.index + 1) % eval_rate == 0
```
Otherwise, the way it is currently (without the "`+ 1`"), I think it will always evaluate on the first episode (when `episode.index == 0`). Is this what we want?

ultra/ultra/evaluate.py
```diff
@@ -53,21 +53,22 @@ def evaluation_check(
...
     ]
 
     # Skip evaluation if there are no agents needing an evaluation.
     if len(agent_ids_to_evaluate) < 1:
         return
 
-    episode.eval_mode()
-    evaluation_data = {}
+    if eval_episodes < 1:
+        return
```
christianjans, line 71: Great idea moving this check to this function!

ultra/ultra/evaluate.py
```diff
@@ -76,37 +77,83 @@ def evaluation_check(
...
+            log_dir=log_dir,
+            eval_mode=False,
+        )
+        evaluation_task_ids[evaluation_task_id] = (
+            episode.get_itr(agent_id),
+            episode,
+            "eval_train",
+        )
+
+    for agent_id in agent_ids_to_evaluate:
```
christianjans, line 100: There were problems before when we had these `evalute.remote(..)` calls in the same `for agent_id in agent_ids_to_evaluate: ...` loop, correct? I wonder if it works now since the evaluations happen asynchronously? I can test this too.

christianjans (commented):
Looks great! Just a few comments/questions.

ultra/ultra/evaluate.py
AlexLewandowski: I am not 100% sure which issue youre referring to but it would be good to test!

ultra/ultra/evaluate.py
AlexLewandowski: I think it's good practiceto evaluate at initialization, it provides a baseline before any learning happens.

ultra/ultra/rllib_train.py
AlexLewandowski: Whoops, now you see my debugging process.

christianjans: Haha, all good!

ultra/ultra/evaluate.py
christianjans: Sorry yeah, see #744 for the changes I was thinking of. Let me know your thoughts. I didn't run into any problems with those changes either.

JenishPatel99 (approved):

christianjans (approved):

PR #740, opened 2021-04-05
Extra ULTRA scenarios for debugging social-vehicle encoders

adreena:
Adding a new feature to maps for adding stopped vehicles on the south route

- for adding stops we need to specify the number of vehicles we need + their positions and lane
- this change requires some minor modifications to the other distributions/config files related to scenarios
- we can have other social vehicles moving as well or keep it simple by just have a few blocks on the road

@AlexLewandowski @JenishPatel99 @christianjans this is the draft for stopped vehicles, please don't push changes to this code for now and lets finalize the idea on our next meeting

TODO:
- [x] Overwrite `traffic/all.rou.xml` file with new vehicles
- [x] Add `stops` to the other task configs
- [x] Ensure ego vehicles do not hijack stopped vehicles
- [x] Address issue in some scenarios not including stopped vehicles when it should

### DIFF
--- a/ultra/tests/scenarios/task00-multiagent/config.yaml
+++ b/ultra/tests/scenarios/task00-multiagent/config.yaml
@@ -13,6 +13,7 @@ levels:
         2lane_t:
            percent: 1
            specs: [[70kmh,p-test,1]]
+           stops: null
     test:
       total: 1
       ego_missions:
@@ -22,6 +23,7 @@ levels:
         2lane_t:
            percent: 1
            specs: [[70kmh,p-test,1]]
+           stops: null
   no-traffic:
     train:
       total: 1
@@ -36,6 +38,7 @@ levels:
         2lane_t:
            percent: 1
            specs: [[70kmh,no-traffic,1]]
+           stops: null
     test:
       total: 1
       ego_missions:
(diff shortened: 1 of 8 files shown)


AlexLewandowski (commented):
Looks good to me. I wonder if the potential "dynamic" config file that Jenish mentioned could help clean up the yaml files, since most of them dont use the stop parameter.

JenishPatel99 (approved):

christianjans:
> Looks good to me. I wonder if the potential "dynamic" config file that Jenish mentioned could help clean up the yaml files, since most of them dont use the stop parameter.

Yeah good call.

If we wanted the `stops` parameter to be optional before a dynamic config is implemented, we can also just check if a `stops` parameter is provided. This is what happens with the offset parameters for ego missions: https://github.com/huawei-noah/SMARTS/blob/6cb8ffda74e235c25d63b74313623b2e03e402f7/ultra/ultra/scenarios/generate_scenarios.py#L71-L82 Let me know your guys' thoughts on if we want to do this check for `stops` as well so that we don't have to add `stops` to already existing task configs.

But yes definitely, perhaps with a dynamic config, the usage of the offset and stops parameters can be cleaned up.

Issue #407, opened 2021-01-13
NGSIM dataset abnormal replay


zbzhu99:
The data replay of the NGSIM data set on SMARTS seems to have some problems:
1. The overall position of the vehicle is offset downwards
2. The front of the vehicle will swing when it first enters the scene
3. The replay speed is obviously too fast

https://user-images.githubusercontent.com/32090492/104401449-2edf4300-558f-11eb-94a1-eee969bd2ae8.mov

Perhaps the way PPUU processes data can be used as a reference.

This is the NSGIM dataset file used by PPUU: http://bit.ly/PPUU-data

The following is some code of PPUU data processing:

https://github.com/Atcold/pytorch-PPUU/blob/2e087d9ef556cf75723ff2c5f5b71c4bb0b358dc/map_i80.py#L18-L41

JingfeiPeng:
Currently there is some mismatch between the vehicle data and map data. Vehicles are off the route when they are initially spawned, causing the the training to be done at first step (as vehicles are already off route). This could also be why the vehicles are offset downwards. It seems either the map data is incorrect or the vehicle position data is incorrect.

liamchzh:
@zbzhu99 Can you share the map.net.xml that you're using. Please upload it in the comment. Thank you.

Gamenot:
@Gamenot @zbzhu99 Reminder to upload the `map.net.xml` file related to the dataset.

zbzhu99:
Apologies for the delayed response, and here is the map file related to the dataset:


[map.net.xml.zip](https://github.com/huawei-noah/SMARTS/files/6003713/map.net.xml.zip)

sah-huawei:
Just from watching the attached video, it appears that the vehicle locations may be "flipped" from top to bottom (i.e., mirrored around the x-axis).  It appears that vehicles are merging from the top down, as opposed from the bottom up.

sah-huawei:
I've finally addressed most of the things in this issue in PR #741, including flipping the coordinate system around the x-axis.  However, the attached `map.net.xml`still doesn't quite match the trajectories in the NGSIM dataset.  

In particular, it has an extra lane at the bottom as compared to:
- the [image of that stretch of highway](https://github.com/Atcold/pytorch-PPUU/blob/master/I-80/cam2.png) from the PPUU repository;
-  the data dictionary ( `trajectory-data-dictionary.htm`) in the [NGSIM dataset from PPUU](http://bit.ly/PPUU-data);
[...]

PR #741, opened 2021-04-06
traffic history improvements for imitation learning

### LINKED ISSUE #732: Traffic Histories interface bug and scalability
**High Level Description**
When a large traffic history dataset (like NGSIM) is used, a few issues are revealed, including samples being missed (the bug).  This is needed to support large-scale imitation learning.

**SMARTS version**
0.4.15

#402 #407 #596 #640 #226 etc.

**Description of Problems**
Traffic history files are currently supplied in a json format, which wastes a lot of space (e.g. redundantly specifying each field name for each vehicle at each timestep) and also time (having to be _entirely_  read through and parsed on every single batch query, roughy per every 300 entries in the file, even if it's not all kept in memory ).  This slows down Smarts considerably on traffic history playback and isn't scalable.

Further, If entries are out of order in the batches (which we cannot easily/efficiently prevent with the current json-to-dictionary parsing) the provider will miss entries when it queries for them at a time if they don't fit into our "two batch" window that we keep in memory.  I am experiencing this with the NGSIM traffic history dataset for i80.

**Plan**
- Change the supported traffic history file format to a `sqlite` database file instead of json
    - Update and checkin the dataset conversion script(s) to generate this format instead of json
    - Update `traffic_history_provider.py` to query this (instead of the `trafffic_history_service`) to:
        - get all unique vehicles and infer their missions (start time+position) at the beginning of the simulation
        - get which vehicles changed at each timestep
- (Possibly) Support multiple (keyed) histories in the same database file (instead of supporting a list of history filenames)
- Provide a script to convert any existing json history files to sqlite3

sah-huawei:
Changes traffic history file format from JSON to an opaque `.shf` format (currently a `SQLite` file database).

This reduces the size of history files by over 60%, speeds up stepping from the `TrafficHistoryProvider` to under 2ms, and fixes bugs related to issue #407 (missed and skipped samples).  

This obsoletes the `Traffic_history_service` (and associated unit tests of it).

Traffic history database files are automatically created when a `Scenario` specifies them, as before, like:
```
gen_scenario(
    t.Scenario(
        traffic_histories=["i80-0400.yaml", "i80-0500.yaml"],
    ),
    output_dir=Path(__file__).parent,
)
```
However we now support a `yaml` "dataset spec" file as input that contains a pointer to the original dataset to be converted as well as parameters related to the conversion (examples have be added to Issue #732).  This converts and imports the data into the sqlite database when scenarios are first built and obsoletes the previous conversion scripts ( `tools/interaction_dataset_converter.py` and `ngsim_dataset_converter.py`).

Along the way:
- added ability to set default agent speed to the imitation learning "agent replacement" example
- added ability to support different lane widths in off-road checks (as required by the NGSIM dataset, whose lanes are ~6x)
- added ability to convert old JSON history files to the new format
- added ability to run SMARTS in "near real-time" such that traffic can be watched in sumo-gui at correct speed

Closes #732, #407


smarts/core/renderer.py
```diff
@@ -58,6 +58,7 @@ def __new__(cls):
             # disable vsync otherwise we are limited to refresh-rate of screen
             loadPrcFileData("", "sync-video false")
             loadPrcFileData("", "model-path %s" % os.getcwd())
+            loadPrcFileData("", "model-cache-dir %s/.panda3d_cache" % os.getcwd())
```
JingfeiPeng, line 61: is this related to traffic data?

smarts/core/sumo_road_network.py
```diff
@@ -86,10 +86,13 @@ def write_glb(self, output_path):
 
 
 class SumoRoadNetwork:
-    def __init__(self, graph, net_file):
+    def __init__(self, graph, net_file, default_lane_width=None):
         self._log = logging.getLogger(self.__class__.__name__)
         self._graph = graph
         self._net_file = net_file
+        self._default_lane_width = (
+            default_lane_width if default_lane_width is not None else 3.2
```
JingfeiPeng, line 94: should we add some explanation for 3.2 here?

smarts/sstudio/genscenario.py
```diff
@@ -403,7 +409,28 @@ def _validate_entry_tactic(mission):
...
 
-def gen_traffic_histories(scenario: str, histories):
-    output_path = os.path.join(scenario, "traffic_histories.pkl")
-    with open(output_path, "wb") as f:
-        pickle.dump(histories, f)
+def gen_traffic_histories(scenario: str, histories_datasets, overwrite: bool):
+    genhistories_py = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "genhistories.py"
+    )
+    for hdsr in histories_datasets:
```
JingfeiPeng, line 416: I am wondering what does `hdsr` and `hds` stand for?...

cli/studio.py
```diff
@@ -146,6 +146,7 @@ def _clean(scenario):
         "social_agents/*",
         "traffic/*",
         "history_mission.pkl",
```
JingfeiPeng, line 148: Since we are using `.shf` files now we can probably remove `"history_mission.pkl"` here

JingfeiPeng (approved):
Nice change!

JingfeiPeng:
where could I take a look at how NGSIM and Interaction dataset scenario folder looks now?

smarts/core/renderer.py
sah-huawei: Oops.  No, it's not.  I accidentally left that in after a test.  I'll take it back out.  Thanks for catching that!

smarts/core/sumo_road_network.py
sah-huawei: Yes, I will add a comment.

smarts/sstudio/genscenario.py
sah-huawei: hds -> "history data set" and hdsr -> "history data set ref" (or something like that!)

cli/studio.py
sah-huawei: Yeah, definitely we need to eventually, but I wanted to leave it there a little bit to get rid of any existing files on the next clean.

smarts/core/renderer.py
sah-huawei: (It did speed up rendering though, so we might consider adding it at some point.)

sah-huawei:
> where could I take a look at how NGSIM and Interaction dataset scenario folder looks now?

I have them locally if you want me to zip them up and send them.  (I don't think we should push them here though.)
Or, you can download NGSIM  from the link in Issue #407 (and INTERACTION from zbzhu99's imitation learning fork) and try out the yaml file examples I put in Issue #732.  That's probably better b/c then you can see if I forgot anything in my instructions.

cli/studio.py
JingfeiPeng: Cool! One other thing we can remove is the `ijson` requirement in dependency. line 64 and 65 in setup.py and line 48 in requirements.txt. ijson was only used for imitation learning

sah-huawei: funny, I did, but I had to add it back.  It's still being used in `genhistories.py` to convert old json files.

smarts/core/smarts.py
```diff
@@ -303,6 +310,11 @@ def reset(self, scenario: Scenario):
     def setup(self, scenario: Scenario):
         self._scenario = scenario
 
+        self._use_realtime_clock = scenario.traffic_history and (
+            not self._traffic_sim.headless or not self._envision.headless
+        )
```
Gamenot, line 315: I see why the real-time option might be useful for imitation learning but I have a few concerns. My first inclination here is that it should not be attached to the traffic history because there are other uses for a realtime clock option and this binds the option to the traffic history. 

Secondly, Envision is already supposed to play back at real-time and I believe this complicates fixing Envision so I do not think Envision should be included with this option.

Otherwise this is useful to `sumo-gui` but there is another alternative. It is possible to send the `--gui-settings-file` option to `sumo-gui` to add step delay and breakpoints via a configuration file: https://sumo.dlr.de/docs/sumo-gui.html#configuration_files.

smarts/core/sumo_road_network.py
```diff
@@ -522,7 +542,8 @@ def point_is_within_road(self, point):
         #
         # return False
 
-        nearest_lanes = self.nearest_lanes(point[:2], radius=5)
+        radius = max(5, 2 * self._default_lane_width)
```
Gamenot, line 545: I think this with the default lane width at 3.2 would change some behaviour, I think these radius values were just arbitrary to begin with though.

smarts/core/traffic_history_provider.py
```diff
@@ -18,44 +18,54 @@
...
 class TrafficHistoryProvider(Provider):
     def __init__(self):
         self._is_setup = False
-        self._traffic_history_service = None
         self._map_location_offset = None
-        self.replaced_vehicle_ids = set()
-        self.start_time_offset = 0
+        self._replaced_vehicle_ids = set()
+        self._start_time_offset = 0
+        self._histories_db = None
```
Gamenot, line 36: Looks like you ordered the instance variables but then added one to put it back out of order. :rofl:

smarts/core/traffic_history_provider.py
```diff
@@ -65,65 +75,71 @@ def sync(self, provider_state):
...
-        time_with_offset = str(round(timestamp + self.start_time_offset, 1))
-        if not self._traffic_history_service.fetch_history_at_timestep(
-            time_with_offset
-        ):
-            return ProviderState(vehicles=[])
+    @staticmethod
+    def _decode_vehicle_type(vehicle_type):
+        # Options from NGSIM and INTERACTION currently include:
+        #  1=motorcycle, 2=auto, 3=truck, 4=pedestrian/bicycle
+        # But we don't yet have glb models for 1 and 4.
```
Gamenot, line 92: I will add this as an issue.

smarts/core/traffic_history_provider.py
```diff
@@ -65,65 +75,71 @@ def sync(self, provider_state):
...
+        if vehicle_type == 3:
+            return "truck"
+        return "passenger"
```
Gamenot, line 95: I see this as a potential issue that we would default to a passenger vehicle when provided a pedestrian since we do not support non-road actors yet nor do we yet want to evaluate pedestrians.

smarts/sstudio/genhistories.py
```diff
@@ -0,0 +1,446 @@
...
+                vid,
+                # time units are in milliseconds for both NGSIM and Interaction datasets, convert to secs
+                round(float(self.column_val_in_row(row, "sim_time")) / 1000, 3),
+                float(self.column_val_in_row(row, "position_x")) * self.scale,
+                float(self.column_val_in_row(row, "position_y")) * self.scale,
+                float(self.column_val_in_row(row, "heading_rad")),
+                float(self.column_val_in_row(row, "speed")) * self.scale,
+                self.column_val_in_row(row, "lane_id"),
+            )
+            if not any(a is not None and np.isnan(a) for a in traj_args):
```
Gamenot, line 160: I think the presence of `isnan` would be a value useful to warn about rather than silently skipping.

smarts/sstudio/genhistories.py
```diff
@@ -0,0 +1,446 @@
...
+
+    @staticmethod
+    def _lookup_agent_type(agent_type):
+        # Try to match the NGSIM types...
+        if agent_type == "car":
+            return 2
+        elif agent_type == "truck":
+            return 3
+        elif agent_type == "pedestrian/bicycle":
+            return 4
```
Gamenot, line 206: Is motorcycle also supposed to be here?

Gamenot (approved):

smarts/core/traffic_history_provider.py
Gamenot: #755, #756

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -32,26 +35,32 @@ def main(scenarios, headless, seed):
...
 
+            # agent vehicle will enter right away...
             modified_mission = replace(mission, start_time=0.0)
             scenario.set_ego_missions({agent_id: modified_mission})
+
             observations = smarts.reset(scenario)
 
             dones = {agent_id: False}
-            while not dones[agent_id]:
+            while not dones.get(agent_id, False):
```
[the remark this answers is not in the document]
sah-huawei: Yeah, I hit that in my testing.  It happened early on and I can't clearly remember the circumstances, but I suspect it was related to my having the `mission.start_time` wrong initially (leading to `agent_id` not being started yet).  So I guess, when things are correct, it's not necessary, but I added it so I wouldn't crash immediately with a key error and could debug other things first.

smarts/core/renderer.py
```diff
@@ -58,6 +58,8 @@ def __new__(cls):
             # disable vsync otherwise we are limited to refresh-rate of screen
             loadPrcFileData("", "sync-video false")
             loadPrcFileData("", "model-path %s" % os.getcwd())
+            # TODO: the following speeds up rendering a bit... might consider it.
+            # loadPrcFileData("", "model-cache-dir %s/.panda3d_cache" % os.getcwd())
```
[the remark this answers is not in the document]
sah-huawei: oh, right, model loading.  (However, as vehicle models can be loaded during the simulation, it can still affect the step time.)

smarts/core/sumo_road_network.py
sah-huawei: ok, done.

smarts/core/smarts.py
```diff
@@ -224,7 +226,12 @@ def _step(self, agent_actions):
         extras = dict(scores=scores)
 
         # 8. Advance the simulation clock.
-        self._elapsed_sim_time += dt
+        self._elapsed_sim_time = round(self._elapsed_sim_time + dt, 3)
```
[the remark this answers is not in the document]
sah-huawei: Good point.  (I added it because the addition of `dt` sometimes caused floating point precision issues, like an `_elapsed_sim_time` equal to, say, `1.9999999` instead of `2.0`.)  I just changed it to never "round too much".

smarts/core/smarts.py
sah-huawei: Re the traffic-history limitation, I agree with you.  I just was hesitant to auto-set it and change the behavior for other things that might have inadvertently set `headless` to False but where no one watches the gui/envision.

Re Envision, there is still a bug there, but you're right that this "masks" the problem.  

I will look into the `--gui-settings-file` option...

smarts/core/sumo_road_network.py
sah-huawei: Yeah, that was my sense too (just from looking at the history).  I did recognize this as a change, but it didn't seem to affect any tests.  I think the reason it's ok is because this just means that the query to the Sumo net file might return a lane when it didn't use to, but we still check (and/r sort)  the lanes that are returned, so the extra-returned lane doesn't have any impact.

smarts/core/traffic_history_provider.py
sah-huawei: hehe, yep!  :)

smarts/core/traffic_history_provider.py
sah-huawei: Thanks.

smarts/core/traffic_history_provider.py
sah-huawei: Yeah, I agree.  I'm adding a change to skip pedestrians for now.

sah-huawei: ... although on second thought, I think we should probably just add a pedestrian model fairly soon because, in an imitation learning scenario, it could cause problems to train from a vehicle that swerves or stops suddenly for seeming no reason (if a pedestrian that triggered this behavior is not part of the simulation).  
So I guess I'm not going to skip them after all.  We should just do Issue #756 relatively soon instead.

smarts/sstudio/genhistories.py
sah-huawei: yeah, in general I agree.  Unfortunately, the rolling window thing method used to calculate the moving averages for `position_x` and `position_y` leaves 1-window-width of NaNs at the end.  This was the easiest way to ignore those.  I'll add a comment to explain that.

smarts/sstudio/genhistories.py
sah-huawei: oops, yes!  good catch!

smarts/core/smarts.py
sah-huawei: Using step delay via a `--gui-settings-file` could work assuming we know (approximately) how long our steps really take.  But if these will be variable, or if we'll be improving our step time, then we may have to tweak this occasionally.

I'm willing to do this, but let's chat about it on Monday first.  I do see other advantages to having a near-real-time mode (in particular, asyncrhonously running alongside ROS nodes), but there may be better ways to achieve that, so I can be persuaded to leave this out or now (and/or use the sumo-gui delay setting).

smarts/core/traffic_history_provider.py
Gamenot: Yes, more-so I was worried about imitation learning attempting to take over a pedestrian because it thinks it is a passenger vehicle.

sah-huawei: Ah oh.  I just added something to prevent that:  only passenger cars will now be included in the agent missions used by `history_vehicle_replacement_for_imitation_learning.py`.

Issue #411, opened 2021-01-13
Vehicle Sensors Inappropriately Replaced


Gamenot:
This comment out will be reverted but I found that somehow vehicles that already had sensors were having their sensors replaced.

Run `python examples/egoless.py scenarios/minicity --max_episode_steps 100000000`

_Originally posted by @Gamenot in https://github.com/huawei-noah/SMARTS/pull/410#discussion_r556533348_

Gamenot:
@Adaickalavan This problem should exist on master as well but if you wish to test this on the `fix-mini-city-problems` it will be reproducible there.

PR #745, opened 2021-04-07
BUG REVIEW: SMARTS `segfault`

Gamenot:
@sah-huawei 
Run this example like `python3.7 examples/low_level.py scenarios/loop`.

### DIFF
--- a/examples/low_level.py
+++ b/examples/low_level.py
@@ -1,0 +1,71 @@
+import logging
+
+from envision.client import Client as Envision
+from examples import default_argument_parser
+from smarts.core.agent import AgentSpec
+from smarts.core.agent_interface import AgentInterface, AgentType
+from smarts.core.scenario import Scenario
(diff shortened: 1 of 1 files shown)


examples/low_level.py
```diff
@@ -0,0 +1,58 @@
...
+        list([]),
+    )
+
+    smarts.reset(next(scenarios_iterator))
+
+    for _ in range(80):
+        smarts.step({})
+        smarts.attach_sensors_to_vehicles(
+            agent_spec, smarts.vehicle_index.social_vehicle_ids()
+        )
```
sah-huawei, line 41: When a social vehicle enters a bubble, sensors are also added to the vehicle, which causes this test to be non-deterministic -- it will sometimes assert from trying to add a sensor twice because of that.

Gamenot: That problem seems to relate to https://github.com/huawei-noah/SMARTS/issues/411

PR #722, opened 2021-03-31
Controller gain improvements

Gamenot:

PR #749, opened 2021-04-09
Proof of concept: move cutin uturn waypoints generation from smarts to open-agent

JingfeiPeng:
Took a stab at decoupling cutin and uturn waypoint generation from smarts. 

In summary:
- Changed the original `AgentBehaviour` class which just has an aggressiveness value to be responsible for uturn and cutin waypoint generation. 
- Kept `AgentBehaviour` as part of AgentInterface. Gave `WaypointsSensor` access to AgentInterface.agent_behavior from vehicle.py
- `WaypointsSensor` calls `apply_custom_behaviour` from AgentBehaviour
- will pass all the needed information for uturn and cutin from `WaypointsSensor` to `AgentBehaviour`, such as `self._road_network`   `self._waypoints`
self._road_network can be accessed from sim
self._waypoints can be accessed from scenario.waypoints

Since PR #722 haven't been merged and I expect there are new PR changes in Behaviour Model soon. I didn't proceed to change the details in `def uturn_waypoints:` `def cut_in_waypoints:` from referencing `self.xxx` in mission_planer to reference things in the new structure. It is better to make these changes after finalizing Behaviour model changes, and first making sure if this structure change makes sense

For now, this is just a proof of concept PR to show whether this new structure works.


junluo-huawei:
Why is it necessary for the AgentInterface to has AgentBehaviuor as part of it? Why should WaypointsSensor have information about custom behavior at all? It seems to me that all that logic of adapting and filtering waypoints belongs to the internal of the agent. Maybe what’s needed instead is actually map observation that covers a broader area, map observation that includes adjacent lanes etc?

From: Jeffer Jingfei Peng ***@***.***
Sent: Thursday, April 8, 2021 8:41 PM
To: huawei-noah/SMARTS ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [huawei-noah/SMARTS] Proof of concept: move cutin uturn waypoints generation from smarts to open-agent (#749)


Took a stab at decoupling cutin and uturn waypoint generation from smarts.

In summary:

  *   Changed the original AgentBehaviour class which just has an aggressiveness value to be responsible for uturn and cutin waypoint generation.
  *   Kept AgentBehaviour as part of AgentInterface. Gave WaypointsSensor access to AgentInterface.agent_behavior from vehicle.py
  *   WaypointsSensor calls apply_custom_behaviour from AgentBehaviour
  *   will pass all the needed information for uturn and cutin from WaypointsSensor to AgentBehaviour, such as self._road_network self._waypoints ( though they are supposed to be private variable, will see what can be done)

Since PR #722<https://github.com/huawei-noah/SMARTS/pull/722> haven't been merged and I expect there are new PR changes in Behaviour Model soon. I didn't proceed to change the details in def uturn_waypoints: def cut_in_waypoints:. It is better to make these changes after finalizing Behaviour model changes.

For now, this is just a proof of concept PR to show whether this new structure works.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/huawei-noah/SMARTS/pull/749

Commit Summary

  *   initial commit
  *   black

File Changes

  *   M smarts/core/agent_interface.py<https://github.com/huawei-noah/SMARTS/pull/749/files#diff-09e41aaa060eed1eb0690b068f085ab0de6a9d7aae708a50c2cef760936d49f3> (10)
  *   M smarts/core/mission_planner.py<https://github.com/huawei-noah/SMARTS/pull/749/files#diff-72668ccabb65e2bfe5e459aae7b67da9e3798e1fdb65014cb2c3c95d435a615f> (322)
  *   M smarts/core/sensors.py<https://github.com/huawei-noah/SMARTS/pull/749/files#diff-fc41655ffc7f7dc46a63dbf123288d7ab6cbc498eabdc7b3c0a99a57edbaad63> (20)
  *   M smarts/core/vehicle.py<https://github.com/huawei-noah/SMARTS/pull/749/files#diff-3e1faf2c970b35f533e2632431d749d9bc9fc29f2b04ec2af6b799046fe8c843> (1)
  *   M zoo/policies/open-agent/open_agent/__init__.py<https://github.com/huawei-noah/SMARTS/pull/749/files#diff-402baf40c60d691242bf4bd56b7d01224dfc8766de87118b7101c272e55ca163> (4)
  *   A zoo/policies/open-agent/open_agent/agent_behaviour.py<https://github.com/huawei-noah/SMARTS/pull/749/files#diff-6040df8281e7819618d31a88dcd973ec73173dd30487e3d87b46ee97156c1462> (342)

Patch Links:

  *   https://github.com/huawei-noah/SMARTS/pull/749.patch
  *   https://github.com/huawei-noah/SMARTS/pull/749.diff

JingfeiPeng:
> Why is it necessary for the AgentInterface to have AgentBehaviuor as part of it? Why should WaypointsSensor have information about custom behavior at all? It seems to me that all that logic of adapting and filtering waypoints belongs to the internal of the agent. Maybe what’s needed instead is actually map observation that covers a broader area, map observation that includes adjacent lanes etc? @Gamenot : You should remember that there has been a long standing request for adding broader map observation (e.g. for ULTRA). I do feel this is a similar need. Need for map observation (with waypoints being one kind of map info) in the context of behavior and mission is the issue here. Let's have a design review first!!!

sure, i guess we should try to have the agent generate its path purely based on observation in the act method. If needed to access `self._road_network` could be given to the OpEnAgent class as part of requirement in OpEnAgent class. 

One other thing is the OpEnAgent class Act method currently still have all the leftover code for using OpEn. We should remove these code

Gamenot:
> Why is it necessary for the AgentInterface to has AgentBehaviuor as part of it? 

We have a few targets regarding the AgentBehaviour: controller, trajectory generator, Agent, Actor. The AgentBehaviour is intended to be an interface that translates into parameters for the controller and trajectory generators. Since these values are 

> Why should WaypointsSensor have information about custom behavior at all?
 
We have a systemic problem of mixing up waypoints and trajectories. The WaypointsSensor is inappropriately named or inappropriately used since its actual role is to generate trajectories `[pose, speed]`. A WaypointSensor should not have any influence by behaviour but perhaps a trajectory generator might. Our problem here is that the WaypointSensor should return to being just a waypoint sensor again but almost everywhere the result is used assumes trajectories instead. Perhaps the start of the mistake was generating the equally spaced waypoints at each step.

> It seems to me that all that logic of adapting and filtering waypoints belongs to the internal of the agent. Maybe what’s needed instead is actually map observation that covers a broader area, map observation that includes adjacent lanes etc? 

The WaypointSensor currently generates branching trajectories from the current lane. And then there is a separate RoadWaypointSensor which finds the nearby static waypoints on adjacent lanes (both forward and oncoming) as well as the static route waypoints (but I believe that the route waypoints also got polluted by the dynamic equally spaced waypoints generator.)

>@Gamenot : You should remember that there has been a long standing request for adding broader map observation (e.g. for ULTRA). I do feel this is a similar need. Need for map observation (with waypoints being one kind of map info) in the context of behavior and mission is the issue here. Let's have a design review first!!!

OK.

junluo-huawei:
> We have a systemic problem of mixing up waypoints and trajectories. The WaypointsSensor is inappropriately named or inappropriately used since its actual role is to generate trajectories `[pose, speed]`. A WaypointSensor should not have any influence by behaviour but perhaps a trajectory generator might. Our problem here is that the WaypointSensor should return to being just a waypoint sensor again but almost everywhere the result is used assumes trajectories instead. Perhaps the start of the mistake was generating the equally spaced waypoints at each step since I believe those augmented waypoints are actually just an attempt to generate appropriate trajectories.

Indeed. We are overloading Waypoints multiple times too much. We need a proper map observation API that is independent of the agent's mission route/goal or its motion trajectory. Design and design review first, please.

Gamenot:
https://github.com/huawei-noah/SMARTS/tree/feature-extract_behavior_models ended up quite similarly although it also highlights some of the mangled use of waypoints.

PR #750, opened 2021-04-09
WIP Multi-agent game of tag demo (Don't merge)

JingfeiPeng:
Creating this WIP PR to keep track of conversation, problems and progress

The only two important files currently is 
`examples/game_of_tag/custom_adapters.py`: file for defining rewards, observations
`examples/game_of_tag/game_of_tag.py`: file for running training using rllib

**Progress:**
Predator vehicle will speed up and catch the prey vehicles if they are following them

**Rewards**
Deduction to prey reward when prey collide with predator (Collision DoneCreteria is turned off)
Positive reward when predator collide with prey. (Collision DoneCreteria is turned off)
discount factor of (0.999)*(current number of timesteps during simulation) applied to all positive and negative rewards
small deduction when agent's action changed (current action != previous action). Kept track of this in TrainingState
When no prey vehicles present (all went off_road), predator receive 0 rewards until end of simulation
positve reward when predator is following prey (calculated by angle between two positions compared to ego heading)
positive reward when predator is following prey and have higher speed than the prey it is following


JingfeiPeng:
one problem I have currently is the code in `TagModelAgent` was never reached, (prints in act method and init method were not showing up while prints in adapter codes are showing)
image

JingfeiPeng:
Closing this PR because it has diverged to attempt to use QMIX for the training. While the training can run, QMIX turns out to be not as good as using PPO with multi-agent. The code for using PPO is on a different branch (`multi_agent_ppo_version_demo`). Will open a clean PR on that branch instead

PR #759, opened 2021-04-12
Add image observations for ULTRA baseline agents

christianjans:
Allows ULTRA baseline agents to use image observations.

TODO:
- [x] Fix error encountered when undergoing learning
- [x] Clean up how state size is calculated for non-image states
- [x] Store image observation stacks more efficiently in replay buffer
- [ ] Clean up how RLlib will handle the new image observation option


AlexLewandowski:
NTS: We should revisit this PR and narrow the scope to get just BDQN compatibile with image obs for now.

christianjans:
Closed in favour of #958.

Issue #760, opened 2021-04-12
ULTRA Option to evaluate agent post graduation


JenishPatel99:
**Is your feature request related to a problem? Please describe.**
To better understand the curriculum learning approach using grade based structure in ULTRA, an option to evaluate the agent after graduating a grade is essential. Thus, this PR will add a ```eval_per_grade``` condition to the curriculum (config.yaml). This will allow for evaluation, on grade X's test scenarios, post graduation.

**Describe the solution you'd like**
Configured ```train,py```, ```evaluate.py```, ```coordinator.py``` to include the new ```eval_per_grade``` condition

PR #761, opened 2021-04-12
ULTRA Evaluate agent after grade completion

JenishPatel99:
Addressing Issue: #760

### DIFF
--- a/ultra/tests/scenarios/grade_based_test_curriculum/config.yaml
+++ b/ultra/tests/scenarios/grade_based_test_curriculum/config.yaml
@@ -4,6 +4,7 @@ curriculum:
     2: [["00-gb",test_grade2]]
     3: [["00-gb",test_grade3]]
   conditions:
+    eval_per_grade: False
     episode_based:
       toggle: True
       cycle: False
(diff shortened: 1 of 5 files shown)


AlexLewandowski (approved):
I feel that something like this could be handled in the train script as well. Of course, we did discuss keeping the train script relatively "vanilla" and per that discussion, this seeems sensible.

christianjans (approved):
Looks good! 👍

I wonder if perhaps some of the logic can be simplified, but maybe that will become more apparent in the context of the entire grade-based pull request and also some of the new changes in the ultra-develop branch.

JenishPatel99:
@AlexLewandowski I completely agree, there are some aspects from the grade-based stuff that can be added to the "vanilla" train later on. But for now, I will try to keep the same template as the vanilla train script, so the only difference would be no-gb and gb.

> I feel that something like this could be handled in the train script as well. Of course, we did discuss keeping the train script relatively "vanilla" and per that discussion, this seeems sensible.

JenishPatel99:
@christianjans I have not explicit refactored the grade-based, but as I continue I will keep in mind of any simplification of the code. But yeah, once the dynamic scenario addition is done, we can start filtering the code

> Looks good! 👍
> 
> I wonder if perhaps some of the logic can be simplified, but maybe that will become more apparent in the context of the entire grade-based pull request and also some of the new changes in the ultra-develop branch.

PR #769, opened 2021-04-15
traffic history issues

sah-huawei:
Main thing here is a fix for a bug introduced by PR #765:  the update to `sim_time` caused some times to be imprecise, meaning that the "between" query done in the `traffic_history_provider.py` sometimes misses.

Also a miscellaneous update to give traffic history vehicles an id that will allow them to be colored as social agents (as opposed to ego agents) in Sumo, allowing for the 1 ego agent used by the history replacement script to stand out as red.

### DIFF
--- a/smarts/sstudio/genhistories.py
+++ b/smarts/sstudio/genhistories.py
@@ -168,7 +168,7 @@ class _TrajectoryDataset:
         self._log.debug("shifting sim_times..")
         mcur = dbconxn.cursor()
         mcur.execute(
-            "UPDATE Trajectory SET sim_time = sim_time - (SELECT min(sim_time) FROM Trajectory)"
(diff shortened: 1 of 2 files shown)


JingfeiPeng (approved):

smarts/sstudio/genhistories.py
```diff
@@ -168,7 +168,7 @@ def create_output(self):
         self._log.debug("shifting sim_times..")
         mcur = dbconxn.cursor()
         mcur.execute(
-            "UPDATE Trajectory SET sim_time = sim_time - (SELECT min(sim_time) FROM Trajectory)"
+            "UPDATE Trajectory SET sim_time = round(sim_time - (SELECT min(sim_time) FROM Trajectory), 3)"
```
Gamenot, line 171: I think that the digit param could at minimum be a descriptive class constant like `TIME_DECIMAL_ACCURACY` since it has a large impact on the behaviour of this class.

Gamenot (approved):

PR #771, opened 2021-04-16
Traffic History dones

### LINKED ISSUE #770: finishing vehicles in `observation_collection_for_imitation_learning.py` should return `done`
(As requested by @zbzhu99 via WeChat.)

Currently, the script attaches sensors to vehicles from traffic history (using `observe_from`) and collects observations from them.  When such a vehicle exits the map (i.e., is no longer in the history), it is desired that `done[agent_id]` is set to `True` for it.

sah-huawei:
For Issue #770.  But rather than use a different `AgentInterface` for each vehicle with different `max_episode_steps` (dependent on their `max(sim_time)`from the `.shf` file (or equivalently, a custom `done_criteria` lambda), I opted for the simpler approach of just keeping track of when history vehicles leave.  

This is slightly less clean -- partly because of the assumption it makes about the default agent names being "Agent-{v}" -- but I think it's probably good enough for this simple example.

Closes #770.

### DIFF
--- a/examples/observation_collection_for_imitation_learning.py
+++ b/examples/observation_collection_for_imitation_learning.py
@@ -36,11 +36,19 @@ def main(scenarios, headless, seed):
         smarts.step({})
 
         current_vehicles = smarts.vehicle_index.social_vehicle_ids()
+        # We explicitly watch for which agent/vehicles left the simulation here
+        # since we don't have a "done criteria" that detects when a vehicle's
+        # traffic history has played itself out.
         done_vehicles = prev_vehicles - current_vehicles
         prev_vehicles = current_vehicles
 
         smarts.attach_sensors_to_vehicles(agent_spec, current_vehicles)
         obs, _, _, dones = smarts.observe_from(current_vehicles)
+        # The `dones` returned above should be empty for traffic histories
+        # where all vehicles are assumed to stay on the road and not collide.
+        # TODO:  add the following assert once the maps are accurate enough that
+        # we don't have any agents accidentally go off-road.
+        # assert not done
         for v in done_vehicles:
             dones[f"Agent-{v}"] = True
         # TODO: save observations for imitation learning
--- a/examples/observation_collection_for_imitation_learning.py
+++ b/examples/observation_collection_for_imitation_learning.py
@@ -20,8 +20,8 @@ def main(scenarios, headless, seed):
 
     smarts = SMARTS(
         agent_interfaces={},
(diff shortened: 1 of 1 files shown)


JingfeiPeng (approved):

examples/observation_collection_for_imitation_learning.py
```diff
@@ -30,12 +30,19 @@ def main(scenarios, headless, seed):
...
-        obs, _, _, _ = smarts.observe_from(smarts.vehicle_index.social_vehicle_ids())
+
+        current_vehicles = smarts.vehicle_index.social_vehicle_ids()
+        done_vehicles = prev_vehicles - current_vehicles
+        prev_vehicles = current_vehicles
+
+        smarts.attach_sensors_to_vehicles(agent_spec, current_vehicles)
+        obs, _, _, dones = smarts.observe_from(current_vehicles)
+        for v in done_vehicles:
+            dones[f"Agent-{v}"] = True
```
Gamenot, line 45: Sorry, I am not going to block with this but I am unsure what we are doing with dones here. Why are we combining the dones of the previous step with the dones of the current step observe?

Gamenot (approved):

examples/observation_collection_for_imitation_learning.py
sah-huawei: The `dones` returned from `observe_from()` do not include vehicles that were removed from the simulation because their trajectory (from the traffic history db) was completed, it only includes vehicles that are done because they met one of the "done criteria" (off road, collision, etc.).  

So if we want to include vehicles that have finished playing out their history, we have to explicitly check for that.  One way -- probably the "proper" way --  would have been to use a different `AgentInterface` for each vehicle, and set its `max_episode_steps` accordingly.   However, I didn't like this because it made the code around the `attach_sensors_to_vehicles()` call ugly, so I opted for the above instead.

What's happening there is that we're explicitly watching for which vehicles left the simulation _since_ the previous step and marking them `done`.  The `dones` list returned by `observe_from()` _should_ be empty on each step, given that all vehicles in the history data stay on the road and there aren't any collisions (afaik).  I originally asserted that, but then we discovered that the map is currently incorrect so we occasionally still have a vehicles show up as off-road (until I fix the map, at which point I'll add the assert).

sah-huawei: I'm going to add a comment to try to make this clearer (given that this is an "example"!).

PR #773, opened 2021-04-16
Fix test notebook ipykernel

Gamenot:
Explicitly install `ipykernel` seems to fix the test problem.


sah-huawei (approved):
This seems fine, although I do wonder why it used to work (i.e., the notebook test used to pass without this).

JingfeiPeng (approved):

PR #777, opened 2021-04-18
Disable ultra test on smarts

JingfeiPeng:
-disabled ultra base tests on smarts branches (branches that don't start with ultra- or ultra_)
-scheduled ultra test to run on develop on every wednesday
-fixed comments in actions since the days in a week is from 0 to 6

### DIFF
--- a/.github/workflows/ci-smarts-run-ultra-tests.yml
+++ b/.github/workflows/ci-smarts-run-ultra-tests.yml
@@ -2,7 +2,7 @@ name: SMARTS scheduled run of ULTRA tests
 
 on:
   schedule:
-    - cron: '25 6 * * 7'
+    - cron: '40 6 * * 6'
       # Time is in UTC
       # Runs at 11.00pm, UTC  , every Wednesday
       # Runs at  6.00pm, UTC-4, every Wednesday
--- a/.github/workflows/ci-smarts-run-ultra-tests.yml
+++ b/.github/workflows/ci-smarts-run-ultra-tests.yml
@@ -2,7 +2,7 @@ name: SMARTS scheduled run of ULTRA tests
 
 on:
   schedule:
-    - cron: '40 6 * * 6'
+    - cron: '0 11 * * 2'
       # Time is in UTC
       # Runs at 11.00pm, UTC  , every Wednesday
       # Runs at  6.00pm, UTC-4, every Wednesday
--- a/.github/workflows/ci-smarts-run-ultra-tests.yml
+++ b/.github/workflows/ci-smarts-run-ultra-tests.yml
@@ -2,16 +2,15 @@ name: SMARTS scheduled run of ULTRA tests
 
 on:
   schedule:
-    - cron: '0 23 * * 3'
+    - cron: '25 6 * * 7'
       # Time is in UTC
       # Runs at 11.00pm, UTC  , every Wednesday
-      # Runs at  6.00pm, UTC-5, every Wednesday
+      # Runs at  6.00pm, UTC-4, every Wednesday
   workflow_dispatch:
 
 jobs:
   test-base:
     runs-on: ubuntu-18.04
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     container: huaweinoah/smarts:v0.4.13-minimal
     steps:
       - name: Checkout
--- a/.github/workflows/ci-smarts-run-ultra-tests.yml
+++ b/.github/workflows/ci-smarts-run-ultra-tests.yml
@@ -1,12 +1,12 @@
-name: ULTRA CI Base Tests
+name: SMARTS scheduled run of ULTRA tests
 
-on: 
-  push:
-    branches:
-      - ultra[-_/]**
-  pull_request:
-    branches:
-      - ultra[-_/]**
+on:
+  schedule:
+    - cron: '0 23 * * 3'
+      # Time is in UTC
+      # Runs at 11.00pm, UTC  , every Wednesday
+      # Runs at  6.00pm, UTC-5, every Wednesday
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-ultra-tests.yml
```diff
@@ -1,6 +1,14 @@
 name: ULTRA CI Base Tests
 
-on: [push, pull_request]
+on: 
+  push:
+    branches:
+      - ultra-**
+      - ultra_**
```
JingfeiPeng, line 7: image
I tried to use ultra[-_]** or 'ultra[-_]**' glob matching but got this error, so I switched to the way above

christianjans: Oh strange, I had included a slash to match branches that begin with `ultra/` (in addition to `ultra-` and `ultra_`), and it seemed to work with the pattern `ultra[-_/]**`. If `ultra/` is an ULTRA branch name that we want to include, maybe give that pattern a try? Or include `ultra/` in the list of branches to include.

JingfeiPeng: yes I should also add a ultra / ** here. Though my main question is the pattern ultra[ - _ /]** might not have worked. I based off ultra-develop branch then pushed a branch named 'smarts-branch-test-ci', the smarts-ci-base-test were not running on this test branch. So I think the glob expression might not been working
image
image

JingfeiPeng: image

christianjans: Oh okay thank you for letting me know. I will investigate this as well.

Gamenot: > image
> I tried to use ultra[ - _ ]** or 'ultra[ - _ ]**' glob matching but got this error, so I switched to the way above

If you are using glob wouldn't it be something along the lines of `[\-_]**` since `"-"` is a reserved character?

JingfeiPeng: ah makes sense

JingfeiPeng: image
image
None of them worked actually. Will switch back to without using glob

Gamenot: This looks like yaml is interacting badly with pattern matching.

Gamenot (approved):

.github/workflows/ci-ultra-tests.yml
christianjans: Okay good to know, I have opened #780 to fix the `ultra-develop` branch.

christianjans (approved):
Looks great!

Adaickalavan (commented):
Please verify that the cron schedule and the accompanying comments are correct by utilising a cron tool such as: https://crontab.guru/#0_23_*_*_3

.github/workflows/ci-smarts-run-ultra-tests.yml
```diff
@@ -0,0 +1,110 @@
...
+          pip install --upgrade numpy
+      - name: Run ultra tests
+        run: |
+          cd ultra
+          . .venv/bin/activate
+          scl scenario build-all ultra/scenarios/pool
+          pytest -v ./tests/
+  test-package-via-setup:
+    runs-on: ubuntu-18.04
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```
Adaickalavan, line 48:
```suggestion
```

If the `test-package-via-setup` test runs on schedule and not on push/pull requests, then the above filter may not be required.

.github/workflows/ci-smarts-run-ultra-tests.yml
```diff
@@ -0,0 +1,110 @@
...
+          pip install --upgrade numpy
+      - name: Run test
+        run: |
+          cd ultra
+          . .venv/bin/activate
+          scl scenario build-all ultra/scenarios/pool
+          pytest -v ./tests/test_ultra_package.py
+  test-package-via-wheel:
+    runs-on: ubuntu-18.04
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```
Adaickalavan, line 78:
```suggestion
```

If the `test-package-via-wheel` test runs on schedule and not on push/pull requests, then the above filter may not be required.

JingfeiPeng:
> Please verify that the cron schedule and the accompanying schedule comments are correct as desired, by utilising a cron tool such as: https://crontab.guru/#0_23_*_*_3

https://docs.github.com/en/actions/reference/events-that-trigger-workflows
according to the github action doc, the days are from 0-6 instead of 1-7, which is different from the cron job tool. This is evident as if I tried to use '0 11 * * 7' I will get an error since 7 is out of range
image

Though on a side note, I wonder where do i check the output of scheduled cron job? I am not sure if it had run on develop branch since we constantly have new commits to it.

Adaickalavan:
Notifications for scheduled workflow runs:
https://docs.github.com/en/actions/guides/about-continuous-integration#notifications-for-workflow-runs

.github/workflows/ci-smarts-run-ultra-tests.yml
```diff
@@ -0,0 +1,110 @@
+name: SMARTS scheduled run of ULTRA tests
+
+on:
+  schedule:
+    - cron: '0 11 * * 2'
+      # Time is in UTC
+      # Runs at 11.00pm, UTC  , every Wednesday
+      # Runs at  6.00pm, UTC-4, every Wednesday
```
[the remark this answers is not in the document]
JingfeiPeng: oh sorry, you were right, Sunday starts at 0, will fix this

PR #781, opened 2021-04-20
Add `step`, `save`, and `load` methods to `smarts.core.agent.Agent`

christianjans:


smarts/core/agent.py
```diff
@@ -58,6 +58,15 @@ def act(self, obs):
 
         raise NotImplementedError
 
+    def step(self, state, action, reward, next_state, done, info):
+        return {}
```
JingfeiPeng, line 62: if the goal is to have users override these methods maybe have `raise NotImplementedError`?

christianjans: I think for these methods, a default implementation may be needed. These are more for RL agents to be able to save/load from a model, and also to step given knowledge of where an action in a state lead. (Sorry I should have specified this in the description! Will update the description.)

I am unsure if this sort of implementation is what the SMARTS team is wanting, but these methods (or similar) are what ULTRA would need.

sah-huawei, line 62: What would an Ultra agent do if `step()` is called? 
Could this be called independently of SMARTS' `step()`?  Or would SMARTS be expected to iterate over all of the agents that it knows about when it's stepped and call `step()` for each of them? 
If not the former, I'm thinking we might want to somehow prevent `step()` from being called externally.  And if the latter, we should include that in this PR.

Note that if `step()` is going to be allowed to be called on an agent independently of SMARTS' `step()`, then we need to figure out what the semantics of this should be.  For example, what is it passed and what does it return?  It won't necessarily be passed `Observations` from SMARTS generated on the last SMARTS' step.  And if `Actions` are returned, these will not have the granularity of "something that happens during the next SMARTS' `step()`". 

(For these reasons, I have less issues with `save()` and `load()` for an Agent -- which I think we could add immediately -- than I do with `step()` for an Agent.)

smarts/core/agent.py
```diff
@@ -58,6 +58,15 @@ def act(self, obs):
 
         raise NotImplementedError
 
+    def step(self, state, action, reward, next_state, done, info):
```
sah-huawei, line 61: This should also indicate the types of its parameters and the type of thing it returns.

smarts/core/agent.py
```diff
@@ -58,6 +58,15 @@ def act(self, obs):
...
+        return {}
+
+    def save(self, path):
```
sah-huawei, line 64: As above, for this and `load()`, the type of the parameter (presumably `str`) and return value (if any) should be explicitly indicated.

sah-huawei (changes requested):

smarts/core/agent.py
christianjans: Good questions, yeah sorry these methods are still a work-in-progress, and if they are deemed needed, then of course more documentation will be added.

> What would an Ultra agent do if `step()` is called?

The `step` method is used to record a transition (state, action, reward, next state) that a reinforcement learning agent can learn from. This sometimes requires storing the transition in a "replay buffer", or 

> Could this be called independently of SMARTS' `step()`?

Yes, the `step()` method of an agent is meant to be called independently of SMARTS' `step()`. Example usage may be in one's own training loop script:
```python
# Initialize agent IDs, agent specs, etc.
...

# An object used to record data to a file.
data_recorder = DataRecorder()

# Dict[str, smarts.core.agent.Agent], a dictionary of agent IDs and their agent instance.
agents = {...}

observations = env.reset()

for _ in range(1000):
  actions = {
    agent_id: agents[agent_id].act(agent_observation)
    for agent_id, agent_observation in observations.items()
  }
  next_observations, rewards, dones, infos = env.step(actions)

  for agent_id, agent in agents.items():
    learning_progress_data = agent.step(
      state=observation[agent_id],
      action=actions[agent_id],
      reward=rewards[agent_id],
      next_state=next_observations[agent_id],
      done=dones[agent_id],
      info=infos[agent_id],
    )
    data_recorder.record(learning_progress_data, agent_id)

  observations = next_observations
```
It allows the agent to observe where it ended up after it took an action in its given state. Additionally, the reward associated with that state, and other information can be given. The agent can then learn from these `(state, action, reward, next_state, done)` transitions.

In ULTRA, `step` returns information about its learning progress (neural network loss) in the form of a dictionary. However, depending on what one wants to record, this dictionary can of course be populated with other information.

> It won't necessarily be passed Observations from SMARTS generated on the last SMARTS' step.

Yes, this is definitely correct, and that's okay. An RL agent's state

PR #783, opened 2021-04-20
ULTRA Improve high density traffic

JenishPatel99:


ultra/ultra/scenarios/generate_scenarios.py
```diff
@@ -444,10 +449,11 @@ def generate_social_vehicles(
         begin_time = begin_times[start_lane_id][begin_time_idx[start_lane_id]]
         begin_time_idx[start_lane_id] += 1
         end_time = begin_time + 3600  # 1 hour
+
         if "stopwatcher" in behavior_idx:
             start_lane_id = route_lanes[stopwatcher_info["direction"][0]] - 1
             end_lane_id = route_lanes[stopwatcher_info["direction"][1]] - 1
-
+            begin_time = random.randint(10, 50)
```
christianjans, line 456: Just curious, why is `begin_time` random for stopwatchers?

ultra/ultra/scenarios/task1/config.yaml
```diff
@@ -152,3 +152,76 @@ levels:
...
+           stops: null
+        3lane_t:
+           percent: 0.5
+           specs: [[50kmh,mid-density,0.33], [70kmh,mid-density,0.34], [100kmh,mid-density,0.33]]
+           stops: null
+    test:
+      total: 150
+      ego_missions:
+        # left turn from south to west
+      - start: soutmid
```
christianjans, line 212: Just curious, is `soutmid` an edge we support? I have never seen it before.

ultra/ultra/config.yaml
```diff
@@ -43,6 +43,12 @@ tasks:
     hard:
       train: "scenarios/task1/train_hard*"
       test:  "scenarios/task1/test_hard*"
+    high-density:
+      train: "scenarios/task1/train_high-density*"
+      test:  "scenarios/task1/test_high-density*"
+    mid-density:
+      train: "scenarios/task1/train_mid-density*"
+      test:  "scenarios/task1/test_mid-density*"
```
christianjans, line 51: What are these extra levels for?

ultra/ultra/scenarios/common/distributions.py
```diff
@@ -671,76 +706,102 @@
...
-                "begin_time_init": {
-                    "func": basic_begin_time_init_func,
-                    "params": {"probability": prob_heavy},
+                    "func": poisson_init,
+                    "params": {"temperature": 10},
                 },
                 "has_turn": True,
                 "deadlock_optimization": True,
             },  # blocking
+            "south-east": None,
```
christianjans, line 722: No traffic from south to east lanes?

christianjans (changes requested):
Looks great! Just a few suggestions and questions. Also, do we have analysis plots to see how these changes compare to the graphs in the original slides?

ultra/tests/test_analysis.py
```diff
@@ -60,7 +60,7 @@ def test_interface_analyze(self):
                 os.makedirs(output)
 
             os.system(
-                f"python ultra/scenarios/interface.py analyze --scenarios {save_dir} --max-steps 600 --end-by-stopwatcher --output {output}"
+                f"python ultra/scenarios/interface.py analyze --scenarios {save_dir} --max-steps 600 --end-by-stopwatcher --output {output} --video 1"
```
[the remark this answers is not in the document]
JenishPatel99: Yes, just to make sure there isn't error in producing the traffic visualizations

ultra/ultra/config.yaml
JenishPatel99: This were for specific density based levels, we can remove them as well

ultra/ultra/evaluate.py
```diff
@@ -60,7 +60,7 @@ def evaluation_check(
     agent_ids_to_evaluate = [
         agent_id
         for agent_id in agent_ids
-        if episode.index % eval_rate == 0
+        if (episode.index + 1) % eval_rate == 0
```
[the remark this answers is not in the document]
JenishPatel99: Yes, I think this was added because I was kind of in the mindset of curriculum learning. And I do not evaluate in the first episode. I can revert that back

ultra/ultra/scenarios/generate_scenarios.py
christianjans: Okay, so like you said, sometimes the stopwatcher wouldn't spawn before the episode ended, so this random begin time ensures that it spawns and has enough time to finish the level. Maybe we could add a comment here to explain this?

JenishPatel99: You are correct, it might be a bit misleading without any comments. I will add an comment

ultra/ultra/scenarios/common/begin_time_init_funcs.py
```diff
@@ -71,16 +76,37 @@ def burst_begin_time_init_func(
...
+                elif int(speed) > 70:
+                    variable_time_between_cluster = (
+                        time_between_cluster[0] - speed_offset - (num_lanes / 2),
+                        time_between_cluster[1] - speed_offset - (num_lanes / 2),
+                    )
+                else:
+                    variable_time_between_cluster = (
+                        time_between_cluster[0],
+                        time_between_cluster[1],
+                    )
```
christianjans, line 95: Actually, I was thinking about this again, and just so that we don't run into any problems when/if we ever add any other speeds to ULTRA, perhaps we can raise an exception if we encounter a speed we don't currently have. Maybe something like:
```suggestion
                if int(speed) == 50:
                    variable_time_between_cluster = (
                        time_between_cluster[0] + speed_offset + (num_lanes / 2),
                        time_between_cluster[1] + speed_offset + (num_lanes / 2),
                    )
                elif int(speed) == 70:
                    variable_time_between_cluster = (
                        time_between_cluster[0],
                        time_between_cluster[1],
                    )
                elif int(speed) == 100:
                    variable_time_between_cluster = (
                        time_between_cluster[0] - speed_offset - (num_lanes / 2),
                        time_between_cluster[1] - speed_offset - (num_lanes / 2),
                    )
                else:
                    raise ValueError(f"Unsupported speed: {speed}.")
```
Let me know your thoughts.

JenishPatel99: Yes, I agree. At least from what I can see that a new speed can only be inducted if and only if a new map is constructed

ultra/ultra/scenarios/common/distributions.py
```diff
@@ -295,57 +295,92 @@
...
+                    "params": {
+                        "vehicle_cluster_size": (1, 2),
+                        "time_between_cluster": (60, 80),
+                        "time_for_each_cluster": 5,
+                    },
                 },
-                "has_turn": True,
-                "deadlock_optimization": True,
+                "has_turn": False,
+                "deadlock_optimization": False,
```
AlexLewandowski, line 306: Is there a specific reason for flipping these flags? I assume they introduce issues with the traffic density, but it would be good to document reasoning.

ultra/ultra/scenarios/common/distributions.py
```diff
@@ -284,8 +284,8 @@
                 "distribution": behavior_distribution,
                 "start_end_on_different_lanes_probability": 0.0,
                 "begin_time_init": {
-                    "func": basic_begin_time_init_func,
-                    "params": {"probability": prob_heavy},
+                    "func": poisson_init,
+                    "params": {},
```
AlexLewandowski, line 288: Would you say the main fix is switching to poisson instead of basic func?

ultra/ultra/scenarios/common/begin_time_init_funcs.py
```diff
@@ -71,16 +76,39 @@ def burst_begin_time_init_func(
...
+                    variable_time_between_cluster = (
+                        time_between_cluster[0] - speed_offset - (num_lanes / 2),
+                        time_between_cluster[1] - speed_offset - (num_lanes / 2),
+                    )
+                else:
+                    raise ValueError(f"Unsupported speed: {speed}.")
+
+                time_to_next_cluster = np.random.uniform(*variable_time_between_cluster)
+            else:
+                time_to_next_cluster = np.random.uniform(*time_between_cluster)
```
AlexLewandowski, line 101: I see, so you design the traffic density subject to the speed.  I wonder if it might be an issue if we introduce more speeds. We could have a piecewise linear function as a function of the speed without the if condition. Something like :

```
if speed >= 70
  speed_factor = (speed - 70)/30
elseif speed < 70 
  speed_factor = (speed-70)/20
```

and then have each component of variable_time_between_cluster use the speed_factor, like so `time_between_cluser[0] + (speed_factor)*(speed_offset - num_lanes/2)`.

We probably won't add more speeds, but it might be good to "future proof"

ultra/ultra/scenarios/common/distributions.py
JenishPatel99: @AlexLewandowski I agree, I did not play around with those flags. I assume that the `has_turn` flag means that there is some type of turn in the specifiied route

ultra/ultra/scenarios/common/distributions.py
JenishPatel99: @AlexLewandowski The `poisson_init` is only used for the ego mission route (which is usually south-west). The poisson distribution gives a nice inflow of social vehicles in the mission route: "nice" in terms it does not clog up the south lanes

ultra/ultra/scenarios/common/begin_time_init_funcs.py
JenishPatel99: @AlexLewandowski Christian and I had a discussion on this as well [https://github.com/huawei-noah/SMARTS/pull/783#discussion_r625777493](url). The current setup takes only into account of the three speeds that we have. In terms of the function of speed that governs the began times for variable speed still has to reconcile with. I like the idea of the piecewise linear function. We can discuss this in the ULTRA meeting on wednesday

ultra/ultra/config.yaml
JenishPatel99: @christianjans I will remove them once this PR is ready to go. Those are only for debugging purposes

ultra/ultra/scenarios/common/distributions.py
christianjans: Should `has_turn` be `True` here then? The route `south-east` should have a turn.

ultra/ultra/scenarios/common/distributions.py
```diff
@@ -295,57 +295,92 @@
...
-                    "params": {"probability": prob_heavy},
+                    "func": burst_begin_time_init_func,
+                    "params": {
+                        "vehicle_cluster_size": (1, 2),
+                        "time_between_cluster": (60, 70),
+                        "time_for_each_cluster": 2,
+                    },
                 },
-                "has_turn": True,
+                "has_turn": False,
```
christianjans, line 383: Also should `has_turn` be `True` here too? Since the `east-south` route has a turn?

christianjans (commented):
Okay sorry just a couple more questions!

ultra/ultra/scenarios/common/distributions.py
JenishPatel99: Should be go now 👌

christianjans (approved):

PR #791, opened 2021-04-23
more map shifting fixes

### LINKED ISSUE #790: Physics off in 0.4.16 pre-release
## **BUG REPORT**
**High Level Description**
Physics is off, causing the ego vehicle to not collide.

**SMARTS version**
SMARTS develop branch 6cb8ffda74e235c25d63b74313623b2e03e402f7 .

This issue does not occur in master (i.e. 0.4.15.)

https://github.com/huawei-noah/SMARTS/issues/716

Run the following `python3.7 examples/single_agent.py cutin_stopped_vehicle_2lane` with this scenario: 
[cutin_stopped_vehicle_2lane.zip](https://github.com/huawei-noah/SMARTS/files/6363278/cutin_stopped_vehicle_2lane.zip)
 

Collision should occur between SMARTS agent and sumo traffic vehicke, ending the vehicle early, traffic vehicles ends up in correct location SUMO GUI but totally different location in SMARTS.

**Error logs and screenshots**
image

sah-huawei:
Fixed bug where the positions of Sumo vehicles were incorrectly being shifted.  After PR #727, they no longer needed to be shifted because the map being used by Sumo now matches the map in the `SumoRoadNetwork` object.  

However, this also exposes a potential problem for external / third-party simulators that might add vehicles directly to Sumo because they may not know about the map shifting and so still have their coordinate systems relative to the unshfited/offset map.  So this PR also adds the `--allow-offset-map` option to `scl scenario build` to support _not_ auto-shifting maps.  Now, maps are only potentially shifted when scenarios are built (as opposed to being auto-shifted whenever `SumoRoadNetwork` opens an offset `map.net.xml` file), and this flag can override that at scenario build time if desired/required.

Also, similar to PR #727, use the possibly-shifted road network map if one has been auto-generated for traffic generation.

Closes #790


cli/studio.py
```diff
@@ -53,9 +64,17 @@ def _build_single_scenario(clean, scenario):
         _clean(scenario)
 
     scenario_root = Path(scenario)
-    map_net = scenario_root / "map.net.xml"
+    map_net = str(scenario_root / "map.net.xml")
+    if not allow_offset_map:
+        SumoRoadNetwork.from_file(map_net, shift_to_origin=True)
+    elif os.path.isfile(SumoRoadNetwork.shifted_net_file_path(map_net)):
```
JingfeiPeng, line 70: I guess this warning is to warn if user set `allow_offset_map` to False but there is an offset_map. 
If so, 
```suggestion
    elif not allow_offset_map and os.path.isfile(SumoRoadNetwork.shifted_net_file_path(map_net)):
```

smarts/core/sumo_traffic_simulation.py
```diff
@@ -597,11 +597,6 @@ def _compute_traffic_vehicles(self) -> List[VehicleState]:
             speed = sumo_vehicle[tc.VAR_SPEED]
             vehicle_type = sumo_vehicle[tc.VAR_VEHICLECLASS]
             dimensions = VEHICLE_CONFIGS[vehicle_type].dimensions
-            # adjust sumo vehicle location if we shifted the map
-            map_offset = self._scenario.road_network.net_offset
-            assert len(map_offset) == 2
-            front_bumper_pos[0] += map_offset[0]
-            front_bumper_pos[1] += map_offset[1]
```
JingfeiPeng, line 604: To confirm, the reason why these are no longer needs is because when building the scenario, it is already using the shifted map thus traffic vehicle coordinates are shifted when built? Please correct me if otherwise. 

This part is also kinda needed when running third party simulators that add traffic directly to sumo right? if so, it is possible to have an configurable option in sumo_traffic_simulation to control shift coordinates or not.

JingfeiPeng (approved):

smarts/core/sumo_traffic_simulation.py
sah-huawei: No, actually the reason is that, as of PR #727 , Sumo is started with the shifted map (thanks to the update to the `--net-file` parameter in `_base_sumo_load_params()` in `sumo_traffic_simulation.py`), so its coordinate system matches the coordinate system in use within Smarts. Vehicle locations coming from Sumo via Traci will be relative to the shifted/offset map already.

However, any third-party simulator that is adding vehicles to Sumo directly (outside of Smarts) may not know about the fact that the scenario map was shifted before Sumo was started. If they were to try to add vehicles to the map via Traci using the original coordinate system, things would seem off to them. So that's what necessitated adding the option to prevent map shifting entirely (i.e., not here, but where the scenario is built) in such situations.

Gamenot (approved):

PR #799, opened 2021-04-27
trigger ego vehicle to be done based on other agent's alive/done situation

### LINKED ISSUE #715: Allow DoneCreteria to be configurable based on the number of agents alive
During multi-agent training, it would be useful to have DoneCreteria based on the number of agents alive currently in the simulation.

Make DoneCreteria configurable based on the agents that are alive in the map.

**Additional context**
When creating game_of_tag, it would useful for the predators to be done when all the preys are done (if they accidentally go off  road). Since it wouldn't help the training of predators if preys are no longer in simulation and vice versia

JingfeiPeng:
resolves #715 

This change is needed in situation of multi-agent training. For example, to make prey vehicles done when there are no more predator vehicles so that prey vehicles don't run alone in the simulation, which creates noise in the training. 

will add unit tests if this change in DoneCreteria makes sense

### DIFF
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -19,7 +19,7 @@
 # THE SOFTWARE.
 from dataclasses import dataclass, field, replace
 from enum import IntEnum
-from typing import Optional, Union
+from typing import Optional, Union, List, Tuple
 
 from .controllers import ActionSpaceType
 from .lidar_sensor_params import BasicLidar
@@ -130,6 +130,31 @@ class AgentType(IntEnum):
 
 
 @dataclass(frozen=True)
+class AgentsListAlive:
+    agents_list: List[str]
+    """The list of agents to check whether they are alive"""
+    minimum_agents_alive_in_list: int
+    """Triggers the agent to be done if the number of alive agents in agents_list falls below the given value"""
+
+
+@dataclass(frozen=True)
+class AgentsAliveDoneCriteria:
+    minimum_ego_agents_alive: Optional[int] = None
+    """If set, triggers the agent to be done if the total number of alive ego agents falls below the given value."""
+    minimum_total_agents_alive: Optional[int] = None
+    """If set, triggers the agent to be done if total number of alive agents falls below the given value."""
+    agent_lists_alive: Optional[List[AgentsListAlive]] = None
+    """A termination criteria based on the ids of agents. If set, triggers the agent to be done if any list of agents fails 
+    to meet its specified minimum number of alive agents.
+    Example: [
+        (['agent1', 'agent2'], 1),
+        (['agent3'], 1)
+    ]
+    This agent's done event would be triggered if both 'agent1' and 'agent2' is done *or* 'agent3' is done.
+    """
+
+
+@dataclass(frozen=True)
 class DoneCriteria:
     """Toggleable conditions on which to trigger episode end."""
 
@@ -149,6 +174,8 @@ class DoneCriteria:
     """End the episode when the agent is not moving for 60 seconds or more. To account
     for controller noise not moving means <= 1 meter of displacement within 60 seconds.
     """
+    agents_alive: Optional[AgentsAliveDoneCriteria] = None
+    """If set, triggers the ego agent to be done based on the number of active agents for multi-agent purposes."""
 
 
 @dataclass
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -148,10 +148,10 @@ class AgentsAliveDoneCriteria:
     to meet its specified minimum number of alive agents.
     Example: [
         AgentsListAlive(
-            agents_list=[AGENT1, AGENT2], minimum_agents_alive_in_list=1
+            agents_list=['agent1','agent2'], minimum_agents_alive_in_list=1
         ),
         AgentsListAlive(
-            agents_list=[AGENT3], minimum_agents_alive_in_list=1
+            agents_list=['agent3'], minimum_agents_alive_in_list=1
         ),
     ]
     This agent's done event would be triggered if both 'agent1' and 'agent2' is done *or* 'agent3' is done.
--- a/smarts/core/agent_interface.py
+++ b/smarts/core/agent_interface.py
@@ -147,8 +147,12 @@ class AgentsAliveDoneCriteria:
     """A termination criteria based on the ids of agents. If set, triggers the agent to be done if any list of agents fails 
     to meet its specified minimum number of alive agents.
     Example: [
-        (['agent1', 'agent2'], 1),
-        (['agent3'], 1)
+        AgentsListAlive(
+            agents_list=[AGENT1, AGENT2], minimum_agents_alive_in_list=1
+        ),
+        AgentsListAlive(
+            agents_list=[AGENT3], minimum_agents_alive_in_list=1
+        ),
     ]
     This agent's done event would be triggered if both 'agent1' and 'agent2' is done *or* 'agent3' is done.
     """
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -26,6 +26,7 @@ from typing import Dict, Iterable, List, NamedTuple, Set, Tuple
 
 import numpy as np
 
+from smarts.core.agent_interface import AgentsAliveDoneCriteria
 from smarts.core.mission_planner import MissionPlanner
 from smarts.core.utils.math import squared_dist, vec_2d
 from smarts.sstudio.types import CutIn, UTurn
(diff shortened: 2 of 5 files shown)


smarts/core/sensors.py
```diff
@@ -342,6 +380,7 @@ def _is_done_with_events(cls, sim, agent_id, vehicle, sensor_state):
             or (is_not_moving and done_criteria.not_moving)
             or (is_off_route and done_criteria.off_route)
             or (is_wrong_way and done_criteria.wrong_way)
+            or agents_alive_done
```
sah-huawei, line 383: Do you want to add it to `Events` too?

smarts/core/sensors.py
```diff
@@ -318,6 +319,40 @@ def observe(sim, agent_id, sensor_state, vehicle):
...
+        cls, agent_manager, agents_alive: AgentsAliveDoneCriteria
+    ):
+        if not agents_alive:
+            return False
+
+        done = False
+        if agents_alive.total_ego_agents_alive:
+            done = (
+                len(agent_manager.ego_agent_ids) < agents_alive.total_ego_agents_alive
+            )
```
sah-huawei, line 333: Here (and in the following) you could just do:
 ```python
if agents_alive.total_ego_agents_alive and len(agent_manager.ego_agent_ids) < agents_alive.total_ego_agents_alive:
    return True
```
(rather than having multiple `if not done...` clauses.)

File: smarts/core/sensors.py [hunk 3]
```diff
@@ -318,6 +319,40 @@ def observe(sim, agent_id, sensor_state, vehicle):
...
+                ), "Please specify a list of agent ids to watch"
+                assert isinstance(
+                    min_alive, int
+                ), "Please specify an int for minimum number of alive agents in the list"
+                agents_alive = [
+                    1 if id in agent_manager.agent_ids else 0 for id in agent_list
+                ]
+                done = agents_alive.count(1) < min_alive
+                if done:
+                    break
```
sah-huawei, line 352: ...and could just `return True` here too.  (with the default "fall through" case below being `return False`.)

sah-huawei (approved):

smarts/core/sensors.py
JingfeiPeng: I see. yea that makes the code simpler, and even removes the need to have a `done` flag

smarts/core/agent_interface.py
```diff
@@ -129,6 +129,23 @@ class AgentType(IntEnum):
...
 
+@dataclass(frozen=True)
+class AgentsAliveDoneCriteria:
+    total_ego_agents_alive: Optional[int] = None
+    """if set, trigggers the agent to be done if total number of alive ego agents falls below total_agents_alive"""
+    total_social_agents_alive: Optional[int] = None
+    """if set, trigggers the agent to be done if total number of alive social agents falls below total_social_agents_alive"""
+    agent_lists_alive: Optional[List[Tuple[List[str], int]]] = None
+    """a list of agent lists to check alive or not, structure is List[Tuple[list_of_agent_ids, minimum_number_of_agents_alive in the list]]
+    if set, triggers the agent to be done if any list of agents fails to meet its specified minimum_number_of_agents_alive
```
Gamenot, line 140:
```suggestion
    """A termination criteria based on the ids of agents. If set, triggers the agent to be done if any list of agents fails to meet its specified minimum number of alive agents.
```

smarts/core/agent_interface.py
```diff
@@ -129,6 +129,23 @@ class AgentType(IntEnum):
     """Agent performs trajectory tracking using model predictive control."""
 
 
+@dataclass(frozen=True)
+class AgentsAliveDoneCriteria:
+    total_ego_agents_alive: Optional[int] = None
+    """if set, trigggers the agent to be done if total number of alive ego agents falls below total_agents_alive"""
```
Gamenot, line 135:
```suggestion
    minimum_ego_agents_alive: Optional[int] = None
    """If set, triggers the agent to be done if the total number of alive ego agents is lower than the given value."""
```

smarts/core/agent_interface.py
```diff
@@ -129,6 +129,23 @@ class AgentType(IntEnum):
...
+    total_social_agents_alive: Optional[int] = None
+    """if set, trigggers the agent to be done if total number of alive social agents falls below total_social_agents_alive"""
```
Gamenot, line 137:
```suggestion
    minimum_social_agents_alive: Optional[int] = None
    """If set, triggers the agent to be done if total number of alive social agents falls below the given value"""
```

smarts/core/agent_interface.py
```diff
@@ -129,6 +129,23 @@ class AgentType(IntEnum):
...
+    agent_lists_alive: Optional[List[Tuple[List[str], int]]] = None
+    """a list of agent lists to check alive or not, structure is List[Tuple[list_of_agent_ids, minimum_number_of_agents_alive in the list]]
+    if set, triggers the agent to be done if any list of agents fails to meet its specified minimum_number_of_agents_alive
+    Example: [
+        (['agent1', 'agent2'], 1),
+        (['agent3'], 1)
+    ]
+    Done on the ego agent would be triggered if both agent1 and agent2 is done or agent3 is done
```
Gamenot, line 145:
```suggestion
    This agent's done event would be triggered if both 'agent1' and 'agent2' is done *or* 'agent3' is done.
```

smarts/core/agent_interface.py
```diff
@@ -149,6 +166,8 @@ class DoneCriteria:
     """End the episode when the agent is not moving for 60 seconds or more. To account
     for controller noise not moving means <= 1 meter of displacement within 60 seconds.
     """
+    agents_alive: Optional[AgentsAliveDoneCriteria] = None
+    """If set, triggers ego agent to be done based on AgentsAliveDoneCriteria"""
```
Gamenot, line 170:
```suggestion
    """If set, triggers the ego agent to be done based on the number of active agents for multi-agent purposes."""
```

smarts/core/agent_interface.py
```diff
@@ -129,6 +129,23 @@ class AgentType(IntEnum):
     """Agent performs trajectory tracking using model predictive control."""
 
 
+@dataclass(frozen=True)
+class AgentsAliveDoneCriteria:
+    total_ego_agents_alive: Optional[int] = None
+    """if set, trigggers the agent to be done if total number of alive ego agents falls below total_agents_alive"""
+    total_social_agents_alive: Optional[int] = None
```
Gamenot, line 136: I wonder if we should make distinction between social agents and agents here?

smarts/core/agent_interface.py
```diff
@@ -129,6 +129,23 @@ class AgentType(IntEnum):
...
+    """if set, trigggers the agent to be done if total number of alive social agents falls below total_social_agents_alive"""
+    agent_lists_alive: Optional[List[Tuple[List[str], int]]] = None
+    """a list of agent lists to check alive or not, structure is List[Tuple[list_of_agent_ids, minimum_number_of_agents_alive in the list]]
```
Gamenot, line 139: nit: I think perhaps that the tuple could be replaced with a class to allow for a descriptive condition.

Gamenot:
Good work!

smarts/core/agent_interface.py
JingfeiPeng: is putting `*` around `or` done on purpose (`*or*`)? If so, I am wondering whats the meaning for this?

smarts/core/agent_interface.py
JingfeiPeng: so instead of having `minimum_social_agents_alive` have `minimum_total_agents_alive`?
I think both approaches are pretty similar, if the user wants X number of ego agents alive while keeping Y number of social agents alive, the user could do `minimum_ego_agents_alive=X` `minimum_total_agents_alive=X+Y` or `minimum_ego_agents_alive=X` `minimum_social_agents_alive=Y`. 
I guess there is a tradeoff between if the user wants just X number of total agents alive or just Y number of social agents alive. Perhaps it is better to just add another option for "minimum_total_agents_alive"?

smarts/core/agent_interface.py
JingfeiPeng: Sounds good! will do

smarts/core/agent_interface.py
Gamenot: No need for it, it was just an emphasis.

smarts/core/agent_interface.py
Gamenot: Sure, my consideration here was just that the number of ego agents is important. I am no not so much sure that the number of social agents is important by itself because in general we treat social agents as just agents but perhaps at a finalised stage. I also note the hole here in that there is no current way to end on reaching a minimum number of total agents.

I think the options I would prefer here is to remove the option so we do not have to explicitly support the social agent distinction or to go with the `minimum_total_agents_alive=X+Y` option.

JingfeiPeng: got it, added the `minimum_total_agents_alive` option

Gamenot (approved):
LGTM

PR #809, opened 2021-04-30
ULTRA Refactor curriculum learning structure

JenishPatel99:
- Remove unnecessary APIs
- Add function headers
- Clean up some logic structure


ultra/ultra/train.py
```diff
@@ -272,11 +276,14 @@ def train(
...
-                    episode, total_scenarios_passed, agents, average_scenarios_passed
+                (
+                    average_reached_goal,
+                    total_reached_goal,
+                ) = calculate_average_reached_goal(
+                    episode, total_reached_goal, agents, average_reached_goal
                 )
                 if (episode.index + 1) % CurriculumInfo.pass_based_sample_rate == 0:
-                    print(f"({episode.index + 1}) ASP: {average_scenarios_passed}")
+                    print(f"({episode.index + 1}) ASP: {average_reached_goal}")
```
christianjans, line 286:
```suggestion
                    print(f"({episode.index + 1}) Average reached goal: {average_reached_goal}")
```
Perhaps change to full name?

ultra/ultra/utils/curriculum/coordinator.py
```diff
@@ -134,73 +170,74 @@ def graduate(self, index, average_scenarios_passed=None):
...
-            total_scenarios_passed += episode.info[episode.active_tag][
-                list(agents.keys())[0]
-            ].data["reached_goal"]
 
     def display(self):
+        """Prints information about succeeding grade
+
+        Raises:
+            AttributeError : Unexpected error due to self.grade_counter
+                             not being defined
```
christianjans, line 234: Does this throw an error? It looks like `self.grade_counter` is initialized in the coordinator's construction, so I think it should always be defined? Let me know your thoughts and/or what you experienced.

ultra/ultra/utils/episode.py
```diff
@@ -193,6 +193,9 @@ def eval_mode(self):
     def gap_mode(self):
         self.active_tag = "Gap"
 
+    def get_mode(self):
+        return self.active_tag
+
```
christianjans, line 198:
```suggestion
    @property
    def active_tag(self):
        return self.active_tag
```

ultra/ultra/utils/curriculum/coordinator.py
JenishPatel99: @christianjans The problem was actually that the self.train_grade was not defined. The problem originated from one of the curriculum test cases not defining the train_grade. I made some changes which eliminates the use of the try-catch statement

ultra/docs/curriculum_learning.md
```diff
@@ -0,0 +1,126 @@
...
+      1: [[<task>,<level>]]
+      2: [[<task>,<level>], [<task>,<level>]]
+      # 3: [[<task>,<level>], ... ,[<task>,<level>]] can have as many combinations of tasks and levels
+    conditions:
+      eval_per_grade: <bool> # Evaluation takes places after completion of grade. An "exam" at the end of year
+      episode_based: # Agent graduates after completing a N number of episodes in each grade
+        toggle: <bool> # Enable the condition
+        cycle: <bool> # Option to keep cycling through grades to episodes limit
+      pass_based: # Agent graduates after getting an average completion rate, the average is taken over the eval-rate (sampling-rate)
+        toggle: <bool> # Enable the condition
```
AlexLewandowski, line 29: Just wondering, what happens if they are both toggled? Does the last one go-through, in this case pass_based? Also, could you clarify cycle: would you go grade 1 -> grade 2 -> grade 1 and so on?

ultra/docs/curriculum_learning.md
```diff
@@ -0,0 +1,126 @@
...
+      conditions:
+        eval_per_grade: True
+        episode_based:
+          toggle: True 
+          cycle: True 
+        pass_based:
+          toggle: False 
+          pass_rate: 0.50 # If the average scenario passed exceeds more than 0.5 then grade is switched
+          sample_rate: 30 # Every 30 episodes slot, the average reached goal (arg) is calculated (arg = total_scenario_passed / sample_rate)
+          warmup_episodes: 100 # Starting at any grade, the first 100 episodes will be subjected to no sampling
```
AlexLewandowski, line 53: These are ignored if toggle = false, correct?

ultra/tests/test_curriculum.py
```diff
@@ -162,10 +161,11 @@ def test_static_train_single_agent(self):
     def test_static_coordinator(self):
         log_dir = os.path.join(CurriculumTest.OUTPUT_DIRECTORY, "logs/")
         curriculum_dir = "../tests/scenarios/static_test_curriculum"
-        num_episodes = 8
+        num_episodes = 6
         etag = "sac-v0"
 
         agent_coordinator = Coordinator(curriculum_dir, num_episodes)
+        agent_coordinator._train_grade = ">>> TEST COORDINATOR <<<"
```
AlexLewandowski, line 168: Is this a placeholder? it seems like later on in the code you have a dict or list `self._train_grade = CurriculumInfo.curriculum["static"]["grades"][counter]`

ultra/ultra/train.py
```diff
@@ -351,6 +356,26 @@ def dynamic_curriculum_setup(curriculum_metadata):
...
+        total_reached_goal += episode.info[episode.active_tag][
+            list(agents.keys())[0]
+        ].data["reached_goal"]
+        average_reached_goal = total_reached_goal / sample_rate
+        total_reached_goal = 0.0
+        return average_reached_goal, total_reached_goal
+    else:
+        total_reached_goal += episode.info[episode.active_tag][
+            list(agents.keys())[0]
+        ].data["reached_goal"]
```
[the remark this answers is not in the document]
AlexLewandowski: That makes sense to me. Although we may want to do multi-agent curriculum learning, i think the single-agent case will be challenging enough.

ultra/ultra/tune.py
```diff
@@ -124,7 +124,7 @@ def tune_train(
 
     for episode in episodes(num_episodes, etag=etag, log_dir=log_dir):
         # Reset the environment and retrieve the initial observations.
-        observations = env.reset()
+        observations, scenario = env.reset()
```
AlexLewandowski, line 127: I think this was leftover from debugging?

ultra/ultra/utils/curriculum/coordinator.py
```diff
@@ -28,23 +28,29 @@
...
-        self.grade_checkpoints = []
-        self.num_episodes = num_episodes
-        self.grade_counter = 0
-        self.episode_per_grade = 1
-        self.warmup_episodes = 1
+        self._train_counter = cycle(tuple([i * 1 for i in range(self.num_grades)]))
+        self._eval_counter = cycle(tuple([i * 1 for i in range(self.num_grades)]))
+        self._num_episodes = num_episodes
+        self._grade_counter = 0
+        self._episodes_per_grade = -1
```
AlexLewandowski, line 43: Just to clarify, is this incremented and used as an index somewhere (so that it starts at 0?)

AlexLewandowski (approved):
I think you can go ahead and merge these changes into your own branch, and I'll do a more thorough review of that branch vs. ultra-develop. I know you changed a lot and the important changes aren't highlighed here (since it is a diff between 2 curriculum learning branches)

ultra/ultra/train.py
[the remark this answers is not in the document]
JenishPatel99: Absolutely, one of the goals in the grade-based structure was to generalize across multiple agents. Maybe as of right now we might not have the idea on how ultra curriculas work in a multi-agent setting, however extending the code to support multi-agent is not a bad idea

ultra/docs/curriculum_learning.md
JenishPatel99: Oh yes, when the two conditions are toggled, then there is an warning that is issued at runtime.
https://github.com/huawei-noah/SMARTS/blob/ecd25bbf5fc853dddb991b40766bd8c0fb2ed642/ultra/ultra/utils/curriculum/curriculum_info.py#L80-L87

If the cycle flag is true then the grades will start back from the beginning, in  this case it will be grade1.

I will clarify those two things in the documentation

ultra/docs/curriculum_learning.md
JenishPatel99: Yes, these are ignored if the pass_based_toggle is true

ultra/ultra/utils/curriculum/coordinator.py
JenishPatel99: Yes, I will initialize that to zero and any other variables as well

ultra/tests/test_curriculum.py
JenishPatel99: Yes, the problem was that the `display()` in coordinator.py uses the `self._grade_counter`. However, in the test cases, I forgot to admit the agent into curriculum (initial `next_train_grade`). Because of that the `self._grade_counter` was not intialized and it produced an error. I have updated the code and the new changes removes line 168

PR #814, opened 2021-05-04
Zoo manager spawned as a subprocess

Adaickalavan:
Changed `multiprocess` to `subprocess` in spawning zoo manager to avoid `daemonic processes are not allowed to have children` error while running SMARTS with XingTian.

### DIFF
--- a/smarts/core/remote_agent_buffer.py
+++ b/smarts/core/remote_agent_buffer.py
@@ -26,7 +26,6 @@ import subprocess
 import sys
 import time
 from concurrent import futures
-from multiprocessing import Process
 from typing import List, Tuple
 
 import grpc
--- a/smarts/core/remote_agent_buffer.py
+++ b/smarts/core/remote_agent_buffer.py
@@ -20,7 +20,10 @@
 
 import atexit
 import logging
+import pathlib
 import random
+import subprocess
+import sys
 import time
 from concurrent import futures
 from multiprocessing import Process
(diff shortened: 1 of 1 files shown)


smarts/core/remote_agent_buffer.py
```diff
@@ -185,9 +188,22 @@ def acquire_remote_agent(self, retries=3) -> RemoteAgent:
...
+            (pathlib.Path(__file__).parent.parent / "zoo" / "manager.py")
+            .absolute()
+            .resolve()
+        ),
+        "--port",
+        str(port),
+    ]
+
+    manager = subprocess.Popen(cmd)
+    if manager.poll() == None:
```
sah-huawei, line 203: Is it ok to do this immediately?  Or might the call to manager.py have some delay before it could potentially fail?

sah-huawei (approved):
Seems ok (although maybe remove that no-longer-necessary dependency).

smarts/core/remote_agent_buffer.py
```diff
@@ -20,7 +20,10 @@
 
 import atexit
 import logging
+import pathlib
 import random
+import subprocess
+import sys
 import time
 from concurrent import futures
 from multiprocessing import Process
```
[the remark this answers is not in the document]
Adaickalavan: Yes, I have removed it now.

smarts/core/remote_agent_buffer.py
Adaickalavan: Thus far, it looks okay to check whether the subprocess started properly soon after calling `subprocess.Popen(cmd)`. If the manager server inside `manager.py` fails later, we should see some grpc communication errors.

Gamenot (approved):

Issue #817, opened 2021-05-05
ULTRA Add goal completion to episode callbacks


JenishPatel99:
**Is your feature request related to a problem? Please describe.**
Not a problem, but this feature gives users/dev a direct access to view if the agent(s) has completed the goal (mission) from the episode callbacks

**Describe the solution you'd like**
Add reached_goal metric from agent_info to callback strings

PR #818, opened 2021-05-05
ULTRA Update callbacks

JenishPatel99:
Addressing Issue: #817

### DIFF
--- a/ultra/ultra/utils/episode.py
+++ b/ultra/ultra/utils/episode.py
@@ -321,7 +321,7 @@ def episodes(n, etag=None, log_dir=None):
             f"Total Steps",
             f"Steps/Sec",
             f"Score",
-            f"Goal completion",
+            f"Goal Completed",
         ],
         width=col_width,
         style="round",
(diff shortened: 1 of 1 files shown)


ultra/ultra/utils/episode.py
```diff
@@ -358,12 +359,20 @@ def episodes(n, etag=None, log_dir=None):
                     )
                     for agent_id, agent_info in e.info[e.active_tag].items()
                 ]
+                agent_goal_completion_strings = [
+                    "{}: {}".format(
+                        agent_id,
+                        int(agent_info.data["reached_goal"]),
```
christianjans, line 365: It might make more sense to cast to a `bool` here? Let me know your thoughts.
```suggestion
                        bool(agent_info.data["reached_goal"]),
```

ultra/ultra/utils/episode.py
```diff
@@ -321,6 +321,7 @@ def episodes(n, etag=None, log_dir=None):
             f"Total Steps",
             f"Steps/Sec",
             f"Score",
+            f"Goal completion",
```
christianjans, line 324:
```suggestion
            f"Goal Completed",
```
So that it matches the table title format of: https://github.com/huawei-noah/SMARTS/blob/4e4957b26b6d56e527c4831eba4c7eb0defa16bb/ultra/ultra/utils/episode.py#L321 Additionally, maybe changing it to "Goal Completed" (instead of "Goal Completion") would imply the True/False result we intend (see other review comment).

christianjans (changes requested):
Cool idea!

christianjans (approved):

Issue #473, opened 2021-01-25
Pybullet Build Log Pollution Regression


Gamenot:
Pybullet build log pollution has returned. This seems to be due to changes in pybullet switching the logs from c_stdout to c_stderr.

https://github.com/bulletphysics/bullet3/pull/3192

Issue #60, opened 2020-10-22
Remove "pybullet build time" stamps


liamchzh:
This is annoying
```
pybullet build time: Sep 20 2020 23:49:22
pybullet build time: Sep 20 2020 23:49:22
│              0/100 │               3.99 │                202 │              39.89 │            4lane_t │      basic.rou.xml │ 189590991351023167 │ 185.39 - Agent-007 │
pybullet build time: Sep 20 2020 23:49:22
pybullet build time: Sep 20 2020 23:49:22
pybullet build time: Sep 20 2020 23:49:22
│              1/100 │               4.72 │                202 │              47.25 │            4lane_t │      basic.rou.xml │ 189590991351023167 │ 185.39 - Agent-007 │
pybullet build time: Sep 20 2020 23:49:22
pybullet build time: Sep 20 2020 23:49:22
pybullet build time: Sep 20 2020 23:49:22
```

Gamenot:
This was addressed with #203.

PR #821, opened 2021-05-06
Quiet pybullet import msgs

sah-huawei:
I was getting "pybullet build time: ..." messages with each import of pybullet.   This seems to be a lingering issue from Issue #473 and PR #475 going back to #60.  This PR successfully quieted them for me by suppressing _both_ `stdout` and `stderr`.


Adaickalavan (approved):

Gamenot (approved):
I took a look at this again to make sure that it is not leaking.

PR #827, opened 2021-05-10
Traffic History Dones

### LINKED ISSUE #848: Goal needed for IL agents
When an agent controlled by an IL model hijacks a vehicle in a scenario generated from traffic history data (cf. `history_vehicles_replacement_for_imitation_learning.py`), it needs a way to determine when it has "completed its mission", which is (roughly) just to traverse the map like the other vehicles in the history data.

A new goal that is satisfied when the agent-controlled vehicle "drives off of the map" at a point where it is supposed to (as opposed to, for example, erroneously driving off of the side of the road).

**Describe alternatives you've considered**
Considered:
- just returning the `off_road` event and leaving it to the client to determine.  Not very helpful.
- using `PositionalGoals` at all of the map exit points.  But this is cumbersome to add to every map.
    - as a variation of this, considered automatically adding a `PositionalGoal`at the point where the hijacked vehicle exits the map in the history data.  But this doesn't allow for the IL model to make other choices.
- using `max_episode_steps` to limit the life of the agent.  But this limits its ability to vary its speed and trajectory.

**Additional context**
- Somewhat related to Issue #770.

sah-huawei:
This PR adds the ability to detect when vehicles and agents are done in traffic history simulations.  (cf. Imitation Learning)

When a non-controlled vehicle no longer shows up in the history, it will now be marked as `done` by the `agent_manager` in the `observe_from()` call.

A `TraverseGoal` goal was also added for history vehicles controlled by an imitation learning agent so they can detect when they've made it through the map (and distinguish this from driving off the side of the road).

Closes #848.


Gamenot:
I think there was some criticism here that running off the road because the map ends should not be a specific goal. I think maybe the same result could be achieved by applying a goal to the end of a particular road segment.

sah-huawei:
> I think there was some criticism here that running off the road because the map ends should not be a specific goal. I think maybe the same result could be achieved by applying a goal to the end of a particular road segment.

I'm not necessarily disagreeing, but I'd like to understand a bit more about what you have in mind here.  Would the user (person training/testing their IL model) need to specify a specific segment of a map as part of their code?  (If so, this seems klunky in that they will have to change their code each time they want to use a different map.)  Or were you thinking of somehow (pre-)specifying the goal in the scenario and/or map itself?  (If so, this doesn't seem flexible enough, since different agents may start at different points on the map and so on different road segments, so such a goal may need change as different agents are replaced.) 

I confess I didn't fully understand the criticism of the `ExitGoal` approach (aside from understanding that the `dones` flags should be implemented too), but maybe my understanding the alternatives better will help me to get that too.

Gamenot (approved):

sah-huawei:
> > I think there was some criticism here that running off the road because the map ends should not be a specific goal. I think maybe the same result could be achieved by applying a goal to the end of a particular road segment.
> 
> I'm not necessarily disagreeing, but I'd like to understand a bit more about what you have in mind here. Would the user (person training/testing their IL model) need to specify a specific segment of a map as part of their code? (If so, this seems klunky in that they will have to change their code each time they want to use a different map.) Or were you thinking of somehow (pre-)specifying the goal in the scenario and/or map itself? (If so, this doesn't seem flexible enough, since different agents may start at different points on the map and so on different road segments, so such a goal may need change as different agents are replaced.)
> 
> I confess I didn't fully understand the criticism of the `ExitGoal` approach (aside from understanding that the `dones` flags should be implemented too), but maybe my understanding the alternatives better will help me to get that too.

This was discussed in a group Zoom meeting.  It was decided it was poorly named (hence I renamed it), but this could go ahead.

CHANGELOG.md
```diff
@@ -20,8 +20,10 @@ process
 - Added `near realtime` mode and `uncapped` mode in Envision
 - Added `--allow-offset-map` option for `scl scenario build` to prevent auto-shifting of Sumo road networks
 - Added options in `DoneCriteria` to trigger ego agent to be done based on other agent's done situation
+- Added an `TraverseGoal` goal for imitation learning agents.
```
Adaickalavan, line 23:
```suggestion
- Added a `TraverseGoal` goal for imitation learning agents.
```

smarts/core/agent_manager.py
```diff
@@ -112,7 +112,7 @@ def remove_pending_agent_ids(self, agent_ids):
         assert agent_ids.issubset(self.agent_ids)
         self._pending_agent_ids -= agent_ids
 
-    def observe_from(self, sim, vehicle_ids):
+    def observe_from(self, sim, vehicle_ids, done_this_step=set()):
```
Adaickalavan, line 115: Consider specifying input/output type hints.

smarts/core/agent_manager.py
```diff
@@ -127,6 +127,12 @@ def observe_from(self, sim, vehicle_ids):
             rewards[agent_id] = vehicle.trip_meter_sensor(increment=True)
             scores[agent_id] = vehicle.trip_meter_sensor()
 
+        # also add agents that were done in virtue of just dropping out
+        for dvid in done_this_step:
+            aid = self._vehicle_with_sensors.get(dvid, None)
+            if aid:
+                dones[aid] = True
```
Adaickalavan, line 134: For `dvid` and `aid`, consider something like `done_v_id`, `a_id`, or `agent_id`, in order to be similar to those used a couple of lines above.

smarts/core/scenario.py
```diff
@@ -97,6 +97,28 @@ def is_reached(self, vehicle):
...
+    """An TraverseGoal is satisfied whenever an Agent-driven vehicle
+    successfully finishes traversing a non-closed (acyclical) map
+    It's a way for the vehicle to exit the simulation successfully,
+    for example, driving across from one side to the other on a
+    straight road and then continuing off the map.  This goal is
+    non-specific about *where* the map is exited, save for that
+    the vehicle must be going the correct direction in its lane
+    just prior to doing so."""
+
+    def __init__(self, road_network):
```
Adaickalavan, line 111: Consider specifying input/output type hints.

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -14,14 +15,14 @@
 
 
 class KeepLaneAgent(Agent):
-    def __init__(self, target_speed=15.0):
+    def __init__(self, target_speed: float = 15.0):
         self._target_speed = target_speed
 
-    def act(self, obs):
+    def act(self, obs) -> Tuple[float, int]:
```
Adaickalavan, line 21: from smarts.core.sensors import Observation
```suggestion
    def act(self, obs: Observation) -> Tuple[float, int]:
```

smarts/core/agent_manager.py
```diff
@@ -112,7 +112,11 @@ def remove_pending_agent_ids(self, agent_ids):
         assert agent_ids.issubset(self.agent_ids)
         self._pending_agent_ids -= agent_ids
 
-    def observe_from(self, sim, vehicle_ids):
+    def observe_from(
+        self, sim, vehicle_ids: Set[str], done_this_step: Set[str] = set()
+    ) -> Tuple[
+        Dict[str, Observation], Dict[str, float], Dict[str, float], Dict[str, bool]
+    ]:
```
Adaickalavan, line 119: from smarts.core.smarts import SMARTS
```suggestion
    def observe_from(
        self, sim: SMARTS, vehicle_ids: Set[str], done_this_step: Set[str] = set()
    ) -> Tuple[
        Dict[str, Observation], Dict[str, float], Dict[str, float], Dict[str, bool]
    ]:
```

smarts/core/scenario.py
```diff
@@ -97,6 +97,28 @@ def is_reached(self, vehicle):
         return dist <= self.radius
 
 
+@dataclass
+class TraverseGoal(Goal):
+    """An TraverseGoal is satisfied whenever an Agent-driven vehicle
```
Adaickalavan, line 102:
```suggestion
    """A TraverseGoal is satisfied whenever an Agent-driven vehicle
```

smarts/core/scenario.py
```diff
@@ -97,6 +97,28 @@ def is_reached(self, vehicle):
...
+    just prior to doing so."""
+
+    def __init__(self, road_network: SumoRoadNetwork):
+        super().__init__()
+        self._road_network = road_network
+
+    def is_endless(self):
+        return True
+
+    def is_reached(self, vehicle):
```
Adaickalavan, line 118: Consider specifying input/output type hints.

smarts/core/sumo_road_network.py
```diff
@@ -546,7 +546,49 @@ def point_is_within_road(self, point):
 
         radius = max(5, 2 * self._default_lane_width)
         nearest_lanes = self.nearest_lanes(point[:2], radius=radius)
-        return any(0.5 * nl.getWidth() + 1e-1 > dist for nl, dist in nearest_lanes)
+        return any(dist < 0.5 * nl.getWidth() + 1e-1 for nl, dist in nearest_lanes)
+
+    def drove_off_map(self, veh_position, veh_heading):
```
Adaickalavan, line 551: Consider specifying input/output type hints.

Adaickalavan (approved):

smarts/core/agent_manager.py
sah-huawei: fwiw, I sort of disagree with this one for complicated reasons, although I'm willing to do it.

smarts/core/scenario.py
sah-huawei: These match the base class (as I think they should), which doesn't have type hints.   If I change these, I think I should also change the base class, which means changing other classes derived from it.  This is a bit more dangerous.

smarts/core/agent_manager.py
sah-huawei: So I tried that (with the import as you specified it) but got a weird error on importing due to circular references:
```
  File "/home/steve/Repos/Huawei/SMARTS/examples/observation_collection_for_imitation_learning.py", line 9, in <module>
  File "/home/steve/Repos/Huawei/SMARTS/smarts/core/smarts.py", line 42, in <module>
    from .agent_manager import AgentManager
  File "/home/steve/Repos/Huawei/SMARTS/smarts/core/agent_manager.py", line 31, in <module>
ImportError: cannot import name 'SMARTS' from 'smarts.core.smarts' (/home/steve/Repos/Huawei/SMARTS/smarts/core/smarts.py)
```

It's probably possible to fix this, but given that I'm already a bit predisposed against specifying `SMARTS` here as the type, I'm going to go ahead and pass on this one.

PR #836, opened 2021-05-11
Multiple replacement of history vehicles

### LINKED ISSUE #816: example of replacing multiple history vehicles
Request from @zbzhu99 in Wechat:

> "I think we need to further implement multi-agent imitation learning on SMARTS, which needs to make all the history replay vehicles to be controlled by outside policies.  Can you provide an example script for this?"

sah-huawei:
Expanded the `history_vehicles_replacement_for_imitation_learning.py` example to replace multiple vehicles at a time, as per Issue #816.

Along the way, factored out a `TrafficHistory` class from `Scenario`.

Closes #816.


examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -21,62 +24,150 @@ def act(self, obs):
...
         traffic_sim=None,
         envision=None if headless else Envision(),
     )
+    random_seed(seed)
+    traffic_history_provider = smarts.get_provider_by_type(TrafficHistoryProvider)
+    assert traffic_history_provider
+
+    scenarios_iterator = Scenario.scenario_variations(scenarios, [])
     for _ in scenarios:
         scenario = next(scenarios_iterator)
```
Gamenot, line 45:
```suggestion
    for scenario in scenarios_iterator:
```

smarts/core/agent_manager.py
```diff
@@ -72,6 +72,7 @@ def teardown(self):
         self.teardown_ego_agents()
         self.teardown_social_agents()
         self._vehicle_with_sensors = dict()
+        self._pending_agent_ids = set()
```
Gamenot, line 75: Interesting that this was not reset on `teardown()` before.

smarts/core/agent_manager.py
```diff
@@ -303,7 +304,7 @@ def send_observations_to_social_agents(self, observations):
             obs = observations[agent_id]
             self._remote_social_agents_action[agent_id] = remote_agent.act(obs)
 
-    def switch_initial_agent(self, agent_interface):
+    def switch_initial_agents(self, agent_interface):
         self._initial_interfaces = agent_interface
```
Gamenot, line 308: The params need to also be updated I think.
```suggestion
    def switch_initial_agents(self, agent_interfaces):
        self._initial_interfaces = agent_interfaces
```

smarts/core/scenario.py
```diff
@@ -516,56 +519,25 @@ def _discover_bubbles(scenario_root):
             bubbles = pickle.load(f)
             return bubbles
 
-    def set_ego_missions(self, ego_mission):
-        self._missions.update(ego_mission)
-
```
Gamenot, line 521: That was a dangerous method.

smarts/core/smarts.py
```diff
@@ -330,8 +332,8 @@ def add_provider(self, provider):
         assert isinstance(provider, Provider)
         self._providers.append(provider)
 
-    def switch_ego_agent(self, agent_interface):
-        self._agent_manager.switch_initial_agent(agent_interface)
+    def switch_ego_agents(self, agent_interface):
+        self._agent_manager.switch_initial_agents(agent_interface)
```
Gamenot, line 336: Plural again.
```suggestion
    def switch_ego_agents(self, agent_interfaces):
        self._agent_manager.switch_initial_agents(agent_interfaces)
```

smarts/core/traffic_history.py
```diff
@@ -0,0 +1,124 @@
...
+    def connect_if_not(self):
+        if not self._db_cnxn:
+            self._db_cnxn = sqlite3.connect(self._db)
+
+    def close_db(self):
+        if self._db_cnxn:
+            self._db_cnxn.close()
+            self._db_cnxn = None
+
+    def _query_val(self, query, params=(), result_type=float):
```
Gamenot, line 43: I think a default `result_type` is dangerous here. A user should always know what type they are going to get back.

smarts/core/traffic_history.py
```diff
@@ -0,0 +1,124 @@
...
+    @cached_property
+    def lane_width(self):
+        return self._query_val("SELECT value FROM Spec where key='map_net.lane_width'")
+
+    @cached_property
+    def target_speed(self):
+        return self._query_val("SELECT value FROM Spec where key='speed_limit_mps'")
+
+    @lru_cache(maxsize=32)
+    def vehicle_last_seen_time(self, vehicle_id):
```
Gamenot, line 73: Perhaps `vehicle_final_exit_time`? I was wondering if this was referring to the most recent time the vehicle was seen.

smarts/core/traffic_history.py
```diff
@@ -0,0 +1,124 @@
...
+
+class TrafficHistory:
+    def __init__(self, db):
+        self._db = db
+        self._db_cnxn = None
+
+    def connect_if_not(self):
+        if not self._db_cnxn:
+            self._db_cnxn = sqlite3.connect(self._db)
+
```
Gamenot, line 37: I find calling connect on the traffic history to be somewhat odd. Perhaps just something closer to `initialize` or `setup`?

smarts/core/traffic_history.py
```diff
@@ -0,0 +1,124 @@
...
+    def close_db(self):
+        if self._db_cnxn:
+            self._db_cnxn.close()
+            self._db_cnxn = None
```
Gamenot, line 41: Same here: `teardown()`?

smarts/core/agent_manager.py
sah-huawei: I know.  I was surprised we never hit bugs related to this one!

smarts/core/traffic_history.py
sah-huawei: Yeah, I considered `setup()`, but I shied away from that since calling it is optional.  This is an optimization to open a persistent connection if a client is going to be doing a large number multiple queries (which avoids the overhead of parsing the sqllite file for each one).  Other clients can call queries without using this and they'll also work.

Maybe something like `connect_for_repeated_queries()` paired with `disconnect()`?

smarts/core/traffic_history.py
sah-huawei: Yeah, agreed -- the user should always know, although the problem was that then `params` cannot be optional either, and I think it's important for it to be, and I also think that re-ordering `params` to be the last argument is not desirable since it's sort of "bound more tightly" to the `query` argument.

I guess it could be the very first argument (sort of like a templatized function in C++).  I'll try that and see how it looks...

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -21,62 +24,150 @@ def act(self, obs):
...
-            # tell the traffic history provider to start traffic
-            # at the point when this agent enters...
-            traffic_history_provider = smarts.get_provider_by_type(
-                TrafficHistoryProvider
+            continue
+        veh_start_times = {
+            vid: mission.start_time for vid, mission in veh_missions.items()
+        }
+
+        k = vehicles_to_replace
```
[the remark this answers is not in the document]
sah-huawei: I changed it to `--replacements-per-episode` (although it still supports `-k` as a shortcut for that).

smarts/core/traffic_history.py
sah-huawei: OK, I think I'm ok with the way it looks as the first argument.  Going with that.

smarts/core/traffic_history.py
Gamenot: I see. Then in that case I understand why this would be needed.

smarts/core/traffic_history.py
```diff
@@ -0,0 +1,139 @@
...
+        self, start_time: float, end_time: float
+    ) -> Generator[
+        Tuple[int, int, float, float, float, float, float, float], None, None
+    ]:
+        query = """SELECT V.id, V.type, V.length, V.width,
+                          T.position_x, T.position_y, T.heading_rad, T.speed
+                   FROM Vehicle AS V INNER JOIN Trajectory AS T ON V.id = T.vehicle_id
+                   WHERE T.sim_time > ? AND T.sim_time <= ?
+                   ORDER BY T.sim_time DESC"""
+        return self._query_list(query, (start_time, end_time))
```
Gamenot, line 116: I would consider wrapping the generator with another generator to get types that way you do not need to wrap it later.
```suggestion
        row = self._query_list(query, (start_time, end_time))
        return (TrafficHistoryProvider._HistoryRow(*row) for row in rows)
```

Gamenot (approved):

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -14,69 +18,164 @@
 
 
 class KeepLaneAgent(Agent):
-    def __init__(self, target_speed=15.0):
+    def __init__(self, target_speed: float = 15.0):
         self._target_speed = target_speed
 
-    def act(self, obs):
+    def act(self, obs) -> Tuple[float, int]:
```
Adaickalavan, line 24: from smarts.core.sensors import Observation
```suggestion
    def act(self, obs: Observation) -> Tuple[float, int]:
```

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -14,69 +18,164 @@
...
             )
-            assert traffic_history_provider
-            traffic_history_provider.start_time = mission.start_time
-
-            # agent vehicle will enter right away...
-            modified_mission = replace(mission, start_time=0.0)
-            scenario.set_ego_missions({agent_id: modified_mission})
+            continue
+        veh_start_times = {
+            vid: mission.start_time for vid, mission in veh_missions.items()
```
Adaickalavan, line 61: For `vid`, perhaps consider `v_id` to be similar with other parts of the code.

smarts/core/agent_manager.py
```diff
@@ -303,8 +304,8 @@ def send_observations_to_social_agents(self, observations):
             obs = observations[agent_id]
             self._remote_social_agents_action[agent_id] = remote_agent.act(obs)
 
-    def switch_initial_agent(self, agent_interface):
-        self._initial_interfaces = agent_interface
+    def switch_initial_agents(self, agent_interfaces):
```
Adaickalavan, line 307: Consider specifying input/output type hints.

smarts/core/traffic_history.py
sah-huawei: Good call.  Done.

smarts/core/scenario.py
```diff
@@ -516,60 +519,29 @@ def _discover_bubbles(scenario_root):
...
+        for row in self._traffic_history.first_seen_times():
             start_time = float(row[1])
-            p_cur = histories_db.cursor()
-            vrow = p_cur.execute(p_query, (int(vid), start_time)).fetchone()
-            assert vrow
-            pos_x, pos_y, heading = vrow
+            pph = self._traffic_history.vehicle_pose_at_time(row[0], start_time)
+            assert pph
+            pos_x, pos_y, heading = pph
+            vid = str(row[0])
```
Adaickalavan, line 532:
```suggestion
            v_id = str(row[0])
```

smarts/core/scenario.py
```diff
@@ -516,60 +519,29 @@ def _discover_bubbles(scenario_root):
...
             vehicle_missions[vid] = Mission(
                 start=Start(
                     (pos_x + map_offset[0], pos_y + map_offset[1]), Heading(heading)
                 ),
                 goal=EndlessGoal(),
                 start_time=start_time,
+                vehicle_id=vid,
```
Adaickalavan, line 539:
```suggestion
                vehicle_id=v_id,
```

smarts/core/scenario.py
```diff
@@ -516,60 +519,29 @@ def _discover_bubbles(scenario_root):
...
             start_time = float(row[1])
-            p_cur = histories_db.cursor()
-            vrow = p_cur.execute(p_query, (int(vid), start_time)).fetchone()
-            assert vrow
-            pos_x, pos_y, heading = vrow
+            pph = self._traffic_history.vehicle_pose_at_time(row[0], start_time)
+            assert pph
+            pos_x, pos_y, heading = pph
+            vid = str(row[0])
             vehicle_missions[vid] = Mission(
```
Adaickalavan, line 533:
```suggestion
            vehicle_missions[v_id] = Mission(
```

smarts/core/smarts.py
```diff
@@ -330,8 +332,8 @@ def add_provider(self, provider):
         assert isinstance(provider, Provider)
         self._providers.append(provider)
 
-    def switch_ego_agent(self, agent_interface):
-        self._agent_manager.switch_initial_agent(agent_interface)
+    def switch_ego_agents(self, agent_interfaces):
```
Adaickalavan, line 335: Consider specifying input/output type hints.

smarts/core/scenario.py
sah-huawei: I'm going to go ahead and make this change, although I want to push back slightly and say that I don't think we _always_ need to use `v_id` for a vehicle id everywhere in the code,  just in functions and/or files where it's _already_ being called `v_id`.

smarts/core/smarts.py
sah-huawei: Yeah, I considered that here too (after I added them to the `switch_initial_agent()` method), but I rejected that here since `agent_interface` is just being forwarded.  (That is, the type for this function doesn't matter, no methods/fields are accessed and it's passed on regardless, to be checked at the receiver.)

smarts/core/traffic_history.py
```diff
@@ -0,0 +1,150 @@
...
+        disconnect() should be called when finished."""
+        if not self._db_cnxn:
+            self._db_cnxn = sqlite3.connect(self._db)
+
+    def disconnect(self):
+        if self._db_cnxn:
+            self._db_cnxn.close()
+            self._db_cnxn = None
+
+    def _query_val(self, result_type, query: str, params=()):
```
Adaickalavan, line 50: Consider specifying input/output type hints.

smarts/core/traffic_history.py
```diff
@@ -0,0 +1,150 @@
...
+        ) as dbcnxn:
+            cur = dbcnxn.cursor()
+            cur.execute(query, params)
+            row = cur.fetchone()
+            cur.close()
+        if not row:
+            return None
+        return row if result_type is tuple else result_type(row[0])
+
+    def _query_list(self, query: str, params=()):
```
Adaickalavan, line 62: Consider specifying input/output type hints.

Adaickalavan (approved):

Issue #826, opened 2021-05-07
Clean-up pytest coverage files


Gamenot:
Coverage files left over after running `make test` causes a large leftover mass of `.coverageXXX` files that then must be cleaned up

AC:
- [x] `make` options `test`, `sanity-test`, `test learning`, `test-memory-growth`, `test-long-determinism` clean up all `.coverage` files after exit.

@sah-huawei this will be passed on to RG when he joins.

PR #838, opened 2021-05-11
added recipe to remove pytest coverage files after all tests. Issue #826

RutvikGupta:


Makefile
```diff
@@ -31,10 +32,12 @@ sanity-test: build-all-scenarios
 		./tests/test_examples.py::test_examples[multi_agent] \
 		./tests/test_examples.py::test_multi_instance_example \
 		./smarts/env/tests/test_social_agent.py::test_social_agents
+	rm -f .coverage*
```
sah-huawei, line 35: I'm not sure this one is needed since we aren't passing `--cov` to pytest here.

Makefile
```diff
@@ -31,10 +32,12 @@ sanity-test: build-all-scenarios
...
 
 .PHONY: test-learning
 test-learning: build-all-scenarios
 	pytest -v -s -o log_cli=1 -o log_cli_level=INFO ./smarts/env/tests/test_learning.py
+	rm -f .coverage*
```
sah-huawei, line 40: ditto

Makefile
```diff
@@ -44,13 +47,15 @@ test-memory-growth: build-all-scenarios
...
 		./smarts/core/tests/test_smarts_memory_growth.py
+	rm -f .coverage*
 
 .PHONY: test-long-determinism
 test-long-determinism: 
 	scl scenario build --clean scenarios/minicity
 	PYTHONHASHSEED=42 pytest -v \
 		--forked \
 		./smarts/env/tests/test_determinism.py::test_long_determinism
+	rm -f .coverage*
```
sah-huawei, line 58: ditto

Makefile
```diff
@@ -13,6 +13,7 @@ test: build-all-scenarios
 		--ignore=./smarts/env/tests/test_benchmark.py \
 		--ignore=./smarts/env/tests/test_learning.py \
 		-k 'not test_long_determinism'
+	rm -f .coverage*
```
sah-huawei, line 16: Just to be a tad more careful... everywhere where we keep such an `rm` like this, we should change the thing being removed to `.coverage.*`.  (Notice the extra dot.)  The more restrictive we can make the pattern, the safer it is.

Makefile
```diff
@@ -152,3 +158,7 @@ wheel: docs
 .PHONY: rm-pycache
 rm-pycache:
 	find . -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -delete
+
+.PHONY: rm-cov
+rm-cov:
+	find . -type f -name ".coverage*" -delete
```
sah-huawei, line 164: same comment here as in the `rm -f` case:  let's use `.coverage.*` (with the second dot as part of the pattern).

sah-huawei (changes requested):

Makefile
```diff
@@ -44,6 +45,7 @@ test-memory-growth: build-all-scenarios
 		--dist=loadscope \
 		-n `nproc --ignore 1` \
 		./smarts/core/tests/test_smarts_memory_growth.py
+	rm -f .coverage*
```
sah-huawei, line 48: I meant to add the dot here too.

Makefile
```diff
@@ -126,6 +128,7 @@ clean:
 	rm -f ./$(scenario)/*.rou.alt.xml
 	rm -rf ./$(scenario)/traffic
 	rm -rf ./$(scenario)/social_agents
+	rm -f .coverage*
```
sah-huawei, line 131: ... and here.

sah-huawei (changes requested):

Makefile
Gamenot: Is that dot needed? I know there is also a file called `.coverage` that is left behind.

sah-huawei: In general, I want to be careful with `rm -f` commands using wildcards to make sure they don't remove more than intended... the more restrictive the match pattern, the better.  So I think it's a bit better to have the dot just to restrict the matching cases a slight bit more.  (I'm imagining the case where someone did something like `cp .coverage.whatever.1234.5678 .coverage` in order to make a backup of a particularly-interesting coverage file.)

I personally haven't had a bare  `.coverage` file auto-created on my own laptop, but if they are then I guess we could probably do without the final dot.  Or, alternatively, we could just add an additional, explicit `rm -f .coverage` line without a wildcard to handle that case separately.  I guess it's not that big of a deal.

sah-huawei (approved):
thanks

Makefile
RutvikGupta: thanks for the suggestion.

PR #839, opened 2021-05-11
Removed scenarios and examples modules from pip installation

RutvikGupta:
…they are removed from python installation package or pip


examples/egoless.py
```diff
@@ -2,7 +2,7 @@
 
 import gym
 
-from examples import default_argument_parser
```
Gamenot, line 5: I figure we assume that we are calling python with the repository root in path.

I do think a relative import would be more useful here in the case that someone attempts to use the examples with a packaged version of SMARTS.

RutvikGupta: Its not only that import statement which I changed that has the absolute path but others as well. Do you want me to change those as well?

sah-huawei (approved):

examples/egoless.py
Gamenot: Yes.

PR #845, opened 2021-05-12
Bugfix remove unused `make` pview

Gamenot:
`pview` is no longer usable since we do not keep around `.egg` files (nor would we add them back.) This is just a small clean-up removing the `pview` option from `make`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Copy and pasting the git commit messages is __NOT__ enough.
 
 ## [Unreleased]
 
+## Removed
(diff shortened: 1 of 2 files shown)


RutvikGupta (approved):
Looks good to me!
 I guess we dont use pview anymore to preview 3D maps because we use envision?

sah-huawei (approved):

Gamenot:
> Looks good to me!
> I guess we dont use pview anymore to preview 3D maps because we use envision?

Sorry, yes, we do not keep around these `.egg` files so pview is not at all useful. There are plenty of 3rd party visualisation tools that can be used to visualise models, even when disregarding envision.

PR #852, opened 2021-05-13
Bugfix memory issues

Gamenot:
Two memory issues here:
 - Sumo memory grows when just reset
 - ThreadPoolExecutor grows relative to the number of CPUs available.


smarts/core/remote_agent_buffer.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, zoo_manager_addrs=None, buffer_size=3):
             conn["channel"], conn["stub"] = get_manager_channel_stub(conn["address"])
 
         self._buffer_size = buffer_size
-        self._replenish_threadpool = futures.ThreadPoolExecutor()
+        self._replenish_threadpool = futures.ThreadPoolExecutor(max_workers=max_workers)
```
Gamenot, line 90: What I found here was that the `ThreadPoolExecutor` would continue to add threads until it used up all CPUs.

smarts/core/sumo_traffic_simulation.py
```diff
@@ -240,20 +242,24 @@ def _base_sumo_load_params(self):
...
-        else:
-            restart_sumo = True
+        restart_sumo = (
+            not (
+                self._scenario
+                and self._scenario.net_file_hash == next_scenario.net_file_hash
+            )
+            or self._current_reload_count >= self._reload_count
+        )
+        self._current_reload_count = self._current_reload_count % self._reload_count + 1
```
Gamenot, line 259: If you only reload SUMO there will be a gradual and uncapped memory growth. 

I will need to change the comment here.

sah-huawei: This seems ok to me, although I have a slight preference to "flatten" the boolean logic, like:
```python
restart_sumo = (
    not self._scenario
    or self._scenario.net_file_hash != next_scenario.net_file_hash
    or self._current_reload_count >= self._reload_count
)
```

If there isn't a Sumo option for resetting, we should probably file an Issue with them letting  them know they might have a leak.

smarts/core/remote_agent_buffer.py
sah-huawei: I'm wondering if we should use something like `os.cpu_count() / 2` here.  (I don't feel strongly about this either way though.  Hard-coding `4` also seems fine to me.)

sah-huawei (approved):

smarts/core/remote_agent_buffer.py
Gamenot: Yes, I am thinking I might do something like `clamp()` to place CPUs at `os.cpu_count() - 1` between `1` and `4`.

Adaickalavan: Actually this is only for buffering purposes to avoid code slowdowns, hence hard limiting to a max of 4 threads is fine.

smarts/core/sumo_traffic_simulation.py
```diff
@@ -240,20 +246,22 @@ def _base_sumo_load_params(self):
 
         return load_params
 
-    def setup(self, scenario) -> ProviderState:
+    def setup(self, next_scenario) -> ProviderState:
```
Adaickalavan, line 249: Consider specifying input/output type hints.

Adaickalavan (approved):

smarts/core/remote_agent_buffer.py
Gamenot: Yes, on review, I think that it is correct to just roll it back to 4.

PR #853, opened 2021-05-13
Fixed relative paths

RutvikGupta:
… installation in setup.py


sah-huawei (approved):
thanks!  I think that should be better.

PR #854, opened 2021-05-13
new Imitation action space

### LINKED ISSUE #844: Action space to directly control vehicle acceleration
Driving datasets like NGSIM only contain driving trajectories without action information. But for most imitation methods, we need expert action guidance. Since we can not directly recover brake and throttle from trajectories, the current continuous agent type can not be used.

A new action space that includes the acceleration and angular velocity is needed so that we can compute the expert action from the expert states(e.g., position, speed, and heading at each timestep).

sah-huawei:
Adds a new `ActionSpaceType.Imitation` type and controller that allows for the user to use `(acceleration, angular_velocity)` pairs as their control actions.

There are two options here:  the default uses `pybullet` physics (using underlying throttle, steering angle, torque, etc.) and an `AckermannChassis`, the other bypasses physics (setting vehicle speed and position directly) and a `BoxChassis`.   (The latter is not yet exposed to users.)

Along the way, also fixed bug that prevented the control of agents having non-dynamic action spaces.

Closes #844.


Gamenot (approved):
I just wonder why there are multiple options for the chassis type?

smarts/core/controllers/imitation_controller.py
```diff
@@ -0,0 +1,70 @@
...
+            else:
+                throttle = 0
+                # necessary torque is N*m = kg*m*acceleration
+                torque_ratio = mass / vehicle.chassis.max_btorque
+                brake = np.clip(-acceleration * torque_ratio, 0, 1)
+            steering = np.clip(
+                dt * -angular_velocity * vehicle.chassis.steering_ratio, -1, 1
+            )
+            vehicle.control(throttle=throttle, brake=brake, steering=steering)
+
```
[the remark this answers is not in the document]
sah-huawei: not really.  I think Jun said something like (paraphrasing) "it's probably not necessary to do _{ however he referred to the pybullet case }_, but it would be good if we could eventually."

I was planning to leave it up to him and/or the imitation learning folks to choose.  

(but also, I wanted to show here that it's not necessary to tightly tie the controller to the chassis type.  but it's klunky like this.   this is part of a bigger architectural discussion we need to have eventually, as well, cf. polymorphism and OO.)

smarts/core/controllers/imitation_controller.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+            vehicle.control(new_pose, target_speed)
+
+        # TAI: consider making this a dynamic_action_space, like...
+        elif isinstance(vehicle.chassis, AckermannChassis):
+            mass = vehicle.chassis.mass_and_inertia[0]  # in kg
+            if acceleration >= 0:
+                # necessary torque is N*m = kg*m*acceleration
+                torque_ratio = mass / vehicle.chassis.max_torque
+                throttle = np.clip(acceleration * torque_ratio, 0, 1)
```
iman512003, line 65: Just a minor thing, since the throttle is dimensionless, may be we should use: 
`torque_ratio = mass / (4*wheel_radius*vehicle.chassis.max_torque)`

iman512003 (approved):

smarts/core/controllers/imitation_controller.py
sah-huawei: I think that in the chassis throttle value (between `0` and `1`) is scaled up to be between `0` nd `max_torque`.  That's why I did it this way.  It seems that is we also normalize here by `4 * wheel_radius`, we're limiting the maximum amount of torque we can provide to the vehicle (and so the maximum acceleration) by something like 80% of its available range.  I didn't think we'd want to do that. 
 
Or am I misunderstanding things?

iman512003: When we use `AckermannChassis` and we go through the bullet. In `control` method of `AckermannChassis`, a list of throttles for each wheel will be created(if we pass a scalar throttle), thus if we want to produce specific acceleration with bullet the equation of motion dictates:

`4*wheel_radius*throttle*max_torque=mass*acceleration`

from which we obtain:

`throttle=mass*acceleration/(4*wheel_radius*max_torque)`

sah-huawei: Ah ha, I see that now.  Thank you!  I will make the change.

Issue #846, opened 2021-05-12
'wrong_way' event detected incorrectly


nisheeth-golakiya:
## **BUG REPORT**
**High Level Description**
For a particular ego-mission in 2lane intersection map, 'wrong_way' event is detected even if the agent is following the correct path.

**SMARTS version**
0.4.15

**Steps to reproduce the bug**
In `scenarios/intersections/2lane/scenario.py`, set the following mission for ego-vehicle
```python
import os

from smarts.sstudio import gen_scenario
from smarts.sstudio.types import (
    Route,
    Mission,
    Scenario,
    TrapEntryTactic,
)

from pathlib import Path

mission = Mission(Route(begin=('edge-west-WE', 0, 10),
                        end=('edge-north-SN', 0, 50)),
                  entry_tactic=TrapEntryTactic(wait_to_hijack_limit_s=0,
                                               default_entry_speed=3.0))

gen_scenario(Scenario(ego_missions=[mission]),
             output_dir=Path(__file__).parent)
```

In the terminal, run the agent as follows
```python
from smarts.core.agent_interface import AgentInterface, AgentType
from smarts.core.agent import AgentSpec, Agent
from smarts.core.utils.episodes import episodes

from examples import default_argument_parser

import gym
N_AGENTS = 1
AGENT_IDS = ["Agent %i" % i for i in range(N_AGENTS)]


class KeepLaneAgent(Agent):
    def act(self, obs):
        return (5, 0)
agent_specs = {
        agent_id: AgentSpec(
            interface=AgentInterface.from_type(
                AgentType.LanerWithSpeed, max_episode_steps=500
            ),
            agent_builder=KeepLaneAgent,
        )

        for agent_id in AGENT_IDS
    }
from smarts.env.hiway_env import HiWayEnv
env = HiWayEnv(scenarios=['scenarios/intersections/2lane'],
        agent_specs=agent_specs,
        headless=False,
        seed=42,
    )
agents = {
    agent_id: agent_spec.build_agent()
    for agent_id, agent_spec in agent_specs.items()
}
observations = env.reset()

for _ in range(500):
    agent_actions = {
        agent_id: agents[agent_id].act(agent_obs)
        for agent_id, agent_obs in observations.items()
    }
    observations, _, _, _ = env.step(agent_actions)
    events = observations['Agent 0'].events
    if events.wrong_way:
        print(observations['Agent 0'].ego_vehicle_state.lane_id)
        break
```

[...]

PR #860, opened 2021-05-17
Fixed incorrect wrong-way event while crossing an intersection. See issue #846 

RutvikGupta:


CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - `history_vehicles_replacement_for_imitation_learning.py` now uses new Imitation action space.  See Issue #844.
 ### Fixed
 - Allow for non-dynamic action spaces to have action controllers.  See PR #854.
-
+- Fixed a bug in `sensors.py` which triggered `wrong_way` event when the vehicle goes into an intersection. See Issue #846.
```
sah-huawei, line 18: we'll probably want to keep the newline between version blocks.

smarts/core/sensors.py
```diff
@@ -503,9 +508,12 @@ def _vehicle_is_off_route_and_wrong_way(cls, sim, vehicle):
             # Lanes from an edge are parallel so any lane from the edge will do for direction check
             # but the innermost lane will be the last lane in the edge and usually the closest.
             lane_to_check = route_edge_or_oncoming.getLanes()[-1]
-            is_wrong_way = cls._vehicle_is_wrong_way(
-                sim, vehicle, lane_to_check.getID()
-            )
+            if lane_to_check.getEdge().isSpecial():
+                is_wrong_way = False
```
sah-huawei, line 512: You should probably add a comment here (referencing the Issue).
(... and in the other spot you did this.)

supervisord.conf
```diff
@@ -1,6 +1,6 @@
 [program:smarts]
 environment=PYTHONUNBUFFERED=1
-command=python examples/single_agent.py scenarios/intersections/6lane
+command=python examples/single_agent.py scenarios/intersections/2lane
```
sah-huawei, line 3: why was this change made?  (I'm not necessarily against it, I'm just curious.)

RutvikGupta: I was trying to reproduce the bug. And the bug involved using 2lane intersection. So i had to change the config file. The workstation I got doesn't seem to be working so i have to push changes to a branch and pull them on remote to test them out there.

CHANGELOG.md
RutvikGupta: ok sure, will add these changes.

smarts/core/sensors.py
RutvikGupta: ok.

Gamenot (commented):
Some comments.

supervisord.conf
sah-huawei: no problem, although I think we should avoid pushing unrelated changes.  we should back this one out now that you've tested your code.

smarts/core/sensors.py
```diff
@@ -522,6 +520,18 @@ def _vehicle_is_wrong_way(sim, vehicle, lane_id):
...
+    def _check_wrong_way_event(cls, lane_to_check, sim, vehicle):
+        # When the vehicle is in an intersection, turn off the `wrong way` check to avoid
+        # false positive `wrong way` events.
+        if lane_to_check.getEdge().isSpecial():
+            is_wrong_way = False
+        else:
+            is_wrong_way = cls._vehicle_is_wrong_way(
+                sim, vehicle, lane_to_check.getID()
+            )
+        return is_wrong_way
```
sah-huawei, line 533: Not that big of a deal, but this can be simplified to:
```python
@classmethod
def _check_wrong_way_event(cls, lane_to_check, sim, vehicle):
    # When the vehicle is in an intersection, turn off the `wrong way` check to avoid
    # false positives 'wrong way' events.  (see Issue #846)
    if lane_to_check.getEdge().isSpecial():
        return False
    return cls._vehicle_is_wrong_way(sim, vehicle, lane_to_check.getID())
```

Issue #867, opened 2021-05-18
ULTRA Update maps to remove extra road segments


JenishPatel99:
Current maps for ULTRA are equipped with a road segment which branches off from the south junction. **An image describing it will be posted shortly. That extra road was meant for analysis purpose, where we needed a stationary ego vehicle (which was positioned on that road) to provide observations to the stopwatcher. Although, it helps in analysis of scenarios, however it is not ideal for training/testing purposes. 

Task:
- [x]  Copy existing maps to `scenarios/pool/analysis` to save those maps for analysis purposes
- [x]  Update existing maps by removing the bottom road segments

PR #868, opened 2021-05-19
ULTRA Update existing maps

JenishPatel99:
Issue: #867 

**Note: The large amount of files that were changed are primarily the map files. I have listed the files below which require revision
- `scenarios/generate_scenarios.py`
- `scenarios/interface.py`
- `tests/test_analysis.py`


ultra/ultra/scenarios/interface.py
```diff
@@ -53,14 +53,20 @@
         "--task", help="type a task id [0, 1, 2, 3]", type=str
     )
     parser_generate_scenarios.add_argument(
-        "--save-dir", help="directory for saving maps", type=str, default=None
+        "--save-dir", help="directory for saving scenarios", type=str, default=None
     )
     parser_generate_scenarios.add_argument(
         "--root-dir",
-        help="directory for saving maps",
+        help="directory for saving scenarios",
```
christianjans, line 60: Good catch, but isn't this the directory where we can find the `task<name>/` folder? Maybe something like:
```suggestion
        help="path to task directory",
```
I think this is what the flag does, but of course, correct me if I'm wrong.

ultra/ultra/scenarios/pool/2lane_c/100kmh/map.net.xml
```diff
@@ -1,226 +1,204 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
christianjans, line 1: How similar are these new map files to the old ones? Is it just the extra road segment that is missing? Are all the other edges the same size and at the same coordinates?

ultra/tests/test_analysis.py
```diff
@@ -54,6 +54,7 @@ def test_interface_analyze(self):
                 stopwatcher_route="south-west",
```
christianjans, line 54: Are the maps with the extra road segment necessary for analysis? I forgot to ask in the meeting, but what did Kimia say about this?

It would be ideal if we could perform the analysis on the same maps we are experimenting on... Do you see any way of doing this while still getting rid of the extra road segment on the maps?

ultra/ultra/scenarios/generate_scenarios.py
```diff
@@ -700,7 +701,11 @@ def build_scenarios(
     level_config = task_config["levels"][level_name]
     scenarios_dir = os.path.dirname(os.path.realpath(__file__))
     task_dir = f"{scenarios_dir}/{task}"
-    pool_dir = f"{scenarios_dir}/pool"
+
+    if pool_dir is None:
+        pool_path = os.path.join(scenarios_dir, "pool")
+    else:
+        pool_path = os.path.join(scenarios_dir, pool_dir)
```
christianjans, line 708: If the analysis scenarios are necessary to keep, I wonder if we can have an `pool/analysis_pool/` directory for the maps used for analysis, and then a default `pool/experiment_pool/`.

christianjans (approved):
Looks great!

Issue #830, opened 2021-05-10
Add General Map API


Gamenot:
**Is your feature request related to a problem? Please describe.**
It is difficult from both SStudio and an agent to view specific aspects of a map.

AC:
- [ ] Map descriptor is known to agent to allow access to map through map API
- [ ] Agent does not require taking map information through constructor
- [ ] Map information (e.g. road length, number of lanes, street name, ...) can be easily accessed from SStudio

Issue #829, opened 2021-05-10
Disentangle waypoint concepts


Gamenot:
**Is your feature request related to a problem? Please describe.**
SMARTS's current conception of waypoints is mangled between number of ideas: Map locations, trajectories, lane paths. 

This complicates refactoring the Behavioural Modeling work to make that more independent of the SMARTS core (e.g., Issue #754).

At a minimum, we should split the waypoints into two types:  static and dynamic.  Static waypoints could be generated for the entire map, once, either at startup or when the scenario is built (preferable), evenly spaced.  Dynamic waypoints would probably be renamed, but would be generated by the motion planner (potentially taking static waypoints as inputs) and also BMs.

Things to consider:
- we probably only need static waypoints for mission planning.  when and how often will a vehicle query for these?
- can we cache _all_ the waypoints for a map (shared across vehicles / agents) or will this be cumbersome for large, complicated maps?  (also see Issue #679)
- what to do about the waypoints sensor(s)?  do we want to expose both the static and dynamic?  (Ultra might like the dynamic too cf. Issue #726.)
- for dynamic waypoints, do we want them to take things like traffic lights into account?

PR #872, opened 2021-05-21
Refactor Waypoints

sah-huawei:
Waypoints have been split into two concepts:  `LanePoints` and `Waypoints`.

(Note the inconsistency in the capitalization of "points" has been left to ease the transition for SMARTS' clients who already use `Waypoints`.  Ideally, we could change them both to CamelCase, but if necessary, I'm willing to change `LanePoints` to `Lanepoints`.)

`LanePoints` are static, based only on the map, and always in a Lane.  `Waypoints` are dynamic, based on the map and a vehicle.  Unlike `LanePoints`, `Waypoints` do not have to snap to the middle of a road-network Lane, but rather they start just in front of a vehicle's present location.  `Waypoints` are usually what is returned through a vehicle's sensors (although it's an open question whether we should also add a `LanePoints` sensor or rather re-purpose the `RoadWaypointsSensor`).  `Waypoints` are guaranteed to be evenly-spaced.

`LanePoints` now live in the `SumoRoadNetwork` (and will be generalized to the more abstract map layer when Issue #830 is done).

`Waypoints` will be generated and distributed solely by the `MissionPlanner` (for now -- this will probably change as we re-architect for behavioural models).

Mostly addresses Issue #829.

**EDIT:**  After further analysis of the `equally_spaced_path()` routine, I realize that I was wrong in what I wrote above.  `Waypoints` do not start directly in _front_ of the vehicle, they start "abreast of the vehicle" in the middle of the nearest lane and (roughly) conform to the lane structure/path from there.  (I think they would only deviate slightly around sharp curves/corners.... i.e., the more non-linear the road curvature, the more deviation we might expect.)


smarts/core/sensors.py
```diff
@@ -957,12 +963,13 @@ class RoadWaypointsSensor(Sensor):
...
         self._sim = sim
+        self._lanepoints = sim.road_network.lanepoints
         self._mission_planner = mission_planner
         self._horizon = horizon
 
     def __call__(self):
-        wp = self._sim.waypoints.closest_waypoint(self._vehicle.pose)
-        road_edges = self._sim.road_network.road_edge_data_for_lane_id(wp.lane_id)
+        lp = self._lanepoints.closest_lanepoint(self._vehicle.pose)
+        road_edges = self._sim.road_network.road_edge_data_for_lane_id(lp.lane_id)
```
Gamenot, line 972: I think it might be better to keep reference to the road network in this case and get the lanepoints through it because we end up having to ask the sim for the road network anyway.

Then we can get rid of the sim reference.

smarts/core/sensors.py
```diff
@@ -419,12 +422,12 @@ def _vehicle_is_off_road(cls, sim, vehicle):
...
-                not sim.scenario.road_network.point_is_within_road(corner_coordinate)
-                for corner_coordinate in vehicle.bounding_box
-            ]
-        )
+        # XXX: this isn't technically right as this would also return True
+        #      for vehicles that are completely off road.
+        for corner_coordinate in vehicle.bounding_box:
+            if not sim.scenario.road_network.point_is_within_road(corner_coordinate):
+                return True
+        return False
```
Gamenot, line 430: I do agree this one is easier to read.

sah-huawei, line 430: The above is an unrelated (and very minor) optimization that I saw while looking at flamegraphs for the waypoints cache.   Calls to `point_is_within_road` are expensive.

smarts/core/vehicle.py
```diff
@@ -276,15 +276,15 @@ def bounding_box(self):
...
+        heading = self.heading
         return [
             rotate_around_point(
-                point=self.position[:2]
-                + numpy.array(corner) * numpy.array([half_width, half_length]),
-                radians=self.heading,
-                origin=self.position[:2],
+                point=origin + numpy.array(corner) * radius,
+                radians=heading,
+                origin=origin,
```
sah-huawei, line 287: This is another unrelated and minor optimization that came out of the flamegraph analysis that I didn't back out again.  The `position` and `heading` properties are not trivial (they involve computing stuff), so no need to do them multiple times due to the list comprehension loop.

smarts/core/mission_planner.py
```diff
@@ -150,62 +177,78 @@ def closest_point_on_lane(self, position, lane_id: str):
...
-                pose.position, lookahead, edge_ids
-            )
+    def waypoint_paths_at(
+        self,
+        pose: Pose,
+        lookahead: int,
+        within_radius: float = 5,
+        filter_from_count: int = 3,
+        constrain_to_route: bool = True,
+    ) -> List[List[Waypoint]]:
```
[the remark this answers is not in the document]
sah-huawei: I agree.  In this PR, I was trying not to change any existing behavior / contracts.  We have been using ~1m spacing for waypoints/lanepoints thus far, so I kept that (and didn't change an lookahead parameters either).   Since there currently appear to be two main uses for `Waypoints` -- sent to agents via sensors and used in the lane-following controller -- If/when we have this discussion, it will need to be with existing users (e.g., Ultra) to ensure that we don't break things for their agents.

smarts/sstudio/generators.py
```diff
@@ -62,10 +57,9 @@ def from_file(cls, net_file: str):
         Args:
             net_file: The path to a '\\*.net.xml' file (generally 'map.net.xml')
         """
-        road_network = SumoRoadNetwork.from_file(net_file)
         # XXX: Spacing is crudely "large enough" so we less likely overlap vehicles
-        waypoints = Waypoints(road_network, spacing=2.0)
-        return cls(waypoints, road_network)
+        road_network = SumoRoadNetwork.from_file(net_file, lanepoint_spacing=2.0)
```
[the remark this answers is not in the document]
sah-huawei: Yeah, this was another case of trying not to change the existing behavior.  Note that the replaced line 68 constructed the `RandomRouteGenerator`with a LanePoint/Waypoints spacing of 2.0 as well, so nothing will change downstream due to this change.

smarts/core/sumo_road_network.py
```diff
@@ -536,17 +560,11 @@ def world_coord_from_offset(self, lane: Lane, offset) -> np.ndarray:
...
-        #
-        # return False
-
         radius = max(5, 2 * self._default_lane_width)
-        nearest_lanes = self.nearest_lanes(point[:2], radius=radius)
-        return any(dist < 0.5 * nl.getWidth() + 1e-1 for nl, dist in nearest_lanes)
+        for nl, dist in self.nearest_lanes(point[:2], radius=radius):
+            if dist < 0.5 * nl.getWidth() + 1e-1:
+                return True
+        return False
```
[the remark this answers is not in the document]
sah-huawei: Actually, this was also part of that attempted optimization that I added and then rolled back (described in my comment on `_vehicle_is_on_shoulder()`).

However, `any` is inherently less efficient when you only care about one instance, even in list comprehensions, so I kept this change in.  For example, consider the following:
```python
def foo(x):
    return x % 3 == 0
any([foo(x) for x in range(100)])
```
Here, `foo()` is called 100 times, even though it only needs to be called 3 times to decide the truth value of the expression.

Similarly, in the original code, the expression `dist < 0.5 * nl.getWidth() + 1e-1` can potentially be called more times than necessary.  This method is called at least 5 times per agent per step which, for traffic history replay at least, adds up.  (Although the real thing I was originally trying to optimize/avoid was so many calls to `getNeighboringLanes()`, I just did this along the way and then didn't roll it back.)  It did make a noticeable (albeit small percentage) difference in my profiling.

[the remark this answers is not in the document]
Gamenot: I think that would be true if it was a list being supplied to `any` but it was a generator so it should still short circuit.

```
def foo(x): 
   return x % 3 == 2 
any(foo(x) for x in range(100))
```
which is semantically the same as:
```
def foo(): 
   for x in range(100):   
      yield x % 3 == 2 
any(foo())
```

smarts/core/controllers/lane_following_controller.py
```diff
@@ -78,8 +78,9 @@ def perform_lane_following(
         state = controller_state
         # This lookahead value is coupled with a few calculations below, changing it
         # may affect stability of the controller.
+        lookahead = 10
```
sah-huawei, line 81: I _think_ this change is safe since when I audited the following code it seems the most `Waypoints` we use here is 10 anyway.

smarts/core/sumo_road_network.py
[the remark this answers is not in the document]
sah-huawei: Yes, you're right about the generator vs. the list comprehension, sorry.  So based on that, I would _expect_ there would be no difference between the two and, in that case, I would prefer the shorter `any` version.  _HOWEVER..._

I'm still confused about this because I really did see a performance bump after the change.  So I just went back and did the test again and, surprisingly enough, it's still there!

Here's a table of results of `single_agent.py` run on the `intersections/4lane` scenario with the `any` version:
```
╭────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────╮
│            Episode │      Sim T / Sys T │     Sim T / Wall T │        Total Steps │     Steps / SysSec │        Steps / Sec │       Scenario Map │    Scenario Routes │     Mission (Hash) │             Scores │
├────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┤
│               0/10 │               3.28 │               7.47 │                156 │              32.82 │              74.74 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               1/10 │               3.62 │               8.33 │                156 │              36.20 │              83.29 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               2/10 │               3.84 │               8.80 │                156 │              38.36 │              87.99 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               3/10 │               3.84 │               8.73 │                156 │              38.41 │              87.33 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               4/10 │               3.82 │               8.72 │                156 │              38.20 │              87.22 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               5/10 │               3.61 │               8.23 │                156 │              36.14 │              82.27 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               6/10 │               3.74 │               8.51 │                156 │              37.36 │              85.13 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               7/10 │               3.82 │               8.80 │                156 │              38.16 │              87.98 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               8/10 │               3.57 │               8.25 │                156 │              35.66 │              82.46 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
│               9/10 │               3.52 │               8.14 │                156 │              35.17 │              81.42 │              4lane │      basic.rou.xml │ -18965767758539274 │ 192.34 - Agent-007 │
╰────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────╯
Elapsed Time: 44.69 seconds.  Collected 4758 stack frames (2256 unique)
```
And here's the same thing with the nested loops:
```
╭────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────┬────────────────────╮
│            Episode │      Sim T / Sys T │     Sim T / Wall T │        Total Steps │     Steps / SysSec │        Steps / Sec │       Scenario Map │    Scenario Routes │     Mission (Hash) │             Scores │
├────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┼────────────────────┤
│               0/10 │               3.39 │               7.60 │                156 │              33.91 │              76.02 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               1/10 │               3.83 │               8.77 │                156 │              38.32 │              87.72 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               2/10 │               3.28 │               7.92 │                156 │              32.80 │              79.15 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               3/10 │               3.88 │               8.84 │                156 │              38.76 │              88.36 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               4/10 │               3.79 │               8.86 │                156 │              37.90 │              88.58 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               5/10 │               3.65 │               8.36 │                156 │              36.51 │              83.64 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               6/10 │               3.83 │               8.81 │                156 │              38.29 │              88.10 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               7/10 │               3.89 │               8.93 │                156 │              38.85 │              89.34 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               8/10 │               3.87 │               8.88 │                156 │              38.67 │              88.79 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
│               9/10 │               3.74 │               8.53 │                156 │              37.44 │              85.27 │              4lane │      basic.rou.xml │ -44837943675001737 │ 192.34 - Agent-007 │
╰────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────┴────────────────────╯
Elapsed Time: 44.14 seconds.  Collected 4656 stack frames (2247 unique)
```

Note that to create that table, I modified `util/episodes.py` to add two columns based on the elapsed CPU time (obtained using `time.process_time`).  The main column to pay attention to then is `Sim T / Sys T`.  In the first table, it's average is  `3.666` and in the second it's `3.715`, which is a `1.3%` speedup.  I saw similar speedups on the `loop` scenario as well.

I'm going to dig a little more to try to understand what's going on.  Will report back...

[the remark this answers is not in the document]
sah-huawei: Just to clarify, the _only_ difference in the code between the two tables above is:
```python
        nearest_lanes = self.nearest_lanes(point[:2], radius=radius)
        return any(dist < 0.5 * nl.getWidth() + 1e-1 for nl, dist in nearest_lanes)
```
vs.
```python
        for nl, dist in self.nearest_lanes(point[:2], radius=radius):
            if dist < 0.5 * nl.getWidth() + 1e-1:
                return True
        return False
```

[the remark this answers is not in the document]
sah-huawei: Upon some thought and experimentation, I think there must be some overhead to set up a generator in the `any` case.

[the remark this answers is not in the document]
Gamenot: That is interesting, I can only assume there is bunch of hidden complexity behind `any` even though it should be functionally the same. I will keep that in mind for the future.

smarts/core/controllers/lane_following_controller.py
sah-huawei: > I _think_ this change is safe since when I audited the following code it seems the most `Waypoints` we use here is 10 anyway.

Nope.  It did matter.  All the wavepoints are used in the road curviness computation.  10 is not enough.  (14 seems to be, but reverting it back to 16 just in case.)

Gamenot (approved):

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+    lp: LanePoint = None
+    is_shape_lp: bool = False
+    nexts: List[LinkedLanePoint] = []  # list of next immediate LanePoint(s)
+    # it's a list of LanePoints because a path may branch at junctions
+
+    def __hash__(self):
+        ## distinguish between different continuations here too
+        ## so the lru_cache on _lanepoint_paths_starting_at_lanepoint() below
+        ## doesn't return the wrong set of LanePoints.
+        return hash(self.lp) + sum(hash(nlp.lp) for nlp in self.nexts)
```
davidrusu, line 131: The sum of uniform random variables is not uniform (it's triangular) so you'll get some nasty hash collisions perf issues if this hash is used heavily.

It'd be safer to hash the tuple of the lp's, something like:
```suggestion
        return hash((self.lp, tuple(nlp.lp for nlp in self.nexts))
```

sah-huawei: @davidrusu Hi David!  Good to see you here!  :)
And thanks for the review.  Yep, you're right.  Will make the change now.

davidrusu: Likewise! good to see you involved with SMARTS!

liu-allan (approved):

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+    """a.k.a. MapPoint:  static, based only on a map, always in a Lane"""
+
+    pos: np.ndarray  # Point positioned on center of lane
+    heading: Heading  # Heading angle of lane at this point (radians)
+    lane_width: float  # Width of lane at this point (meters)
+    speed_limit: float  # Lane speed in m/s
+    lane_id: str  # ID of lane under lanepoint
+    lane_index: int  # Index of the lane this lanepoint is over. 0 is the outer(right) most lane.
+
+    def dist_to(self, p) -> float:
```
Adaickalavan, line 66: Type of `p` ?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+        """
+        assert isinstance(
+            h, Heading
+        ), "Heading h ({}) must be an instance of smarts.core.coordinates.Heading".format(
+            type(h)
+        )
+
+        return self.heading.relative_to(h)
+
+    def signed_lateral_error(self, p) -> float:
```
Adaickalavan, line 84: Type of `p` ?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+                *self.pos,
+                self.heading,
+                self.lane_width,
+                self.speed_limit,
+                self.lane_id,
+                self.lane_index,
+            )
+        )
+
+    def __eq__(self, other):
```
Adaickalavan, line 108: Type of `other` ?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+            for lane_id, l_lps in self._lanepoints_by_lane_id.items()
+        }
+
+        self._lanepoints_kd_tree_by_edge_id = {
+            edge_id: LanePoints._build_kd_tree(l_lps)
+            for edge_id, l_lps in self._lanepoints_by_edge_id.items()
+        }
+
+    @staticmethod
+    def _build_kd_tree(linked_lps):
```
Adaickalavan, line 167: Type hints?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+        return KDTree(np.array([l_lp.lp.pos for l_lp in linked_lps]), leaf_size=50)
+
+    @staticmethod
+    def _interpolate_shape_lanepoints(shape_lanepoints, spacing):
```
Adaickalavan, line 171: Type hints?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+        for shape_lp in shape_lanepoints:
+            _, new_lanepoints = LanePoints._interpolate_from_shape_lp(
+                shape_lp, spacing, interp_memo
+            )
+            linked_lanepoints += new_lanepoints
+
+        return linked_lanepoints
+
+    @staticmethod
+    def _interpolate_from_shape_lp(shape_lp, spacing, interp_memo):
```
Adaickalavan, line 186: Type hints?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+                    spacing,
+                    newly_created_lanepoints,
+                )
+
+                shape_queue.put((current_shape_lp, next_shape_lp))
+
+        return initial_lanepoint, newly_created_lanepoints
+
+    @staticmethod
+    def _process_interp_for_lane_lp(
```
Adaickalavan, line 230: Type hints?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+        for edge in edges:
+            for lane in edge.getLanes():
+                _, new_lps = self._shape_lanepoints_along_lane(
+                    lane, lanepoint_by_lane_memo
+                )
+                shape_lanepoints += new_lps
+
+        return shape_lanepoints
+
+    def _shape_lanepoints_along_lane(self, lane, lanepoint_by_lane_memo):
```
Adaickalavan, line 317: Type hints?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+                # Use internal lanes of junctions (if we're at a junction)
+                via_lane_id = out_connection.getViaLaneID()
+                if via_lane_id:
+                    out_lane = self._road_network.lane_by_id(via_lane_id)
+
+                lane_queue.put((out_lane, curr_lanepoint))
+
+        return initial_lanepoint, shape_lanepoints
+
+    def _edge(self, lp: LanePoint):
```
Adaickalavan, line 411: Return type hints?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+        lane = self._road_network.lane_by_id(lp.lane_id)
+        return lane.getEdge()
+
+    @staticmethod
+    def _closest_linked_lp_in_kd_tree_with_pose_batched(
```
Adaickalavan, line 416: Full type hints?

smarts/core/lanepoints.py
```diff
@@ -0,0 +1,564 @@
...
+            sorted(
+                l_lps,
+                key=lambda _llp: squared_dist(poses[idx].position[:2], _llp.lp.pos)
+                + abs(poses[idx].heading.relative_to(_llp.lp.heading)),
+            )
+            for idx, l_lps in enumerate(linked_lanepoints)
+        ]
+
+    @staticmethod
+    def _closest_linked_lp_in_kd_tree_batched(points, linked_lps, tree, k: int = 1):
```
Adaickalavan, line 471: Full type hints?

smarts/core/mission_planner.py
```diff
@@ -20,35 +20,64 @@
...
-from dataclasses import replace
-from typing import Optional
+from dataclasses import dataclass, replace
+from typing import List, Optional, Sequence, Tuple
 
 import numpy as np
 
 from .agent_interface import AgentBehavior
 from .coordinates import Heading, Pose
+from .lanepoints import LanePoints, LanePoint, LinkedLanePoint
```
Adaickalavan, line 30: Using full import paths is better than relative import paths.

Adaickalavan:
1. Consider specifying input/output type hints.
1. Full import paths are better than relative import paths.

smarts/core/lanepoints.py
sah-huawei: Other than knowing it's potentially some "higher order" vector, I honestly have no idea!  (I didn't write this code originally, just moved it from Waypoints.)  I tried tracing a bit, both up and down the callstack, but the type wasn't obvious, so I left it as it was.

smarts/core/lanepoints.py
sah-huawei: same with this one.

smarts/core/lanepoints.py
sah-huawei: I don't think it's appropriate to put type hints on the "other" parameter of a python `__eq__` function.  Unless we want to put `Any` (which I'm not really a fan of doing)?

smarts/core/lanepoints.py
sah-huawei: For this one and a lot of the others in this class, I did not add them because (a) this is pre-existing code that just got refactored here and (b) these are private methods.  I'd be fine with your making a new Issue to go audit all function prototypes in our existing codebase to add type hints to all of them, but that's not what this PR was about and it is frankly too time consuming and distracting to track down each of them on someone else's pre-existing code when I'm already trying to make a fairly large change.  

As a compromise though, for this review, I'll add some more type hints where I know for sure what they are.

smarts/core/mission_planner.py
sah-huawei: I disagree with this one for two reasons:  

First, the existing pattern here (as can be seen just above the line I added) is for relative imports.  So to make that change would imply either being inconsistent within the file (which is worse, imo), or snowballing to do all files in the project.  Again, I don't have a problem with filing a new Issue to do all of these at once, but it should be part of its own, standalone PR.

Second, absolute imports reduce portability of files within a module.

PR #878, opened 2021-05-26
examples and bug fixes for imitation learning

sah-huawei:
Original purpose here was to show an example of how one might record data for Imitation Learning such that the trained model could use the `Imitation` action space.  This led to changing that action space to expect an acceleration _vector_ instead of scalar (since that's what our `AccelerometerSensor` returns).  Since we now want this action space to use `BoxChassis` and avoid `pybullet`, this also required computing linear and angular velocities for vehicles with `BoxChassis` (as used by traffic history vehicles).

Agents in the `history_vehicles_replacement_for_imitation_learning.py` now draw their actions from `.pkl` files saved by the `observation_collection_for_imitation_learning.py` script.  Initial speeds are now part of their missions.

Other bugs fixed along the way:
- Fixed inconsistencies related to `yaw_rate`, which had a type hint of `float` in `EgoVehicleObservation`, but was being returned as a vector for `AckermannChassis`.   Since we already have it as a vector in the `velocity_vectors` property on `Chassis`, went with the scalar float for `yaw_rate` so we can get either.
- Fixed two bugs in the `AccelerometerSensor`: 
    - the sign of accelerations was backwards
    - we need to divide by the `timestep_sec`
- Fixed bug in implementation of the `Imitation` action space where, when the `orientation` was replaced, the `heading` property was not recalculated.

EDIT:  As Adai pointed out, I forgot to mention an interface change (the addition of a field) to the `Observation` class here.  See discussion below in the context of Adai's comment.


examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -14,50 +14,52 @@
...
-        # simulate a "cushion" we might find in the real data
-        avg_following_distance_s = 2
-        acceleration = (nn.speed - me.speed) / avg_following_distance_s
-        angular_velocity = heading_delta / avg_following_distance_s
-        return (acceleration, angular_velocity)
+        assert self._data
+        obs_time = round(obs.sim_time + self._time_offset, 1)
+        data = self._data.get(obs_time, None)
+        if not data:
+            return ((0.0, 0.0, 0.0), 0.0)
```
Adaickalavan, line 47: `return ((0.0, 0.0, 0.0), 0.0)`

Is the return type still `Union[Tuple[float, float], float]` ?

smarts/core/traffic_history.py
```diff
@@ -34,13 +34,17 @@ def __init__(self, db: str):
         self._db = db
         self._db_cnxn = None
 
+    @property
+    def name(self):
+        return self._db.name[:-4]  # all history files have ".shf" extension
```
Adaickalavan, line 39: It is better to use some filename truncation methods, rather than doing `[:-4]` to remove the file extension.

examples/history_vehicles_replacement_for_imitation_learning.py
sah-huawei: oops, fixed.  (This action space is actually `Union[Tuple[Tuple[float, float, float], float], float]`, but I removed the example code using the `Union`, so I simplified it in this example since that's just so ugly!)

smarts/core/scenario.py
```diff
@@ -119,9 +119,12 @@ def is_reached(self, vehicle):
         return self._road_network.drove_off_map(vehicle.position, vehicle.heading)
 
 
-def default_entry_tactic():
+def default_entry_tactic(default_entry_speed: float = None):
```
Adaickalavan, line 122: Missing return type.

smarts/core/controllers/imitation_controller.py
```diff
@@ -32,43 +37,51 @@ def perform_action(cls, dt, vehicle, action):
...
             if isinstance(chassis, BoxChassis):
-                vehicle.control(vehicle.pose, action)
+                vehicle.control(vehicle.pose, action, dt)
             elif isinstance(chassis, AckermannChassis):
                 chassis.speed = action  # hack that calls control internally
             return
 
         assert isinstance(action, (list, tuple)) and len(action) == 2
-        # acceleration in m/s^2, angluar_velocity in rad/s
+        # acceleration is vector in m/s^2, angluar_velocity is scalar in rad/s
```
Adaickalavan, line 46:
```suggestion
        # acceleration is vector in m/s^2, angular_velocity is scalar in rad/s
```

smarts/core/sensors.py
```diff
@@ -127,6 +127,7 @@ class Vias:
 
 @dataclass
 class Observation:
+    sim_time: float
```
Adaickalavan, line 130: Considered breaking SMARTS interface due to introducing a new field `sim_time` in class `Observation`?

sah-huawei: Yes, that was the intent, although I'm sorry I forgot to write about it in the PR description.  (Will edit there too, as that is significant.)

In the short term, this was useful in the `history_replacement_for_imitation_learning.py` script, however there would have been other ways to achieve the goal there without this change.  However, I still feel that it is warranted and defensible... indeed, it should have been there all along.  This way, Observations are more "portable" and can be used in more flexible ways by (possibly remote) learning agents.  For example, the can save them to a database and train with them later, perhaps filtering them based on the simulation time (for example, train on every third `Observation`), but more importantly, replay them in an external environment "in real time", that is at the clock rate they were generated at, without SMARTS.  (For example, as data inputs to a trained model that is being tested.)  For these reasons (and a few other more minor ones), I feel we want this in our Observations.

Regarding this being a "breaking" change, it's a relatively minor sort of breaking.  ;)   That is, the addition of a field to the class will not break mere users of that class who can safely ignore the additional field.  I admit that it will, however, potentially break (rigidly-typed) code that marshals these.  But the only such code of which I am aware is code that we maintain.  When I made the change I intended to go change our marshaling code too, but I forgot.  I can do so now.  However, before I push anything related to that, I'll wait for more discussion here, just in case the group feels that we should hold off on this.

smarts/core/sensors.py
```diff
@@ -1022,25 +1026,39 @@ def teardown(self):
...
 class AccelerometerSensor(Sensor):
     def __init__(self, vehicle, sim):
-        self.linear_accelerations = deque(maxlen=3)
-        self.angular_accelerations = deque(maxlen=3)
+        self._dt = sim.timestep_sec
+        self.linear_velocities = deque(maxlen=3)
+        self.angular_velocities = deque(maxlen=3)
 
-    def __call__(self, linear_velocity, angular_velocity):
+    def __call__(self, linear_velocity, angular_velocity, heading):
```
Gamenot, line 1033: Looks like heading is not used.

smarts/core/chassis.py
```diff
@@ -176,12 +188,23 @@ def bullet_id(self) -> str:
...
+        # in m/s
         return self._speed
 
     @property
     def velocity_vectors(self):
-        return (None, None)
+        # linear_velocity in m/s, angular_velocity in rad/s
+        vh = radians_to_vec(self.pose.heading)
+        if self._speed is not None:
+            linear_velocity = np.array((vh[0] * self._speed, vh[1] * self._speed, 0.0))
```
Gamenot, line 200: Likely less optimised but a bit easier to look at:
```suggestion
            linear_velocity = np.array((vh[0], vh[1], 0.0)) * self._speed
```

smarts/core/chassis.py
```diff
@@ -151,10 +159,14 @@ def __init__(
...
         )
+        self._pose = None
         self.control(pose, speed)
         self._client = bullet_client
 
-    def control(self, pose: Pose, speed: float):
+    def control(self, pose: Pose, speed: float, dt: float = 0):
+        if self._pose:
+            self._last_heading = self._pose.heading
+        self._last_dt = dt
```
Gamenot, line 169: I somewhat worry about holding onto the last timestep opens the chance of de-synchronization between the simulator state and vehicle state.

smarts/core/controllers/imitation_controller.py
```diff
@@ -32,43 +37,51 @@ def perform_action(cls, dt, vehicle, action):
...
+                    vehicle.position[1] + dt * speed_y,
+                    vehicle.position[2],
+                )
+            )
+            new_pose = Pose(
                 orientation=fast_quaternion_from_angle(target_heading),
                 position=new_position,
             )
-            vehicle.control(new_pose, target_speed)
+            vehicle.control(new_pose, np.linalg.norm((speed_x, speed_y)), dt)
```
Gamenot, line 65: I do not see a reason for normalising the speed here. Shouldn't it just be the speed?

Gamenot (commented):
Sorry for the delay, a few comments.

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -14,50 +14,52 @@
...
+
+        dtime = round(obs_time - self._timestep_sec, 1)
```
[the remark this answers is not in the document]
sah-huawei: Yeah, it probably would have been a better choice on my part to use strings for the dictionary keys in the observations script.  But here I wanted to get the "previous" on on line 49, so using floats was a quick way.  Rounding to the first place only works with the default SMARTS `timestep_sec` of 0.1, so it's not ideal, but I was in a hurry.  I'll go back now and make this more general.

smarts/core/traffic_history_provider.py
```diff
@@ -107,8 +107,10 @@ def step(self, provider_actions, dt, elapsed_sim_time) -> ProviderState:
             return ProviderState(vehicles=[])
         vehicles = []
         vehicle_ids = set()
-        history_time = self._start_time_offset + elapsed_sim_time
-        rows = self._histories.vehicles_active_between(history_time - dt, history_time)
+        dec_digits = len("{}".format(dt)) - 2
```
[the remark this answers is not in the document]
sah-huawei: Good point.  I will try to come up with something more general here and the other places it's used (in `smarts.py` for example).

smarts/core/smarts.py
```diff
@@ -604,6 +611,10 @@ def _pybullet_provider_step(self, agent_actions) -> ProviderState:
     def _nondynamic_provider_step(self, agent_actions) -> ProviderState:
         self._perform_agent_actions(agent_actions)
 
+        self._bullet_client.stepSimulation()
+
+        self._process_collisions()
```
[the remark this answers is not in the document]
sah-huawei: Yeah, when we (soon) refactor things to make `pybullet` optional we should clean things up here.  There are actually two dimensions to this flaw:  the `pybullet` one and the `BoxChassis` / `dynamic_action_space` conflation.

smarts/core/sensors.py
sah-huawei: Yep, good catch.  I had added it when I tried to do something a different way in the accelerometer, but then (incompletely) rolled that back.  removing.

smarts/core/controllers/imitation_controller.py
iman512003: @Gamenot : I think this is correct since `np.linalg.norm` is representing the norm of speed vector not its normalization

sah-huawei: This is just a way to get back to a scalar speed from the two vector components.  We're just using the magnitude of the vector.  But we haven't yet computed this new scalar value yet in this function.   (Note that `speed_x` and `speed_y` have been updated using the acceleration so that this will be a different value now than `vehicle.speed` and there is no more `target_speed` variable.)

smarts/core/traffic_history_provider.py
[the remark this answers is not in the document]
sah-huawei: I think have a good solution for this now.

examples/history_vehicles_replacement_for_imitation_learning.py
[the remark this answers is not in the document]
sah-huawei: I think I have a good solution for this now.

smarts/core/chassis.py
sah-huawei: Yeah, I didn't like this either (for a different reason).  I'll try to improve this...

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -14,50 +14,52 @@
...
+    ):
+        self._vehicle_id = vehicle_id  # for debugging
+        self._time_offset = time_offset
+        datafile = f"data_{scenario.name}_{scenario.traffic_history.name}_Agent-history-vehicle-{vehicle_id}.pkl"
+        # We read actions from a datafile previously-generated by the
+        # observation_collection_for_imitation_learning.py script.
+        # This allows us to test the action space to ensure that it
+        # can recreate the original behaviour.
+        with open(datafile, "rb") as pf:
+            self._data = pickle.load(pf)
```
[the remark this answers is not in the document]
sah-huawei: Yeah, and also real imitation learning users will neither be using this "fake" agent, nor using the code in the observations script to write the `.pkl` files in the first place (because they will presumably modify that example to generate the data as they need it for their model training.

smarts/core/chassis.py
sah-huawei: ... actually, after looking at it again, I don't see any obvious ways to easily clean this up.  Do you have any suggestions?

smarts/core/sensors.py
Gamenot: I think that the principled approach is that we do not add to the observations unless absolutely necessary until we decide to do a larger version update. If you feel this is absolutely needed we could merge this into develop and cut the release at the previous commit.

Then we would make the next release a breaking version.

sah-huawei: OK, since it's not absolutely necessary right now, I will work around it in `history_vehicles_replacement_for_imitation_learning.py` and back out the change to `Observations` for now.  Then I'll add an issue and a separate PR  for it so we can decide about it later.

smarts/core/chassis.py
Gamenot: I am also a bit stumped other than getting the value from a simulation time utility which opens up its own problems.

smarts/core/controllers/imitation_controller.py
Gamenot: Oh, sorry, I keep forgetting that `np.linalg.norm` is for scalar. :facepalm:

smarts/core/utils/math.py
```diff
@@ -269,3 +269,22 @@ def inplace_unwrap(wp_array):
...
+    decimal = strep.find(".")
+    if decimal >= len(strep) - 1:
+        return 1 - decimal
+    return len(strep) - decimal - 1
+
+
+def rounder_for_dt(dt):
+    """return a rounding function appropriate for timestepping."""
+    rp = round_param_for_dt(dt)
+    return lambda f: round(f, rp)
```
Gamenot, line 290: Smart

Gamenot (approved):
No further comments on my part.

smarts/core/utils/math.py
sah-huawei: thanks!  except now I look at it, I see I forgot the type hints!  (Adai has me super sensitive about them now!) :)

Adaickalavan (approved):

smarts/core/sensors.py
sah-huawei: Added Issue #884.

smarts/core/chassis.py
sah-huawei: Yeah, I agree.  I think I'm going to leave it as is then.

PR #880, opened 2021-05-26
Support for Python 3.8

RutvikGupta:
This involved minor changes to check that python3.8 is supported by SMARTS.


setup.py
```diff
@@ -28,14 +28,10 @@
...
         "rtree",  # Used by sumolib
-        "filelock",
-        "lz4",
-        "networkx",
-        "opencv-python",
         "pandas",
         "psutil",
         "visdom",
-        "pybullet",
+        "pybullet==3.0.6",
```
sah-huawei, line 34: Why was this needed?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - Added README_pypi.md to update to the general user installation PyPI instructions. See Issue #828 
 ### Changed
 - `history_vehicles_replacement_for_imitation_learning.py` now uses new Imitation action space.  See Issue #844.
+- Updated and removed some package versions to ensure that Python3.8 is supported by SMARTS. See issue #266.
```
sah-huawei, line 18: I assume you've tested that everything still works fine with 3.7 as well, right?

After this goes in, should we update our official stance on what versions of Python we support (for example, as we state in `README.md` and potentially elsewhere)?  If so, we should update what  we say in those places too as part of this PR.

Also, my understanding of the conversation we had about this in our last meeting was that our stance was going to change from "only version 3.7.x" to "at _least_ version 3.7".  If that's the case, do we want/need to change some of the `==` symbols in `requirements.txt` below to be `>=` as well (for Python ersions newer than 3.8)?

setup.py
```diff
@@ -28,14 +28,10 @@
         "numpy",
         "rich",
         "rtree",  # Used by sumolib
-        "filelock",
-        "lz4",
-        "networkx",
-        "opencv-python",
```
sah-huawei, line 34: Glad we're able to get rid of some things.  Did you also audit the list of things in `requirements.txt`?  The more we can get rid of there, the better!

setup.py
```diff
@@ -71,7 +67,6 @@
             "pytest",
             "pytest-benchmark",
             "pytest-cov",
-            "ipykernel",
```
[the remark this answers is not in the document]
RutvikGupta: `ipykernel`was listed twice so I removed the duplicate.

setup.py
RutvikGupta: I removed the above unused packages and then ran the setup again to see which packages were installed as dependencies. And then updated the `requirements.txt` from the output of `pip freeze`. I can explicitly remove some packages listed in there but not sure if that's going to break things or not.

setup.py
RutvikGupta: `python 3.8` updates `pybullet` which uses  the latest version of `numpy 1.20.3` which is not supported by `tensorflow`. So this is to ensure that pybullet is not updated during the setup.

CHANGELOG.md
RutvikGupta: Yeah there are bunch of places where the `at least` parts needs to be updated. I will modify those. I dont think `requirements.txt` requires >= , cause i have updated the packages  version based on what is supported by both `python3.7` and `python 3.8`.

setup.py
Gamenot, line 34: I think we can fully get rid of opencv but it is used in `benchmark` and `ultra`. We can either move it to a `[benchmark]` or `[rendering]` extra or give `benchmark` a setup file of its own. 

I also think it would be great if we could move panda3d out into an extra but that is enough work to make it a separate PR. @RutvikGupta could you make an issue for that?

RutvikGupta: Ok sure.

Gamenot (approved):
Good work testing out python 3.8.

sah-huawei (approved):
Good.

PR #885, opened 2021-05-27
ULTRA add `--durations` flag to ULTRA CI tests

JenishPatel99:
Motive: To analyze the duration of ULTRA tests

Side task:
- [x] Activate PyPI ULTRA CI test

### DIFF
--- a/.github/workflows/ci-ultra-tests.yml
+++ b/.github/workflows/ci-ultra-tests.yml
@@ -154,31 +154,31 @@ jobs:
           --durations=0 \
           ./tests/test_ultra_package.py \
 
-  # test-package-via-pypi:
-  #   runs-on: ubuntu-18.04
-  #   if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-ultra-tests.yml
```diff
@@ -146,7 +150,9 @@ jobs:
           cd ultra
           . .venv/bin/activate
           scl scenario build-all ultra/scenarios/pool
-          pytest -v ./tests/test_ultra_package.py
+          pytest -v \
+          --durations=0 \
+          ./tests/test_ultra_package.py \
 
   # test-package-via-pypi:
```
christianjans, line 157: Just curious, are we using this test anymore?

christianjans (approved):

.github/workflows/ci-ultra-tests.yml
JenishPatel99: Actually, I have uncommented it and ran the test. It works with the latest version of ultra (ultra-rl==0.2.0). We might want to keep this test to keep  a sanity check on our ultra pypi package

PR #886, opened 2021-05-28
history agent vehicle sizing

sah-huawei:
This PR allows agent vehicles' sizes to be set to match whatever was specified in the traffic history dataset for the vehicle being replaced.  This is necessary for accurate collision detection when training/testing imitation learning models.

### DIFF
--- a/examples/history_vehicles_replacement_for_imitation_learning.py
+++ b/examples/history_vehicles_replacement_for_imitation_learning.py
@@ -52,17 +52,17 @@ class ReplayCheckerAgent(Agent):
         assert math.isclose(
             cur_state.heading, exp["heading"], abs_tol=1e-09
         ), f'vid={self._vehicle_id}: {cur_state.heading} != {exp["heading"]} @ {obs_time}'
-        # Note: the speed check can't be as tight b/c we lose some accuracy (due to angular acceleration)
+        # Note: the other checks can't be as tight b/c we lose some accuracy (due to angular acceleration)
         # by converting the acceleration vector to a scalar in the observation script,
         # which compounds over time throughout the simulation.
         assert math.isclose(
             cur_state.speed, exp["speed"], abs_tol=0.1
         ), f'vid={self._vehicle_id}: {cur_state.speed} != {exp["speed"]} @ {obs_time}'
         assert math.isclose(
-            cur_state.position[0], exp["ego_pos"][0], rel_tol=0.01
+            cur_state.position[0], exp["ego_pos"][0], abs_tol=2
         ), f'vid={self._vehicle_id}: {cur_state.position[0]} != {exp["ego_pos"][0]} @ {obs_time}'
         assert math.isclose(
-            cur_state.position[1], exp["ego_pos"][1], rel_tol=0.01
+            cur_state.position[1], exp["ego_pos"][1], abs_tol=2
         ), f'vid={self._vehicle_id}: {cur_state.position[1]} != {exp["ego_pos"][1]} @ {obs_time}'
 
         # Then get and return the next set of control inputs
--- a/smarts/core/controllers/imitation_controller.py
+++ b/smarts/core/controllers/imitation_controller.py
@@ -19,6 +19,7 @@
 # THE SOFTWARE.
 import math
 import numpy as np
+from dataclasses import dataclass
 
 from smarts.core.chassis import AckermannChassis, BoxChassis
 from smarts.core.coordinates import Pose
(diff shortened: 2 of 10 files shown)


examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -52,17 +52,17 @@ def act(self, obs: Observation, obs_time: float) -> Tuple[float, float]:
...
-        # Note: the speed check can't be as tight b/c we lose some accuracy (due to angular acceleration)
+        # Note: the other checks can't be as tight b/c we lose some accuracy (due to angular acceleration)
         # by converting the acceleration vector to a scalar in the observation script,
         # which compounds over time throughout the simulation.
         assert math.isclose(
             cur_state.speed, exp["speed"], abs_tol=0.1
         ), f'vid={self._vehicle_id}: {cur_state.speed} != {exp["speed"]} @ {obs_time}'
         assert math.isclose(
-            cur_state.position[0], exp["ego_pos"][0], rel_tol=0.01
+            cur_state.position[0], exp["ego_pos"][0], abs_tol=2
```
sah-huawei, line 62: Note that this change, and the similar one just below are for an unrelated reason.  Basically, in testing, I discovered that when a vehicle is chosen for replacement that does a lot of non-straight driving (lane changes, etc.), position errors can compound more for it through the course of the simulation.  I'll discuss this with the Imitation learning folks, but for now, this prevents this demo code from choking.

(FWIW, the way to mitigate this is to make the Imitation action space use acceleration _vectors_ instead of an acceleration scalar, as we had yesterday.  But they asked for a scalar to reduce their model complexity.  There is already a comment about this in `observation_collection_for_imitation_learning.py`.)

smarts/core/scenario.py
```diff
@@ -559,10 +567,10 @@ def discover_missions_of_traffic_histories(self) -> Dict[str, Mission]:
...
+            veh_type = self._traffic_history.vehicle_type(v_id)
+            veh_length, veh_width, veh_height = self._traffic_history.vehicle_size(v_id)
             # missions start from front bumper, but pos is center of vehicle
-            # vehicle_length, _ = self._traffic_history.vehicle_size(v_id)
-            vehicle_length = 3.68  # TODO: agent vehicles need to be resizable!
-            hhx, hhy = radians_to_vec(heading) * (0.5 * vehicle_length)
+            hhx, hhy = radians_to_vec(heading) * (0.5 * veh_length)
             vehicle_missions[v_id] = Mission(
                 start=Start(
                     (pos_x + map_offset[0] + hhx, pos_y + map_offset[1] + hhy),
```
sah-huawei, line 576: Note: as I commented on earlier, the type `Start` should probably be replaced with `Pose` for the initial position in Missions, which would clean up the above a bit.  But I'm trying to avoid that sort of snowballing right now...

smarts/core/controllers/imitation_controller.py
```diff
@@ -53,19 +53,12 @@ def perform_action(cls, dt, vehicle, action):
...
-            )
-            target_speed = vehicle.speed + acceleration * dt
+            dpos = radians_to_vec(vehicle.heading) * vehicle.speed * dt
             new_pose = Pose(
-                orientation=fast_quaternion_from_angle(target_heading),
-                position=new_position,
+                position=vehicle.position + np.append(dpos, 0.0),
+                orientation=fast_quaternion_from_angle(vehicle.heading),
             )
+            target_speed = vehicle.speed + acceleration * dt
```
sah-huawei, line 61: These changes are also for a different reason.  This is just a simplification/cleanup of the code which _shouldn't_ change any of the functionality or computations here.  It's just to make it easier to read.

smarts/core/controllers/imitation_controller.py
```diff
@@ -53,19 +53,12 @@ def perform_action(cls, dt, vehicle, action):
...
-                )
-            )
-            target_speed = vehicle.speed + acceleration * dt
+            dpos = radians_to_vec(vehicle.heading) * vehicle.speed * dt
             new_pose = Pose(
+                position=vehicle.position + np.append(dpos, 0.0),
                 orientation=fast_quaternion_from_angle(target_heading),
-                position=new_position,
             )
+            target_speed = vehicle.speed + acceleration * dt
```
sah-huawei, line 61: These changes are also for a different reason. This is just a simplification/cleanup of the code which shouldn't change any of the functionality or computations here. It's just to make it easier to read.

smarts/core/controllers/imitation_controller.py
```diff
@@ -53,19 +53,14 @@ def perform_action(cls, dt, vehicle, action):
...
-            target_speed = vehicle.speed + acceleration * dt
+            heading_vec = radians_to_vec(vehicle.heading)
+            ortho_vec = radians_to_vec(vehicle.heading + math.pi / 2)
+            dpos = (heading_vec * vehicle.speed + ortho_vec * angular_velocity) * dt
             new_pose = Pose(
+                position=vehicle.position + np.append(dpos, 0.0),
                 orientation=fast_quaternion_from_angle(target_heading),
-                position=new_position,
             )
+            target_speed = vehicle.speed + acceleration * dt
```
sah-huawei, line 63: These changes are also for a different reason. This is mainly a simplification/cleanup of the code to make it easier to read, but also not take `angular_velocity` into account for position updates.

Gamenot (approved):
I have no particular issues.

PR #891, opened 2021-05-31
ULTRA Remove start/end offsets

JenishPatel99:
Minor update: Remove ego vehicle's start and end offsets from the simple level. 

Note: By default levels from tasks 1 & 2 would be not be constrained by any destination/arrival positions (start/end offsets)

### DIFF
--- a/ultra/ultra/scenarios/task1/config.yaml
+++ b/ultra/ultra/scenarios/task1/config.yaml
@@ -66,8 +66,6 @@ levels:
         # left turn from south to west
       - start: south-SN
         end:   west-EW
-        start_offset: [110, 120]
-        end_offset:   [10, 10]
       intersection_types:
         2lane_t:
           percent: 0.2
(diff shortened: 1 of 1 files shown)


christianjans (commented):
Cool change. I like how it homogenizes Task 1 and Task 2 levels. I thought we had added the offset so that the ego vehicle could have a better change of colliding with the social vehicles, is it okay if we remove the offset now because we aren't using the "___-interaction" distributions? Or is there another reason?

JenishPatel99:
> Cool change. I like how it homogenizes Task 1 and Task 2 levels. I thought we had added the offset so that the ego vehicle could have a better change of colliding with the social vehicles, is it okay if we remove the offset now because we aren't using the "___-interaction" distributions? Or is there another reason?

So previously we had the offsets specifically for the _-interaction distributions, and now that we have revert back with the _-density distribution in the simple level the offsets must return to its default values. So yeah, answering your question: We are removing the start/end offsets because we are no longer using the _-interactions.

The _-interaction distributions will used in creating sub-tasks for curriculum learning experiments, or even create a separate task which contain only those types of distributions. However, for now **task1 and **task2 should only contain {low,mid,high}-density or no-traffic distributions.

christianjans (approved):

Issue #900, opened 2021-06-02
Fix internal holes due to Crude Map Geometry


RutvikGupta:
We need to fix the internal holes that are created at sharp turns due to crude SUMO map geometry. These holes may potentially cause on-shoulder & off-road check misfires.
Currently we can get holes in the road due to rough lane geometry shapes.

These holes can be observed in `scenarios/loop`.

![image](https://user-images.githubusercontent.com/47334253/120518275-68c6af80-c39f-11eb-95b6-b153f35ae5ce.png)

The holes appear between the two lanes during a sharp turn. We do not combine the geometry of lanes so each lane is generated separately and hence the holes are not part of any lane unlike the inner edges mentioned in Issue #273. 

![image](https://user-images.githubusercontent.com/47334253/120518561-b4795900-c39f-11eb-9123-a78a22bd8dc2.png)

RutvikGupta:
See PR #300 for a workaround on how to snap the internal gaps to there surrounding nearest lanes

PR #901, opened 2021-06-02
Fixed internal holes in sumo road network created due to crude map geometry

RutvikGupta:
See issue #900. See PR #300.
These holes were observed before in `scenarios/loop`, `scenarios/minicity` and `scenarios/cloverleaf` between two adjacent lanes at multiple sharp curves.  This PR fixes these holes. 

Before:

`scenarios/loop`:

image

`scenarios/cloverleaf` 

image

`scenarios/minicity`

image

After:

`scenarios/loop`

image

`scenarios/minicity`

image

`scenarios/cloverleaf`

image


smarts/core/sumo_road_network.py
```diff
@@ -230,7 +230,10 @@ def _compute_road_polygons(self):
 
                 lane_to_poly[lane.getID()] = shape
 
+        self._snap_internal_holes(lane_to_poly)
+        self._snap_external_holes(lane_to_poly)
```
sah-huawei, line 234: You could factor the following out of the 2 functions you're calling here, like:
```python
    for lane_id in lane_to_poly:
        lane = self.lane_by_id(lane_id)
        if lane.getEdge().isSpecial():
            self._snap_internal_holes(lane)
        else:
            self._snaap_external_holes(lane)
```

(although, you may end up not wanting to do this if you change to a pattern for updating the dictionary outside of the loop, as suggested below.)

smarts/core/sumo_road_network.py
```diff
@@ -269,6 +272,58 @@ def _snap_internal_edges(self, lane_to_poly, snap_threshold=2):
...
+        for lane_id in lane_to_poly:
+            lane = self.lane_by_id(lane_id)
+
+            # Only do snapping for external edge lane holes
+            if lane.getEdge().isSpecial():
+                continue
+
+            incoming = self.lane_by_id(lane_id).getIncoming()
+            if incoming:
+                if incoming[0].getEdge().isSpecial():
```
sah-huawei, line 304: should just use `and` here, like:
```python
    if incoming and incoming[0].getEdge().isSpecial():
        continue
```

smarts/core/sumo_road_network.py
```diff
@@ -269,6 +272,58 @@ def _snap_internal_edges(self, lane_to_poly, snap_threshold=2):
...
+                    (x, y),
+                    max(10, 2 * self._default_lane_width),
+                    include_junctions=False,
+                )
+                if nl:
+                    if nl.getEdge().isSpecial():
+                        continue
+                    nl_shape = lane_to_poly.get(nl.getID())
+                    lane_shape = Polygon(snap(lane_shape, nl_shape, snap_threshold))
+                    lane_to_poly[lane_id] = lane_shape
```
sah-huawei, line 325: Same two comments as above:
- is it ok if `nl_shape` is `None` in the call to `snap()`?
- not good to update dict while iterating over it

smarts/core/sumo_road_network.py
```diff
@@ -269,6 +272,58 @@ def _snap_internal_edges(self, lane_to_poly, snap_threshold=2):
...
+            lane_shape = lane_to_poly[lane_id]
+            for x, y in lane_shape.exterior.coords:
+                nl = self.nearest_lane(
+                    (x, y),
+                    max(10, 2 * self._default_lane_width),
+                    include_junctions=False,
+                )
+                if nl:
+                    if nl.getEdge().isSpecial():
+                        continue
```
sah-huawei, line 322: I would prefer flattening the above 3 lines to:
```python
    if nl and not nl.getEdge().isSpecial():
```

Also, was a similar check like this not needed in the internal version of this above?

smarts/core/sumo_road_network.py
```diff
@@ -269,6 +272,58 @@ def _snap_internal_edges(self, lane_to_poly, snap_threshold=2):
...
+            lane_shape = lane_to_poly[lane_id]
+            for x, y in lane_shape.exterior.coords:
+                nl = self.nearest_lane(
+                    (x, y),
+                    max(10, 2 * self._default_lane_width),
+                    include_junctions=False,
+                )
+                if nl:
+                    nl_shape = lane_to_poly.get(nl.getID())
+                    lane_shape = Polygon(snap(lane_shape, nl_shape, snap_threshold))
```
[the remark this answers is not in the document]
RutvikGupta: Oh, Nice catch! I will add the condition.

smarts/core/sumo_road_network.py
```diff
@@ -269,6 +272,58 @@ def _snap_internal_edges(self, lane_to_poly, snap_threshold=2):
...
+                    lane_to_poly[lane_id] = lane_shape
```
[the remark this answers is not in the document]
RutvikGupta: I will store the value in a temp_variable and then update the dict outside the loop. Thanks for the suggestion!

smarts/core/sumo_road_network.py
RutvikGupta: It seems like this will work only when the internal edges are snapped first and then the external edges later. So I will have to iterate over the two edges separately.

smarts/core/sumo_road_network.py
RutvikGupta: Ok, sure!

smarts/core/sumo_road_network.py
RutvikGupta: Yup, will make these changes.

smarts/core/sumo_road_network.py
RutvikGupta: No, seems like the internal edges can be snapped with any nearby edges to ensure they are seamed tightly. When I add a similar checker for them above, I still get the same result. So adding this check there seems redundant to me.

Though I have to add a checker for external edges, otherwise it seems to cut out external lanes out of the map if they get snapped to internal lanes. 
This is because if the holes are between an internal and a external lane, then only the internal lane needs to be snapped to the nearby lanes to remove that hole. If we snap the external lanes to internal lanes, the external lane itself will be removed from the map. While if the two adjacent lanes are external (i.e. there are only two lanes), then both lanes need to be snapped to each other to remove the holes completely.

RutvikGupta:
I seemed to have missed some holes before that were at almost perpendicular turns as seen in the `scenario/minicity` screenshots above. I have fixed them as well.

smarts/core/sumo_road_network.py
```diff
@@ -269,6 +274,60 @@ def _snap_internal_edges(self, lane_to_poly, snap_threshold=2):
...
+                    (x, y),
+                    max(10, 2 * self._default_lane_width),
+                    include_junctions=False,
+                ):
+                    if nl:
+                        nl_shape = lane_to_poly.get(nl.getID())
+                        if nl_shape:
+                            lane_shape = Polygon(
+                                snap(lane_shape, nl_shape, snap_threshold)
+                            )
```
Gamenot, line 296: I could suggest using continue here to help alleviate the tabbing.

```suggestion
                    if not nl:
                        continue
                    nl_shape = lane_to_poly.get(nl.getID())
                    if nl_shape:
                        lane_shape = Polygon(
                            snap(lane_shape, nl_shape, snap_threshold)
                        )
```

Gamenot (approved):
One minor comment otherwise looks good!

sah-huawei (approved):

PR #903, opened 2021-06-02
ULTRA Update `docs/setup.md`

JenishPatel99:
Motive: Remove previous reference to SMARTS installation

### DIFF
--- a/ultra/docs/setup.md
+++ b/ultra/docs/setup.md
@@ -14,10 +14,7 @@ $ python3.7 -m venv .venv
 # 3 - Activate virtual environment to install all dependencies.
 $ source .venv/bin/activate
 
-# 4 - Install black for formatting (if you wish to contribute).
-$ pip install black
-
-# 5 - Install dependencies.
+# 4 - Install dependencies.
 $ pip install -e .
 ```
 
--- a/ultra/docs/setup.md
+++ b/ultra/docs/setup.md
@@ -8,13 +8,20 @@ ULTRA can be run natively on your machine, or through a Docker container. See in
 # 1 - Navigate to the ULTRA directory.
 $ cd <path_to_SMARTS>/SMARTS/ultra
 
-# 2 - Create a virtual environment.
+# 2 - Install overhead dependencies (python3.7 & sumo)
+./install_deps.sh
+
+# 3 - verify sumo is >= 1.5.0
+# if you have issues see ./docs/SUMO_TROUBLESHOOTING.md
+sumo
+
+# 4 - Create a virtual environment.
 $ python3.7 -m venv .venv
 
-# 3 - Activate virtual environment to install all dependencies.
+# 5 - Activate virtual environment to install all dependencies.
 $ source .venv/bin/activate
 
-# 4 - Install dependencies.
+# 6 - Install dependencies.
 $ pip install -e .
 ```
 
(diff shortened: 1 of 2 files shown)


christianjans (commented):
I think we can also maybe get rid of https://github.com/huawei-noah/SMARTS/blob/be15eaf0765782ba33e240ed5a48a37b01ac6d12/ultra/docs/setup.md#L17-L18 because it seems like `black` is part of the installation requirements now: https://github.com/huawei-noah/SMARTS/blob/c5a16088152e840dbaa9afe302e8bb9338bc6a93/ultra/setup.py#L51

Also, were we going to add anything about the Sumo dependency?

JenishPatel99:
> I think we can also maybe get rid of
> 
> https://github.com/huawei-noah/SMARTS/blob/be15eaf0765782ba33e240ed5a48a37b01ac6d12/ultra/docs/setup.md#L17-L18
> 
> because it seems like `black` is part of the installation requirements now:
> https://github.com/huawei-noah/SMARTS/blob/c5a16088152e840dbaa9afe302e8bb9338bc6a93/ultra/setup.py#L51
> 
> Also, were we going to add anything about the Sumo dependency?

Oh yeah thank you, I have removed the black installation step now and also have added the `install_deps.sh" to the base ultra directory.

ultra/docs/setup.md
```diff
@@ -4,27 +4,26 @@ ULTRA can be run natively on your machine, or through a Docker container. See in
...
 # 1 - Navigate to the ULTRA directory.
 $ cd <path_to_SMARTS>/SMARTS/ultra
 
-# 2 - Create a virtual environment.
+# 2 - Install overhead dependencies (python3.7 & sumo)
+./install_deps.sh
+
+# 3 - verify sumo is >= 1.5.0
+# if you have issues see ./docs/SUMO_TROUBLESHOOTING.md
+sumo
```
christianjans, line 16:
```suggestion
$ ./install_deps.sh

# 3 - verify sumo is >= 1.5.0
# if you have issues see ./docs/SUMO_TROUBLESHOOTING.md
$ sumo
```
The other commands in this setup have a '$' preceding them. It definitely doesn't have to be there, but maybe we should just keep a consistent notation (either with or without the '$' for all commands).

JenishPatel99: No that is right, we should keep them consistent, nice catch!!

christianjans (approved):
Looks great!!

Now that we have an install script, maybe we should have a CI test that tries to install ULTRA's dependencies and run ULTRA on Linux and MacOS. Not in this PR of course, but maybe another thing to add later? What are your thoughts?

JenishPatel99:
> Looks great!!
> 
> Now that we have an install script, maybe we should have a CI test that tries to install ULTRA's dependencies and run ULTRA on Linux and MacOS. Not in this PR of course, but maybe another thing to add later? What are your thoughts?

I was thinking about including the installation of those overhead dependencies from `install_deps.sh` in our `ci-ultra-tests.yml`. However, it seems like we are using a docker container (`huaweinoah/smarts:v0.4.13-minimal`) that probably comes with the python3.7 and sumo deps pre-installed. Are you suggesting to use a fresh Linux (Ubuntu)/MacOS container where we can install **all** (overhead & pip packages) dependencies and then run a basic ULTRA test?

christianjans:
Yeah, that's what I had in mind. Like to use an Ubuntu 18.04 container for the Linux test and something like MacOS 10.15 for the Mac test. Looking at it again though, it doesn't look like SMARTS has a test like this, so maybe it's not necessary?

PR #904, opened 2021-06-02
ULTRA Minimalize `train.py` script

JenishPatel99:
Refactor `train.py` to minimalize the train() method. The train() method contains elements that are responsible for building agents from given policy classes and also performing the training loops. To adhere to the single responsibility design principle we must isolate the responsibilities of the train() method. Here are some changes which might solve this issue

- [x] Create `build_agent` method to transfer the agent creation functionality from `train()` 
- [x] Create `save_agent_metadata` method to save the agent information into an `agent_metadata.pkl`
- [x] Create `create_parser` method to transfer the argument parsing from main method


ultra/ultra/train.py
```diff
@@ -45,24 +45,77 @@
...
-    headless,
-    seed,
-    log_dir,
-    policy_ids=None,
-):
-    torch.set_num_threads(1)
-    total_step = 0
-    finished = False
-    evaluation_task_ids = dict()
+def create_parser():
```
christianjans, line 48: Perhaps just for clarity?
```suggestion
def create_argument_parser():
```
I noticed in SMARTS they have something similar here: https://github.com/huawei-noah/SMARTS/blob/master/examples/__init__.py

If we go with this change, then of course the calls to this function should be changed as well.

ultra/ultra/train.py
```diff
@@ -45,24 +45,77 @@
...
+    parser.add_argument(
+        "--policy-ids",
+        help="Name of each specified policy",
+        default=None,
+        type=str,
+    )
+    return parser
+
 
+def build_agent(policy_classes, policy_ids, max_episode_steps):
```
christianjans, line 118:
```suggestion
def build_agents(policy_classes, policy_ids, max_episode_steps):
```
(And then change calls to this function if we go with this change.)

ultra/ultra/train.py
```diff
@@ -94,6 +147,53 @@ def train(
...
+        with open(f"{experiment_dir}/agent_metadata.pkl", "wb") as metadata_file:
+            dill.dump(
+                {
+                    "agent_ids": agent_ids,
+                    "agent_classes": agent_classes,
+                    "agent_specs": agent_specs,
+                },
+                metadata_file,
+                pickle.HIGHEST_PROTOCOL,
+            )
```
christianjans, line 172: What are your thoughts on if we check if `agent_metadata.pkl` exists outside of the function? So `_save_agent_metadata` would look something like:
```suggestion
def _save_agent_metadata(experiment_dir, filename, agent_ids, agent_classes, agent_specs):
    # Save relevant agent metadata.
    if not os.path.exists(experiment_dir):
        os.makedirs(experiment_dir)
    with open(os.path.join(experiment_dir, filename), "wb") as metadata_file:
        dill.dump(
            {
                "agent_ids": agent_ids,
                "agent_classes": agent_classes,
                "agent_specs": agent_specs,
            },
            metadata_file,
            pickle.HIGHEST_PROTOCOL,
        )
```
Then we would have to check to see if the file exists before we call this function.

If we do it like this, the function may be more descriptive of its name, but let me know what you think.

ultra/ultra/train.py
```diff
@@ -94,6 +147,52 @@ def train(
...
 
+    # Define an 'etag' for this experiment's data directory based off policy_classes.
+    # E.g. From a ["ultra.baselines.dqn:dqn-v0", "ultra.baselines.ppo:ppo-v0"]
+    # policy_classes list, transform it to an etag of "dqn-v0:ppo-v0".
+    etag = ":".join([policy_class.split(":")[-1] for policy_class in policy_classes])
+
+    return agent_ids, agent_classes, agent_specs, agents, etag
+
+
+def _save_agent_metadata(experiment_dir, agent_ids, agent_classes, agent_specs):
```
christianjans, line 158: I'm thinking maybe we should pass in the desired filename to this function too (see original comment on this function), so that we keep the filename in the calling function, rather than have the filename be part of the calling and called function. It may just be cleaner and easier if the string literal ("agent_metadata.pkl") is kept in one place. Let me know your thoughts.

JenishPatel99: This will also generalize the filename itself, thus no hard coding would be require. That makes sense. Let me know the changes in 450a91b

christianjans: Looks good, but I don't think we use `_AGENT_METADATA_FILENAME` anywhere. Let's just get rid of it.

christianjans (approved):

PR #907, opened 2021-06-03
Ensure agents have the required interface for the ULTRA environment

christianjans:
A possible solution to address removing the NeighborhoodVehicles from the default reward adapter's required interface.

### DIFF
--- a/ultra/ultra/baselines/agent_spec.py
+++ b/ultra/ultra/baselines/agent_spec.py
@@ -111,8 +111,6 @@ class BaselineAgentSpec(AgentSpec):
                     "neighborhood_vehicles"
                 ] = NeighborhoodVehicles(radius=200.0)
 
-            print("requirements:", adapter_interface_requirements)
-
             spec = AgentSpec(
                 interface=AgentInterface(
                     **adapter_interface_requirements,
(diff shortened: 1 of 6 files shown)


ultra/ultra/baselines/agent_spec.py
```diff
@@ -104,6 +104,15 @@ def __new__(
             )
             reward_adapter = adapters.adapter_from_type(adapter_type=reward_type)
 
+            if "waypoints" not in adapter_interface_requirements:
+                adapter_interface_requirements["waypoints"] = Waypoints(lookahead=20)
```
JenishPatel99, line 108: Wouldn't the waypoints class always be initialized through the `default_reward_adapter`, thus the first clause of the if statement would never be true. I mean when taking the UNION of the agent interface classes through the `required_interface_from_type` the waypoint class will also be part of it due to the `default_reward_adapter`

christianjans: Right, provided that the agent uses the default reward adapter, then this statement will always be false. But if a baseline agent was not using the default reward adapter and had it so that its required interface did not include Waypoints in the Observation, then this would take care of that problem. Do you think we should get rid of this?

JenishPatel99: Discussions regarding the above review has been taken offline. Safe to close comment

christianjans:
We decided to go with #902 instead of this change.

PR #911, opened 2021-06-07
Replaced supervisord with scl

RutvikGupta:
Added an experiment utility file for the `run` context to replace context given by `supervisord.conf`.


RutvikGupta:
image

I can't seem to remove supervisor package from requirements.txt without failing the test-pip-packages CI test. So I added it back for now.

cli/run.py
```diff
@@ -0,0 +1,63 @@
...
+)
+@click.argument(
+    "script_path", type=click.Path(exists=True), metavar="<script>", required=True
+)
+@click.argument("script_args", nargs=-1, type=click.UNPROCESSED)
+def run_experiment(envision, script_path, script_args):
+    with kill_process_group_afterwards():
+        if envision:
+            subprocess.Popen(
+                ["scl", "envision", "start", "-s", "./scenarios", "-p", "8081"],
```
liu-allan, line 55: Does this open the web browser as well?

RutvikGupta: Yes, Its the same command which `scl envision start` calls. You just have to visit http://localhost:8081/ in your browser to view your experiment.

liu-allan: Maybe include this so it opens automatically?
image

RutvikGupta: ok, sure.

README.md
```diff
@@ -86,30 +86,30 @@ make sanity-test
...
-#
-# [program:smarts]
-# command=python examples/single_agent.py scenarios/loop
-# ...
+scl run --envision <examples/script_path> <scenarios/path> 
 ```
 
+Pass in the agent example path and scenarios folder path above to run an experiment like the one mentioned above.
+
+Toy
```
sah-huawei, line 112: Is "Toy" supposed to be there?

cli/run.py
```diff
@@ -0,0 +1,69 @@
...
```
sah-huawei, line 60: I'm not sure how hard this is to do with "click", but if it's not hard, it would be nice to allow specifying (over-riding) the envision port too.  It could be like:
`scl --envision=8082 run ...`
(And if no `=8082` is specified, it just uses the default port.)

cli/run.py
```diff
@@ -0,0 +1,69 @@
...
+    is_flag=True,
+    default=False,
+    help="Start up Envision server when running an experiment",
+)
+@click.argument(
+    "script_path", type=click.Path(exists=True), metavar="<script>", required=True
+)
+@click.argument("script_args", nargs=-1, type=click.UNPROCESSED)
+def run_experiment(envision, script_path, script_args):
+    with kill_process_group_afterwards():
```
sah-huawei, line 57: nice!

cli/run.py
```diff
@@ -0,0 +1,69 @@
...
+                ["scl", "envision", "start", "-s", "./scenarios", "-p", "8081"],
+            )
+            # Just in case: give Envision a bit of time to warm up
+            time.sleep(2)
+            webbrowser.open_new_tab("http://localhost:8081")
+
+        script = subprocess.Popen(
+            [sys.executable, script_path, *script_args],
+        )
+        script.communicate()
```
sah-huawei, line 69: We're going to need to see `stdout` and `stderr` (and probably want to see it "live" / unbuffered).

cli/run.py
RutvikGupta: Sure, I will add that.

cli/run.py
RutvikGupta: You do get the stdout and stderr live, when you run the command. They are getting directed to STDOUT by default not to PIPE.

README.md
RutvikGupta: Seems like a typo. I will remove it. Thanks.

cli/run.py
RutvikGupta: I don't think its possible to give one option two different types (Boolean and INT).  In fact after some research I think its not possible with Click. So I just added another option `--port` so that users can mention the port they want to run the server on.

sah-huawei: Ok, yeah that's what I was afraid of.  From a skim of the `click` documentation, it _appears_ like something like this _should_ work for what you were trying to do originally:
```python
@click.option("--envision", type=int, is_flag=False, flag_value=8081, default=None)
```
... but I just tried it and it annoyingly doesn't.

So I guess adding the `--port` option is ok, although I think it's currently not clear that this is _only_ for Envision and has no meaning if used without `--envision`.   So I might call it something like `--envision-port` instead of just `--port`.  And also warn (or fail) if it's specified without specifying `--envision`.

cli/run.py
sah-huawei: oops, sorry, yeah I see you're right!  cool.

cli/run.py
RutvikGupta: Yeah, I had already tried that, but it didn't seem to work.
I will add the port changes. Thanks!

sah-huawei (approved):
good

sah-huawei (approved):

README.md
```diff
@@ -263,11 +262,11 @@ $ python examples/single_agent.py scenarios/loop
...
 ```bash
 # For this to work, your account needs to be added to the huaweinoah org
-$ cd /path/to/SMARTS
-export VERSION=v0.4.16
-docker build --no-cache -f ./etc/docker/Dockerfile -t smarts:$VERSION .
-docker tag smarts:$VERSION huaweinoah/smarts:$VERSION
 docker login
+
+export VERSION=v0.4.3-pre
+docker build --no-cache -t smarts:$VERSION .
```
Adaickalavan, line 268: Why were these lines changed?

RutvikGupta: These were changed when I was merging it with develop. I will revert them back.

README.md
```diff
@@ -262,13 +262,13 @@ $ python examples/single_agent.py scenarios/loop
...
-export VERSION=v0.4.3-pre
-docker build --no-cache -t smarts:$VERSION .
+$ cd /path/to/SMARTS
+export VERSION=v0.4.16
+docker build --no-cache -f ./etc/docker/Dockerfile -t smarts:$VERSION .
 docker tag smarts:$VERSION huaweinoah/smarts:$VERSION
+docker login
 docker push huaweinoah/smarts:$VERSION
-```
+ ```
```
Adaickalavan, line 271: Remove the extra space before ` ``` `.

Adaickalavan (approved):

PR #915, opened 2021-06-07
ULTRA Evaluate specific models

JenishPatel99:
Motive: Evaluate an explicit model rather than evaluating all possible models, as previously implemented inside of `evaluate.py`.

- [x] Remove the `models` flag which was previously used to check for compatibility of models and retrieving the agent ids
- [x] Use the `experiment-dir` flag instead to fulfil the purpose of the `models` flag
https://github.com/huawei-noah/SMARTS/blob/a5d204c2b3276a4ea7c3116c3d5e5fc3779f519b/ultra/ultra/evaluate.py#L257-L258
- [x] Add `agents` flag to state which agent(s) is/are to be evaluated. Currently only one agent is able to be evaluated at a time, more investigation is needed
- [x] Add `models-to-evaluate` flag to clarify which agent's model (e.g. `000/12345` -> `<agent-id>/<model_name>`) that you want to evaluate, if no value is set then all models of that agent will be evaluated
- [x] Construct a logic structure to extract the `agent_checkpoint_directories` if an explicit model is given (i.e. `models-to-evaluate != None`). The `agent_checkpoint_directories` is a dictionary where the keys represent the agent ids and the values are a list of the agent's (key's) checkpoint (model) directories.
- [x] Need to update the `evaluate_saved_models` call inside of `ultra/tune.py`

### DIFF
--- a/ultra/docs/getting_started.md
+++ b/ultra/docs/getting_started.md
@@ -78,20 +78,26 @@ After training your agent, your models should be saved under `logs/<timestamped_
 - Re-run the evaluation with `ultra/evaluation.py`. Available arguments include:
   - `--task`: The task number to run (default is 1).
   - `--level`: The level of the task (default is easy).
-  - `--models`: The path to the saved model (default is models/).
+  - `--agents`: The ids of agents that you wish to evaluate (default is None; all agents are subject to evaluation).
   - `--episodes`: The number of evaluation episodes (default is 200).
   - `--max-episode-steps`: The option to limit the number of steps per epsiodes (default is 200).
   - `--timestep`: The environment timestep in seconds (default is 0.1).
   - `--headless`: Provide this flag to run evaluation without Envision.
   - `--experiment-dir`: The path to the spec file that includes adapters and policy parameters.
   - `--log-dir`: The log directory location (default is logs/).
+  - `--models-to-evaluate`: Provide this flag if you wish to evaluate agent(s) at desired checkpoint(s)
 
   For example, let's re-run our DQN's evaluation with the following command:
   ```sh
-  $ python ultra/evaluate.py --task 1 --level easy --models logs/<timestamped_experiment_name>/models/ --experiment-dir logs/<timestamed_experiment_name>/ --episodes 5
+  $ python ultra/evaluate.py --task 1 --level easy --agents <agent_ids> --experiment-dir logs/<timestamed_experiment_name>/ --episodes 5
   ```
   > This will produce another experiment directory under `logs/` containing the results of the evaluation.
 
+  The `--models-to-evaluate` flag will provide you the functionality to evaluate agent(s) at specific checkpoint(s). The `--models-to-evaluate` flag takes a list of arguments in this format: <agent_id>/<model>. For example, if we want to evaluate an agent with id="000" at checkpoints 0, 33, 87, 120 (these values represent the steps at which the checkpoint was saved) we can run the following command:
+  ```sh
+  $ python ultra/evaluate.py --task 1 --level easy --agents 000 --experiment-dir logs/<timestamed_experiment_name>/ --episodes 5 --models-to-evaluate 000/0 000/33 000/87 000/120
+  ```
+  
 ## View Tensorboard Results
 
 To view the Tensorboard results of this experiment, run the command below:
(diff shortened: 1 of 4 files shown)


ultra/docs/getting_started.md
```diff
@@ -78,20 +78,26 @@ After training your agent, your models should be saved under `logs/<timestamped_
...
+  - `--models-to-evaluate`: Provide this flag if you wish to evaluate agent(s) at desired checkpoint(s)
 
   For example, let's re-run our DQN's evaluation with the following command:
   ```sh
-  $ python ultra/evaluate.py --task 1 --level easy --models logs/<timestamped_experiment_name>/models/000/ --experiment-dir logs/<timestamed_experiment_name>/ --episodes 5
+  $ python ultra/evaluate.py --task 1 --level easy --agents <agent_ids> --experiment-dir logs/<timestamed_experiment_name>/ --episodes 5
   ```
   > This will produce another experiment directory under `logs/` containing the results of the evaluation.
 
+  The `--models-to-evaluate` flag will provide you the functionality to evaluate agent(s) at specific checkpoint(s). The `--models-to-evaluate` flag takes a list of arguments in this format: <agent_id>/<model>. For example, if we want to evaluate an agent with id="000" at checkpoints 0, 33, 87, 120 (these values represent the steps at which the checkpoint was saved) we can run the following command:
```
christianjans, line 96: Should the word "steps" be replaced with "episodes" now?

ultra/ultra/evaluate.py
```diff
@@ -292,13 +302,35 @@ def evaluate_saved_models(
...
-    assert all(
-        len(checkpoint_directory) == number_of_checkpoints
-        for checkpoint_directory in directories_iterator
-    ), "Not all agents have the same number of checkpoints saved"
+    # directories_iterator = iter(agent_checkpoint_directories.values())
+    # number_of_checkpoints = len(next(directories_iterator))
+    # assert all(
+    #     len(checkpoint_directory) == number_of_checkpoints
+    #     for checkpoint_directory in directories_iterator
+    # ), "Not all agents have the same number of checkpoints saved"
```
christianjans, line 333: Can all this be removed now?

ultra/ultra/evaluate.py
```diff
@@ -308,41 +340,40 @@ def evaluate_saved_models(
...
-            )[0]
-            episode.record_tensorboard(recording_step=episode.index)
-            episode.eval_count += 1
-    finally:
-        time.sleep(1)
-        ray.shutdown()
+    for agent_id, checkpoint_directories in agent_checkpoint_directories.items():
+        num_of_checkpoints = len(checkpoint_directories)
+        ray.init()
+        try:
```
christianjans, line 346: Just curious, why was the ray initialization and `try` statement moved inside the loop?

ultra/ultra/evaluate.py
```diff
@@ -292,57 +302,65 @@ def evaluate_saved_models(
...
+            agent_id = model.split("/")[0]
+            if agent_id in agent_checkpoint_directories.keys():
+                for model_directory in agent_checkpoint_directories[agent_id]:
+                    if model.split("/")[-1] == model_directory.split("/")[-1]:
+                        if agent_id in custom_checkpoint_directories:
+                            custom_checkpoint_directories[agent_id].append(
+                                model_directory
+                            )
+                        else:
+                            custom_checkpoint_directories[agent_id] = [model_directory]
```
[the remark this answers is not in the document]
JenishPatel99: I do understand, however my motive was to first establish the experiment directory and then list the agents that the user wants to evaluate. Then after, use the `models-to-evaluate` flag to hand pick the agent and its checkpoint directory (`<agent_id>/<episode_number>`). Previously, we had the issue where the `models` contained the experiment directory as well the `experiment-dir`. 

I actually thought about the exception handling of incorrect models. Currently, models that do not exist are not included inside the custom_checkpoint_directories, hence no error is produced. Do you think we can keep the `<agent_id>/<episode_number>` method and add an exception handler to deal with incorrect models?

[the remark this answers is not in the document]
christianjans: Sure yeah, if we throw an exception if the `<agent_id>` is not in the specified agent IDs or if `<episode_number>` is not an episode of one of the agent's saved models, then it should be good.

ultra/ultra/evaluate.py
JenishPatel99: Discussed offline

ultra/ultra/evaluate.py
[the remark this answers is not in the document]
JenishPatel99: That is a good idea to throw exception for invalid agent_ids and episode number (observation number). The changes are reflected in this commit 7bfc83e

christianjans (approved):

PR #916, opened 2021-06-08
Multi-agent FrameStack wrapper

Adaickalavan:
+ Added a `FrameStack` wrapper which returns stacked observations for each agent.
+ Added a `FrameStack` test case.


smarts/env/wrappers/frame_stack.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class FrameStack(gym.Wrapper):
+    """ By default, this wrapper will stack 3 consecutive frames as an agent observation"""
+
+    def __init__(self, env: gym.Env, num_stack: int = 3):
+        assert num_stack > 1, f"Expected num_stack > 1, but got {num_stack}."
+        super(FrameStack, self).__init__(env)
+        self.num_stack = num_stack
+        self.frames = {
```
sah-huawei, line 37: Does `self.frames` need to be "public"?  (If not, I would rename it to `self._frames`.)

smarts/env/wrappers/frame_stack.py
```diff
@@ -0,0 +1,77 @@
...
+        self, frame: sensors.Observation
+    ) -> Dict[str, Deque[sensors.Observation]]:
+        """Update and return frames stack with given latest single frame."""
+
+        new_frames = dict.fromkeys(frame)
+
+        for agent_id, observation in frame.items():
+            self.frames[agent_id].append(observation)
+            frames_list = list(self.frames[agent_id])
+            new_frames[agent_id] = copy.deepcopy(frames_list)
```
sah-huawei, line 51: I understand why you're doing it, but this deep copy is sort of expensive (especially if `num_stack` is large).  I would recommend adding a comment to the class explaining this (that is, explaining that the stack will incur the cost of deep copying each frame `num_stack` times).

While you're at it, it should also be explained that the stack is a "moving window" on the frames... that is, the same frame will be in more than one result, just shifted over by one spot each time.

smarts/env/wrappers/frame_stack.py
```diff
@@ -0,0 +1,77 @@
...
+
+        return self._get_observations(env_observations), rewards, dones, infos
+
+    def reset(self) -> Dict[str, Deque[sensors.Observation]]:
+        env_observations = super(FrameStack, self).reset()
+        for agent_id, observation in env_observations.items():
+            [
+                self.frames[agent_id].append(observation)
+                for _ in range(self.num_stack - 1)
+            ]
```
sah-huawei, line 76: Why use a list comprehension here instead of a simple loop?  (You're not using the resulting list.)
Like:
```python
for agent_id, observation in env_observation.items():
    for _ in range(self.num_stack - 1):
        self.frames[agent_id].append(observation)
```

Also if there are more than `num_stack` observations for an agent, wouldn't you want the later ones instead of the earlier ones?

sah-huawei (approved):
I think this is fine to merge in now as is, but you might also want to check out the newly-filed issue #921.  That _probably_ warrants a separate pull request, but if you thought it wasn't too hard, you could also include it in this one.  Either way is up to you.

Gamenot (approved):
Good work

smarts/env/tests/test_frame_stack.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.fixture
+def agent_specs():
+    def observation_adapter(env_observation):
+        return env_observation.top_down_rgb.data
+
+    return {
+        "AGENT_"
+        + agent_id: AgentSpec(
+            interface=AgentInterface(
+                rgb=True,
```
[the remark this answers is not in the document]
Adaickalavan: Agreed, I think it is easier to understand. I have changed it to be explicit `rgb=RGB()`.

smarts/env/tests/test_frame_stack.py
```diff
@@ -0,0 +1,105 @@
...
+            env = FrameStack(env, num_stack)
+        return
+
+    # Wrap env with FrameStack to stack multiple observations
+    env = FrameStack(env, num_stack)
+    agents = {
+        agent_id: agent_spec.build_agent()
+        for agent_id, agent_spec in agent_specs.items()
+    }
+    rgb = RGB()
```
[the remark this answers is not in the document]
Adaickalavan: Changed to obtain `rgb` value directly from `AgentInterface.rgb`.

Issue #603, opened 2021-02-19
Add `scl zoo install`


Gamenot:
New smarts command line to install a zoo agent.

- [x] `scl zoo install <agent>`
  - [x] Attempts to install the specified agent from the given path/uri
- [x] `scl zoo install <agent> ...`
  - [x] Attempts to install all listed agents

PR #926, opened 2021-06-09
Added `scl zoo install` command to install agents at specified paths.

RutvikGupta:
Closes Issue #603.


cli/zoo.py
```diff
@@ -76,5 +76,45 @@ def manager(port):
...
+        # nargs=-1 in combination with a default value is not supported
+        # if agent_paths is not given, set the known two zoo agent paths as default
+        agent_paths = ["zoo/policies/open-agent", "zoo/policies/rl-agent"]
+
+    for agent_path in agent_paths:
+        pip_install_cmd = [
+            "pip",
+            "install",
+            ".",
+        ]
```
sah-huawei, line 101: This should be outside the loop.

cli/zoo.py
```diff
@@ -76,5 +76,45 @@ def manager(port):
...
+        proc = subprocess.Popen(
+            pip_install_cmd,
+            stderr=subprocess.PIPE,
+            cwd=goal_dir,
+        )
+        proc.wait()
+        proc.communicate()
+
+        if proc.returncode != 0:
+            click.echo(f"{agent_path} already installed")
```
sah-huawei, line 113: Is "already installed" the only reason that the return code might be non-zero?   (For example, what if they don't have internet at the time they try this?)

cli/zoo.py
RutvikGupta: Yeah make sense.

sah-huawei (approved):

cli/zoo.py
```diff
@@ -76,5 +76,55 @@ def manager(port):
...
+    for agent_path in agent_paths:
+        goal_dir = os.path.join(os.getcwd(), agent_path)
+        clean(goal_dir)
+        proc = subprocess.Popen(
+            pip_install_cmd,
+            stderr=subprocess.PIPE,
+            cwd=goal_dir,
+        )
+        proc.wait()
+        stdout, std_err = proc.communicate()
```
Adaickalavan, line 117: Is `stdout` used?

cli/zoo.py
```diff
@@ -76,5 +76,55 @@ def manager(port):
...
+)
+@click.argument(
+    "agent_paths",
+    type=click.Path(exists=True),
+    metavar="<script>",
+    nargs=-1,
+    required=True,
+)
+def install_agents(agent_paths):
+    if not agent_paths:
```
Adaickalavan, line 91: Why do `if not agent_paths:` when `agent_paths` has `required=True`?

Adaickalavan (approved):

cli/zoo.py
```diff
@@ -76,5 +76,55 @@ def manager(port):
...
+    ]
+
+    def clean(exec_dir):
+        subprocess.check_call(
+            [sys.executable, "setup.py", "clean", "--all"],
+            cwd=exec_dir,
+        )
+
+    for agent_path in agent_paths:
+        goal_dir = os.path.join(os.getcwd(), agent_path)
```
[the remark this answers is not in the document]
RutvikGupta: ok, sure!

cli/zoo.py
RutvikGupta: No, but you need a variable there since `proc.communicate()` returns a tuple.

cli/zoo.py
RutvikGupta: Apparently even is we set `required=True`, the command runs without it so I just added a safety net to be safe.

cli/zoo.py
Adaickalavan: For unused variables, the following will also work: 
```
import subprocess

proc = subprocess.Popen(['echo', 'Hello'], stdout=subprocess.PIPE)
stdout = proc.communicate()[0]
print(stdout)

proc = subprocess.Popen(['echo', 'Hello'], stdout=subprocess.PIPE)
stdout, _ = proc.communicate()
print(stdout)
```

PR #927, opened 2021-06-09
Add flag to disable mission shuffling when generating scenarios in ULTRA

christianjans:
By default, mission shuffling is still done when creating scenarios. However, the `--no-mission-shuffle` flag is now added so that scenarios without shuffled missions can be generated.

### DIFF
--- a/ultra/tests/test_scenarios.py
+++ b/ultra/tests/test_scenarios.py
@@ -125,6 +125,106 @@ class ScenariosTest(unittest.TestCase):
             print(err)
             self.assertTrue(False)
 
+    def test_interface_generate_with_shuffled_missions(self):
+        TASK = "00-multiagent"
+        LEVEL = "shuffle_test"
+        ROOT_DIR = "tests/scenarios"
+        SAVE_DIR = os.path.join(
+            ScenariosTest.OUTPUT_DIRECTORY, "maps/shuffle_test-with-shuffle/"
+        )
+
+        if os.path.exists(SAVE_DIR):
+            shutil.rmtree(SAVE_DIR)
+
+        generate_command = (
+            "python ultra/scenarios/interface.py generate "
+            f"--task {TASK} "
+            f"--level {LEVEL} "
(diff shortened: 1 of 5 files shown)


ultra/tests/test_scenarios.py
```diff
@@ -125,6 +125,106 @@ def test_interface_generate_multiagent(self):
...
+            os.system(generate_command)
+        except Exception as error:
+            print(error)
+            self.assertTrue(False)
+
+        first_missions = None
+        num_scenario_missions_that_differ = 0
+
+        for dirpath, _, files in os.walk(SAVE_DIR):
+            if "missions.pkl" in files and "train" in dirpath:
```
JenishPatel99, line 158: I may think adding a comment which states that the test scenarios only will contain 1 mission per scenario thus shuffling will take no affect on the agent's mission, where as train scenarios can have multiple missions thus shuffling will alter agents' missions. I do think the comment is a "must have" but can provide a gentle reminder

ultra/ultra/scenarios/generate_scenarios.py
```diff
@@ -832,6 +838,11 @@ def build_scenarios(
     parser.add_argument(
         "--root-dir", help="directory of maps", type=str, default="ultra/scenarios"
     )
+    parser.add_argument(
+        "--no-mission-shuffle",
+        help="Do not shuffle ego missions.",
+        action="store_false",
+    )
```
JenishPatel99, line 845: Do we need a parser inside of the main() in `generate_scenarios.py`? I believe the CLI arguments are taken into affect by the `interface.py`. I have not utilize the `generate_scenarios.py` for generating scenarios from the command line

JenishPatel99: I think we can create a new branch and run ours tests after removing the parser (I guess we might not even a main method) and see if any errors occurred. I am quite sure that there wouldn't be any issue

JenishPatel99 (approved):

Issue #928, opened 2021-06-09
Save checkpoint directories in terms of episode index


JenishPatel99:
Previously, we were saving the checkpoint directories as the agent's iteration (total number of steps taken). However, in recent times we have moved over to episodic training and evaluation, which means that it should be appropriate to start saving the checkpoint directories as the agent's episode index.

Let me know your thoughts on this @christianjans @AlexLewandowski

christianjans:
I think this would be a fine change. If we go this route however, would it make sense to move all other step-based values to be episode-based? Such as the data that is collected during training/evaluation.

JenishPatel99:
> I think this would be a fine change. If we go this route however, would it make sense to move all other step-based values to be episode-based? Such as the data that is collected during training/evaluation.

Absolutely, I had been collecting data episodically for both training and evaluation for the curriculum learning branches. I forgot that those changes were not implemented inside of ultra-develop. Those changes can be found here: #932

PR #929, opened 2021-06-09
ULTRA Save checkpoint directory via episode index

JenishPatel99:
Addressing issue: #928 

- [x] Evaluation results will be recorded via tensorboard in terms of the episode index. Thus, the x-axis of evaluation graphs will represent episode indices

### DIFF
--- a/ultra/ultra/evaluate.py
+++ b/ultra/ultra/evaluate.py
@@ -73,9 +73,7 @@ def evaluation_check(
 
     for agent_id in agent_ids_to_evaluate:
         # Get the checkpoint directory for the current agent and save its model.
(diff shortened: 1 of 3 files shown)


ultra/ultra/evaluate.py
```diff
@@ -338,7 +338,7 @@ def evaluate_saved_models(
                     )
                 ]
             )[0]
-            episode.record_tensorboard()
+            episode.record_tensorboard(recording_step=1)
```
christianjans, line 341: If this loop were to iterate multiple times, would this overwrite existing data?

christianjans (commented):
Should the `episode.record_tensorboard()` calls in `train.py` and `tune.py` also be changed?

PR #932, opened 2021-06-10
ULTRA Migrate from step-based to episode-based sampling

JenishPatel99:
Addressing Issue: #928

- [x] Sample data episodically in `train.py` and `tune.py`
- [x] Save checkpoint directory in terms of episode index
- [x] Add an unit test inside of `test_evaluate.py` to make sure that the evaluation data is recorded at proper episode indices

### DIFF
--- a/ultra/tests/test_evaluate.py
+++ b/ultra/tests/test_evaluate.py
@@ -363,7 +363,7 @@ class EvaluateTest(unittest.TestCase):
         # do, then the evaluation training data is properly saved under the results.pkl
         # and also correctly added to the tensorboard
         for index in evaluation_training_results[AGENT_ID].keys():
-            self.assertFalse((index) % eval_rate)
+            self.assertEqual((index) % eval_rate, 0)
 
     def test_extract_policy_from_path(self):
         paths = [
--- a/ultra/tests/test_evaluate.py
+++ b/ultra/tests/test_evaluate.py
@@ -313,6 +313,58 @@ class EvaluateTest(unittest.TestCase):
             print(err)
             self.assertTrue(False)
 
+    def test_record_evaluation_at_proper_episode_indices(self):
+        """Due to parallelization, there might arise a situation where the episode
+        object at the beginning of an evaluation would not match the episode
+        object when recording to tensorboard. This test ensures that the evaluation data
+        (for both test and train scenarios) is recorded at the proper episode index.
+        """
+        AGENT_ID = "000"
+        log_dir = os.path.join(
+            EvaluateTest.OUTPUT_DIRECTORY, "output_eval_episode_check_log/"
+        )
+
+        # Arbitary values for evaluation rate and number of training episodes
+        eval_rate = 4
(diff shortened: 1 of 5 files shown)


ultra/ultra/evaluate.py
```diff
@@ -138,7 +136,9 @@ def collect_evaluations(evaluation_task_ids: dict):
             episode.eval_train_mode()
 
         episode.info[episode.active_tag] = ray.get(ready_evaluation_task_id)
-        episode.record_tensorboard(recording_step=agent_iteration)
+        episode.record_tensorboard(
+            recording_step=episode.index
+        )  # Record evaluation episodically
```
christianjans, line 141: Oh wait okay sorry I just realized that this might not work the way it is currently implemented. Have you tested it by looking at the Tensorboard and results pkls? Since we are running evaluation in parallel, `episode.index` here is most likely not the same episode index when we began evaluation. Therefore, it will record the data at the wrong episode.

Perhaps an easy way to check would be to run evaluation every 5 episodes or something, and then see if the Tensorboard and the results pkls record data at episodes indices that are multiples of 5.

ultra/tests/test_evaluate.py
```diff
@@ -313,6 +313,58 @@ def test_evaluate_multiagent(self):
...
+            ),
+            "rb",
+        ) as handle:
+            evaluation_results = dill.load(handle)
+
+        # Check if the episode indices are divisible by the evaluation rate. If they
+        # do, then the evaluation data is properly saved under the results.pkl
+        # and also correctly added to the tensorboard
+        for index in evaluation_results[AGENT_ID].keys():
+            self.assertFalse((index) % eval_rate)
```
christianjans, line 352:
```suggestion
            self.assertEqual((index) % eval_rate, 0)
```

ultra/tests/test_evaluate.py
```diff
@@ -313,6 +313,58 @@ def test_evaluate_multiagent(self):
...
+            ),
+            "rb",
+        ) as handle:
+            evaluation_training_results = dill.load(handle)
+
+        # Check if the episode indices are divisible by the evaluation rate. If they
+        # do, then the evaluation training data is properly saved under the results.pkl
+        # and also correctly added to the tensorboard
+        for index in evaluation_training_results[AGENT_ID].keys():
+            self.assertFalse((index) % eval_rate)
```
christianjans, line 366:
```suggestion
            self.assertEqual((index) % eval_rate, 0)
```

christianjans (commented):
Looks great, just a suggestion for the test:

ultra/tests/test_evaluate.py
JenishPatel99: Yes, that is correct. Thanks

christianjans (approved):

christianjans:
Does Alex know about these changes?

AlexLewandowski:
This all looks good to me!

PR #937, opened 2021-06-14
ULTRA Run tests in forked subprocesses

JenishPatel99:
Fork unit tests inside of `test_evaluate.py` and `test_train.py` to prevent segmentation faults and other termination signal (i.e. SIGABRT)

### DIFF
--- a/.github/workflows/ci-ultra-tests.yml
+++ b/.github/workflows/ci-ultra-tests.yml
@@ -45,7 +45,7 @@ jobs:
           --ignore=./tests/test_rllib_train.py \
           --ignore=./tests/test_tune.py \
           --ignore=./tests/test_train.py \
-          --ignore=./tests/test_evaluate.py \
+          --ignore=./tests/test_evaluate.py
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-ultra-tests.yml
```diff
@@ -44,6 +44,13 @@ jobs:
           --ignore=./tests/test_social_vehicles.py \
           --ignore=./tests/test_rllib_train.py \
           --ignore=./tests/test_tune.py \
+          --ignore=./tests/test_train.py \
+          --ignore=./tests/test_evaluate.py
+          pytest -v \
```
christianjans, line 49: Perhaps we can add a comment that explains why we have to divide up the heavy tests?

JenishPatel99: Of course, that is indeed needed 👍

christianjans (approved):
Awesome, seems to work great! Thanks!

Issue #883, opened 2021-05-27
Make Panda3D optional


RutvikGupta:
`Panda3D` is only required for visualization and rendering in SMARTS. 
We need to  move `Panda3D` to extra in `setup.py` so that we can remove its dependency from other packages during installation and should only be installed if the user need to have rendering in the simulations for agent camera observations.

- [x] Panda3D is fully optional
- [x] Warning to install the `extra` or otherwise named dependency when trying to use cameras
- [x] Updated documentation about rendering and camera based observations.

PR #940, opened 2021-06-15
Made Panda3D optional

RutvikGupta:
Closes Issue #883.


smarts/core/lidar.py
```diff
@@ -90,8 +89,44 @@ def _compute_rays(self):
             for yaw, roll in itertools.product(yaws, rolls):
                 rot = pybullet.getQuaternionFromEuler((roll, 0, yaw))
                 origin = np.array([0, 0, 0])
-                direction = np.array(
-                    Quat(rot).xform((0, self._sensor_params.max_distance, 0))
+
+                def mult_quat(q1, q2):
```
Gamenot, line 93: I could suggest moving this and the quaternion rotation method to `utils/math`.

smarts/core/renderer.py
```diff
@@ -158,7 +169,16 @@ def __init__(self, simid: str):
...
+
         self._showbase_instance = _ShowBaseInstance()
+        try:
+            print(self._showbase_instance.test_init())
+        except Exception as e:
+            # Known reasons for this failing:
+            raise AttributeError(
+                f"You may not have installed the [render] dependencies required to render the camera sensor observations. "
+                "Install them first using the command `pip install -e .[render]` at the source directory."
+            ) from e
```
Gamenot, line 181: OK, rather than adding this `test_init()` method it might be better to throw the `RendererException` within the `_ShowBaseInstance.__init()__`. Then we may not need to even need to do this

```suggestion
        self._showbase_instance = _ShowBaseInstance()
```

smarts/core/renderer.py
```diff
@@ -60,91 +47,115 @@ def required_to(cls, thing):
...
-        with self._render_lock:
-            for np in self.render.children:
-                if np != sim_root and not np.isHidden():
-                    np.hide()
-                    hidden.append(np)
-            self.taskMgr.mgr.poll()
-            for np in hidden:
-                np.show()
+    class _ShowBaseInstance:
+        pass
```
Gamenot, line 158: Then we can do this:
```suggestion
    class _ShowBaseInstance:
        def __init__():
           raise RendererException(f"You may not have installed the [render] dependencies required to render the camera sensor observations. "
                "Install them first using the command `pip install -e .[render]` at the source directory."
           )
```

smarts/core/renderer.py
```diff
@@ -60,91 +47,115 @@ def required_to(cls, thing):
...
+        FrameBufferProperties,
+        GraphicsOutput,
+        GraphicsPipe,
+        NodePath,
+        OrthographicLens,
+        Shader,
+        Texture,
+        WindowProperties,
+        loadPrcFileData,
+    )
```
Gamenot, line 63: I could suggest leaving a comment on why we are wrapping this in the try block.

setup.py
```diff
@@ -86,6 +84,10 @@
             "sphinx-rtd-theme",
             "sphinxcontrib-apidoc",
         ],
+        "render": [
```
Gamenot, line 87: I am unsure what the name should be: `render`? `renderer`? `camera`? `camera-obs`? `rendering`?

Gamenot (changes requested):

smarts/core/lidar.py
RutvikGupta: oh ok, will do.

smarts/core/renderer.py
RutvikGupta: Yeah this looks more neat. Thanks.

smarts/core/renderer.py
RutvikGupta: Sure.

setup.py
RutvikGupta: Maybe, this is something we can ask others,

sah-huawei (commented):
I think there's a cleaner way to make the `Renderer` stuff optional (which I'd planned to do before, but never finished).  Rather than explain it as a comment here, I found it easier to just do and push it.  _However I did not test it, so please feel free to review and tweak that as you see fit._

RutvikGupta:
> I think there's a cleaner way to make the `Renderer` stuff optional (which I'd planned to do before, but never finished). Rather than explain it as a comment here, I found it easier to just do and push it. _However I did not test it, so please feel free to review and tweak that as you see fit._

Yes, this seems more better than the try block which seemed less elegant. Thanks for the help!. 
P.S.: It had a minor issue, which is also fixed.

sah-huawei (approved):
good

Gamenot (approved):

PR #943, opened 2021-06-16
tested ultra for various scenarios

Dikshuy:
tested ultra on single-agent with traffic, multi-agent without traffic, and multi-agent with traffic


Dikshuy:
nvm

PR #944, opened 2021-06-16
Update scenario building documentation in ULTRA

christianjans:
@JenishPatel99, can you think of any other documentation that has to be changed as a result of #868?

### DIFF
--- a/ultra/ultra/scenarios/README.md
+++ b/ultra/ultra/scenarios/README.md
@@ -9,9 +9,15 @@
   - `social_vehicle_definitions.py`: Contains different social behaviors.
   - `visualization.py`: A tool for visualizing scenarios.
 - **pool/**: Contains different maps with various...
-  - number of lanes (2, 3, 4, 5, and 6),
-  - max speeds (50kmh, 70kmh, 100kmh), and
-  - shapes (c: cross-intersection, t: t-intersection).
+  - **analysis_pool/**: Contains different maps for analysis with various...
+    - number of lanes (2, 3, 4, 5, and 6),
+    - max speeds (50kmh, 70kmh, 100kmh), and
+    - shapes (c: cross-intersection, t: t-intersection).
+  - **experiment_pool/**: Contains different maps for experiments with various...
+    - number of lanes (2, 3, 4, 5, and 6),
+    - max speeds (50kmh, 70kmh, 100kmh), and
+    - shapes (c: cross-intersection, t: t-intersection).
+  > Note that the analysis pool and experiment pool contain the exact same maps except that the analysis pool has a side road on its maps that is used for analysis.
 - **task1/**: 
   - `config.yaml`: Defines the ego mission and traffic patterns for Task 1 (generalize over intersections by training on t-intersections, testing on cross-intersections).
 - **task2/**:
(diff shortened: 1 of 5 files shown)


JenishPatel99 (approved):

ultra/ultra/scenarios/interface.py
```diff
@@ -65,7 +65,7 @@
         "--pool-dir",
         help="directory for locating maps",
         type=str,
-        default=None,
+        default="pool/experiment_pool",
```
christianjans, line 68: I think this has to be relative to the directory we run `interface.py` from (the main `ultra/` directory). Or have you tested it? Does this work?

ultra/ultra/scenarios/README.md
```diff
@@ -54,14 +60,19 @@
...
+    # Generate the analysis scenarios from the analysis maps in the analysis pool.
+    $ python ultra/scenarios/interface.py generate --task 1 --level easy --pool-dir ultra/scenarios/pool/analysis_pool/
     ```
     Available parameters:
     - `--task`: Selects the task to generate (based on the task's config.yaml).
     - `--level`: Selects the level of the task from the task's config.yaml.
     - `--stopwatcher`: Include a stopwatcher (a vehicle that records the number of steps taken to do the left turn) in the scenarios.
     - `--save-dir`: The directory for saving the completed scenarios (completed scenarios will be put in the task directory if not specified).
     - `--root-dir`: The directory containing the task to be created (default is ultra/scenarios).
+    - `--pool-dir`: The directory containing the maps for the scenarios (default is pool/experiment_pool).
```
christianjans, line 75: This would be changed too if the default value for the `--pool-dir` argument is changed.

ultra/ultra/scenarios/interface.py
JenishPatel99: Yes, so the path for the pool directory is relative in terms of the scenarios folder. I have used a previously defined `scenarios_dir` path (as shown below) to join the `pool_dir``
https://github.com/huawei-noah/SMARTS/blob/125da655b4e28a5ebd00dd1283442d0828d6db09/ultra/ultra/scenarios/generate_scenarios.py#L707-L713

Also inside of `tests/test_analysis` we use the `pool_dir` argument to explicitly define the `pool/analysis_pool` path. The test does pass as well
https://github.com/huawei-noah/SMARTS/blob/125da655b4e28a5ebd00dd1283442d0828d6db09/ultra/tests/test_analysis.py#L50-L58

christianjans: Ahh yes, I see.

In order to be consistent with where we are defining our directories from, perhaps we should make this relative to the base ULTRA folder: `pool/experiment_pool` -> `ultra/scenarios/pool/experiment_pool`. For example, the `--root-dir` argument of `interface.py` has a default value relative to the base ULTRA directory (`ultra/scenarios`).

JenishPatel99: I do agree with that change. I have made those changes in the commits below

JenishPatel99: I do agree with that change. I have made those changes in the commits below

ultra/ultra/scenarios/README.md
JenishPatel99: a1766c9

ultra/ultra/scenarios/interface.py
christianjans: Oh awesome, thanks! Just made a small change to your change.

JenishPatel99 (approved):
Looks good!!

Issue #921, opened 2021-06-08
Implement frame stacking through SMARTS simulator rather than environment wrappers?


christianjans:
**Is your feature request related to a problem? Please describe.**
I have seen #916 and the [frame stack wrapper](https://github.com/huawei-noah/SMARTS/blob/6cb8ffda74e235c25d63b74313623b2e03e402f7/benchmark/wrappers/rllib/frame_stack.py) used in the RLlib benchmarks, but was wondering if frame stacking could be implemented without the use of an environment wrapper. That is, it would be done internally in SMARTS, allowing for frame stacking to be specified for stackable elements (such as images, lidar, etc.) of each agent's observation.

**Describe the solution you'd like**
Say, for example, a user wants their agent to have their RGB observation be the current RGB image plus the previous three RGB images (for a total of four images in the stack). They could define their AgentSpec like:
```python
my_agent_spec = AgentSpec(
  interface=AgentInterface(
    rgb=RGB(width=64, height=64, stack_size=4, resolution=(50 / 64)),
    # ... other parts of the agent interface ...
  ),
  # ... other parts of the agent spec ...
}
```
And then when `step` is called on an environment like "hiway-v0" (without a wrapper), SMARTS would return the agent's observation with a stacked `TopDownRGB` of some form. I am unsure of exactly what form this part of the observation should take. Whether `observation.top_down_rgb` should be a `Sequence` of `TopDownRGB` instances, or perhaps `TopDownRGB`'s attributes, `metadata` and `data`, should be `Sequence`s?

**Describe alternatives you've considered**
Alternatives would be similar to the Gym wrapper in #916 and the RLlib wrapper used in the SMARTS' benchmarking. Perhaps a concern with these wrapper methods however is that not every agent may need, or expect, a stacked observation. For example, in #916 it seems like an environment that is wrapped by `FrameStack` will return an agent's observation in the form of a `Sequence` type (a `Deque`). Modifications to the agents' observation adapters may be required to handle this change.

Additionally, if not every agent requires a stacked observation, the returned observations from SMARTS will be considerably larger than needed as the environment will stack _all_ of the agent's observations.

Finally, these wrappers don't provide the same "control" over the observation as this proposed method would when it comes to stacking. For example, with the environment wrappers, if one agent wants a stacked observation, then all other agents must have a stacked observation. And if an agent just wants the images in their observation to be stacked, the environment wrappers would actually stack everything in the agent's observation.

**Additional context**
[...]

PR #948, opened 2021-06-16
Add frame stacking to ULTRA

christianjans:
After considering many options, this may be the solution to go with for frame stacking in the ULTRA environment. It is not as extensible as using environment wrappers, however it provides the ability to add frame stacking to specific observations, and provides agents with consistently-sized observations (provided their observation adapter produces a consistently-sized observation).

Environment wrappers cannot be used in ULTRA without some major refactoring. It would require the baselines to be changed along with how training and evaluation use the environments.

Frame stacking could not be done in the agents either due to the agents having both `act` and `step` methods. There was no guarantee that the observation passed to `step` was the observation observed directly after the previous time `step` was called. This made not possible to keep track of a stack of observations within the agent.

Finally, the comment on #921 outlines reasons why frame stacking cannot be implemented in the SMARTS simulator to be used by agents that specify a stack size in their `AgentInterface`.

Currently, only `TopDownRGB` elements of the observation are stacked by changing the `data` attribute of this class. This can similarly be done for any other element of the observations that uses a NumPy array like the `DrivableAreaGridMap` and the `OccupancyGridMap`.

TODO
- [x] Mention differences between ULTRA's Gym and RLlib environments now
- [x] Address TODO in `ultra_env.py`
- [x] Trim docstring of stacking method in `ultra_env.py`
- [x] Rename stacking method in `ultra_env.py` to be more general
- [x] Pass CI tests

### DIFF
--- a/ultra/docs/ultra_environments.md
+++ b/ultra/docs/ultra_environments.md
@@ -1,0 +1,14 @@
+# ULTRA Environment
+
+ULTRA's Gym and RLlib environments inherit from SMARTS' `smarts.env.hiway_env.HiWayEnv`
+and `smarts.env.rllib_hiway_env.RLlibHiWayEnv` respectively. However, they differ in two
+ways:
+1. ULTRA's gym environment takes a `scenario_info` parameter that is used to specify a
+specific task and level that describe the type of scenarios the environment will use.
+2. ULTRA's gym environment performs automatic framestacking of certain parts of the
+environment observation it receives from SMARTS. Currently, if an agent has a
+`TopDownRGB` in its observation, the ULTRA environment will change the `data` attribute
+of this class to still be a NumPy array, but with shape
+`(_STACK_SIZE, HEIGHT, WIDTH, 3)` where `_STACK_SIZE` is the number of frames to stack
+(its value can be found in `ultra.env.ultra_env`), `HEIGHT` is the height of the RGB
+image, and `WIDTH` is the width of the RGB image.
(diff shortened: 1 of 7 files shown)


ultra/docs/ultra_environments.md
```diff
@@ -0,0 +1,20 @@
...
+and `smarts.env.rllib_hiway_env.RLlibHiWayEnv` respectively.
+
+ULTRA's Gym environment differs in two ways from SMARTS' HiWayEnv:
+1. ULTRA's Gym environment takes a `scenario_info` parameter that is used to specify a
+specific task and level that describe the type of scenarios the environment will use.
+2. ULTRA's Gym environment performs automatic framestacking of certain parts of the
+environment observation it receives from SMARTS. Currently, if an agent has a
+`TopDownRGB` in its observation, the ULTRA environment will change the `data` attribute
+of this class to still be a NumPy array, but with shape
+`(_STACK_SIZE, HEIGHT, WIDTH, 3)` where `_STACK_SIZE` is the number of frames to stack
```
JenishPatel99, line 13: I am jus wondering what attribute does the "3" represent?

JenishPatel99 (commented):
Looks good 👍

ultra/docs/ultra_environments.md
christianjans: Yeah good question, each pixel in the image has a red, green, and blue value and the 3 in the last position just indicates that the image's RGB values are in this dimension of the array.

PR #956, opened 2021-06-21
ULTRA Update `MANIFEST.in`

JenishPatel99:
The current version of the `MANIFEST.in` requires update to include the following files:
- `params.yaml` for all baselines
- `config.yaml` for taskX
- `envision_bash` inside `ultra/env`
- fix ```recursive-include ultra/scenarios/pool ``` by adding asterisk at the end

@christianjans Are their any other non-python files that may need to be included in a .whl file?

### DIFF
--- a/ultra/MANIFEST.in
+++ b/ultra/MANIFEST.in
@@ -1,6 +1,17 @@
 include sac/checkpoint/checkpoint
 include ultra/agent_pool.json
 include config.yaml
+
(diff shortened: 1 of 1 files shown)


christianjans (approved):
I think that is everything from what I can see 👍

JenishPatel99:
> I think that is everything from what I can see 👍

Perfect, thanks. We can always update it later 👍

Issue #784, opened 2021-04-21
Envision multi-instance display


Adaickalavan:
## **BUG REPORT**
Unable to view multi-instance display when "All Simulations" is selected in Envision while running multiple SMARTS instances.

**SMARTS version**
0.4.15

**Previous associated issues**
None

**Steps to reproduce the bug**
1) Start envision in a terminal
    ```
    $ scl envision start -s ./scenarios -p 8081
    ```
2) Start SMARTS simulation in another terminal
    ```
    $ python3.7 ./examples/multi_agent.py ./scenarios/intersections/6lane/
    ```

**Resulting and expected behaviour**
Expected: Able too see multi-instance display in Envision when "All Simulations" options is selected.

![image](https://user-images.githubusercontent.com/8764693/122244324-cc1e0a80-ce92-11eb-9692-440c801cad5c.png)

**Error logs and screenshots**
None
 
**System information**- Operating system & Environment [Please note if using Docker]: Ubuntu 16.04
- Number of system cores available: 16
- Python version [Please note if using Jupyter]: Python3.7
- Python packages [e.g. from `pip freeze`]
- SUMO version: 1.8.0
- Date when bug occurred: 19 April 2021

**Impact [If known]**
Unable to view multi-instance display in Envision.

RutvikGupta:
Something related to this was addressed in PR #363. But that was fixed before this issue was noticed.

RutvikGupta:
After tracking down, the git commit history, it seems like the multi-instance display stopped working after SMARTS version 0.4.14 release. (https://github.com/huawei-noah/SMARTS/commit/c8473049b4b52ad3eb29f7c7eb93753c2d70df55)

RutvikGupta:
After some more investigation, its safe to say that the issue is arising from changes done to the `.js` files in this release. Changes to the `client.py` or other files dont seem to be the issue.

RutvikGupta:
PR #575 and #547 were made to the JavaScript files after SMARTS version 0.4.13 release and before SMARTS version 0.4.14 release which caused the multi-instance display to stop working.

RutvikGupta:
Actually PR #580 is the cause of this issue.

RutvikGupta:
Also for PR #545, the following code snippet in `app.js` tries to reconnect to a single simulation every time, you switch to "All Simulations".
[...]

PR #957, opened 2021-06-22
Fixed the Multi Instance Display of Envision

RutvikGupta:
Fixes Issue #784. Edited some changes of PR #545 so that envision does not reconnects to a running simulation when switched to "All Simulation" or multi-instance display.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,7 @@ clean:
 .PHONY: format
 format:
 	# pip install isort==5.7.0
-	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./benchmark/ ./cli ./envision ./examples/ ./extras/ ./scenarios/ ./smarts ./ultra ./zoo
+	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/benchmark/ ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./ultra ./zoo
 	# pip install black==20.8b1
 	black .
 	# npm install prettier
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,7 @@ clean:
 .PHONY: format
 format:
 	# pip install isort==5.7.0
-	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/benchmark/ ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./ultra ./zoo
+	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/marl_benchmark/ ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./ultra ./zoo
 	# pip install black==20.8b1
 	black .
 	# npm install prettier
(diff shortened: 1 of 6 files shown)


envision/web/src/components/app.js
```diff
@@ -67,20 +67,20 @@ function App({ client }) {
...
       setSimulationIds(ids);
-    };
+    })();
 
     // checks if there is new simulation running every 3 seconds.
-    const interval = setInterval(fetchRunningSim, 3000);
-    return () => clearInterval(interval);
-  }, [matchedSimulationId]);
+    // const interval = setInterval(fetchRunningSim, 3000);
+    // return () => clearInterval(interval);
```
RutvikGupta, line 82: I had to comment out these two lines since they kept on reconnecting the server to a running Simulation when you switch to "All Simulation" or multi-instance display page. The downside now is that the user needs to reload the server every time a new Simulation is added to the server port.

RutvikGupta: One thing we can do is to check if there is a change in the number of ids. If there is, we can call the `fetchRunningSim`, else we don't set the interval for `fetchRunningSim`.

RutvikGupta: Ok , I tried this and I don't think this can work.

RutvikGupta: UPDATE: This has been fixed now!

Gamenot:
I have tested this out, it appears to be working. I think there are still a few problems remaining here but it is in much better shape than before.

Gamenot (approved):

envision/web/src/components/app.js
```diff
@@ -61,26 +61,29 @@ function App({ client }) {
   const recorderRef = useRef(null);
   const { addToast } = useToasts();
   const history = useHistory();
+  const [totalSimIds, setTotalSimIds] = useState(0);
+
+  let getSimIdsLength = async () => {
+    let ids = await client.fetchSimulationIds();
+    setTotalSimIds(ids.length);
+  };
+  setInterval(getSimIdsLength, 3000);
```
Gamenot, line 70: I believe this works better. The one remaining part of it is that we will need to fix the automatic navigation to a simulation page that occurs when the number of simulation IDs change but I also think we can leave this to a different PR.

RutvikGupta: Hmm, I when I try, it doesn't navigates to simulation page when no. of sim IDs changes but remains on the simulation you were originally on.

RutvikGupta: Oh, you meant when you are on "All Simulation" page. Yeah well that makes sense because for a new simulation ID it pushes it to the `history` and the server reconnects envision to that simulation if you are not connected to any simulation i.e. if you are on the multi-instance display.

Gamenot: #961

Makefile
```diff
@@ -130,7 +130,7 @@ clean:
 .PHONY: format
 format:
 	# pip install isort==5.7.0
-	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./benchmark/ ./cli ./envision ./examples/ ./extras/ ./scenarios/ ./smarts ./ultra ./zoo
+	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/benchmark/ ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./ultra ./zoo
```
Adaickalavan, line 133: Is it `./baselines/marl_benchmark` ?

RutvikGupta: Yes. Fixed it.

Adaickalavan (approved):

PR #958, opened 2021-06-22
ULTRA baselines with new adapters

christianjans:
Refactors the existing baselines to allow some to be compatible with multiple observation and action types.
| Baseline | Default Vector Observation | Default Image Observation | Default Discrete Action | Default Continuous Action |
|-----------|-------------------------------|------------------------------|----------------------------|-----------------------------|
| BDQN | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | |
| DQN | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| PPO | :heavy_check_mark: | | | :heavy_check_mark: |
| SAC | :heavy_check_mark: | | | :heavy_check_mark: |
| TD3 | :heavy_check_mark: | :heavy_check_mark: | | :heavy_check_mark: |

Questions:
- Should we get rid of the BDQN's `explore.py` and `network.py`? There was no difference between the BDQN and DQN's versions of these files, and the changes in this PR currently have the BDQN using the DQN's versions of these files.


JenishPatel99 (approved):
@christianjans looks good. I tried my best to spot any syntax and logic error, there weren't though 👍. However, I think we might need to test each baseline to expose an runtime errors. I had a version of a test suite which ran training and evaluation for each baseline policy. I will create an issue for that and push a PR as well

ultra/tests/test_replay_buffer.py
```diff
@@ -0,0 +1,341 @@
...
+                self.assertIn(action, actions)
+                self.assertIn(reward, rewards)
+                self.assertIn(done, dones)
+                self.assertTrue(np.array_equal(state, states[index_of_state]))
+                self.assertTrue(np.array_equal(next_state, next_states[index_of_state]))
+                self.assertTrue(np.array_equal(action, actions[index_of_state]))
+                self.assertEqual(reward, rewards[index_of_state])
+                self.assertEqual(done, dones[index_of_state])
+
+    # def test_vector_replay_buffer(self):
```
JenishPatel99, line 98: Are we going to implement the vector-based replay buffer test later on?

ultra/ultra/baselines/dqn/dqn/policy.py
```diff
@@ -22,7 +22,7 @@
 import torch
 from torch import nn
 import numpy as np
-from ultra.baselines.dqn.dqn.network import *
+from ultra.baselines.dqn.dqn.network import DQNCNN, DQNWithSocialEncoder
```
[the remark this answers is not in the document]
christianjans: Great catch, thanks!

ultra/tests/test_replay_buffer.py
christianjans: Ideally yes, although I think it is a lower priority now with everything else that is going on. We (or someone else) can implement it in the future.

JenishPatel99 (approved):
New changes looks good

PR #962, opened 2021-06-23
ULTRA refactor `ultra_env.py`

JenishPatel99:
The `ultra_env.py` needs a bit of refactoring:
- [x] Remove `get_tasks()` because the method name and its purpose do not compliment each other properly. Instead add a new method `get_scenarios()` which takes in the `scenario_info` tuple and returns a list of either all training or testing scenarios that corresponds to the specified task and level

### DIFF
--- a/ultra/ultra/env/ultra_env.py
+++ b/ultra/ultra/env/ultra_env.py
@@ -56,13 +56,14 @@ class UltraEnv(HiWayEnv):
         eval_mode=False,
         ordered_scenarios=False,
     ):
-        self.agent_specs = agent_specs
         self.scenario_info = scenario_info
         self.headless = headless
         self.timestep_sec = timestep_sec
         self.seed = seed
         self.eval_mode = eval_mode
-        self.ordered_scenarios = ordered_scenarios
+
+        agent_specs = agent_specs
+        ordered_scenarios = ordered_scenarios
 
         self.scenarios = self.get_scenarios(scenario_info)
 
@@ -157,7 +158,7 @@ class UltraEnv(HiWayEnv):
 
         Args:
             scenario_info Tuple[str, str]: The first index represents
-            task id and second index represents task's level
+                task id and second index represents task's level
 
         Returns:
             List[str]: Absolute path of scenarios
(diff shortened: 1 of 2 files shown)


ultra/ultra/env/ultra_env.py
```diff
@@ -143,17 +145,41 @@ def reset(self):
...
+        To obtain the path of scenarios:
+            1. Get the filename patterns of the train and test scenarios from
+               ultra/config.yaml from the given task and level.
+            2. Change the filename patterns from relative to absolute path
+            3. Depending on the evaluation mode, we can provide glob.glob with
+               the appropriate filename pattern to return a list of scenario dirs
+
+        Args:
+            scenario_info Tuple[str, str]: The first index represents
+            task id and second index represents task's level
```
christianjans, line 160:
```suggestion
                task id and second index represents task's level
```

ultra/ultra/env/ultra_env.py
```diff
@@ -55,23 +56,24 @@ def __init__(
...
-            _scenarios = glob.glob(f"{self.scenarios['train']}")
-        else:
-            _scenarios = glob.glob(f"{self.scenarios['test']}")
+        self.headless = headless
+        self.timestep_sec = timestep_sec
+        self.seed = seed
+        self.eval_mode = eval_mode
+        self.ordered_scenarios = ordered_scenarios
+
+        self.scenarios = self.get_scenarios(scenario_info)
```
christianjans, line 67: Do these have to be instance members? I don't think it matters too much if the `ordered_scenarios` is a member of the instance (since it is only a boolean type), but I can imagine the `scenarios` variable can be quite large; and if we only use it in the constructor, it may just be wasted space. Let me know your thoughts.

JenishPatel99: Yes, I was thinking that the headless, timestep_sec and seed parameters are defined to env params, thus I thought it would right to include them as instance members. However for the ordered_scenarios and the agent_specs it would be okay to keep them as regular variables

christianjans: I believe `self.scenarios` can also be changed to just a local variable?

christianjans (approved):

PR #965, opened 2021-06-24
ULTRA Sanity check on baselines

JenishPatel99:
To ensure that all of the ultra baselines (bdqn, dqn, ppo, sac, td3) are working properly, we can run a end-to-end training run for each baseline. This way we can expose any run-time issues across all of the baselines, and not just the default (i.e is sac)

@christianjans Do you think of any others tests across all baselines which might expose any issues?


christianjans (commented):
I think this is great! The only concern I have is that we are not testing the `save` and `load` methods of the baselines since we are not doing evaluation. But this might be difficult to do. What are your thoughts?

christianjans (approved):

PR #982, opened 2021-06-30
ULTRA Add task id to generated scenario directories

JenishPatel99:
Previously, we had the below format of scenarios:
For train scenarios: `<path_to_scenarios>/train_<level_name>_<intersection_type>_<speed>_<traffic_density>`
For test scenarios: `<path_to_scenarios>/test_<level_name>_<intersection_type>_<speed>_<traffic_density>`

As we add more task, such as task 3, we might encounter tasks with similar level names (e.g. task 1 and 3 have easy, medium and hard levels). So when we generate the scenarios for them we might have ambiguity between task 1 and task 3 if they were saved in the same directory. To prevent that, adding a task_id to the scenario directory will clarify which task does a scenario belong to. The new path will look like this:
For train scenarios: `<path_to_scenarios>/train_<task_id>_<level_name>_<intersection_type>_<speed>_<traffic_density>`
For test scenarios: `<path_to_scenarios>/test_<task_id>_<level_name>_<intersection_type>_<speed>_<traffic_density>`


christianjans (approved):
Good idea! Wait why do the test tasks not need their paths changed?

JenishPatel99:
> Good idea! Wait why do the test tasks not need their paths changed?

I was having a bit of trouble with updating test tasks, the ci tests kept failing. However, I manage to do fix it now. The test tasks will also follow the *_task_level_* format

christianjans (approved):

Issue #823, opened 2021-05-07
ULTRA Load existing models for training


JenishPatel99:
**Is your feature request related to a problem? Please describe.**
Currently models at different checkpoints can be used to be evaluated on test scenarios. However, ULTRA does not support training existing models. This feature can allow the trained models to be further trained on additional tasks and levels. 

Lets say that an agent saves its model after successfully training on the no-traffic and the blocking grades. Using the proposed feature we can continuing training that model on more harder scenarios. Instead of training the model from scratch we can now begin training from an experienced state. This can reduce the time taken to learn the target task.   

**Describe the solution you'd like**
The method essentially uses the agent's metadata information from a stored model in the experiment directory and retrieves the `agent_ids`, `agent_policies`. Using the `checkpoint_directory` we can also retrieve the `agent_spec`. The solution will be more clear once I have linked the PR with this issue with comments.

**Additional context**
We might have to check if the model's network (model that we want to continue training) is updating properly.

**Notice**
A common misconception is that training an existing model is the same as continuing (or resuming) training. In the current implementation, the models are loaded to extract only the agent information (agent_ids, agent_classes). Using the checkpoint directory, a new agent_spec is created and then after the a **new agent** is built using that agent_spec. In layman's term, we are transferring the previous agent's knowledge to a new agent. 

This is not the same as the previous agent resuming the training from where it last left off. To accomplish that the replay buffer and the optimizer must also be saved and loaded back to resume training

PR #988, opened 2021-07-02
ULTRA Load models to train

JenishPatel99:
Similar changes to those in PR #834 which were push to the curriculum learning branch. Although it seems that it will be a nice feature to have inside of core ULTRA. Especially, if you want to train a previos model. More information can be found on Issue #823

### DIFF
--- a/ultra/tests/test_train.py
+++ b/ultra/tests/test_train.py
@@ -24,6 +24,7 @@ import os
 import shutil
 import sys
 import unittest
+import glob
 
 import gym
 import ray
@@ -31,7 +32,7 @@ import ray
 from smarts.core.agent import AgentSpec
 from smarts.zoo.registry import make
 from ultra.baselines.sac.sac.policy import SACPolicy
-from ultra.train import train
+from ultra.train import train, load_model
 
 AGENT_ID = "001"
 seed = 2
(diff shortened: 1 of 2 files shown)


ultra/docs/training_models.md
```diff
@@ -0,0 +1,19 @@
+# Training existing models
+
+ULTRA has supported the process to seperately **evaluate** saved models through the `evaluate.py` script. 
+However, ULTRA has now extended support **training** of previously saved models. This will allow users to
```
christianjans, line 4:
```suggestion
However, ULTRA has now extended support to **training** of previously saved models. This will allow users to 
```

ultra/docs/training_models.md
```diff
@@ -0,0 +1,19 @@
+# Training existing models
+
+ULTRA has supported the process to seperately **evaluate** saved models through the `evaluate.py` script. 
+However, ULTRA has now extended support **training** of previously saved models. This will allow users to 
+continue training their saved models from previous experiment.
```
christianjans, line 5:
```suggestion
continue training their saved models from a previous experiment. 
```

ultra/docs/training_models.md
```diff
@@ -0,0 +1,19 @@
...
+However, ULTRA has now extended support **training** of previously saved models. This will allow users to 
+continue training their saved models from previous experiment. 
+
+To accomodate this feature the experiment directory must include the **model** needed to be further trained and
+the **replay buffer** used by the agent. 
+
+> Although, it is not necessary to save the replay buffer inside the experiment directory, the performance of training the model will be affected due to no prior experience replay.
+
+As an example, we have trained a SAC agent for 10000 episodes and at every 200th episodic checkpoint a model is saved.
+he replay buffer is also saved at the end of the experiment. The experiment directory has been stored in our local
```
christianjans, line 13:
```suggestion
The replay buffer is also saved at the end of the experiment. The experiment directory has been stored in our local
```

ultra/docs/training_models.md
```diff
@@ -0,0 +1,19 @@
...
+directory. By the running the following command, we can train a single model from the experiment directory. 
+```sh
+    python ultra/train.py --task 1 --level easy --episodes 10 --experiment-dir <path-to-experiment-dir>
+```
+
+> We have chosen to train only one model from the experiment directory to ensure simplicity. If there are multiple models inside the experiment directory then last model which was saved will be used.
```
christianjans, line 19:
```suggestion
> We have chosen to train only one model from the experiment directory to ensure simplicity. If there are multiple models inside the experiment directory then the last model which was saved will be used.
```

ultra/ultra/train.py
```diff
@@ -115,6 +122,76 @@ def create_argument_parser():
...
+        for agent_id, agent_spec in agent_specs.items()
+    }
+
+    try:
+        _ = [
+            agents[agent_id].load_replay_buffer(
+                os.path.join(experiment_dir, "replay_buffers", agent_id)
+            )
+            for agent_id in agent_ids
+        ]
```
christianjans, line 184: Will this work for agents where some have a replay buffer and some do not have a replay buffer?

For example, say the `agent_ids` are `["000", "001"]` where 000 does not have a replay buffer, but 001 does have a replay buffer. As we iterate through the IDs, the `NotImplementedError` will be thrown on the first iteration since 000 does not have a replay buffer, and we will be immediately redirected to the `except` block without 001 ever having loaded its replay buffer.

Is this a possibility?

JenishPatel99: Will do 👍

ultra/tests/test_train.py
```diff
@@ -41,6 +42,21 @@ class TrainTest(unittest.TestCase):
...
+        try:
+            os.system(
+                f"python ultra/train.py --task 00 --level easy --episodes 3 --eval-rate 2 --max-episode-steps 2 --log-dir {model_log_dir} --eval-episodes 1 --headless"
+            )
+        except Exception as err:
+            print(err)
+            self.assertTrue(False)
+
+        if not os.path.exists(model_log_dir):
+            self.assertTrue(False)
```
christianjans, line 58: `self` doesn't seem to be an attribute of this method. I wonder if we can use this setup method to just call a multi-agent training run and a single agent training run, and then all tests that need a training run, can just use the experiment directories that were created by this method (like how `test_evaluate.py` does it).

I can take a look at this too, but let me know if you have any ideas.

christianjans (commented):
Sorry! I just realized I had started some of this review earlier, here is what I have so far.

Issue #969, opened 2021-06-28
Abnormal collision with imitation learning


zbzhu99:
## **BUG REPORT**
**High Level Description**
The history replacement agent will terminate with collision frequently

**Details**
I train the imitation learning agent on NGSIM map using SMARTS of 4c829edd7535eaf0e79b57d384fdaf8711226978. When I update SMARTS to the newest develop branch and test the model, the trained agent seems to have much more collisions than expected. 

From my test, the abnormal collision is likely related to 8bdbba22864b2f13e80dbd04bfb17d98f76724d5.

zbzhu99:
This is the map file I used:

[map.net.xml.zip](https://github.com/huawei-noah/SMARTS/files/6739167/map.net.xml.zip)

sah-huawei:
@zbzhu99 Thanks.  The good news is that you're using the most current one, which is at 1:1 scale.  Assuming both (a) the history dataset (`.shf`) file you are using was re-generated (`scl build --clean...`) with the `.yaml` file in your zip folder, and (b) the model you are testing now was trained using this same dataset (`.shf`)  and map (`.net.xml`) at 1:1 scale, then everything should work.  But if either (a) or (b) are incorrect assumptions, then this could explain the behavior you are seeing.

Please let me know about (a) and (b) and then I will try to debug accordingly.  Thanks for your patience!

zbzhu99:
When I rebuild the scenario, the map now looks weird:

sah-huawei:
How odd!  This appears to be a problem with the `map.glb` created for and used by Envision.  I'm able to reproduce it here.  The `map.net.xml` file seems fine (both in `netedit` and in `sumogui`), so I suspect this is just a visualization problem (i.e., the data should still be ok), but I will dig into it further (espcially to see if it the problem might relate to collisions too) and report back.

sah-huawei:
@RutvikGupta This (the `map.glb` problem,not the collision issue) appears to be related to PR #901.  (I reverted to just before that and the `map.glb` appears fine in Envision.)  Could you please help us to fix that part?  Meanwhile, I'll continue to investigate the collision issue.  Thanks!

RutvikGupta:
Sure, I will have a look.

RutvikGupta:
Which scenario was used to run the test? I don't see the visualization issue in existing set of scenarios.

RutvikGupta:
[...]

PR #992, opened 2021-07-05
Fix NGSIM map holes

RutvikGupta:
Logic fixes to the `_snap_internal_holes` and  `_snap_external_holes` methods in `sumo_road_network.py` which caused wrong snapping of the merging lanes to its nearest lanes in NGSIM map. See PR #901.


smarts/core/sumo_road_network.py
```diff
@@ -231,8 +231,11 @@ def _compute_road_polygons(self):
                 lane_to_poly[lane.getID()] = shape
 
         # Remove holes created at tight junctions due to crude map geometry
+        # NOTE: If you don't want snapping of a particular edge, set its type to "ignore_snapping" in
+        # the scenario's map.net.xml file. See scnenario/ngsim/map.net.xml for example.
```
Gamenot, line 235: I am curious, is this inbuilt or a custom type?

RutvikGupta: No, obviously custom type.

Gamenot (commented):
I think I am waiting on @sah-huawei 's comments on this. I am not entirely confident in this change.

sah-huawei (changes requested):
I think that instead of adding the `ignore_snapping` type we should try to fix the NGSIM map (and then share it directly with Zhengbang rather than pushing it to our repo).  As I understand it, this map has issues due to the way it was (custom) made -- potentially messing up the labeling of internal lanes and/or junctions --  and also then how I scaled it down to 1:1 scale (potentially creating more holes).  Both of these things should be fixable by editing the map.  I will try to do that for Issue #969 asap.

Meanwhile, although it may or may not be a problem with respect to Issue #969, the logic behind snapping lanes to nearby lanes is problematic (for both internal and external).  For example, on line 327 of `sumo_road_network.py` (as it is in this review), I believe there should be a `break` statement to exit the loop after the nearest lane has been snapped-to (rather than continuing through the other nearby, but further-away lanes and possibly snapping to them as well, potentially undoing the original snap).  We should probably fix that while we're here...

The other changes in this PR (related to the iteration over scenarios) do seem worth keeping, no matter what happens with the holes.

smarts/core/sumo_road_network.py
```diff
@@ -946,28 +955,41 @@ def _snap_external_holes(self, lane_to_poly, snap_threshold=2):
...
+                    ):
                         continue
                     nl_shape = lane_to_poly.get(nl.lane_id)
                     if nl_shape:
-                        lane_shape = Polygon(snap(lane_shape, nl_shape, snap_threshold))
-            lane_to_poly[lane_id] = lane_shape
+                        p, np = nearest_points(shPoint(x, y), nl_shape)
+                        if p.distance(np) < snap_threshold:
+                            new_coords.append(np)
+                            break
```
Gamenot, line 988: Is this to express snapping to an edge instead of points? I think we should filter out duplicate points here if the snap threshold repeats points in the polygon. I think this also still might run into a similar problem since we may have some points fall out of the `snap_threshold` that should still be used.

I think one likely problem that we have run into is that with curved roads the outer points of a internal lane might not line up with inner points of the outer lane. 

I think one option we have if the edges are parallel is to construct the outermost lane `lane0` in entirety and then reuse the inner edge of the previous lane `lane(n-1)` on each pass as we move towards the innermost lane `lane(n)`.

smarts/core/sumo_road_network.py
```diff
@@ -946,28 +955,41 @@ def _snap_external_holes(self, lane_to_poly, snap_threshold=2):
...
                     include_junctions=False,
                 ):
-                    if (not nl) or (nl and nl.in_junction):
+                    if not nl or nl.in_junction or nl.lane_id == lane_id:
+                        continue
+                    ll, _ = nl.lane_to_left
+                    rl, _ = nl.lane_to_right
+                    if (not ll or ll.lane_id != lane_id) and (
+                        not rl or rl.lane_id != lane_id
+                    ):
```
Gamenot, line 981: Is there any way to just use `lane_to_left|lane_to_right` exclusively instead of also calling `nearest_lanes`? I feel as though left/right by itself would be faster to resolve.

I think this also might now be bypassing the case of outgoing and in-going lanes although I suspect they should not be a problem in.

RutvikGupta:
With @sah-huawei's new algorithm for internal and external snapping of map holes and the new NGSIM map, the hole issue is resolved for all existing set of maps. Though this was for only the visualization issue, they don't seem to cause any on-shoulder or off-road events for the ego or social vehicles.
The PR is ready for review now.

sah-huawei:
> With @sah-huawei's new algorithm for internal and external snapping of map holes and the new NGSIM map, the hole issue is resolved for all existing set of maps. Though this was for only the visualization issue, they don't seem to cause any on-shoulder or off-road events for the ego or social vehicles. The PR is ready for review now.

Also, I have attached an improved map for the NGSIM i80 datasets to Issue #969.

smarts/core/sumo_road_network.py
```diff
@@ -975,17 +979,36 @@ def _snap_internal_holes(self, lane_to_poly, snap_threshold=2):
...
+                        nl_shape = lane_to_poly.get(nl.lane_id)
+                        if nl_shape:
+                            _, np = nearest_points(p, nl_shape)
+                            if p.distance(np) < thresh:
+                                p = np  # !!!! :)
+                                # allow vertices to snap to more than one thing, but
+                                # try to avoid infinite loops and making things worse instead of better here...
+                                moved = True
+                                snapped_to.add(nl)
+                                thresh *= 0.75
```
sah-huawei, line 1006: I should probably add a comment about this magic number.  (and below too)

sah-huawei (approved):

smarts/core/sumo_road_network.py
sah-huawei: Done.

smarts/sstudio/genhistories.py
```diff
@@ -406,7 +406,9 @@ def _transform_all_data(self):
 
         map_width = self._dataset_spec["map_net"].get("width")
         if map_width:
-            valid_x = (df["position_x"] * self.scale).between(0, map_width)
+            valid_x = (df["position_x"] * self.scale).between(
+                df["length"] / 2, map_width - df["length"] / 2
+            )
```
Gamenot, line 411: What is the intention with this change?

Gamenot (changes requested):
Please add entries to the `CHANGELOG.md` for the changes made.

smarts/sstudio/genhistories.py
sah-huawei: I made that one.  This is to prevent agent vehicles from immediately being flagged as being off the map ("off-road" or "on-shoulder") when they first appear.  Each vehicle is  a different size in the NGSIM dataset, and they are coming from outside of the map in the "real world".   The first data-point we get for them is when their bbox centroid crosses into the map box.  But at that point in time, the vehicle's rear-end may not have made it onto the map yet, so the back tires will be marked as off road.

I originally made this change because I wanted to test the modified maps to see if vehicles were flagged for driving in holes and I was seeing "false positives" based on the map edges.  I only realized later that the map holes  would not cause off-road events anyway, but I think this change should remain (in particular because the imitation learning group uses our Events as feedback in the model training and testing).

Gamenot (approved):

PR #996, opened 2021-07-07
Raise Exception for failing renderer tests due to missing [camera-obs] dependencies

RutvikGupta:
This stems from PR #940. If you run `make sanity-test` before, `test_smarts_doesnt_leak_tasks_after_reset`  in `smarts/core/test/test_smarts.py` because of missing `Panda3D` and its dependencies packages required to use the `Renderer` Class. So I added exceptions to warn users to install the `[camera-obs]` dependency if they run these following tests. 

Additionally I could have added `Panda3D` to test dependencies but this will make `[camera-obs]` obsolete.


sah-huawei (approved):

saulfield:
Should we update the README as well? In the setup section it lists `[camera-obs]` as optional, but it seems it's required for `make sanity-test` to pass.

RutvikGupta:
> Should we update the README as well? In the setup section it lists `[camera-obs]` as optional, but it seems it's required for `make sanity-test` to pass.

Yeah, this was a dilemma I was facing. I am not sure if `make sanity-test` is considered optional or not. So I think I will update the `README.md` asking users to install it if it fails.

smarts/core/tests/test_renderers.py
```diff
@@ -109,7 +116,12 @@ def test_optional_renderer(smarts, scenario):
     assert not smarts.is_rendering
     for _ in range(10):
         smarts.step({})
-    renderer = smarts.renderer
+
+    try:
+        renderer = smarts.renderer
+    except Exception as e:
+        raise RendererException.required_to("run test_renderer.py")
```
Gamenot, line 123: I do not think this does as you expect. The smarts.renderer does not throw an exception.

```suggestion
    renderer = smarts.renderer
    if not renderer:
        raise RendererException.required_to("run test_renderer.py")
```

Gamenot (changes requested):

smarts/core/tests/test_renderers.py
RutvikGupta: image
I tested it and it was showing the Exception and the warning for me in stdout. It appears to be working for me.

Gamenot (approved):

PR #997, opened 2021-07-07
Deterministic route computation

Adaickalavan:
Previous use of
```
next_via_lane_ids = set(conn.getViaLaneID() for conn in via_edge.getOutgoing()[end_edge])
via_lane_id = list(next_via_lane_ids)[0]
```
resulted in random selection of via lane id from the unordered `set`. This caused non-deterministic waypoint observation.

This is fixed by using sorted unique array. When multiple via lane ids are present, the first via lane id from the sorted unique array is used.


sah-huawei (approved):

Gamenot (approved):
Please merge when you can

Issue #455, opened 2021-01-20
Minimum Framerate Test


Gamenot:
- [x] Framerate tests to ensure minimum of 30 fps
- [x] 5 SV and 1 EV
- [x] Display fps for comparison
- [x] On low end machine

PR #1002, opened 2021-07-08
Added a minimum Framerate test

RutvikGupta:
Closes Issue #455.


smarts/core/tests/test_env_frame_rate.py
```diff
@@ -0,0 +1,112 @@
...
+                with FrameMonitor(40):
+                    observations, rewards, dones, infos = env.step(
+                        {AGENT_ID: agent_action}
+                    )
+            except Exception as e:
+                runtime_warnings += 1
+                pass
+            episode.record_step(observations, rewards, dones, infos)
+    env.close()
+    assert runtime_warnings > 0
```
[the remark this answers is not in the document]
RutvikGupta: I will add the logging part of each part of simulation step in another PR since that is a different issue. As for the log of each delta-time,  the minimum, average, and maximum step lengths, I will add them  into the logs of the test for the time being. as you mentioned above.

smarts/core/tests/test_smarts_frame_rate.py
```diff
@@ -0,0 +1,108 @@
...
+    yield smarts
+    smarts.destroy()
+
+
+def test_smarts_framerate(smarts, scenarios):
+    scenario = next(scenarios)
+    smarts.reset(scenario)
+
+    for _ in range(10):
+        with FrameMonitor(30):
```
sah-huawei, line 107: clever!

smarts/core/utils/frame_monitor.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def stop(self):
+        if self._start_time_ms is None:
+            raise FramerateException()
+
+        now = self._time_now()
+        delta = now - self._start_time_ms
+        actual_fps = round(1000 / delta, 2)
+        if round(1000 / delta, 2) > self._desired_fps:
+            raise FramerateException.above_threshold(self._desired_fps, delta)
```
sah-huawei, line 66: Is there a reason why we're throwing/raising for a failing test instead of asserting?

RutvikGupta: The idea was to catch it wherever it was used and then warn the user about the Exception.  `AssertionError` would have worked in the same way.

smarts/core/utils/frame_monitor.py
```diff
@@ -0,0 +1,67 @@
...
+        self._start_time_ms = self._time_now()
+
+    def stop(self):
+        if self._start_time_ms is None:
+            raise FramerateException()
+
+        now = self._time_now()
+        delta = now - self._start_time_ms
+        actual_fps = round(1000 / delta, 2)
+        if round(1000 / delta, 2) > self._desired_fps:
```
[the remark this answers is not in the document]
RutvikGupta: I thought "fps" here measures the time each smarts step is taking. So we want to check if it is lower than a certain threshold. Higher than that and we can comment about the experiment's performance. At least this is what I understood from the issue context.

smarts/core/utils/frame_monitor.py
sah-huawei: It's just that our other tests use asserts, so I think it should be consistent.

smarts/core/utils/frame_monitor.py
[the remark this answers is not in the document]
sah-huawei: My understanding is that "fps" stands for "frames _per second_", so it's the inverse of the amount of time taken per step.  Higher FPS rates mean that SMARTS is faster (which is good).  I think this issue is about making sure SMARTS is "fast enough", in other words that the FPS is _above_ a certain threshold.

smarts/core/tests/test_env_frame_rate.py
```diff
@@ -0,0 +1,117 @@
...
+            )
+
+            episode.record_step(observations, rewards, dones, infos)
+            step_counter += 1
+        if step_counter != 0:
+            avg_frame_rate = fps_sum / step_counter
+        test_logger.info(
+            f"Episode {episode_counter}, Minimum fps: {minimum_frame_rate}, Maximum fps: {maximum_frame_rate}, Average fps: {avg_frame_rate}."
+        )
+        assert (minimum_frame_rate >= 10) and (20 <= avg_frame_rate <= 30)
```
sah-huawei, line 116: I don't think we need (or want) the "`<= 30`" part of the assert above.

RutvikGupta: ok

smarts/core/utils/frame_monitor.py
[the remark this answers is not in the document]
RutvikGupta: Oh yes, my bad! I will invert the comparison. Thanks!

smarts/core/utils/frame_monitor.py
RutvikGupta: Ok, i will change it to AssertionError.

scenarios/tests/multi_agents_loop/requirements.txt
```diff
@@ -0,0 +1,3 @@
+--extra-index-url http://localhost:8080
+
+open-agent==0.1.195
```
Gamenot, line 3: I think we should avoid using `open-agent` in tests because of its cost.

smarts/core/tests/test_env_frame_rate.py
```diff
@@ -0,0 +1,117 @@
...
+            minimum_frame_rate = min(minimum_frame_rate, step_fps)
+            fps_sum += step_fps
+            test_logger.info(
+                f"The time delta at episode {episode_counter}, step {step_counter+1} is {delta} milliseconds which is {step_fps} fps."
+            )
+
+            episode.record_step(observations, rewards, dones, infos)
+            step_counter += 1
+        if step_counter != 0:
+            avg_frame_rate = fps_sum / step_counter
```
Gamenot, line 112: The following also works:

```suggestion
        avg_frame_rate = fps_sum / (step_counter or 1)
```

Gamenot (commented):
CHANGELOG entry is also needed.

smarts/core/utils/frame_monitor.py
```diff
@@ -0,0 +1,67 @@
...
+# THE SOFTWARE.
+
+import time
+
+
+class FramerateException(Exception):
+    """An exception raised if frame rate has increased higher than the desired threshold."""
+
+    def __str__(self):
+        return "The monitor has not started yet."
```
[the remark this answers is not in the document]
Gamenot: Reading the docs again with a fresh mind, it says that it will override the default stringification of the exception arguments.

Gamenot (approved):

PR #1004, opened 2021-07-12
Use one argument to specify scenarios in ULTRA env

christianjans:
Hi @JenishPatel99, I decided to make a separate PR for these changes. Let me know what you think.

The CI test I added doesn't really check much, but it is modelled after the existing CI test for the task-level scenario info (`tests/test_env.py::EnvTest::test_scenario_task`). I have made sure (by simply printing out the scenarios on line 65 of `ultra_env.py`) that the scenarios are being obtained correctly.

### DIFF
--- a/ultra/tests/test_env.py
+++ b/ultra/tests/test_env.py
@@ -44,6 +44,8 @@ timestep_sec = 0.1
 seed = 2
 task_id = "00"
 task_level = "easy"
+prebuilt_scenario1 = "tests/task/test_task00_easy_2lane_t_70kmh_p-test-flow-1"
+prebuilt_scenario2 = "tests/task/train_task00_easy_2lane_t_70kmh_p-test-flow-0"
 
 
 class EnvTest(unittest.TestCase):
@@ -60,6 +62,22 @@ class EnvTest(unittest.TestCase):
         ray.shutdown()
         self.assertTrue(task_id1 == task_id)
         self.assertTrue(task_level1 == task_level)
+
+    def test_scenario_sequence(self):
+        @ray.remote(max_calls=1, num_gpus=0, num_cpus=1)
+        def run_experiment():
+            _, env = prepare_test_env_agent(
+                scenario_info=[prebuilt_scenario1, prebuilt_scenario2]
+            )
+            scenarios = env.scenario_info
+            env.close()
+            return scenarios
+
+        ray.init(ignore_reinit_error=True)
+        scenarios = ray.get(run_experiment.remote())
+        ray.shutdown()
+        self.assertTrue(scenarios[0] == prebuilt_scenario1)
+        self.assertTrue(scenarios[1] == prebuilt_scenario2)
 
     def test_headless(self):
         @ray.remote(max_calls=1, num_gpus=0, num_cpus=1)
@@ -184,7 +202,7 @@ class TestLaneAgent(Agent):
         return "keep_lane"
 
 
-def prepare_test_env_agent(headless=True):
+def prepare_test_env_agent(scenario_info=("00", "easy"), headless=True):
     timestep_sec = 0.1
     # [throttle, brake, steering]
     policy_class = "ultra.baselines.ppo:ppo-v0"
@@ -192,7 +210,7 @@ def prepare_test_env_agent(headless=True):
     env = gym.make(
         "ultra.env:ultra-v0",
         agent_specs={AGENT_ID: spec},
-        scenario_info=("00", "easy"),
+        scenario_info=scenario_info,
         headless=headless,
         timestep_sec=timestep_sec,
         seed=seed,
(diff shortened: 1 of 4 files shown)


ultra/tests/test_env.py
```diff
@@ -61,6 +63,22 @@ def run_experiment():
         self.assertTrue(task_id1 == task_id)
         self.assertTrue(task_level1 == task_level)
 
+    def test_scenario_sequence(self):
+        @ray.remote(max_calls=1, num_gpus=0, num_cpus=1)
+        def run_experiment():
+            _, env = prepare_test_env_agent(
+                scenario_info=[prebuilt_scenario1, prebuilt_scenario2]
+            )
```
JenishPatel99, line 71: Do we need to initialize an environment to retrieve the scenarios? I am wondering that the `get_scenarios_from_scenario_info` is a static method, then why don't we just test that method without initializing an env. This way we can ensure that the `get_scenarios_from_scenario_info` works from  a static method POV

ultra/tests/test_env.py
```diff
@@ -61,6 +63,22 @@ def run_experiment():
...
+            scenarios = env.scenario_info
```
JenishPatel99, line 72: Is it necessary to include a getter() for the scenario_info?

ultra/ultra/env/ultra_env.py
```diff
@@ -148,41 +140,43 @@ def reset(self):
...
+            scenario_paths = tasks[f"task{task}"][level]
+
+            scenario_paths["train"] = os.path.join(base_dir, scenario_paths["train"])
+            scenario_paths["test"] = os.path.join(base_dir, scenario_paths["test"])
+
+            if not eval_mode:
+                scenarios = glob.glob(f"{scenario_paths['train']}")
+            else:
+                scenarios = glob.glob(f"{scenario_paths['test']}")
+        except Exception:
```
JenishPatel99, line 175: I believe we should identify the appropriate exception type (in the case it would an KeyError exception)

JenishPatel99 (commented):
I do like this approach, especially making the `get_scenarios()` a static method

ultra/tests/test_env.py
christianjans: We currently don't have a getter for the `scenario_info` in the ULTRA env. And I think it is fine if we continue to not have a getter as we only access the env's scenario info in this test as far as I'm aware. We can also obtain the `scenario_info` from the env's `info` attribute: https://github.com/huawei-noah/SMARTS/blob/2c1979e5b290c4267d8e3347e12ad4e2ff0727ef/ultra/ultra/env/ultra_env.py#L158-L164

If the env's `scenario_info` attribute is used elsewhere (other than the tests) then it might be good to make a getter (using the `@property` decorator).

Let me know your thoughts.

ultra/ultra/env/ultra_env.py
christianjans: Yes, a KeyError could occur in line 166, but also a ValueError could occur in line 160 if there are not exactly 2 elements in the `scenario_info` sequence. Are you thinking we should specify the specific exceptions in case we get an unexpected exception?

ultra/tests/test_env.py
christianjans: If we include the static method test, should we get rid of the `test_scenario_sequence` test?

ultra/ultra/env/ultra_env.py
JenishPatel99: Yeah, something like that. I think there are two cases here, one is that there can be a ValueError at line 160, however if there are exactly **2** scenario directories passed in, then we might also get a KeyError as well.  We can have something like this
on line 175

```python
except (ValueError, KeyError) as e:
```

ultra/tests/test_env.py
JenishPatel99: Hmm, I do not know if we do use the `scenario_info` object elsewhere. I am a bit of keen to implement getters/setters when it comes to retrieving/modify class members, but I believe it is fine for now. Yes, we can also use the `info` attribute to get the scenario info if we need it.

ultra/tests/test_env.py
JenishPatel99: Yeah, I do not think the `test_scenario_sequence` test is needed

ultra/ultra/env/ultra_env.py
```diff
@@ -148,41 +140,46 @@ def reset(self):
...
-        To obtain the path of scenarios:
-            1. Get the filename patterns of the train and test scenarios from
-               ultra/config.yaml from the given task and level.
-            2. Change the filename patterns from relative to absolute path
-            3. Depending on the evaluation mode, we can provide glob.glob with
-               the appropriate filename pattern to return a list of scenario dirs
+    @staticmethod
+    def get_scenarios_from_scenario_info(
+        scenario_info: Union[Tuple[str, str], Sequence[str]], eval_mode: bool
+    ) -> List[str]:
```
JenishPatel99, line 146: I was thinking that we should make the `eval_mode` a keyword argument because it is not necessary when there are explicit scenarios passed in. And also if we are planning to use it as a static method, it will be better to inform that the second argument is indeed the evaluation mode param

christianjans: Ahh yes, thanks, good call.

JenishPatel99 (approved):

PR #1006, opened 2021-07-14
Optimize image replay buffer storage in ULTRA

christianjans:
For a given Transition holding states in the form of stacked images, the replay buffer now stores only newest frame of the `next_state` element of the Transition.

```python
              TRANSITION                               TRANSITION                                 TRANSITION
      state                next_state             state           next_state       state                        next_state
[[f1, f2, f3, f4], ..., [f2, f3, f4, f5]] -> [[f1, f2, f3, f4], ..., [f5]] -> [[f1, f2, f3, f4], ..., concatenate(([f2, f3, f4], [f5]))]
     (BEFORE ENTERING REPLAY BUFFER)                (IN REPLAY BUFFER)                    (EXITING THE REPLAY BUFFER)
```


ultra/ultra/baselines/common/replay_buffer.py
```diff
@@ -188,6 +189,13 @@ def add(
 
         self._memory.append(new_experience)
 
+    def __getitem__(self, index):
+        if torch.is_tensor(index):
+            index = index.tolist()
+        state, action, reward, next_state, done, others = tuple(self._memory[index])
+        next_state = torch.cat((state[1:], next_state))  # Reattach the previous frames.
+        return state, action, reward, next_state, done, others
```
JenishPatel99, line 197: Hey @christianjans I was just wondering where do we use this method? I am trying to spot an value access from the replay buffer objects which will call the __getitem__()?

JenishPatel99 (approved):
Really like the idea 👍

Issue #661, opened 2021-03-09
Log SMARTS simulation step


Gamenot:
**Is your feature request related to a problem? Please describe.**
It is unclear at times where the simulation performance is taking a hit during development.

**Describe the solution you'd like**
I would like the addition of logging of sections of the scenario step to give an easy way to evaluate smarts performance problems.

**Describe alternatives you've considered**
A CI test that will evaluate performance automatically.

**Additional context**
Difficulty in finding the origin of a slow agent.act() call brought the need for more detailed performance logging up.

PR #1008, opened 2021-07-15
Log SMARTS simulation step to help evaluate smarts performance problems.

RutvikGupta:
Closes Issue #661.


smarts/core/smarts.py
```diff
@@ -205,35 +214,41 @@ def _step(self, agent_actions):
...
+                self._renderer.render()
 
-        response_for_ego = self._agent_manager.filter_response_for_ego(
-            (observations, rewards, scores, dones)
-        )
+        with timeit("calculating observations and rewards", self._log):
+            observations, rewards, scores, dones = self._agent_manager.observe(self)
+            response_for_ego = self._agent_manager.filter_response_for_ego(
+                (observations, rewards, scores, dones)
+            )
```
Gamenot, line 231: nit: This is not important but I could suggest that the returned values sit on the outer level.
```suggestion
        observations, rewards, scores, dones = None, None, None, None
        response_for_ego = None
        with timeit("calculating observations and rewards", self._log):
            observations, rewards, scores, dones = self._agent_manager.observe(self)
            response_for_ego = self._agent_manager.filter_response_for_ego(
                (observations, rewards, scores, dones)
            )
```

Gamenot (approved):
This can go in as is.

smarts/core/smarts.py
```diff
@@ -205,35 +214,44 @@ def _step(self, agent_actions):
...
+                self._renderer.render()
 
-        observations, rewards, scores, dones = self._agent_manager.observe(self)
+        observations, rewards, scores, dones = None, None, None, None
+        response_for_ego = None
 
-        response_for_ego = self._agent_manager.filter_response_for_ego(
-            (observations, rewards, scores, dones)
-        )
+        with timeit("calculating observations and rewards", self._log):
```
sah-huawei, line 230: All of the other messages were capitalized except this one??  Scandalous!!!  ;)

smarts/env/hiway_env.py
```diff
@@ -164,7 +165,11 @@ def step(self, agent_actions):
             agent_id: self._agent_specs[agent_id].action_adapter(action)
             for agent_id, action in agent_actions.items()
         }
-        observations, rewards, agent_dones, extras = self._smarts.step(agent_actions)
+        observations, rewards, agent_dones, extras = None, None, None, None
+        with timeit("SMARTS Simulation/Scenario Step", self._log):
+            observations, rewards, agent_dones, extras = self._smarts.step(
+                agent_actions
+            )
```
sah-huawei, line 172: I don't really have a huge problem with all of these (and I think I originally suggested the issue), but I'm also not a huge fan of what this does to our code readability.

Another option throughout would be to ensure that our standard log messages all include a timestamp (e.g., by using the `format` param to `logging.basicConfig()`) and then just add more log messages before and after operations like:
    self._log.info("stepping all providers...")
    self._provider_step()
    self._log.info("doing the next thing...")
    self._next_thing()
    self._log.info("etc.") 
These often can serve as (and replace) code comments too.

Anyway, I'm not saying that's necessarily better (b/c they require you to actually do math when you look at the log file to see how long things took), but at least it doesn't change variable scoping.

sah-huawei (commented):
as I wrote, I think there's another pattern that might not be as obtrusive, but I also don't feel too strongly either way.

smarts/env/hiway_env.py
RutvikGupta: I thought the idea here was to precisely measure how long each step will take here which the current changes allow use to by wrapping those steps in a `contextmanager`. The changes you are proposing may improve code readability but won't give precise measuring since each step takes milliseconds to finish.

sah-huawei: Well, if we want precision, we can profile it (and/or use flamegraph).  I originally suggested this issue (to Tucker, who then filed it) as a sort of heuristic "poor-man's profiling" to allow us to see delays in the logs without actually going all the way to profiling.  So it wasn't meant for precision originally, although it's not a bad idea.

I could go either way on this PR -- I certainly won't block it.  Maybe let's see what others think.

RutvikGupta: I just changed the logging to your suggestion above since it makes sense not to change variable scoping.

Adaickalavan (approved):

PR #1010, opened 2021-07-16
ULTRA Fix 5lane_t and 6lane_t map error

JenishPatel99:


christianjans (commented):
I just tried these new maps out and they are working great. There's no longer a problem with the 5 and 6 lane t-intersection scenarios!

I'm not sure why the diff between the analysis pool maps is so large, but I assume it's okay?

Additionally, I noticed that the `info.json` in the analysis pool's 5 and 6 lane t-intersection has been made empty? Is there any reason why this was done? I noticed that we currently have some stuff in these files on our master branch (https://github.com/huawei-noah/SMARTS/blob/master/ultra/ultra/scenarios/pool/analysis_pool/5lane_t/info.json, https://github.com/huawei-noah/SMARTS/blob/master/ultra/ultra/scenarios/pool/analysis_pool/6lane_t/info.json).

JenishPatel99:
> I just tried these new maps out and they are working great. There's no longer a problem with the 5 and 6 lane t-intersection scenarios!
> 
> I'm not sure why the diff between the analysis pool maps is so large, but I assume it's okay?
> 
> Additionally, I noticed that the `info.json` in the analysis pool's 5 and 6 lane t-intersection has been made empty? Is there any reason why this was done? I noticed that we currently have some stuff in these files on our master branch (https://github.com/huawei-noah/SMARTS/blob/master/ultra/ultra/scenarios/pool/analysis_pool/5lane_t/info.json, https://github.com/huawei-noah/SMARTS/blob/master/ultra/ultra/scenarios/pool/analysis_pool/6lane_t/info.json).

Hmm, that is weird, I checked inside the current branch, and the `info.json` are non-empty:
https://github.com/huawei-noah/SMARTS/blob/195504cf1e102eef5a8965bf53292bf84007869f/ultra/ultra/scenarios/pool/analysis_pool/5lane_t/info.json
https://github.com/huawei-noah/SMARTS/blob/195504cf1e102eef5a8965bf53292bf84007869f/ultra/ultra/scenarios/pool/analysis_pool/6lane_t/info.json

It may be git switching the files from one node to another? I am not exactly sure. I believe the analysis pool maps are larger because they include data about the extra lane at the bottom of the south lane

christianjans (approved):

PR #1012, opened 2021-07-19
ULTRA option to provide variable hijack time limit

JenishPatel99:
Provide variation to the ego vehicle hijack time limit by supporting tuple values

### DIFF
--- a/ultra/ultra/scenarios/generate_scenarios.py
+++ b/ultra/ultra/scenarios/generate_scenarios.py
@@ -437,7 +437,17 @@ def generate_left_turn_missions(
         speed_m_per_s = float("".join(filter(str.isdigit, speed))) * 5.0 / 18.0
         hijacking_params = route_distributions["ego_hijacking_params"]
         zone_range = hijacking_params["zone_range"]
-        waiting_time = hijacking_params["wait_to_hijack_limit_s"]
+
+        wait_to_hijack_limit_s = hijacking_params["wait_to_hijack_limit_s"]
+        waiting_time = (
+            np.random.randint(
+                wait_to_hijack_limit_s[0],
+                wait_to_hijack_limit_s[1],
+            )
+            if type(wait_to_hijack_limit_s) is tuple
+            else wait_to_hijack_limit_s
+        )
+
         start_time = (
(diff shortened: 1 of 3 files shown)


ultra/ultra/scenarios/generate_scenarios.py
```diff
@@ -437,7 +437,17 @@ def generate_left_turn_missions(
...
         zone_range = hijacking_params["zone_range"]
-        waiting_time = hijacking_params["wait_to_hijack_limit_s"]
+
+        wait_to_hijack_limit_s = hijacking_params["wait_to_hijack_limit_s"]
+        waiting_time = (
+            np.random.randint(
+                wait_to_hijack_limit_s[0],
+                wait_to_hijack_limit_s[1],
+            )
+            if type(wait_to_hijack_limit_s) is tuple
```
christianjans, line 447: Perhaps change to an `isinstance(...)` to be more inclusive?
```suggestion
            if isinstance(wait_to_hijack_limit_s, (list, tuple))
```
I noticed in `distributions.py` there are both lists and tuples for some of the values in the dictionaries.

ultra/ultra/scenarios/generate_scenarios.py
```diff
@@ -437,7 +437,17 @@ def generate_left_turn_missions(
         speed_m_per_s = float("".join(filter(str.isdigit, speed))) * 5.0 / 18.0
         hijacking_params = route_distributions["ego_hijacking_params"]
         zone_range = hijacking_params["zone_range"]
-        waiting_time = hijacking_params["wait_to_hijack_limit_s"]
+
```
christianjans, line 440: Wait sorry I might have missed why we need this change, is there something it could help with?

JenishPatel99: Actually, this change is not very significant in the grand scheme of generating scenarios. Currently, the only way that we can change the ego vehicle starting "time" is by altering the hijack limit inside the `distributions.py`.  This feature will allow users to set a range to when can the ego vehicle be hijacked. It is not necessary for this feature to be added.

christianjans (approved):

PR #1013, opened 2021-07-19
ULTRA decrease large amounts of vehicles from traffic distributions

JenishPatel99:
There were instances inside `distributions.py` where the number of vehicles for a single route was set to 1000. However, during a normal run with out baselines, we have never exhausted all 1000 of vehicles. I do feel it is unnecessary to have these many. This will also help decrease the scenario generation as we have less vehicles to construct and ID. Let me know your thoughts on this @christianjans

### DIFF
--- a/ultra/ultra/scenarios/common/distributions.py
+++ b/ultra/ultra/scenarios/common/distributions.py
@@ -146,7 +146,7 @@ t_patterns = {
             },
             "south-north": None,  # blocking
             "west-east": {
-                "vehicles": 1000,
+                "vehicles": 200,
                 "start_end_on_different_lanes_probability": 0.0,
                 "distribution": behavior_distribution,
                 "begin_time_init": {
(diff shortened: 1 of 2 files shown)


ultra/ultra/scenarios/common/distributions.py
```diff
@@ -146,7 +146,7 @@
             },
             "south-north": None,  # blocking
             "west-east": {
-                "vehicles": 1000,
+                "vehicles": 200,
```
christianjans, line 149: Interesting, okay good call. How did you decide on the number 200?

JenishPatel99: The 200 social vehicles is still large amount, however I sort of accounted for some worst case scenarios. It is a bit hard to explain here, but I can reduce the number to 100 even?

christianjans: Oh okay, I see. Even with the changes of #1014, the scenario will not run out of social vehicles?

JenishPatel99: Even with `endless_traffic` turned off, we have a sufficient amount of vehicles to entertain for more 200 secs (2000 timesteps)

christianjans (approved):

PR #1014, opened 2021-07-19
ULTRA set `endless_traffic` flag to false inside ULTRA

JenishPatel99:
The `endless_traffic` flag is used inside the `sumo_traffic_simulation.py` to control whether social vehicles are allowed to teleport back to their starting position once completed their mission route. Teleportation of the social vehicles is not necessary inside of ULTRA's environment because the scenarios that are generated by the traffic distribution does not account for that behavior. Thus, to disable this feature, we must pass `endless_traffic=False` to the parent constructor in `ultra_env.py`. I have opted to not make this feature optional in ULTRA because it goes against the core scenario setup, thus users will not be able to pass the `endless_traffic` flag

### DIFF
--- a/ultra/docs/ultra_environments.md
+++ b/ultra/docs/ultra_environments.md
@@ -18,8 +18,8 @@ of this class to still be a NumPy array, but with shape
 (its value can be found in `ultra.env.ultra_env`), `HEIGHT` is the height of the RGB
 image, and `WIDTH` is the width of the RGB image.
 4. ULTRA's Gym environment sets the `endless_traffic` flag to be False. This stops
-social vehicles from teleporting back to their starting position once completion of
-mission route. 
+social vehicles from teleporting back to their starting position once they complete
+their mission route.
 
 ULTRA's RLlib environment differs in one way from SMARTS' RLlibHiWayEnv:
 1. ULTRA's RLlib environment takes a `scenario_info` key as part of its config. This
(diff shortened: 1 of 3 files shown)


ultra/docs/ultra_environments.md
```diff
@@ -17,6 +17,9 @@ of this class to still be a NumPy array, but with shape
 `(_STACK_SIZE, HEIGHT, WIDTH, 3)` where `_STACK_SIZE` is the number of frames to stack
 (its value can be found in `ultra.env.ultra_env`), `HEIGHT` is the height of the RGB
 image, and `WIDTH` is the width of the RGB image.
+4. ULTRA's Gym environment sets the `endless_traffic` flag to be False. This stops
+social vehicles from teleporting back to their starting position once completion of
+mission route.
```
christianjans, line 22:
```suggestion
social vehicles from teleporting back to their starting position once they complete
their mission route.
```

ultra/ultra/scenarios/README.md
```diff
@@ -61,7 +61,7 @@
 3. Generate scenarios for training:
     ```sh
     # Remove existing easy scenarios.
-    $ rm -rf ultra/scenarios/task1/t*_easy_*
+    $ rm -rf ultra/scenarios/task1/t*_<task>_<level>_*
```
christianjans, line 64: I think this is meant to be specifically for Task 1's easy level.
```suggestion
    $ rm -rf ultra/scenarios/task1/t*_task1_easy_*
```

christianjans (approved):

PR #1018, opened 2021-07-22
Bugfix missing smarts egg info

Gamenot:
This is to fix an issue where mapping the repo as in the instructions to the /src directory would then further require a `smarts.egg-info` to be generated to use the repository code (which requires pip.)

### DIFF
--- a/utils/docker/Dockerfile
+++ b/utils/docker/Dockerfile
@@ -59,7 +59,8 @@ RUN pip install --no-cache-dir -r /tmp/requirements.txt
 ENV PYTHONPATH=/src
 COPY . /src
 WORKDIR /src
-RUN pip install --no-cache-dir -e .[train,test,dev,camera-obs]
+RUN pip install --no-cache-dir -e .[train,test,dev,camera-obs] \
+    && cp /src/smarts.egg-info /media/smarts.egg-info
 
 # For Envision
 EXPOSE 8081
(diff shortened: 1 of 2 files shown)


RutvikGupta:
Looks fine to me!

utils/docker/Dockerfile
```diff
@@ -59,7 +59,8 @@ RUN pip install --no-cache-dir -r /tmp/requirements.txt
 ENV PYTHONPATH=/src
 COPY . /src
 WORKDIR /src
-RUN pip install --no-cache-dir -e .[train,test,dev,camera-obs]
+RUN pip install --no-cache-dir -e .[train,test,dev,camera-obs] \
+    && cp /src/smarts.egg-info /media/smarts.egg-info
```
Adaickalavan, line 63:
```suggestion
    && cp -r /src/smarts.egg-info /media/smarts.egg-info
```

Adaickalavan (approved):

utils/docker/Dockerfile
```diff
@@ -59,7 +59,8 @@ RUN pip install --no-cache-dir -r /tmp/requirements.txt
 ENV PYTHONPATH=/src
 COPY . /src
 WORKDIR /src
-RUN pip install --no-cache-dir -e .[train,test,dev,camera-obs]
+RUN pip install --no-cache-dir -e .[train,test,dev,camera-obs] \
```
RutvikGupta, line 62: Does this line guarantees to create `/src/smarts.egg-info`because we immediately try to copy its content to another folder after this line?

RutvikGupta (approved):

utils/docker/Dockerfile
Gamenot: This `pip install -e .` is guaranteed to create `/src/smarts.egg-info`. The reason why it is being copied to a temporary location is because if you map in the SMARTS repository (e.g docker run -v $SMARTS_REPO:/src huaweinoah/smarts:v<major>.<minor>.<micro>) the SMARTS repository in blank state does not have `smarts.egg-info` which is used for python to understand the editable package.

Then in order to get the `smarts.egg-info` you would need to run `pip install -e .` inside the container which requires `PyPI` or redirect to another package index. That can cause an issue when working with proxies since `PyPI` access can be denied. This is just a hack to get around it since we generate this file before-hand anyway so it is not a problem to copy it back if it does not already exist as long as people are using the same micro version of editable source to match the Docker container.

PR #1020, opened 2021-07-22
ULTRA set `sumo_seed` to same as the scenario seed

JenishPatel99:
I am not quite sure why the `sumo_seed` was set to a random integer between 1 and 5, indicated below:
https://github.com/huawei-noah/SMARTS/blob/9fbf3d6195d7eb409fe0b2107155e00ed235b408/ultra/ultra/scenarios/generate_scenarios.py#L339

To ensure complete reproducibility of scenarios it is best to keep the `sumo_seed` same as the scenario seed.

Side task:
- [x] Remove unnecessary numpy import

### DIFF
--- a/ultra/ultra/scenarios/generate_scenarios.py
+++ b/ultra/ultra/scenarios/generate_scenarios.py
@@ -34,6 +34,7 @@ from multiprocessing import Manager, Process
 from shutil import copyfile
 from typing import Any, Dict, Sequence
 
+import numpy as np
 import yaml
 
 from smarts.core.utils.sumo import sumolib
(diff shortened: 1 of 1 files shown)


ultra/ultra/scenarios/generate_scenarios.py
```diff
@@ -346,7 +345,6 @@ def generate_left_turn_missions(
         }
 
     random.seed(seed)
-    np.random.seed(seed)
```
christianjans, line 349: Does removing this still allow for the scenarios to be created deterministically?

christianjans (commented):
This will be a great fix, thanks!

ultra/ultra/scenarios/generate_scenarios.py
JenishPatel99: Sorry, it does make a difference, thank you very much for questioning 😄  @christianjans

christianjans (approved):

PR #1027, opened 2021-07-23
Upgrade ULTRA's `setup.py`

christianjans:
Upgrade ULTRA's `setup.py` to use a SMARTS version that does not require X11 and can be used for the release of the next version.

TODO: Upgrade the SMARTS dependency to 0.4.19 once it is released (due to an inconsistency in SMARTS 0.4.18).

Will require a review from Jenish and Adai before merging.


christianjans:
@Adaickalavan, do you know what could be causing the problem in ULTRA's light Tune test? It seems it occurs when we are trying to instantiate the `RemoveAgentBuffer` for SMARTS' `AgentManager` (https://github.com/huawei-noah/SMARTS/pull/1027/checks?check_run_id=3162260242).

Adaickalavan:
Hi @richardliaw

+ The [test case](https://github.com/huawei-noah/SMARTS/blob/ultra_upgrade_setup/ultra/tests/test_tune.py), which makes use of the source [code](https://github.com/huawei-noah/SMARTS/blob/ultra_upgrade_setup/ultra/ultra/tune.py), fails.

+ We get an error message:
_"ValueError: signal only works in main thread"_

+ Raw log file of the failure can be obtained from [here](https://github.com/huawei-noah/SMARTS/pull/1027/checks?check_run_id=3162260242).

Do you know of any methods to overcome this?

richardliaw:
@Adaickalavan it looks like `signal.signal` is not supported in Ray. Can you try instead using an exit handler (atexit)?

ultra/setup.py
```diff
@@ -39,13 +35,13 @@
...
     long_description_content_type="text/markdown",
-    version="0.2.0",
+    version="0.2.1",
     packages=find_packages(exclude=["tests"]),
     include_package_data=True,
     zip_safe=True,
     python_requires=">=3.7",
     install_requires=[
-        "smarts[train,test]==0.4.14",
+        "smarts[train,test,camera-obs] @ git+https://git@github.com/huawei-noah/SMARTS@social-agents#egg=smarts",
```
JenishPatel99, line 44: The smarts version specified on branch `social-agents` is v0.4.18, are we still going for upgrading smarts dep. to v0.4.19 for ULTRA?

JenishPatel99 (commented):
Looks good 👍

ultra/setup.py
Adaickalavan: 1. We will eventually update ULTRA to depend on the latest SMARTS release, namely v0.4.19, when it becomes available. 
2. Using SMARTS==0.4.18 causes `./tests/test_tune.py` to fail due to Ray not supporting `signal.signal`.
3. `signal.signal` occurs within the `RemoteAgentBuffer` class inside SMARTS and it is always instantiated by default upon SMARTS startup.
4. I modified SMARTS such that the `RemoteAgentBuffer` is only instantiated when social agents are needed. See #1037.
5. Since ULTRA does not use social agents currently, `RemoteAgentBuffer` would not be instantiated, and thus the `signal.signal` would not hinder our usage of Ray's `tune`.
6. In order to pass the tests and to keep up with the latest SMARTS optimizations, ULTRA's `setup.py` installs SMARTS from the branch `social-agents`. 
7. Currently the branch `social-agents` = [SMARTS==0.4.18 + some additional changes].
8. In summary, this is a handy method for ULTRA to use the latest SMARTS code, while waiting for the next release.

JenishPatel99: Thank you @Adaickalavan for that information, it clearly make sense 👍 We might need to update ULTRA 's `setup.py` once SMARTS v0.4.19 (w/ `social-agents` branch updates) is released

PR #1028, opened 2021-07-25
Add Dockerfile that installs ULTRA and all its dependencies

christianjans:
The current Dockerfile requires users to install dependencies in the Docker container after the container is built. This new Dockerfile will install all dependencies when the Docker image is being created.

Requested to be added by @Adaickalavan.


JenishPatel99 (approved):
There might be some merge conflicts when merging #1028 and #1027, but otherwise looks good

christianjans:
I have updated the documentation for setting up ULTRA using Docker, but was unable to test out the commands due to the Huawei proxy on the dev box. Do the commands updated `setup.md` instructions look right @Adaickalavan?

If so, feel free to merge.

PR #1030, opened 2021-07-26
ROS wrapper fixes

sah-huawei:
A couple of necessary minor fixes revealed by integration testing.
One of them (the injection of `.sim`) should be temporary.

### DIFF
--- a/examples/ros/src/smarts_ros/scripts/ros_driver.py
+++ b/examples/ros/src/smarts_ros/scripts/ros_driver.py
@@ -474,8 +474,8 @@ class ROSDriver:
         agents = AgentsStamped()
         agents.header.stamp = rospy.Time.now()
         for agent_id, agent_obs in observations.items():
-            pose = Pose.from_center(veh_state.position, veh_state.heading)
             veh_state = agent_obs.ego_vehicle_state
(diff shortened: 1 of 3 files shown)


examples/ros/src/smarts_ros/scripts/ros_driver.py
```diff
@@ -502,7 +502,10 @@ def _do_agents(self, observations: Dict[str, Observation]) -> Dict[str, Any]:
             for agent_id, agent in self._agents_to_add.items():
                 spec, mission = agent[0], agent[1]
                 assert agent_id not in self._agents
-                self._agents[agent_id] = spec.build_agent()
+                agent = spec.build_agent()
+                # XXX: hack! in the future this injection must be removed or done another way...
+                agent.sim = self._smarts
```
Gamenot, line 507: Honestly, looking at it again I would still hope to not put this here.

Gamenot (approved):

examples/ros/src/smarts_ros/scripts/ros_driver.py
sah-huawei: I agree! 
Let's try to think of a better way in the near future.

PR #1031, opened 2021-07-27
Fix scenario exclusion prefixes in ULTRA

christianjans:
For missions and bubbles, the exclusion prefixes that determine which vehicles to not hijack was passed in as a list. This caused a problem when we attempt to hash either a mission or a bubble. For example, when the SMARTS gym environment attempts to hash the missions in its `scenario_log` property: https://github.com/huawei-noah/SMARTS/blob/e3681e7a04b7f4cc448525900fe65b5f76415bcb/smarts/env/hiway_env.py#L138-L160 However, if the missions contained an exclusion prefix that was of the list type, it would throw `TypeError: unhashable type: 'list'`.

To fix this, we can provide the `exclusion_prefixes` as a tuple (this is the type that it should be according to the type hinting).


JenishPatel99 (approved):
Nice catch 👍

PR #1032, opened 2021-07-27
ULTRA Add missing ego parameters to intersection specific traffic distributions

JenishPatel99:
- Add missing ego parameters
- Fix comment inside `generate_scenarios.py`

### DIFF
--- a/ultra/ultra/scenarios/common/distributions.py
+++ b/ultra/ultra/scenarios/common/distributions.py
@@ -426,7 +426,12 @@ t_patterns = {
             "east-south": None,
             "east-north": None,  # blocking
             "east-west": None,
-        }
+        },
+        "ego_hijacking_params": {
+            "zone_range": [5, 10],
+            "wait_to_hijack_limit_s": 2,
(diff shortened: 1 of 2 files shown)


ultra/ultra/scenarios/common/distributions.py
```diff
@@ -1061,7 +1071,12 @@
                     "end": (150, 160),
                 },
             },
-        }
+        },
+        "ego_hijacking_params": {
+            "zone_range": [5, 10],
+            "wait_to_hijack_limit_s": 2,
+            "start_time": 3,  # any value or default for LANE_LENGTH / speed_m_per_s
```
christianjans, line 1078: I noticed the `start_time` is different for these additions than the existing `ego_hijacking_params` for other distributions, is this okay?

JenishPatel99: The `start_time` is a bit different for the intersection-specific scenarios. The other distribution do not require for the ego to spawn immediately thus `start_time` there is 10

christianjans (approved):

PR #1033, opened 2021-07-28
Mission start_pose support

sah-huawei:
Adding support for specifying a Mission `start_pose` as a `Pose` object instead of `start` as a `Start` object. 

This fixes a bug where a newly-added ROS agent's position was offset by half of the vehicle length.


sah-huawei:
Nevermind.  I changed my mind about this one.  Will do it a different way.

PR #1034, opened 2021-07-28
ROS agent mission start fix

sah-huawei:
Fixes ROS agent position being off by half car-length because Mission starts were assumed to be from the front-bumper.

Also fixed a bug with a bad reference to `self._log` in a static method that I saw along the way.

Also removed some unnecessary checks from the test node (since they were risky due to race conditions and since it's not being used primarily to "test" sending the initial agents). 

(replaces obsolete/cancelled PR #1033)

### DIFF
--- a/smarts/core/scenario.py
+++ b/smarts/core/scenario.py
@@ -29,7 +29,7 @@ from dataclasses import dataclass, field
 from functools import lru_cache
 from itertools import cycle, product
 from pathlib import Path
-from typing import Any, Dict, Sequence, Tuple
+from typing import Any, Dict, Optional, Sequence, Tuple
 
 import numpy as np
 from cached_property import cached_property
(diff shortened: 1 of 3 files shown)


smarts/core/scenario.py
```diff
@@ -52,12 +52,14 @@
 class Start:
     position: Tuple[int, int]
     heading: Heading
+    from_front_bumper: Optional[bool] = True
```
Gamenot, line 55: This seems overly specific to me since we are now explicit about the vehicle feature in relation to the "start line". Common formats I remember are: front bumper, center, center of gravity, and back axle. As is, this is an interface change that will require a future change.

I think to go through it will require a new issue at minimum.

Gamenot (approved):

smarts/core/scenario.py
sah-huawei: Yeah, I agree with this being less than ideal.  (And you can look at the PR that I originally opened and then rejected for an even more less-than-ideal solution!)  So I'm definitely open to suggestions as to alternatives you think are better!

However, re it being an interface change, I have a couple of thoughts related to that:
- it's technically not an interface change because it's optional (and defaults to True, which is what it was before)so no existing users will have to change their code.
- but even so, recall this is already for the `0.5` branch (which has significant interface changes)

sah-huawei: Do we want to just scrap the `Start` class altogether and switch `Mission` over to use `Pose` instead?

PR #1037, opened 2021-07-30
Instantiate social agent buffer only if needed

Adaickalavan:
Instantiate social agent buffer only if needed.

### DIFF
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -456,6 +456,10 @@ class AgentManager:
         self._social_agent_data_models[agent_id] = agent_model
 
     def start_social_agent(self, agent_id, social_agent, agent_model):
+        if not self._remote_agent_buffer:
+            self._remote_agent_buffer = RemoteAgentBuffer(
+                zoo_manager_addrs=self._zoo_addrs
+            )
         remote_agent = self._remote_agent_buffer.acquire_remote_agent()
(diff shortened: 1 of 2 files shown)


sah-huawei (approved):

smarts/core/agent_manager.py
```diff
@@ -319,7 +321,12 @@ def switch_initial_agents(self, agent_interfaces: Dict[str, AgentInterface]):
 
     def setup_agents(self, sim):
         self.init_ego_agents(sim)
-        self.setup_social_agents(sim)
+        if sim.scenario.social_agents:
+            if not self._remote_agent_buffer:
+                self._remote_agent_buffer = RemoteAgentBuffer(
+                    zoo_manager_addrs=self._zoo_addrs
+                )
+            self.setup_social_agents(sim)
```
Gamenot, line 329: I think the `RemoteAgentBuffer` instance creation should sit in the `setup_social_agents()` to eliminate this private call dependency for the public `setup_social_agents()`.

Gamenot (changes requested):
One change needed but definitely an improvement

smarts/core/agent_manager.py
Adaickalavan: The `RemoteAgentBuffer` class creation has been moved into `setup_social_agents()` function.

Gamenot (approved):

PR #1038, opened 2021-07-30
ULTRA Update done criteria

JenishPatel99:
To stop vehicles from sliding into the oncoming traffic lanes, the `wrong_way` criteria should be enforced. This will terminate an episode if the vehicle goes into another lane. I do believe this necessary for our continuous policies to properly complete a scenario without driving in the opposite lane @christianjans Let me know your thoughts on this one

### DIFF
--- a/ultra/ultra/baselines/agent_spec.py
+++ b/ultra/ultra/baselines/agent_spec.py
@@ -24,7 +24,7 @@ import numpy as np
 import torch, yaml, os, inspect, dill
 
 from smarts.core.agent import AgentSpec
-from smarts.core.agent_interface import AgentInterface
+from smarts.core.agent_interface import AgentInterface, DoneCriteria
 from ultra.baselines.common.yaml_loader import load_yaml
 import ultra.adapters as adapters
 
@@ -111,6 +111,7 @@ class BaselineAgentSpec(AgentSpec):
                 interface=AgentInterface(
                     **adapter_interface_requirements,
                     max_episode_steps=max_episode_steps,
+                    done_criteria=DoneCriteria(wrong_way=True)
                     debug=True,
                 ),
                 agent_params=dict(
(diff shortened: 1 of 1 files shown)


christianjans:
I think the `wrong_way` actually terminates the episode if the agent is moving in the wrong direction (backwards from the direction of the mission I assume?). https://github.com/huawei-noah/SMARTS/blob/e3681e7a04b7f4cc448525900fe65b5f76415bcb/smarts/core/agent_interface.py#L181-L184 This can happen even if the agent is in the correct lane.

I'm not sure how this would impact and/or restrict our baselines. I don't think I have ever seen any drive in the wrong direction before.

JenishPatel99:
> I think the `wrong_way` actually terminates the episode if the agent is moving in the wrong direction (backwards from the direction of the mission I assume?).
> 
> https://github.com/huawei-noah/SMARTS/blob/e3681e7a04b7f4cc448525900fe65b5f76415bcb/smarts/core/agent_interface.py#L181-L184
> 
> This can happen even if the agent is in the correct lane.
> I'm not sure how this would impact and/or restrict our baselines. I don't think I have ever seen any drive in the wrong direction before.

Yeah, you are correct. For some reason, when I first checked scenarios on envision with wrong_way=True, it seems to terminate when ego was in opposite lanes. However, I checked it again, and now the wrong_way is actually terminating scenarios at intersections where it seems that the ego vehicle is doing nothing wrong. 

However, I am still kind of puzzled why does the ego vehicle go off the mission route  (i.e. in the west lane post turn) but no termination is registered. However, I will close this pull request.

christianjans:
> However, I am still kind of puzzled why does the ego vehicle go off the mission route (i.e. in the west lane post turn) but no termination is registered. However, I will close this pull request.

Yes I agree, I would have thought that the `off_route` criteria would cover this problem: https://github.com/huawei-noah/SMARTS/blob/e3681e7a04b7f4cc448525900fe65b5f76415bcb/smarts/core/agent_interface.py#L173-L174

Issue #651, opened 2021-03-08
Simplify ULTRA reward function


AlexLewandowski:
Current ULTRA reward function has many different components. Should try to simplify as much as possible.

PR #1039, opened 2021-07-30
Update default reward documentation in ULTRA

christianjans:
Forgot to mention this in #651.

### DIFF
--- a/ultra/docs/adapters.md
+++ b/ultra/docs/adapters.md
@@ -277,27 +277,24 @@ custom_reward = (
     ego_speed_reward +
     ego_distance_from_center_reward +
     ego_angle_error_reward +
-    ego_reached_goal_reward +
     ego_step_reward +
     environment_reward
 )
 ```
 
 Where:
-- `ego_goal_reward` is `0.0`
+- `ego_goal_reward` is `1.0` if the ego vehicle has reached its goal, else `0.0`
 - `ego_collison_reward` is `-1.0` if the ego vehicle has collided, else `0.0`.
 - `ego_off_road_reward` is `-1.0` if the ego vehicle is off the road, else `0.0`.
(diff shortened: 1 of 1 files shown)


ultra/docs/adapters.md
```diff
@@ -277,33 +277,30 @@ custom_reward = (
...
     ego_angle_error_reward +
-    ego_reached_goal_reward +
     ego_step_reward +
     environment_reward
 )
 ```
 
 Where:
-- `ego_goal_reward` is `0.0`
+- `ego_goal_reward` is `1.0` if the ego vehicle has reached its goal, else `0.0`
```
JenishPatel99, line 286: Would this be the `ego_reached_goal` as we have it in the `default_reward_adapter.py`
https://github.com/huawei-noah/SMARTS/blob/ecbc5cb2b90a4063777ed017331a07e406e3c717/ultra/ultra/adapters/default_reward_adapter.py#L134-L152

JenishPatel99 (commented):
Sorry, @christianjans I should have updated the documentation as well when I reshaped the reward. Thank you 👍

ultra/docs/adapters.md
christianjans: Do you mean `ego_reached_goal_reward`? Yeah, in the documentation, the names of some of the reward components are different than those that appear in the adapter in order to provide more clarity (e.g. `ego_distance_from_center_reward` instead of `ego_dist_center_reward`, `environment_reward` instead of `env_reward`). I think readers of these docs should be able to deduce what each one refers to even if the names are not the exact same?

JenishPatel99: Ohh I understand. That is a good idea 👍

JenishPatel99 (approved):

PR #1040, opened 2021-07-31
fix ROS dt issues

sah-huawei:
This came out of plotting the velocity received by the chassis in update_state with the position delta divided by `dt` (as discussed in our meeting earlier).

I discovered that the `dt` for the `ProviderState` object wasn't being merged.

Along the way, I (very slightly) improved the accuracy of the `dt` computation in `ros_driver.py` (and added the moving-average code back in), fixed a race condition at startup, as well as fixed another minor bug where the `updated` state wasn't being set correctly for Agent-controlled vehicles.

### DIFF
--- a/examples/ros/src/smarts_ros/scripts/ros_driver.py
+++ b/examples/ros/src/smarts_ros/scripts/ros_driver.py
@@ -370,7 +370,7 @@ class ROSDriver:
         lin_acc_slope = (vs.linear_acceleration - prev_lin_acc) / prev_dt
         ang_acc_slope = (vs.angular_acceleration - prev_ang_acc) / prev_dt
 
-        # The following 4 lines are a hack b/c I'm too stupid to figure out
+        # The following ~10 lines are a hack b/c I'm too stupid to figure out
         # how to do calculus on quaternions...
         heading = vs.pose.heading
         heading_delta_vec = staleness * (
@@ -389,7 +389,7 @@ class ROSDriver:
         vs.pose.position += staleness * (
             vs.linear_velocity
             + 0.5 * vs.linear_acceleration * staleness
-            + lin_acc_slope * staleness * staleness / 6.0
+            + lin_acc_slope * staleness ** 2 / 6.0
         )
 
         vs.linear_velocity += staleness * (
@@ -402,7 +402,7 @@ class ROSDriver:
         vs.linear_acceleration += staleness * lin_acc_slope
         vs.angular_acceleration += staleness * ang_acc_slope
 
-    def _update_smarts_state(self, step_delta: float) -> bool:
+    def _update_smarts_state(self) -> float:
         with self._state_lock:
             if (
                 not self._recent_state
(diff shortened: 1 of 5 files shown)


smarts/core/provider.py
```diff
@@ -37,6 +37,7 @@ def merge(self, other: "ProviderState"):
         assert our_vehicles.isdisjoint(other_vehicles)
 
         self.vehicles += other.vehicles
+        self.dt = max(self.dt, other.dt, key=lambda x: x if x else 0)
```
sah-huawei, line 40: This is the key bug fix here.  When this was missing, line 591 of smarts.py did not function as intended.

smarts/core/smarts.py
```diff
@@ -608,7 +608,6 @@ def _pybullet_provider_sync(self, provider_state: ProviderState):
                         pybullet_vehicle = self._vehicle_index.vehicle_by_id(vehicle_id)
                         assert isinstance(pybullet_vehicle.chassis, BoxChassis)
                         pybullet_vehicle.update_state(vehicle, dt=dt)
-                        pybullet_vehicle.updated = True
```
sah-huawei, line 611: This line was also a (relatively-minor) bug.  It should have been `vehicle.updated = True`.

smarts/core/vehicle.py
```diff
@@ -538,6 +538,7 @@ def control(self, *args, **kwargs):
         self._chassis.control(*args, **kwargs)
 
     def update_state(self, state: VehicleState, dt: float):
+        state.updated = True
```
sah-huawei, line 541: This is the fix for the above bug.

examples/ros/src/smarts_ros/scripts/ros_driver.py
```diff
@@ -348,29 +345,80 @@ def _entity_to_vs(entity: EntityState) -> VehicleState:
...
+            )
+
+        @property
+        def stamp(self):
+            return self.vector[0]
+
+        def average_with(self, other_vect: np.ndarray):
+            self.vector += other_vect
+            self.vector /= 2
+            self.update_vehicle_state()
```
Gamenot, line 371: nit: From what I can see `vector`, `update_vehicle_state`, `average_with`, should be private members.

Gamenot (approved):
This looks good to merge!

smarts/core/provider.py
sah-huawei: Yes.  (Because either argument may be None, so the code got ugly using `if`-statements.)

examples/ros/src/smarts_ros/scripts/ros_driver.py
sah-huawei: The sub-class containing those three methods (`_VehicleStateVector`) is already marked as private, and these 3 things are being accessed by the "external" parent/owner class code (not just by methods in the subclass).  In such cases, I usually leave them public to denote that the owner class can access them too.

PR #1041, opened 2021-08-01
re-apply control forces with each step of pybullet

sah-huawei:
As we discovered in testing today, when we break the pybullet step into substeps (for non-fixed time-stepping), it appears to be necessary to reapply the control forces to the vehicle with each pybullet substep.

### DIFF
--- a/smarts/core/chassis.py
+++ b/smarts/core/chassis.py
@@ -79,6 +79,9 @@ class Chassis:
     def control(self, *args, **kwargs):
         raise NotImplementedError
 
+    def reapply_last_control(self):
+        raise NotImplementedError
+
     def teardown(self):
         raise NotImplementedError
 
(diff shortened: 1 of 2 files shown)


smarts/core/chassis.py
```diff
@@ -631,6 +639,10 @@ def control(self, throttle=0, brake=0, steering=0):
         self._apply_throttle(throttle_list)
         self._apply_brake(brake)
 
+    def reapply_last_control(self):
+        assert self._last_control
+        self.control(*self._last_control)
```
Gamenot, line 644: I remember the cost of a `pybullet` call can be fairly high it looks like there is no way around it.

Gamenot (approved):

smarts/core/chassis.py
sah-huawei: Yeah, but I haven't been able to think of another way around this yet unfortunately.  I also think once everything is functional and stable, we should probably do another profiling pass.

PR #1042, opened 2021-08-03
ULTRA Modify `max_episode_steps` value's origin during evaluation

JenishPatel99:
Instead of getting the value of the `max_episode_steps` from the agent_spec in the `agent_metadata.pkl` (i.e. in experiment directory), use the `max_episode_step` argument which is passed via the instance creator (i.e.` __new__()`) in `agent_spec.py`

Side tasks:
- [ ] Change default `max_episode_steps` value in `evaluate.py` from 10000 to 1200


ultra/docs/getting_started.md
```diff
@@ -80,7 +80,7 @@ After training your agent, your models should be saved under `logs/<timestamped_
   - `--level`: The level of the task (default is easy).
   - `--agents`: The ids of agents that you wish to evaluate (default is None; all agents are subject to evaluation).
   - `--episodes`: The number of evaluation episodes (default is 200).
-  - `--max-episode-steps`: The option to limit the number of steps per epsiodes (default is 200).
+  - `--max-episode-steps`: The option to limit the number of steps per epsiodes (default is 1200).
```
christianjans, line 83: I think this also has to be changed in line 54?

ultra/tests/test_evaluate.py
```diff
@@ -385,6 +385,25 @@ def extract(path):
...
+                max_episode_steps=_MAX_EPISODE_STEPS,
+                experiment_dir=experiment_dir,
+                agent_id=_AGENT_ID,
+            )
+        }
+
+        self.assertEqual(
+            agent_spec[_AGENT_ID].interface.max_episode_steps, _MAX_EPISODE_STEPS
+        )
+
```
christianjans, line 406: 👍👍 awesome.

ultra/docs/getting_started.md
JenishPatel99: Oh yeah, perfect thank you @christianjans

christianjans (approved):

PR #1043, opened 2021-08-03
trap vehicle check

sah-huawei:
check all vehicles' (not just agent vehicles') positions when a trap will introduce a new vehicle

### DIFF
--- a/smarts/core/trap_manager.py
+++ b/smarts/core/trap_manager.py
@@ -29,7 +29,7 @@ from shapely.geometry import Point, Polygon
 from smarts.core.mission_planner import Mission, MissionPlanner
 from smarts.core.scenario import Start, default_entry_tactic
 from smarts.core.utils.math import clip, squared_dist
-from smarts.core.vehicle import VehicleState
+from smarts.core.vehicle import VehicleState, VEHICLE_CONFIGS
 from smarts.sstudio.types import MapZone, TrapEntryTactic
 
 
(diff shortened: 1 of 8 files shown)


smarts/core/trap_manager.py
```diff
@@ -193,15 +188,18 @@ def largest_vehicle_plane_dimension(vehicle):
...
+                    vehicle_comp.sort(
                         key=lambda v: squared_dist(v[0], mission.start.position)
                     )
 
-                    # Make sure there is not an agent vehicle in the same location
-                    pos, largest_dimension, _ = agent_vehicle_comp[0]
-                    if squared_dist(pos, mission.start.position) < largest_dimension:
-                        continue
+                    # Make sure there is not a vehicle in the same location
+                    for pos, largest_dimension, _ in vehicle_comp[:3]:
```
Gamenot, line 197: There was a minor assumption that checking against the nearest 3 vehicles would be enough to see if any are intruding. A higher number (e.g. 4-6) is probably better if we are preempting issues with small vehicles like bikes.

Gamenot (approved):

smarts/core/trap_manager.py
sah-huawei: Alternatively, I don't think the performance hit would be that bad to just check them all.  What do you think?

Gamenot (approved):
This looks reasonable.

smarts/core/trap_manager.py
sah-huawei: as discussed offline, just changed it to check all of them.  also, fixed a bug here with the continue scope!!

smarts/core/trap_manager.py
```diff
@@ -192,19 +187,18 @@ def largest_vehicle_plane_dimension(vehicle):
...
-                if len(agent_vehicle_comp) > 0:
-                    agent_vehicle_comp.sort(
-                        key=lambda v: squared_dist(v[0], mission.start.position)
-                    )
-
-                    # Make sure there is not an agent vehicle in the same location
-                    pos, largest_dimension, _ = agent_vehicle_comp[0]
+                overlapping = False
+                for pos, largest_dimension, _ in vehicle_comp:
                     if squared_dist(pos, mission.start.position) < largest_dimension:
```
Gamenot, line 194: Truthfully, this should be more like:
```suggestion
                    if squared_dist(pos, mission.start.position) < (largest_dimension+<new_vehicle_dimensions>) ** 2:
```

RutvikGupta: What should be that dimension? The length of the new vehicle?

sah-huawei: Yeah, which isn't really completely specified at this point (since this is where we _decide_ whether to add it).  However, I just pushed an update which uses something similar to what's done in `build_agent_vehicle()` in `vehicle.py` to do this.

Gamenot (approved):
That looks to be more accurate

RutvikGupta (approved):

PR #1047, opened 2021-08-04
Add dt to Observation class

sah-huawei:
As the title says.  This will be useful to external agents when running with a non-fixed timestep.

This is marked as a breaking changes since it's an interface change.  So it is targetted for the v0.5 release with the other timestep-related stuff.


Gamenot (approved):
Let's get this in!

Issue #877, opened 2021-05-25
Display episode/scenario info on envision output


JenishPatel99:
**Is your feature request related to a problem? Please describe.**
Running an ULTRA experiment (Headless is False) we can visually observe the agent and its behavior (speed, heading, position) through Envision. However, when we view the agent on Envision we are not sure about which episode index ("episode number") the agent is on.

**Describe the solution you'd like**
Is there is an Envision API that allows us to display the episode index and/or scenario information that the agent is currently training/testing on? 

![image](https://user-images.githubusercontent.com/55423380/119575339-246e5a80-bd74-11eb-88ae-9939a84d47da.png)

The above image is something what we were thinking about

Gamenot:
This is a useful visualisation, we can add it to the next version I believe.

JenishPatel99:
> This is a useful visualisation, we can add it to the next version I believe.

@Gamenot Sweet, thank you Tucker. 

Sorry for being a bit greedy 😄 but is there a way to patch the agent details (speed, heading, episode #) onto the frames of the generated MP4 file, If not it is all good.

PR #1049, opened 2021-08-05
Display scenario/map name on Envision output

JenishPatel99:
Addressing issue: #877 

Add the scenario or map name on each frame (state) during simulation, so it can then be viewed on Envision. 

NOTE:
I have included an updated web distribution with these new changes, so there is no need to build a new dist/ using npm as suggested in the [https://github.com/huawei-noah/SMARTS/blob/master/envision/README.md](url)

Few questions:
- Is this a valid solution to implement?
- Can this be included in the next smarts release?
- Is there way to display the scenario name when the application starts (like the score is displayed), instead of manually checking the option on the control panel.


smarts/core/smarts.py
```diff
@@ -1053,9 +1053,15 @@ def _try_emit_envision_state(self, provider_state, obs, scores):
             for bubble in self._bubble_manager.bubbles
         ]
 
+        scenario_name = {
+            agent_id: (self.scenario._root).split("/")[-1]
```
Adaickalavan, line 1057: A path split library could be used instead of using `.split("/")[-1]`.

JenishPatel99: Appropriate changes implemented. Thanks 👍

smarts/core/smarts.py
```diff
@@ -1053,9 +1053,15 @@ def _try_emit_envision_state(self, provider_state, obs, scores):
             for bubble in self._bubble_manager.bubbles
         ]
 
+        scenario_name = {
+            agent_id: os.path.split(self.scenario._root)[1]
+            for agent_id in list(self._agent_manager.ego_agent_ids)
+        }
+
```
christianjans, line 1060: I wonder if this could somehow be just a string since the scenario is the same for all agents. I wouldn't know how to handle this on the JS side though. Other than that it looks good!

christianjans (approved):

smarts/core/smarts.py
JenishPatel99: The file changes do address this issue. If you run the simulation with the new web dist/ then you will see only the scenario name and not an agent_id tag beside it

Adaickalavan (approved):

PR #1050, opened 2021-08-06
Add appendix to adapter docs in ULTRA

christianjans:
Add an appendix to the adapter documentation in ULTRA to provide more information for users looking to use the current adapters, or to create their own adapters.


JenishPatel99 (approved):
Looks great 👍

PR #1053, opened 2021-08-09
ULTRA remove double instances of evaluation check inside `train.py`

JenishPatel99:
When merging the commit 32de433, the evaluation check function call was duplicated twice inside the train.py. This PR will remove the duplicated function call


christianjans (approved):
Good catch 👍

PR #1055, opened 2021-08-10
ULTRA Add wheel installation step in setup doc

JenishPatel99:
It seems like we do not include the wheel installation step in our setup docs. This is needed for ULTRA deps, and it is not provided through smarts installation.


christianjans:
Oh interesting, thanks for finding this. I recall though that `pip install -e .` had worked before, but now it does not unless we install the wheel package?

christianjans (approved):

Issue #1045, opened 2021-08-04
TypeError: 'NoneType' object is not callable


Adaickalavan:
## **BUG REPORT**
**High Level Description**
Following the instructions in `getting_started.md` to run ULTRA, produces `TypeError: 'NoneType' object is not callable` message at the end of the training.

**SMARTS version**
0.4.18

**Steps to reproduce the bug**
```bash
$ cd <path/to/SMARTS>/SMARTS/ultra
$ scl scenario build-all ultra/scenarios/pool/experiment_pool/
$ python ultra/scenarios/interface.py generate --task 1 --level easy
$ python ultra/train.py --task 1 --level easy --episodes 10 --eval-episodes 2 --eval-rate 5 --policy dqn
```

**Resulting and expected behaviour**
Preferably no error messages at the end of training.

**Error logs and screenshots**
```bash
/usr/lib/python3.7/subprocess.py:883: ResourceWarning: subprocess 11322 is still running
  ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
Exception ignored in: <function SMARTS.__del__ at 0x7f48b83b9830>
Traceback (most recent call last):
  File "/home/workspaces/SMARTS/.venv/lib/python3.7/site-packages/smarts/core/smarts.py", line 466, in __del__
  File "/home/workspaces/SMARTS/.venv/lib/python3.7/site-packages/smarts/core/smarts.py", line 444, in destroy
  File "/home/workspaces/SMARTS/.venv/lib/python3.7/site-packages/smarts/core/smarts.py", line 423, in teardown
  File "/home/workspaces/SMARTS/.venv/lib/python3.7/site-packages/smarts/core/utils/cache.py", line 130, in wrapper
  File "/home/workspaces/SMARTS/.venv/lib/python3.7/site-packages/smarts/core/vehicle_index.py", line 314, in teardown
  File "/home/workspaces/SMARTS/.venv/lib/python3.7/site-packages/smarts/core/vehicle_index.py", line 696, in _build_empty_controlled_by
TypeError: 'NoneType' object is not callable
```
 
**System information**
 
- Operating system & Environment [Please note if using Docker]: Ubuntu 16.04
- Python version [Please note if using Jupyter]: 3.7
- SUMO version: 1.9.1

PR #1056, opened 2021-08-10
Remove Smarts __del__()

Adaickalavan:
Fixes issue #1045 .

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -481,9 +481,6 @@ class SMARTS:
             self._bullet_client.disconnect()
             self._bullet_client = None
 
-    def __del__(self):
-        self.destroy()
-
     def _teardown_vehicles(self, vehicle_ids):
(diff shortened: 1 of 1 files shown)


Adaickalavan:
There were two parts to the error in #1045 
(i) ResourceWarning: subprocess 11322 is still running = Addressed by #1058 
(ii) TypeError: 'NoneType' object is not callable - Addressed by #1056

Gamenot:
> There were two parts to the error in #1045
> (i) ResourceWarning: subprocess 11322 is still running - Addressed by #1058
> (ii) TypeError: 'NoneType' object is not callable - Addressed by #1056

Is this occurring because of the order which modules are being deleted? This seems odd to me. I am nervous about merging this without having a better grasp of the reason why it fails. I think the immediate tradeoff of merging this is that some resources may not clean up unless the user of a `SMARTS` object calls `destroy()` explicitly.

Gamenot:
I have found another way of fixing the issue. I am closing this in favour of #1190.

PR #1057, opened 2021-08-10
ULTRA WSL support

JenishPatel99:
ULTRA can support Windows subsystem for Linux (WSL) on the condition that there are no social-agents implemented (i.e. no instances of RemoteBufferAgent)

Both versions of WSL (i.e. WSL 1 and WSL 2) do run ULTRA.

WSL supports many Linux distributions, however for ULTRA, only Ubuntu 18.04 be supported. The `install_deps` script needs to be modify to include seperate installation for WSL

### DIFF
--- a/ultra/install_deps.sh
+++ b/ultra/install_deps.sh
@@ -8,6 +8,13 @@ function check_python_version_gte_3_7 {
     || hash python3.9 2>/dev/null;
 }
 
+function check_python_version_gte_3_8 {
+    echo "Checking for >=python3.8"
+    # running through current minor verions
+    hash python3.8 2>/dev/null \
+    || hash python3.9 2>/dev/null;
+}
+
 function do_install_for_linux {
     echo "Installing sumo (used for traffic simulation and road network)"
     sudo add-apt-repository ppa:sumo/stable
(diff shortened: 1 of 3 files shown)


ultra/install_deps.sh
```diff
@@ -74,9 +81,95 @@ function do_install_for_macos {
...
+    echo ""
+}
+
 if [[ "$OSTYPE" == "linux-gnu" ]]; then
     echo "Detected Linux"
-    do_install_for_linux
+    # If kernel release is Microsoft-based then proceed with WSL
+    # installation otherwise native linux installation
+    DISTRIB=$(awk -F= '/^NAME/{print $2}' /etc/os-release)
+    if [[ $DISTRIB == '"Ubuntu"' ]]; then
```
christianjans, line 158: I wonder if this if statement should be in `do_install_for_WSL`? I think the way it is currently, a non-ubuntu distribution cannot natively install ULTRA (previously we didn't require the Linux distribution to be Ubuntu).

ultra/install_deps.sh
```diff
@@ -74,9 +81,90 @@ function do_install_for_macos {
     fi
 }
 
+function do_install_for_WSL {
+    # We currently only support Ubuntu distribution for WSL (Windows subsystem
+    # for Linux). The main difference between WSL and native installation is that
+    # WSL does not support python3.7, instead python3.8 and its dependencies will be
+    # installed
```
christianjans, line 88: Sorry just saw this, is Python 3.8 an issue for 18.04 and 20.04 on WSL? It seems we are installing 3.7 on 18.04 for WSL.

christianjans (approved):

ultra/install_deps.sh
christianjans: Resolved offline.

PR #1059, opened 2021-08-13
ground collisions clarification

sah-huawei:
Earlier, I was confused and mistaken about there being a bug in `_query_bullet_contact_points()` as I was expecting it to also return contacts with the ground plane.  

Once I figured out that it does not, I added comments and tweaked `test_collisions.py` to clarify that ground collisions will never be included in `contact_points`.

### DIFF
--- a/smarts/core/chassis.py
+++ b/smarts/core/chassis.py
@@ -59,9 +59,13 @@ with open(controller_filepath, "r") as controller_file:
 def _query_bullet_contact_points(bullet_client, bullet_id, link_index):
     contact_objects = set()
 
-    min_, max_ = bullet_client.getAABB(bullet_id, link_index)
     # `getContactPoints` does not pick up collisions well so we cast a fast box check on the physics
+    min_, max_ = bullet_client.getAABB(bullet_id, link_index)
+    # note that getAABB returns a box around the link_index link only,
+    # which means it's offset from the ground (min_ has a positive z)
+    # if link_index=0 (the chassis link) is used.
     overlapping_objects = bullet_client.getOverlappingObjects(min_, max_)
+    # the pairs returned by getOverlappingObjects() appear to be in the form (body_id, link_idx)
     if overlapping_objects is not None:
         contact_objects = set(oo for oo, _ in overlapping_objects if oo != bullet_id)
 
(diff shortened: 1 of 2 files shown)


smarts/core/chassis.py
```diff
@@ -59,9 +59,13 @@
...
-    min_, max_ = bullet_client.getAABB(bullet_id, link_index)
     # `getContactPoints` does not pick up collisions well so we cast a fast box check on the physics
+    min_, max_ = bullet_client.getAABB(bullet_id, link_index)
+    # note that getAABB returns a box around the link_index link only,
+    # which means it's offset from the ground (min_ has a positive z)
+    # if link_index=0 (the chassis link) is used.
     overlapping_objects = bullet_client.getOverlappingObjects(min_, max_)
+    # the pairs returned by getOverlappingObjects() appear to be in the form (body_id, link_idx)
     if overlapping_objects is not None:
         contact_objects = set(oo for oo, _ in overlapping_objects if oo != bullet_id)
```
Gamenot, line 70: Is this still a problem? I think you were looking at doing the following:

```suggestion
        contact_objects = set(oo for _, oo in overlapping_objects if oo != bullet_id)
```

smarts/core/chassis.py
```diff
@@ -59,9 +59,13 @@
 def _query_bullet_contact_points(bullet_client, bullet_id, link_index):
     contact_objects = set()
 
-    min_, max_ = bullet_client.getAABB(bullet_id, link_index)
     # `getContactPoints` does not pick up collisions well so we cast a fast box check on the physics
+    min_, max_ = bullet_client.getAABB(bullet_id, link_index)
+    # note that getAABB returns a box around the link_index link only,
+    # which means it's offset from the ground (min_ has a positive z)
```
Gamenot, line 65: I see, that should mean that the only way the chassis hits the ground is if there is some serious loss of control or hitting inclined ground.

Gamenot: Perhaps this is a mistake of thinking that the wheels are part of the chassis. The wheels cause elevation of the actual chassis so it never touches the ground. I believe this current behavior might be correct.

smarts/core/chassis.py
sah-huawei: No, it turned out it wasn't a bug, I was just misunderstanding the thing that I wrote in the comment I added.  (That's why I just added the comment instead.)

smarts/core/chassis.py
sah-huawei: Yep, that's what I just did a variation of (on the `map_abstraction` branch, as discussed in chat).

PR #1060, opened 2021-08-16
Waymo-SMARTS replay simulation

RutvikGupta:
SMARTS support to import traffic histories from the Waymo motion dataset to replay in a SMARTS simulation.


.gitignore
```diff
@@ -107,6 +107,9 @@ bubbles.pkl
 missions.pkl
 traffic_histories.pkl
 **/social_agents/*
+*.pkl
```
Gamenot, line 110: I think I agree with this filter but if we add this then I think we can remove the more specific `.pkl` filters (e.g. `missions.pkl`.)

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -52,14 +52,15 @@ def act(self, obs: Observation) -> Tuple[float, float]:
...
         assert math.isclose(
-            cur_state.heading, exp["heading"], abs_tol=1e-09
+            cur_state.heading, exp["heading"], abs_tol=1e-2
         ), f'vid={self._vehicle_id}: {cur_state.heading} != {exp["heading"]} @ {obs_time}'
         # Note: the other checks can't be as tight b/c we lose some accuracy (due to angular acceleration)
         # by converting the acceleration vector to a scalar in the observation script,
         # which compounds over time throughout the simulation.
         assert math.isclose(
-            cur_state.speed, exp["speed"], abs_tol=0.1
+            cur_state.speed, exp["speed"], abs_tol=1
```
Gamenot, line 63: `abs_tol=1` means we can be a whole `m/s` off which is quite a bit since it compounds.

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -99,8 +100,13 @@ def main(
     assert traffic_history_provider
 
     scenarios_iterator = Scenario.scenario_variations(scenarios, [])
-    for scenario in scenarios_iterator:
+    total_scenario_variations = Scenario.discover_traffic_history_scenarios_count(
+        scenarios
+    )
+    for _ in range(total_scenario_variations):
+        scenario = next(scenarios_iterator)
```
Gamenot, line 107: This can just be shortened to using the iterator via `for`.
```suggestion
    for scenario in scenarios_iterator:
```

examples/observation_collection_for_imitation_learning.py
```diff
@@ -49,40 +49,42 @@ def main(scenarios: Sequence[str], headless: bool, seed: int):
...
+
+    total_scenario_variations = Scenario.discover_traffic_history_scenarios_count(
+        scenarios
     )
 
-    scenario = next(scenarios_iterator)
-    obs = smarts.reset(scenario)
+    scenarios_iterator = Scenario.scenario_variations(scenarios, [])
+    for _ in range(total_scenario_variations):
+        scenario = next(scenarios_iterator)
```
Gamenot, line 59: Again same can be done here by consuming the iterator with `for`.
```suggestion
    for scenario in scenarios_iterator:
```

smarts/core/scenario.py
```diff
@@ -115,15 +115,13 @@ class TraverseGoal(Goal):
     the vehicle must be going the correct direction in its lane
     just prior to doing so."""
 
-    def __init__(self, road_network: SumoRoadNetwork):
-        super().__init__()
-        self._road_network = road_network
+    road_network: SumoRoadNetwork
```
Gamenot, line 118: While I am slightly confused about having an initialiser in a frozen dataclass I do not think we should be exposing the `road_network` through this object.

smarts/core/traffic_history.py
```diff
@@ -54,9 +56,7 @@ def disconnect(self):
             self._db_cnxn.close()
             self._db_cnxn = None
 
-    def _query_val(
-        self, result_type: TypeVar["T"], query: str, params: Tuple = ()
-    ) -> T:
+    def _query_val(self, result_type: T, query: str, params: Tuple = ()) -> T:
```
Gamenot, line 59: Neat.

examples/history_vehicles_replacement_for_imitation_learning.py
RutvikGupta: `scenarios_iterator` is cyclic. If we iterate over it in a for loop, we will go into infinite loop. Thats why we had to workaround finding the total number of scenario variations.

examples/history_vehicles_replacement_for_imitation_learning.py
RutvikGupta: I think it should be 0.1. We must have missed that.

smarts/core/scenario.py
saulfield: I agree, what would be a better way? Keeping the initializer with the frozen attribute throws an error. Removing the frozen attribute breaks Python 3.8 since it inherits from a frozen dataclass. We could remove frozen on both this class and the `Goal` class, but that would break immutability and hashing (we would have to add `unsafe_hash=True`).

Since we just use the `road_network` for `drove_off_map()` perhaps we could just pass a closure that wraps that function to avoid passing the whole network?

smarts/core/scenario.py
```diff
@@ -364,6 +362,27 @@ def variations_for_all_scenario_roots(
...
+                scenario_roots.extend(Scenario.discover_scenarios(root))
+
+        total_traffic_history_variations = 0
+        for scenario_root in scenario_roots:
+            traffic_histories = Scenario.discover_traffic_histories(scenario_root) or [
+                None
+            ]
+            if traffic_histories != [None]:
+                total_traffic_history_variations += len(traffic_histories)
+        return total_traffic_history_variations
```
Gamenot, line 384: My problem with this method is that the information comes at non-trivial cost and then that cost must be incurred again in order to make use of the scenarios.

The exact number of scenarios is useful but the locations of those scenarios are also tied in for the majority of uses.

smarts/core/scenario.py
Gamenot: Yes, I think the problem here I am seeing is that we are not using a `dataclass` not as just data. The `road_network` is not serialisable and I would be somewhat wary of providing any reference value to a `dataclass`. The solution I can think of is forcing `is_reached()` to also take the `road_network` as input but I have some fear that will require the `road_network` in places it should not show up. Perhaps that fear is unfounded though. 

Otherwise we could consider completely stripping the `dataclass` decorator. I think a large part of the purpose of the `core.scenario` versions of the goals were to operate as objects rather than just data or else it is not worth much to spend the effort converting from `sstudio` to `core.scenario` scenarios.

smarts/sstudio/genhistories.py
```diff
@@ -450,6 +460,159 @@ def column_val_in_row(self, row, col_name):
...
+                        row["speed"] = lerp(prev_row["speed"], row["speed"], t)
+                        row["position_x"] = lerp(
+                            prev_row["position_x"], row["position_x"], t
+                        )
+                        row["position_y"] = lerp(
+                            prev_row["position_y"], row["position_y"], t
+                        )
+                        row["heading_rad"] = lerp(
+                            prev_row["heading_rad"], row["heading_rad"], t
+                        )
```
sah-huawei, line 571: We might need to shift this back to be between 0 and 2pi (or -pi and pi?).

smarts/sstudio/genhistories.py
```diff
@@ -450,6 +460,159 @@ def column_val_in_row(self, row, col_name):
...
+                timestep = 0.1
+                time_current = row["sim_time"]
+                time_expected = round(j * timestep, 3)
+                time_error = time_current - time_expected
+
+                if time_error == 0:
+                    continue
+
+                # Align to expected timestep
+                row["sim_time"] = time_expected
```
sah-huawei, line 550: In general, for all of the `row` updates that follow, since you're replacing `row[j][...]` immediately, if you have to interpolate the _next_ value based on this one, on so on, errors could compound.  It might be better to only update the `row[j][...]` entries after we've interpolated everything.  (I realize it's not as convenient in terms of code though.)

smarts/sstudio/genhistories.py
```diff
@@ -450,6 +460,159 @@ def column_val_in_row(self, row, col_name):
...
+                        row["speed"] = lerp(row["speed"], next_row["speed"], t)
+                        row["position_x"] = lerp(
+                            row["position_x"], next_row["position_x"], t
+                        )
+                        row["position_y"] = lerp(
+                            row["position_y"], next_row["position_y"], t
+                        )
+                        row["heading_rad"] = lerp(
+                            row["heading_rad"], next_row["heading_rad"], t
+                        )
```
sah-huawei, line 592: ditto

sah-huawei (commented):
Great work, guys!  It's looking good.  Just a few very minor things flagged...

smarts/core/traffic_history.py
```diff
@@ -89,6 +94,12 @@ def target_speed(self) -> float:
         query = "SELECT value FROM Spec where key='speed_limit_mps'"
         return self._query_val(float, query)
 
+    @cached_property
+    def ego_vehicle_id(self) -> int:
+        query = "SELECT id FROM Vehicle WHERE is_ego_vehicle = 1"
+        ego_id = self._query_val(int, query)
```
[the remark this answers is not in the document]
RutvikGupta: We added the `is_ego_vehicle`entry to 0 as default for datasets not having this entry.

smarts/sstudio/genhistories.py
RutvikGupta: `row["heading_rad"] = obj_state.heading - math.pi / 2`.
They are shifted back first and then interpolated.

sah-huawei: But I think the interpolation can still cause values to be outside of this range.  For example, if the value at time `t` is `1.5*pi` and the value at time `t+0.08` is `2*pi`, then the interpolated value for time `t+0.1` will be bigger than `2*pi`.

smarts/core/traffic_history.py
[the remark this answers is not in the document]
sah-huawei: No, I mean, when there are no vehicles with `is_ego_vehicle = 1`, the `_query_val` call will return `None` (just confirmed by looking at it), which means `ego_id` will be returned to the caller as `None`.   It looks like you're immediately trying to convert it to a `str` in the history vehicles replacement script.  We should probably handle the case of `None` there too (and anywhere else this is called).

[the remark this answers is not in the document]
RutvikGupta: This query in particular is only used in Waymo dataset which will always record which vehicle was Waymo ego (since it was recorded by it). We can add a flag checking if the value returned is not None, but I guess we will just hijack a random vehicle if by chance it is not mentioned in the dataset.

.github/workflows/ci-base-tests.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           pip install --upgrade pip
           pip install --upgrade wheel
           pip install --upgrade -r requirements.txt
-          pip install --upgrade -e .[train,test,camera-obs]
+          pip install --upgrade -e .[train,test,camera-obs,waymo]
```
Adaickalavan, line 43: Is there any `waymo` related tests which requires it to be installed here?

RutvikGupta: No actually we just needed to `requirements.txt` to `scenario/waymo` folder to prevent CI tests to fail when building the waymo scenario. So i added that. Thanks for the reminder.

scenarios/waymo/gen_sumo_map.py
```diff
@@ -0,0 +1,168 @@
...
+import subprocess
+import xml.dom.minidom
+import xml.etree.ElementTree as ET
+from pathlib import Path
+from typing import List, Tuple
+
+from smarts.sstudio.genhistories import Waymo
+from waymo_open_dataset.protos import scenario_pb2
+
+os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"
```
Adaickalavan, line 13: Why was the "TF_CPP_MIN_LOG_LEVEL" set to 3? Error messages are likely to be suppresed at log level 3. Perhaps, can use log level 2.

scenarios/waymo/gen_sumo_map.py
```diff
@@ -0,0 +1,168 @@
...
+        return f
+
+    def _create_node(self, node_id, x, y):
+        node = ET.SubElement(self.nodes_root, "node")
+        node.set("id", node_id)
+        node.set("type", "priority")
+        node.set("x", str(x))
+        node.set("y", str(y))
+
+    def _create_edge(self, edge_id, start_id, end_id, shape_str, width=5):
```
Adaickalavan, line 84: If possible, type hints for all function inputs and outputs would be helpful.

scenarios/waymo/gen_sumo_map.py
RutvikGupta: Its not needed now. Removed!

scenarios/waymo/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Setup
+
+Follow the instructions for the setup of SMARTS in the main [README](https://github.com/huawei-noah/SMARTS/). Then install the `[waymo]` dependencies to install the `waymo-dataset` package:
+
+```bash
+pip install -e .[waymo]
+```
+
+Next, download the dataset files to the folder `scenarios/waymo/waymo_dataset`. It is recommended to download the dataset files from the `training_20s` folder as they have the full traffic capture for each scenario.
```
Adaickalavan, line 15: @RutvikGupta Perhaps, the exact steps to download dataset `training_20s` can be better explained, similar to how you had explained to me which I think was helpful.

smarts/sstudio/genhistories.py
```diff
@@ -216,8 +228,7 @@ def rows(self):
             self._next_row = None
             yield last_row
 
-    @staticmethod
-    def _lookup_agent_type(agent_type):
+    def _lookup_agent_type(self, agent_type):
```
Adaickalavan, line 231: Consider type hinting.

smarts/sstudio/genhistories.py
```diff
@@ -410,8 +421,7 @@ def rows(self):
                 for state in states.values():
                     yield (t, state)
 
-    @staticmethod
-    def _lookup_agent_type(agent_type):
+    def _lookup_agent_type(self, agent_type):
```
Adaickalavan, line 424: Consider type hinting.

smarts/sstudio/genhistories.py
```diff
@@ -450,6 +460,180 @@ def column_val_in_row(self, row, col_name):
         return None
 
 
+class Waymo(_TrajectoryDataset):
+    def __init__(self, dataset_spec, output):
```
Adaickalavan, line 464: Consider input/output type hinting for all the functions which we modify or write.

smarts/core/vehicle_index.py
```diff
@@ -665,6 +665,10 @@ def begin_rendering_vehicles(self, renderer):
     def sensor_states_items(self):
         return map(lambda x: (self._2id_to_id[x[0]], x[1]), self._sensor_states.items())
 
+    def check_vehicle_id_has_sensor_state(self, vehicle_id):
```
Adaickalavan, line 668: Consider type hints.

PR #1062, opened 2021-08-17
Refactoring changes to ULTRA load models

christianjans:
Some minor refactoring changes to ULTRA load models.

### DIFF
--- a/ultra/tests/test_train.py
+++ b/ultra/tests/test_train.py
@@ -32,7 +32,7 @@ import ray
 from smarts.core.agent import AgentSpec
 from smarts.zoo.registry import make
 from ultra.baselines.sac.sac.policy import SACPolicy
-from ultra.train import train, load_model
+from ultra.train import train, load_agents
 
 AGENT_ID = "001"
 seed = 2
@@ -177,7 +177,7 @@ class TrainTest(unittest.TestCase):
         experiment_dir = glob.glob(
             os.path.join(TrainTest.OUTPUT_DIRECTORY, "model_logs/*")
         )[0]
-        agent_ids, agent_classes, agent_specs, agents = load_model(experiment_dir)
+        agent_ids, agent_classes, agent_specs, agents = load_agents(experiment_dir)
 
         self.assertEqual(agent_ids[0], "000")
         self.assertEqual(agent_classes[agent_ids[0]], "ultra.baselines.sac:sac-v0")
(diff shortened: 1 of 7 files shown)


ultra/ultra/train.py
```diff
@@ -134,20 +134,24 @@ def load_agents(experiment_dir):
...
+        latest_checkpoint_directories = {
+            agent_id: sorted(
+                glob.glob(os.path.join(experiment_dir, "models", agent_id, "*")),
+                key=lambda x: int(x.split("/")[-1]),
+            )[-1]
+            for agent_id in agent_ids
+        }
+    else:
+        print("No models found, building agent_spec without any checkpoint directories")
+        latest_checkpoint_directories = {agent_id: None for agent_id in agent_ids}
```
christianjans, line 147: I know if one agent has a checkpoint directory, then all other agents should also have a checkpoint directory (because evaluation is done at the same episodic rate for all agents). However, for robustness sake and to be consistent, perhaps we can check, for each agent, whether they have a checkpoint directory:
```suggestion
    latest_checkpoint_directories = {}
    for agent_id in agent_ids:
        checkpoint_directories = sorted(
            glob.glob(os.path.join(experiment_dir, "models", agent_id, "*")),
            key=lambda x: int(x.split("/")[-1]),
        )
        latest_checkpoint_directory = checkpoint_directories[-1] if len(checkpoint_directories) > 0 else None
        latest_checkpoint_directories[agent_id] = latest_checkpoint_directory
```
I haven't tested this yet, but did you have a test that you ran when you originally caught this error?

ultra/tests/test_train.py
```diff
@@ -41,6 +42,38 @@ class TrainTest(unittest.TestCase):
...
+        try:
+            os.system(
+                f"python ultra/train.py --task 00 --level easy --episodes 3 --eval-rate 2 --max-episode-steps 2 --log-dir {single_agent_model_log_dir} --eval-episodes 1 --headless"
+            )
+        except Exception as err:
+            print(err)
+            self.assertTrue(False)
+
+        if not os.path.exists(single_agent_model_log_dir):
+            self.assertTrue(False)
```
christianjans, line 60: There is no `self` attribute in this method (see also the multi-agent training command).

JenishPatel99: Sorry, I forgot to commit those changes. It should be there now

JenishPatel99 (approved):
I made a few documentation changes, but everything else seems perfect 👍 PR #1067 can also be merged later on

ultra/ultra/train.py
JenishPatel99: Changes applied

PR #1070, opened 2021-08-23
Add scenario interface generation seed in ULTRA

christianjans:


JenishPatel99 (approved):
This looks good. I believe a range of 2^31-1 will be more than sufficient to diversified the scenarios

Issue #1054, opened 2021-08-09
fuzzy behavior when `wrong_way` done criteria is enabled


JenishPatel99:
**High Level Description**
What is the purpose of the wrong way done criteria and when should it be triggered?

**Desired SMARTS version**
v0.4.18

**Operating System**
 ubuntu 20.04.1 LTS

**Problems**
I am a bit confused as to how does the wrong way done criteria actually work. In the `sensors.py`, there is a special case for which the wrong way check is disabled at the intersection, as mention below:
```python
 @classmethod
    def _check_wrong_way_event(cls, lane_to_check, sim, vehicle):
        # When the vehicle is in an intersection, turn off the `wrong way` check to avoid
        # false positive `wrong way` events.
        if lane_to_check.getEdge().isSpecial():
            return False

        return cls._vehicle_is_wrong_way(sim, vehicle, lane_to_check.getID())

```

However, during a visual inspection on Envision, I noticed that there was an unexpected termination at the middle of the intersection when I enable the wrong way criteria. Additionally, when I disabled the criteria, the ego vehicle was able to finish the mission properly. Here are the screen recordings

A. Wrong_way=True

https://user-images.githubusercontent.com/55423380/128784633-d6e140bf-e70e-4c56-8c25-449115abe75b.mp4

B. Wrong_way=False

https://user-images.githubusercontent.com/55423380/128784663-a6cee181-fea6-4fe5-b6e0-d6df660c4b5c.mp4

There were also issues when the car was not in the intersection. The wrong way criteria do not seem to trigger when the car is in the opposing lane. However, when I enable on_shoulder done criteria then it seems to be picking up when the ego vehicle is fully into the other lane

C. Wrong_way=True, scenario ends with off_road trigger

https://user-images.githubusercontent.com/55423380/128785904-9c7c4af5-c69c-4739-8bb0-b7f08ed4777f.mp4

D. Wrong_way=True, On_shoulder=True, scenarios ends with on_shoulder trigger

https://user-images.githubusercontent.com/55423380/128785943-e1b4db3c-00fa-455a-8f47-be722e46b566.mp4

What we want for ULTRA and the competition is that when the ego vehicle is driving (or idling) in the opposing lane (i.e. egos heading is opposite of the direction of the lane) the scenario should be terminated, However, I am not sure how to fix the following issue

I can grant the scenarios used for debugging upon request

PR #1071, opened 2021-08-24
Fix wrong way detection

Adaickalavan:
Fixes wrong way detection issue #1054 .

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -489,27 +489,30 @@ class Sensors:
         if not nearest_lane:
             return (True, False)
 
+        # Check whether vehicle is in wrong-way
+        is_wrong_way = cls._check_wrong_way_event(nearest_lane, vehicle)
+
+        # Check whether vehicle has no-route or is on-route
         if (
-            not route_roads
-            or nearest_lane.road in route_roads
+            not route_roads  # Vehicle has no-route. E.g., endless mission with a random route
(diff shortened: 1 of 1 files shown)


smarts/core/sensors.py
```diff
@@ -489,40 +489,43 @@ def _vehicle_is_off_route_and_wrong_way(cls, sim, vehicle):
...
 
         veh_offset = nearest_lane.offset_along_lane(vehicle_pos)
 
-        # so we're not obviously on the route, but we might have just gone
+        # so we're obviously not on the route, but we might have just gone
         # over the center line into an oncoming lane...
         for on_lane in nearest_lane.oncoming_lanes_at_offset(veh_offset):
             if on_lane.road in route_roads:
-                is_wrong_way = cls._check_wrong_way_event(on_lane, sim, vehicle)
                 return (False, is_wrong_way)
```
JenishPatel99, line 509: Does this case handle when the vehicle is not completely "off route" but part of the vehicle is on the opposing lane? 

So I was wondering if the car goes off the route even **partially** then the wrong way event is triggered (i.e. wrong_way=True), however if I have (wrong_way=False, off_route=True) then will the event trigger will occur when the car is **completely** off the route?

JenishPatel99 (approved):
The changes looks good. I also verified the changes by looking at the envision results for models that previously exhibited wrong way behaviour

Issue #270, opened 2020-12-03
`macos` Tests


Gamenot:
Current CI tests run on `linux v18` but there are no tests currently to make sure that SMARTS continues to work on `macos`.

AC:
- [x] Test installation steps on `macos`.
- [x] CI runs the `make test` test for `macos` os on merge to master.
- [x] CI tests dependency requirements for `macos` on merge to master.

Related: #265

liamchzh:
@liamchzh bring this up with Nancy.

PR #1072, opened 2021-08-24
MacOS CI/CD Base and dependency tests

RutvikGupta:
Closes Issue #270 . 

PS: For some reason the Panda3D and its dependencies are not being detected on the mac CI tests even though the tests pass locally on macOS computer. So I have ignored all tests related to `Renderer` class until we found a solution for this dependency.


Gamenot (approved):

envision/utils/multiprocessing_queue.py
```diff
@@ -0,0 +1,92 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# Both the classes comes entirely from Eli Bendersky's blog:
+# http://eli.thegreenplace.net/2012/01/04/shared-counter-with-pythons-multiprocessing/sing
+# Please refer to https://eli.thegreenplace.net/pages/about for more info about the author
+
+import multiprocessing
+import multiprocessing.queues
```
Adaickalavan, line 28: Is it necessary to import both `multiprocessing` and `multiprocessing.queues` ?

.github/workflows/ci-base-tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           - ./smarts/core
           - ./smarts/env --ignore=./smarts/env/tests/test_rllib_hiway_env.py
           - ./smarts/env/tests/test_rllib_hiway_env.py
-          - ./smarts/sstudio
+          - ./smarts/sstudio --ignore=./smarts/sstudio/genhistories.py
```
Adaickalavan, line 31: Why `--ignore=./smarts/sstudio/genhistories.py` in linux test?

.github/workflows/ci-base-tests.yml
```diff
@@ -54,4 +54,59 @@ jobs:
...
+
+  mac:
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    strategy:
+      matrix:
+        tests:
+          - ./envision
+          - ./smarts/contrib
+          - ./tests
+          - ./smarts/sstudio --ignore=./smarts/sstudio/genhistories.py
```
Adaickalavan, line 67: Why ignore `--ignore=./smarts/sstudio/genhistories.py` a non test file in mac test?

.github/workflows/ci-base-tests.yml
Adaickalavan, line 31: Why ignore `--ignore=./smarts/sstudio/genhistories.py` a non test file in linux test?

.github/workflows/ci-base-tests.yml
```diff
@@ -54,4 +54,59 @@ jobs:
...
+          pip install -e .[train,test,camera-obs]
+      - name: Run smoke tests
+        run: |
+          . ${{env.venv_dir}}/bin/activate
+          export SUMO_HOME="/usr/local/opt/sumo/share/sumo"
+          open -g -a XQuartz.app
+          make build-all-scenarios
+          PYTHONHASHSEED=42 pytest -v \
+            --doctest-modules \
+            -n auto \
```
Adaickalavan, line 103: Why were the options removed in mac test?
```
--forked \
--dist=no \
```

Adaickalavan, line 103: What lead to the removal of these options in mac test?
```
--forked \
--dist=no \
```

RutvikGupta: Some of the core tests pass for Macos CI.

.github/workflows/ci-base-tests.yml
RutvikGupta: Its because of the Waymo import in `genhistories.py`. Pip package provided by Waymo for the dataset is only for Linux. You need to compile your own packages for other OS. So I ignored the file here since it checks for dependencies for all files in `sstudio` folder.

RutvikGupta: I can also put the import statement in a try block if that's a better solution.

envision/utils/multiprocessing_queue.py
RutvikGupta: Sorry, forgot to remove it.

Adaickalavan:
1) The new `requirements.txt` for mac appears to have (i) updated package versions, and (ii) several additional packages.
2) What lead to the need for a different `requirements.txt` for mac tests?
3) How were the new packages for mac deduced? 
4) Will the same updated package versions be applicable to linux too?

RutvikGupta:
> 1. The new `requirements.txt` for mac appears to have (i) updated package versions, and (ii) several additional packages.
> 2. What lead to the need for a different `requirements.txt` for mac tests?
> 3. How were the new packages for mac deduced?
> 4. Will the same updated package versions be applicable to linux too?

2. Some packages in linux don't exist for mac so, i had to create a new `requirements.txt` for comparing this with packages installed. Basically for dependency test.
3. I got them from the pip freeze output on the CI.
4. Yes but i didn't want to update them since it causes the Linux dependencies to fail.

Adaickalavan:
Previously the ci test runs in ~20 minutes, but with the macos test added it runs in ~40 minutes. This is because macos test starts from scratch and not from a base docker image. 

So should we enable macos test for master branch only?

RutvikGupta:
> Previously the ci test runs in ~20 minutes, but with the macos test added it runs in ~40 minutes. This is because macos test starts from scratch and not from a base docker image.
> 
> So should we enable macos test for master branch only?

I was thinking the same. Maybe we should do it only for master branch

Adaickalavan (approved):

PR #1073, opened 2021-08-24
Organize `replay_buffer_dataset` by episodes

JenishPatel99:
Changes from PR #1068 are also applied. Do not worry about the other file changes. Only files changes are: `test_train.py` and `common.py`


ultra/ultra/utils/common.py
```diff
@@ -343,3 +345,62 @@ def clip_angle_to_pi(angle):
 def compute_grad(x, prev_x, dt):
     grad = (x - prev_x) / dt if prev_x is not None else x / dt
     return grad
+
+
+def combine_replay_buffer_dataset_with_episodes_results(
+    experiment_dir, agent_id, active_tag
+):
```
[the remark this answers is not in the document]
christianjans: Hmm okay, I tried it myself and it looks like there is a circular import problem. Perhaps a docstring for this function would be useful instead?

ultra/ultra/utils/common.py
```diff
@@ -353,9 +353,30 @@ def compute_grad(x, prev_x, dt):
...
+    Args:
+        experiment_dir (str): Directory which contains the latest_replay_buffer.pkl
+                              and results.pkl
+        agent_id (str): Representing which agent's replay buffer dataset and episode results
+                        to be used
+        active_tag (str): Identifies the mode of the experiment (Train, Evaluation-Training,
+                          Evaluation), which is needed to get the correct results.pkl
+
+    Return:
+        episodes_data (Dict[int, Dict[str, Union[List[ReplayBufferDataset], Episode]]])):
```
christianjans, line 376: Okay thanks, the docstring definitely helps (me at least).

One thing I wonder is if these types are correct? I think for the transitions part, it should be a `Transition` object (`ultra.baselines.common.replay_buffer.Transition`), and not a `ReplayBufferDataset`?

Additionally, I believe the `Episode` part of the hint may be incorrect as well? I do not think we are not retrieving the Episode object itself when we load in the results, are we? For the `episode_results` key, its value is actually a `Dict`?

If so, maybe the specific type-hinting is too complicated and we can just do something like:
```suggestion
        episodes_data (Dict[int, Dict[str, Any]])):
```
The example you provide gives lots of information already about the types. But if I am wrong please let me know.

JenishPatel99: Yeah, that's right Christian. I was thinking that if I type hinted the variables to their primitive types then it might be too confusing, thus I simplified to the following:
 - a set of transitions is a replay buffer dataset
 - the episode results practically represents the attributes of the episode

If the example alone provides adequate information then it is okay to generalize the type to be `any` 👍

christianjans: Okay great, and sorry one more thing I just noticed, is it supposed to be:
```suggestion
        episodes_data (Dict[str, Dict[str, Union[List[ReplayBufferDataset], Episode]]])):
```
I think the episode index is actually a string?
https://github.com/huawei-noah/SMARTS/blob/5da9aa6d09bf455394f14783e2ea3a8981adbd06/ultra/ultra/utils/common.py#L361

JenishPatel99: The episode index is indeed an integer value, this is the same value that is saved during `record_tensorboard` where our `recording_step` is initialized to `episode.index` (i.e. an integer). I did verify it by printing out the keys of the `episodes_data` which was given as `dict_keys([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])`. 

The confusion was in the example given. I have updated the docstring

christianjans (approved):

JenishPatel99:
One of the ULTRA CI tests (ULTRA CI Base Tests / test-package-via-pypi) fails because to the ultra-rl pypi does not support the updated version of smarts. Once updated, the ci test should passed

PR #1075, opened 2021-08-26
Translate traffic history vehicle positions on import for shifted maps

saulfield:
For SUMO maps that have been shifted, we now shift the positions of the traffic history vehicles on import (before writing them to the SQLite db). This lets us remove the `xy_offset` property from `RoadMap` and eliminates the need to do this shifting at runtime.

### DIFF
--- a/smarts/core/road_map.py
+++ b/smarts/core/road_map.py
@@ -56,14 +56,6 @@ class RoadMap:
         raise NotImplementedError()
 
     @property
-    def xy_offset(self) -> Tuple[float, float]:
-        """optional: this is the amount that external coordinates must
-        be shifted in order to have the coordinate system (bounding_box)
-        start at the origin.  Will be (0. 0) for most maps."""
-        # TAI:  get rid of this, fix traffic_history_provider instead
-        return (0, 0)
-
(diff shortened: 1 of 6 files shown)


Gamenot (approved):

smarts/core/road_map.py
```diff
@@ -55,14 +55,6 @@ def bounding_box(self) -> BoundingBox:
...
         raise NotImplementedError()
 
-    @property
-    def xy_offset(self) -> Tuple[float, float]:
-        """optional: this is the amount that external coordinates must
-        be shifted in order to have the coordinate system (bounding_box)
-        start at the origin.  Will be (0. 0) for most maps."""
-        # TAI:  get rid of this, fix traffic_history_provider instead
-        return (0, 0)
-
```
sah-huawei, line 65: While I'm happy to see this go, I'm a bit nervous by something that Jun said in a recent tech meeting about keeping things as close to the data as possible.  I think we should still go ahead with this, but also be prepared to bring this back in the future if necessary.

sah-huawei (approved):
Looks good!

Issue #1079, opened 2021-08-27
Traffic Lights


Gamenot:
**Is your feature request related to a problem? Please describe.**
Traffic lights and priority are not currently provided in observation.

**Describe the solution you'd like**
Waymo traffic light simulation provided to the observations. 

**Describe alternatives you've considered**
Implemented native traffic light simulation.

**Additional context**
Recent requests for traffic light information. Removed previous traffic light information: #645

Gamenot:
This has to go in 0.6 because it is an interface change.

saulfield:
Addressed by #1529.

PR #1083, opened 2021-09-02
Waymo traffic lights simulation provided as observations to SMARTS

RutvikGupta:
See Issue #1079.
Adds traffic light data from Waymo as observations in Smarts. The observations include the light state and the (x, y) position of the "stopping point" of the vehicle in the lane with respect to the traffic light (i.e. no other lane information).


Adaickalavan:
Why do we get the following while running make test?

https://pipelines.actions.githubusercontent.com/Nzwgt7wQOg9IlSQCyepePeuX1c2DLh9vH8MfEbWSwqK8xomSVz/_apis/pipelines/1/runs/18091/signedlogcontent/5?urlExpires=2021-09-03T22%3A19%3A23.5641057Z&urlSigningMethod=HMACV1&urlSignature=yGQlqQkwQxa2UeOZhC90xjgDIgyJxinFNPaCGqXth80%3D

image

RutvikGupta:
> Why do we get the following while running make test?
> 
> See `test (./smarts/core)` --> `Run smoke tests`
> https://github.com/huawei-noah/SMARTS/runs/3507259322
> 
> image

Its because the Waymo dataset is missing. We can ignore building this scenario in the tests since its not possible to do so without the dataset anyways.

cli/studio.py
```diff
@@ -143,6 +143,8 @@ def build_all_scenarios(clean, allow_offset_maps, scenarios):
         path = Path(scenarios_path)
         for p in path.rglob("*.net.xml"):
             scenario = f"{scenarios_path}/{p.parent.relative_to(scenarios_path)}"
+            if scenario == f"{scenarios_path}/waymo":
+                continue
```
sah-huawei, line 147: I'm not sure about this "short circuit" here.  In addition to wanting to avoid "special cases" like this in general, I think the danger is that `_build_single_scenario()` also (perhaps eventually) does something necessary aside from (mostly) Sumo stuff.  In fact, right now, it can be used to clean the scenario folder as well as install additional pip requirements.  Even though "_our_" Waymo use cases don't require these, other users' might.

Could we move this check further down the call stack and/or make it more generic?  (For example, lower down the stack, maybe we can skip things related to Sumo in general if there is no `map.net.xml` file.)

smarts/core/sensors.py
```diff
@@ -123,6 +130,33 @@ class Vias:
...
+    GO = 6
+
+    # Flashing light signals.
+    FLASHING_STOP = 7
+    FLASHING_CAUTION = 8
+
+
+@dataclass(frozen=True)
+class TrafficLightData:
+    stop_point: Tuple[float, float]
```
sah-huawei, line 153: We could use `Point` here (from `smarts.core.coordinates`).  ,,,
But is it really just a point and not a line segment?  If so, this seems to imply that a traffic light can only control a single lane.  I still feel like we must be missing something about the waymo dataset if this is the case!

smarts/core/sensors.py
```diff
@@ -1050,3 +1091,33 @@ def closest_point_on_lane(position, lane_id):
 
     def teardown(self):
         pass
+
+
+class TrafficLightSensor(Sensor):
+    def __init__(
+        self, traffic_history: TrafficHistory, map_location_offset: Tuple[float, float]
+    ):
+        self._traffic_history = traffic_history
```
sah-huawei, line 1100: fwiw, I think we're eventually going to want to generalize this away from just coming from history data so that SMARTS can manage its own traffic lights, or rely on some other co-simulator (such as Sumo) to do so.  So I'm wondering if it's more appropriate to add some sort of intermediate layer between the sensor and the history data, like a `TrafficLightManager` class or something.  But I guess we can wait on that for now...

smarts/core/vehicle.py
```diff
@@ -526,6 +527,16 @@ def attach_sensors_to_vehicle(sim, vehicle, agent_interface, plan):
             )
         )
 
+        if agent_interface.traffic_lights:
+            assert (
+                sim.scenario.traffic_history
+            ), "Traffic history required for traffic lights sensor"
+            vehicle.attach_traffic_lights_sensor(
+                TrafficLightSensor(
+                    sim.scenario.traffic_history, sim.scenario.road_map.xy_offset
```
sah-huawei, line 536: Of course, when this gets merged with PR #1075, this will need to change.  Just flagging it so we don't miss it.

smarts/sstudio/genhistories.py
```diff
@@ -117,6 +128,14 @@ def _create_tables(self, dbconxn):
                    FOREIGN KEY (vehicle_id) REFERENCES Vehicles(id)
                ) WITHOUT ROWID"""
         )
+        ccur.execute(
+            """CREATE TABLE Traffic_Lights (
```
sah-huawei, line 132: The convention we've been using is to have singular table names.  And I think it's better to treat table names like class names.  So I suggest this table be named `TrafficLight`.

smarts/sstudio/genhistories.py
```diff
@@ -117,6 +128,14 @@ def _create_tables(self, dbconxn):
...
+                   sim_time REAL NOT NULL,
```
sah-huawei, line 133: I think we're going to want an index on the `sim_tiime` column given that this is queried on every step (potentially for multiple vehicles).

smarts/core/sensors.py
RutvikGupta: This for now is just specific to Waymo since Asad wanted this data to be read in. But when TrafficLights will be added and supported by SMARTS in general, we will generalize this as well.

cli/studio.py
RutvikGupta: currently we don't build any scenario with `map.net.xml` file missing. Here i skip building the waymo scenario since it tries to load in the Waymo dataset which is not added to the repo and hence throws the above error. The only we can stop it is to skip building the Waymo scenario during `build-all` command is by the above check or remove its `map.net.xml` file. So we decided to go with this one. If we move it further below the call stack it means we will be putting it in the  `_build_single_scenario()` where we don't want to since we still want to build it individually.

smarts/core/sensors.py
saulfield: Thanks for the tip, I'll use `Point` instead.

For your other comment -- the traffic light data is given on a per-lane basis rather than per-light, so each point represents the "stop point" in a single lane if the relevant light is in a "stop" state. There is no concept of a "light" on its own. This is how the Waymo dataset defines it, we are simply passing it on.

Here's an image from a scenario to illustrate:
image

cli/studio.py
RutvikGupta: After some discussion we decided to remove the `map.net.xml` since it for an external dataset that we don't have available available and makes the scenario incomplete (Like NGSIM). So this takes care of the above issue.

smarts/core/sensors.py
```diff
@@ -17,7 +17,9 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
+from enum import Enum
 import logging
+from smarts.core.traffic_history import TrafficHistory
 import time
 from collections import deque, namedtuple
```
Adaickalavan, line 24: Ordering `import` and `from` statements alphabetically improves code presentation.

smarts/sstudio/genhistories.py
```diff
@@ -26,15 +26,22 @@
...
 from typing import Dict, Generator, Union
 
+try:
+    from waymo_open_dataset.protos import scenario_pb2
+except ImportError:
+    print(sys.exc_info())
+    print(
+        "You may not have installed the [waymo] dependencies required to use the waymo replay simulation. Install them first using the command `pip install -e .[waymo]` at the source directory."
+    )
+    pass
```
Adaickalavan, line 43: Printing `sys.exc_info()` directly, without re-formatting it according to our needs, will appear a little messy. `pass` may not be needed at the end.
 
```suggestion
try:
    from waymo_open_dataset.protos import scenario_pb2
except ImportError:
    print(
        "You may not have installed the [waymo] dependencies required to use the waymo replay simulation. Install them first using the command `pip install -e .[waymo]` at the source directory."
    )
```

smarts/sstudio/genhistories.py
```diff
@@ -487,18 +527,7 @@ def read_dataset(path: str) -> Generator[bytes, None, None]:
...
-            return t * (b - a) + a
-
-        def constrain_angle(angle):
-            """Constrain to [-pi, pi]"""
-            angle = angle % (2 * math.pi)
-            if angle > math.pi:
-                angle -= 2 * math.pi
-            return angle
-
+    def _get_scenario(self):
```
Adaickalavan, line 530: Missing return type hint.

RutvikGupta: It was removed because its return type is `scenario_pb2.Scenario` which when shown as a return type hint throws a `NotImplementedError` if the waymo dependencies fail to install above in the import try block.

smarts/sstudio/genhistories.py
```diff
@@ -26,18 +26,17 @@
 import sqlite3
 import struct
 import sys
-
 import ijson
 import numpy as np
 import pandas as pd
 import yaml
+import traceback
```
Adaickalavan, line 33: `traceback` appears to be not used.

PR #1084, opened 2021-09-03
Conditional Sumo Import

sah-huawei:
Made all core imports of Sumo stuff conditional.

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -52,7 +52,6 @@ from .provider import Provider, ProviderState
 from .road_map import RoadMap
 from .scenario import Mission, Scenario
 from .sensors import Collision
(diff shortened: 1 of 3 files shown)


smarts/core/smarts.py
```diff
@@ -80,7 +79,7 @@ class SMARTS:
     def __init__(
         self,
         agent_interfaces,
-        traffic_sim: SumoTrafficSimulation,
+        traffic_sim,  # SumoTrafficSimulation
```
Gamenot, line 82: I am not sure we need to specify that it is `SumoTrafficSimulation` any more since we are getting away from that.

sah-huawei: Yeah, agreed.  But I'll remove that in a future PR.

Gamenot (approved):

PR #1087, opened 2021-09-03
Parallel environments

Adaickalavan:
+ Added `ParallelEnv` class to simulate multiple SMARTS environments in parallel, with synchronous or asynchronous episodes.
+ Added `parallel_environment.py` example on running multiple environments in parallel.


smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,342 @@
...
+        observation_space, action_space = self._get_spaces()
+
+        # TODO: This dummy `observation_space` and `action_space` should be removed after they
+        # are properly specified in SMARTS/smarts/env/hiway_env.py:__init__() function.
+        action_space = gym.spaces.Dict(
+            {
+                agent_id: gym.spaces.Box(
+                    np.array([0, 0, -1]), np.array([+1, +1, +1]), dtype=np.float32
+                )  # throttle, break, steering
+                for agent_id in ["Agent1", "Agent2"]
```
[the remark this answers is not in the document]
Adaickalavan: https://github.com/huawei-noah/SMARTS/blob/833dc8dbdb2d78096ca53163097e35bfa6dcba9d/smarts/env/wrappers/parallel_env.py#L126-L147.

Our current SMARTS's observation and action space on line L127 do not conform to gym spaces.

Hence, as noted on lines L129-L130, the observation and action space on lines L131-L141 are temporary dummy spaces compliant with gym spaces. These lines L131-L141 are to be removed later when SMARTS's spaces become gym compliant.

Unfortunately, `None` is not a valid gym space.

These temporary dummy spaces will not affect the operation of this parallel environment wrapper.

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,342 @@
...
+        Args:
+            timeout (Union[int, float, None], optional): Seconds to wait before timing out.
+                Defaults to None, and never times out.
+
+        Raises:
+            NoAsyncCallError: If `reset_wait` is called without calling `reset_async`.
+            mp.TimeoutError: If response is not received from pipe within `timeout` seconds.
+
+        Returns:
+            Sequence[Dict[str, Any]]: A batch of observations from the vetorized environment.
```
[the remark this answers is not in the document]
Adaickalavan: Spelling corrected.

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,342 @@
...
+
+class ParallelEnv(AsyncVectorEnv):
+    """Batch together multiple environments and step them in parallel. Each
+    environment is simulated in an external process for lock-free parallelism
+    using `multiprocessing` processes, and pipes for communication.
+
+    Note:
+        Number of environments should not exceed the number of available CPU
+        logical cores.
+    """
```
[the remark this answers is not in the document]
Adaickalavan: Users will now be warned of possible simulation slow down if requested number of parallel environments exceed number of available cpu cores.

smarts/env/tests/test_parallel_env.py
```diff
@@ -0,0 +1,197 @@
...
+            assert observations.keys() == agent_specs.keys()
+
+        # Compute batched actions
+        batched_actions = []
+        for observations in batched_observations:
+            actions = {
+                agent_id: agents[agent_id].act(agent_obs)
+                for agent_id, agent_obs in observations.items()
+            }
+            batched_actions.append(actions)
```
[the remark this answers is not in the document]
Adaickalavan: Refactored `_get_batched_actions()` function.

examples/parallel_environment.py
```diff
@@ -0,0 +1,251 @@
...
+            agent_builder=ChaseViaPointsAgent,
+        )
+        for agent_id in agent_ids
+    }
+
+    # Create a callable env constructor. Here, for illustration purposes, each environment is
+    # wrapped with a FrameStack wrapper which returns stacked observations for each environment.
+    env_frame_stack = lambda env: FrameStack(
+        env=env,
+        num_stack=2,
```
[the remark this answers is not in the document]
Adaickalavan: All parameters are user configurable now.

smarts/env/tests/test_parallel_env.py
```diff
@@ -0,0 +1,197 @@
...
+    )
+    for process in env.processes:
+        assert process.is_alive() == True
+    assert len(env.processes) == len(env_constructors)
+
+    # Check seed number of each environment
+    first_seed = 7
+    seeds = env.seed(first_seed)
+    for index, seed in enumerate(seeds):
+        assert seed == first_seed + index
```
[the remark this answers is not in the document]
Adaickalavan: The test `./smarts/env/tests/test_parallel_env.py::test_create_parallel_env` was simulated with an erroneous seed function, resulting in the assertion error message shown below.

image

Here, the message shows that we got seed=23 but expected seed=7+1. The assertion error message coupled with the four lines of code directly above it informs us that we expect environments to be seeded in sequentially increasing order beginning from the `first_seed` number for the first environment.

Consider popular open sourced repositories such as 
+ TensorFlow, for example see [tf_wrappers_test.py](https://github.com/tensorflow/agents/blob/v0.9.0/tf_agents/environments/tf_wrappers_test.py)
+ Gym, for example see [test_filter_observation.py](https://github.com/openai/gym/blob/master/gym/wrappers/test_filter_observation.py)

where their unit tests with `self.assertEqual()` and `assert` statements are devoid of error messages.

When `assert` statements and surrounding code is easily understandable, comments or error messages simply bloat the code and thus may be unnecessary.

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,355 @@
...
+        )
+        observation_space = gym.spaces.Box(
+            low=-1, high=1, shape=(256, 256, 3), dtype=np.float32
+        )
+
+        super(AsyncVectorEnv, self).__init__(
+            num_envs=len(env_constructors),
+            observation_space=observation_space,
+            action_space=action_space,
+        )
```
sah-huawei, line 151: This seems dangerous -- that is, to "skip over" calling `super(ParralelEnv, self).__init__` and instead call `super(AsyncVectorEnv,  self).__init__`.  I.e., to call the init function from `VectorEnv` instead of calling the one from the class's parent, `AsyncVectorEnv`.   At a minimum, there should be a comment here as to why this was done (and warning future coders of the anomaly, which they might overlook).

Is it to avoid having to support members of the base class such as `copy` and `shared_memory`?

The danger here is that this is a way of "breaking encapsulation".  You have to look at the source code of the parent class(es) to know whether what we've done here is ok (it seems to be right now), and it may not continue to be so if either of them change.   I see that we've pinned to this version of `gym` (`0.18.3`) so we are probably able to get away with this for now, but code like this increases the burden of doing any `gym` upgrade in the future because the person doing that may not understand the subtleties of what you've done here.   Also, by skipping its `__init__`, we are inheriting an "incomplete" version of the base class (since its `__init__` method does set fields that we don't), which may confuse any code that wants to reference our `ParallelEnv` as a `AsyncVectorEnv`.

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def seed(self, seed: int) -> Tuple[int]:
+        """Sets unique seed for each environment.
+
+        Args:
+            seed (int): Seed number.
+
+        Raises:
+            AlreadyPendingCallError: If `seed` is called while another function call is pending.
+        """
```
sah-huawei, line 164: This should document what is returned as well (especially since this is the key difference between this implementation of `seed()` and the implementation in the parent class).

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,355 @@
...
+            raise mp.TimeoutError(
+                "The call to `reset_wait` has timed out after "
+                "{0} second{1}.".format(timeout, "s" if timeout > 1 else "")
+            )
+
+        observations, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        self._raise_if_errors(successes)
+        self._state = AsyncState.DEFAULT
+
+        return observations
```
sah-huawei, line 217: I see this, as well as other implementations here like `step_wait()`, are very similar to the implementations in the base class, but avoid using the `self.copy` and `self.shared_memory` members there (which would be public in this child class as welif the  `__init__` call did the normal thing and called `__init__` on the base class).  

What is the reason we're avoiding these optimizations?

examples/parallel_environment.py
```diff
@@ -0,0 +1,282 @@
...
+            return (obs.waypoint_paths[0][0].speed_limit, 0)
+
+        nearest = obs.via_data.near_via_points[0]
+        if nearest.lane_index == obs.ego_vehicle_state.lane_index:
+            return (nearest.required_speed, 0)
+
+        return (
+            nearest.required_speed,
+            1 if nearest.lane_index > obs.ego_vehicle_state.lane_index else -1,
+        )
```
Gamenot, line 34: I think we should use as primitive agent as possible.

```suggestion
class ChaseViaPointsAgent(Agent):
    def act(self, obs: Sequence[Observation]) -> Tuple[float, int]:
        # Here, we only utilise the newest frame from the stacked observations.
        newest_obs = obs[-1]
        speed_limit = newest_obs.waypoint_paths[0][0].speed_limit
        return (speed_limit, 0)
```

examples/parallel_environment.py
```diff
@@ -0,0 +1,282 @@
...
+        sim_name=args.sim_name,
+        headless=args.headless,
+        seed=args.seed,
+        num_agents=args.num_agents,
+        num_stack=args.num_stack,
+        num_env=args.num_env,
+        auto_reset=False,
+        max_episode_steps=args.max_episode_steps,
+        num_episodes=args.episodes,
+    )
```
Gamenot, line 282: I think we should not repeat this and just add a cli option for `auto_reset` instead.

smarts/core/remote_agent_buffer.py
```diff
@@ -216,8 +215,8 @@ def spawn_local_zoo_manager(port):
 def get_manager_channel_stub(addr):
     channel = grpc.insecure_channel(f"{addr[0]}:{addr[1]}")
     try:
-        # Wait until the grpc server is ready or timeout after 30 seconds
-        grpc.channel_ready_future(channel).result(timeout=30)
+        # Wait until the grpc server is ready or timeout after 10 seconds
+        grpc.channel_ready_future(channel).result(timeout=10)
```
Gamenot, line 219: I think the timeout should not be decided at this level. I would suggest requiring it as an argument.

smarts/env/hiway_env.py
```diff
@@ -127,6 +131,14 @@ def __init__(
         if visdom:
             visdom_client = VisdomClient()
 
+        # Action space
+        # TODO : `self.action_space` needs to be added and be automatically
+        # modified depending on `action_adapter` function used.
+
+        # Observation space
+        # TODO : `self.observation_space` needs to be added after the
+        # SMARTS observations are properly typed to be gym compliant.
```
Gamenot, line 140: Please reference issues here.

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,358 @@
...
+    env = env_constructor(sim_name=name)
+
+    # Environment setup complete
+    pipe.send((None, True))
+
+    try:
+        while True:
+            # Short block for keyboard interrupts
+            if not pipe.poll(polling_period):
+                continue
```
Gamenot, line 326: How does this raise keyboard interrupts that are not seen by any of the other `Pipe` methods?

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+class ParallelEnv(AsyncVectorEnv):
+    """Batch together multiple environments and step them in parallel. Each
+    environment is simulated in an external process for lock-free parallelism
+    using `multiprocessing` processes, and pipes for communication.
+
+    Note:
+        Simulation might slow down when number of parallel environments
+        requested exceed number of available CPU logical cores.
```
Gamenot, line 50: Generally there is the suggestion to leave 1 CPU open for the main thread. Performance also generally improves by the number of physical cores rather than logical ones, with some exceptions.
```suggestion
        requested exceed number of available CPU logical cores -1.
```

smarts/env/wrappers/parallel_env.py
```diff
@@ -0,0 +1,358 @@
...
+
+        # Wait for all environments to successfully startup
+        _, successes = zip(*[pipe.recv() for pipe in self.parent_pipes])
+        self._raise_if_errors(successes)
+
+        # Get and check observation and action spaces
+        observation_space, action_space = self._get_spaces()
+
+        # TODO: This dummy `observation_space` and `action_space` should be removed after they
+        # are properly specified in SMARTS/smarts/env/hiway_env.py:__init__() function.
```
Gamenot, line 134: Please reference an issue.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - Allow specifying "-latest" as a version suffix for zoo locator strings.
 - Extended Imitation Learning codebase to allow importing traffic histories from the Waymo motion dataset and replay in a SMARTS simulation. See PR #1060.
 - Added `ros` extension rule to `setup.py`.
+- Added `ParallelEnv` class to simulate multiple SMARTS environments in parallel, with synchronous or asynchronous episodes.
```
Gamenot, line 22: I think you should also mention `CloudpickleWrapper` and the added parallel environment example.

examples/parallel_environment.py
Adaickalavan: Simplified the agent.

smarts/env/wrappers/parallel_env.py
Adaickalavan:
```python
# Short block for keyboard interrupts
if not pipe.poll(polling_period):
    continue
```

This is to make the code more responsive to a `CTRL-C` command.

CHANGELOG.md
Adaickalavan: Changelog has been changed accordingly.

smarts/env/wrappers/parallel_env.py
Adaickalavan: Now, `observation_space` and `action_space` inside `ParallelEnv` class are obtained from the single environment.

smarts/env/wrappers/parallel_env.py
Adaickalavan: Dropped the terms physical and logical.

smarts/env/tests/test_parallel_env.py
```diff
@@ -0,0 +1,197 @@
...
+            for outputs in batched_outputs:
+                outputs.pop("__all__", None)
+                assert outputs.keys() == agents.keys()
+
+        for batched_outputs, kind in zip(
+            (batched_rewards, batched_dones), [float, bool]
+        ):
+            for outputs in batched_outputs:
+                outputs.pop("__all__", None)
+                assert all(isinstance(val, kind) for val in outputs.values())
```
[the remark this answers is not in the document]
Adaickalavan: See previous reply: https://github.com/huawei-noah/SMARTS/pull/1087#discussion_r714148212

smarts/env/tests/test_parallel_env.py
```diff
@@ -0,0 +1,197 @@
...
+        batched_observations, _, batched_dones, _ = env.step(batched_actions)
+        batched_actions = get_batched_actions(batched_observations)
+        if auto_reset:
+            assert all(dones["__all__"] == False for dones in batched_dones)
+        else:
+            assert all(dones["__all__"] == True for dones in batched_dones)
+        # Step 5
+        batched_observations, _, batched_dones, _ = env.step(batched_actions)
+        batched_actions = get_batched_actions(batched_observations)
+        assert all(dones["__all__"] == True for dones in batched_dones)
```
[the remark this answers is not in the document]
Adaickalavan: See previous reply: https://github.com/huawei-noah/SMARTS/pull/1087#discussion_r714148212

smarts/zoo/manager_servicer.py
```diff
@@ -88,6 +90,10 @@ def stop_worker(self, request, context):
         return manager_pb2.Status()
 
     def destroy(self):
+        with self._destroy_lock:
```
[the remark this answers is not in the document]
Adaickalavan: On further thought, given `server = grpc.server(futures.ThreadPoolExecutor(max_workers=1))` in `smarts/zoo/manager.py`, the thread locks are removed.

smarts/env/wrappers/parallel_env.py
Adaickalavan: Obsolete.

PR #1088, opened 2021-09-08
Waymo Dynamic History Vehicle Hijacking Script 

RutvikGupta:
This script allows user to hijack history vehicles dynamically through a trigger event (based on time or a specific behavioral event) in a Waymo Scenario. This will be useful for scenario generation and training agents in the future.


examples/dynamic_history_vehicles_replacement.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Create a table of vehicle trajectory lengths, filtering out non-moving vehicles
+        traj_lens = []
+        for v_id in scenario.traffic_history.all_vehicle_ids():
+            traj = list(scenario.traffic_history.vehicle_trajectory(str(v_id)))
+            if [row for row in traj if row.speed != 0]:
+                traj_lens.append((v_id, len(traj)))
+
+        # Filter out trajectories with less than 100 timesteps
+        vehicle_candidates = [str(t[0]) for t in traj_lens if t[1] > 100]
```
Gamenot, line 124: Is it worth it for the timestep requirement to be configurable?

examples/dynamic_history_vehicles_replacement.py
```diff
@@ -0,0 +1,212 @@
...
+            "agents": {},
+            "elapsed_sim_time": 0.0,
+            "vehicle_candidates": vehicle_candidates,
+            "vehicles_to_replace_randomly": k,
+        }
+        trigger = Trigger(should_trigger, on_trigger, context=context)
+
+        dones = {}
+        observations = smarts.reset(scenario)
+        while not dones or not all(done for done in dones.values()):
```
Gamenot, line 148: We can shorten this syntax since `all(())==True` .

```suggestion
        while not all(dones.values()):
```

examples/dynamic_history_vehicles_replacement.py
```diff
@@ -0,0 +1,212 @@
...
+                for agent_id, agent_obs in observations.items()
+            }
+            logger.debug(
+                "stepping @ sim_time={} for agents={}...".format(
+                    smarts.elapsed_sim_time, list(observations.keys())
+                )
+            )
+            observations, rewards, dones, infos = smarts.step(actions)
+
+            for agent_id in agents.keys():
```
Gamenot, line 167: In theory this could be reduced to the following but I still prefer `keys()`.
```suggestion
            for agent_id in agents:
```

examples/dynamic_history_vehicles_replacement.py
```diff
@@ -0,0 +1,212 @@
...
+                if dones.get(agent_id, False):
+                    if not observations[agent_id].events.reached_goal:
+                        logger.warning(
+                            "agent_id={} exited @ sim_time={}".format(
+                                agent_id, smarts.elapsed_sim_time
+                            )
```
Gamenot, line 173: I tend to prefer f-strings but using format is always fine as well as not not have to double check for py2.7.
```suggestion
                            f"agent_id={agent_id} exited @ sim_time={smarts.elapsed_sim_time}"
```

examples/dynamic_history_vehicles_replacement.py
```diff
@@ -0,0 +1,212 @@
...
+            # Select vehicles and map to agent ids & specs
+            vehicles_to_trap = {}
+            for veh_id in sample:
+                agent_id = f"agent-{veh_id}"
+                vehicles_to_trap[veh_id] = (agent_id, agent_spec)
+                ctx["agents"][agent_id] = agent_spec.build_agent()
+
+            # Create missions for selected vehicles
+            veh_missions = scenario.create_dynamic_traffic_history_missions(
+                vehicles_to_trap.keys(), ctx["elapsed_sim_time"]
```
sah-huawei, line 110: Just a note that `vehicles_to_trap.keys()` is `sample`.  (I was wondering why `vehicles_to_trap` needs to be a `dict`, but I think it's fine to leave it that way.)

smarts/core/smarts.py
```diff
@@ -389,6 +391,28 @@ def add_agent_with_mission(
...
+        for veh_id, (agent_id, agent_spec) in vehicles_to_trap.items():
+            # Create trap to be triggered immediately
+            try:
+                self.add_agent_with_mission(
+                    agent_id, agent_spec.interface, veh_missions[veh_id][0]
+                )
+            except PlanningError:
+                self._log.warning(
+                    f"Unable to create PositionalGoal for vehicle {veh_id}, falling back to TraverseGoal"
+                )
```
sah-huawei, line 408: This should be very rare, right?  I mean, given that the real vehicle from the history data was able to go from the mission start to the mission end, we should be able to find a route to get there too.   I guess this would indicate that we have an incomplete reconstruction of the map.

Did you guys actually encounter this exception in your testing?   Are there other reasons (aside from an incomplete map) that it might be thrown?

RutvikGupta: Yes, we encountered it only for one of the vehicle we hijacked, though the reason for that was the incomplete map and detecting wrong end lane due it being hijacked at middle of the intersection. Since its mission was very short, we decided to fall back to Traversal Goal for it,

examples/dynamic_history_vehicles_replacement.py
saulfield: Thanks. I'll shorten the right expression, but we still need the `not dones` expression to avoid finishing on the first step.

smarts/core/smarts.py
```diff
@@ -389,6 +391,28 @@ def add_agent_with_mission(
...
+        self,
+        vehicles_to_trap: Dict[str, Tuple[str, AgentSpec]],
+        veh_missions: Dict[str, Tuple[Mission, Mission]],
+    ):
+        for veh_id, (agent_id, agent_spec) in vehicles_to_trap.items():
+            # Create trap to be triggered immediately
+            try:
+                self.add_agent_with_mission(
+                    agent_id, agent_spec.interface, veh_missions[veh_id][0]
+                )
```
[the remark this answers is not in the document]
saulfield: Based on this feedback and discussion with Tucker we refactored this to do direct hijacking without the need for a trap, with a interface in `smarts.py`. We also generalized it to work with history vehicles automatically, and lifted the mission assignment logic out to the script itself.

Gamenot:
@RutvikGupta What is blocking this from being merged?

RutvikGupta:
> @RutvikGupta What is blocking this from being merged?

Another approval lol.

PR #1089, opened 2021-09-09
ROS reset refactor

sah-huawei:
Simple refactor to change the name of `SmartsControl` to `SmartsReset` to clarify that it is only used in reset situations.

Also, updated `test_smarts_ros.py` to re-send any test agents on reset.

Also fixed minor race condition related to reset in `ros_driver.py`.

### DIFF
--- a/examples/ros/src/smarts_ros/scripts/ros_driver.py
+++ b/examples/ros/src/smarts_ros/scripts/ros_driver.py
@@ -18,7 +18,7 @@ from smarts_ros.msg import (
     AgentsStamped,
     EntitiesStamped,
     EntityState,
-    SmartsControl,
+    SmartsReset,
 )
 from smarts_ros.srv import SmartsInfo, SmartsInfoResponse, SmartsInfoRequest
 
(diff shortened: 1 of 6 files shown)


examples/ros/src/smarts_ros/scripts/ros_driver.py
```diff
@@ -303,7 +303,7 @@ def _agent_spec_callback(self, ros_agent_spec: AgentSpec):
                 ),
             ),
         )
-        with self._control_lock:
+        with self._reset_lock:
```
Gamenot, line 306: It seems we generate a mission above outside of the lock. This means that the ordering of applying to self._agents_to_add is not deterministic if multiple `smarts_reset()` get called. It should be extremely unlikely to cause a problem unless there is already a problem of reset calls.

Gamenot (approved):

examples/ros/src/smarts_ros/scripts/ros_driver.py
sah-huawei: Good point!  I guess this could happen if someone double-clicks the Reset button (in  a hypothetical co-simulation environment that might have one).  But in that case, the agents being added will be the same and it shouldn't matter.  It could also happen, with different agents, if there are multiple "co-operating" clients trying to add agents at roughly the same time, but I think in such a situation the goal of determinism may not matter.  So I agree, I don't think it's worth the overhead to try to add a fix for this very-unlikely situation yet.

PR #1092, opened 2021-09-16
Project Along Route

sah-huawei:
Basic implementation of `project_along` for `RoadMap.Route` (as requested by Iman for BM work).

### DIFF
--- a/smarts/core/sumo_road_network.py
+++ b/smarts/core/sumo_road_network.py
@@ -30,7 +30,7 @@ from shapely.geometry import Polygon
 from shapely.ops import snap, triangulate
 from subprocess import check_output
 from trimesh.exchange import gltf
-from typing import List, Sequence, Tuple
+from typing import List, Set, Sequence, Tuple
 
 from .coordinates import BoundingBox, Heading, Point, Pose, RefLinePoint
 from .road_map import RoadMap, Waypoint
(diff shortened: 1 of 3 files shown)


smarts/core/sumo_road_network.py
```diff
@@ -794,6 +821,33 @@ def distance_between(self, start: Point, end: Point) -> float:
...
+                        continue
+                    started = True
+                    lane_pt = cand_start_lane.to_lane_coord(start)
+                    start_offset = lane_pt.s
+                else:
+                    start_offset = 0
+                if distance > road.length - start_offset:
+                    distance -= road.length - start_offset
+                    continue
+                return {(lane, distance) for lane in road.lanes}
```
liu-allan, line 848: How does this handle if there are forking roads to project along? ex.

image

From my understanding, it will return lanes of one of the forked roads in the example above

sah-huawei: With a `Route` object, there won't be forks because the whole route is determined from start to end.

However you question is a good one for the _other_ version of this in the `Lane` class.  And my assumption there was that the call to `waypoint_paths_at_offset()` there would deal with this correctly, but now I think there's a bit more work to do.  `waypoint_paths_at_offset()` for a lane gives you paths for all of the places you could go _from this lane only_, but it doesn't consider what would happen if you changed lanes to another lane on the same road first.  So I need to add that in.

So thanks for asking this question!

iman512003 (approved):

Gamenot (approved):

saulfield (approved):

smarts/core/sumo_road_network.py
```diff
@@ -387,6 +389,29 @@ def offset_along_lane(self, world_point: Point) -> float:
         def width_at_offset(self, offset: float) -> float:
             return self._width
 
+        @lru_cache(maxsize=8)
+        def project_along(
+            self, start_offset: float, distance: float
+        ) -> Set[Tuple[RoadMap.Lane, float]]:
+            lookahead = int(np.ceil(distance / self._map._lanepoints.spacing)) + 1
```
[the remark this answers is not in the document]
sah-huawei: removed!  good catch.

PR #1096, opened 2021-09-21
Python 3.8 fixes

saulfield:
This PR fixes some issues from using SMARTS with Python 3.8. The tests now pass with Python 3.8 and we can add this to the CI if we want.

- Added `unsafe_hash=True` to various `Goal` classes since they must be hashable, but can't be `frozen` because of 3.8-specific differences with dataclass inheritance
- Added a hash function that is consistent across Python versions for types that have their `id`s compared to an expected value in the sstudio tests

### DIFF
--- a/smarts/core/plan.py
+++ b/smarts/core/plan.py
@@ -60,7 +60,7 @@ class Start:
         )
 
 
-@dataclass
+@dataclass(unsafe_hash=True)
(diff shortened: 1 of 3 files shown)


Gamenot (approved):
This seems generally reasonable.

PR #1113, opened 2021-10-29
ROS logging

sah-huawei:
Added basic utils to route standard logging msgs to rospy.log.
Also quieted ros frequency warning in `ros_driver.py`.

### DIFF
--- a/examples/ros/src/smarts_ros/scripts/ros_driver.py
+++ b/examples/ros/src/smarts_ros/scripts/ros_driver.py
@@ -3,6 +3,7 @@
 from collections import deque
 import os
 import json
+import logging
 import math
 import rospy
 import sys
@@ -44,6 +45,8 @@ from smarts.core.utils.math import (
 from smarts.core.vehicle import VehicleState
 from smarts.zoo import registry
 
+from utils.ros import log_everything_to_ROS
+
 
 class ROSDriver:
     """Wraps SMARTS as a ROS (v1) node.
@@ -65,6 +68,7 @@ class ROSDriver:
         self._state_publisher = None
         self._agents_publisher = None
         self._most_recent_state_sent = None
+        self._warned_about_freq = False
         with self._state_lock:
             self._recent_state = deque(maxlen=3)
         with self._reset_lock:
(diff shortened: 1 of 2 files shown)


examples/ros/src/smarts_ros/scripts/ros_driver.py
```diff
@@ -640,9 +647,13 @@ def run_forever(self):
 
                 if self._target_freq:
                     if rate.remaining().to_sec() <= 0.0:
-                        rospy.logwarn(
-                            f"SMARTS unable to maintain requested target_freq of {self._target_freq} Hz."
-                        )
+                        msg = f"SMARTS unable to maintain requested target_freq of {self._target_freq} Hz."
```
Gamenot, line 650: I think it might be useful to log the timestamp if possible.

Gamenot (approved):

liu-allan (approved):

examples/ros/src/smarts_ros/scripts/ros_driver.py
```diff
@@ -121,6 +125,8 @@ def setup_ros(
             assert target_freq > 0.0
             self._target_freq = target_freq
 
+        log_everything_to_ROS(level=logging.WARNING)
```
liu-allan, line 128: if I want to log to ros in the bm code, is this all i need to call in the init of the agent? or would this need to be called in act

examples/ros/src/smarts_ros/scripts/ros_driver.py
sah-huawei: Actually, rospy log messages always include the (ros) timestamp as part of the message.  It's those numbers on the left prior to the message.

examples/ros/src/smarts_ros/scripts/ros_driver.py
sah-huawei: No, actually you shouldn't have to do anything.  This line here will apply to all existing and all future loggers too, including your agent's.  You can look at my current agent for an example of how I configure it's logging.  Once this PR gets merged, those log messages will show up both in the Smarts logfile and in the rospy stream to stdout.

PR #1118, opened 2021-11-03
Ignore `ros.py` in CI

Gamenot:
`glob` filter `ros.py` to fix `ci-base-tests` failure noted here: https://github.com/huawei-noah/SMARTS/runs/4053307155?check_suite_focus=true


sah-huawei (approved):
looks good to me.  thanks!

Adaickalavan (approved):

PR #1120, opened 2021-11-05
Fix CI failures

Gamenot:
This change-set moves `examples` to `smarts/examples` to fix the CI issue where the `aiohttp` package installs its own `examples` module.

- Other changes have made their way in as a way to fix the pipelines.

See CHANGELOG.md for list of changes.


Adaickalavan:
`aiohttp` was installed due to `ray`.
```
$ pipdeptree --reverse --packages aiohttp
aiohttp==3.8.0
  - aiohttp-cors==0.7.0 [requires: aiohttp>=1.1]
    - ray==1.0.1.post1 [requires: aiohttp-cors]
  - ray==1.0.1.post1 [requires: aiohttp]
```

Gamenot:
> 1. `aiohttp` was installed due to `ray`.
>    ```
>    $ pipdeptree --reverse --packages aiohttp
>    aiohttp==3.8.0
>      - aiohttp-cors==0.7.0 [requires: aiohttp>=1.1]
>        - ray==1.0.1.post1 [requires: aiohttp-cors]
>      - ray==1.0.1.post1 [requires: aiohttp]
>    ```
> 2. How did we deduce that the `examples` module conflict arose from `aiohttp` package?
> 3. Its strange. The `pytest ./tests/` appear to pass with `aiohttp==4.0.0a0` and `aiohttp==3.7.4`. But it fails with `aiohttp==3.8.0`.

1. Thanks, I did not dig into the origin of the dependency. It looks like we cannot get rid of it.
2. I took the `requirements.txt` and installed it into a virtual environment then checked in `venv/lib/python3.7/site-packages/examples` and found that aiohttp had installed it.
3. It looks like `aiohttp` accidentally included their examples directory in one version of their package. I think we should still move the examples directory under the `smarts` directory to put it in the `smarts` module.

saulfield (approved):

Makefile
```diff
@@ -129,9 +130,9 @@ clean:
 
 .PHONY: format
 format:
-	# pip install isort==5.7.0
+	echo "isort, version `isort --version-number`"
 	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/marl_benchmark/ ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./ultra ./zoo
-	# pip install black==20.8b1
+	black --version
```
RutvikGupta, line 135: If we are not depended on a particular version of black, then we should remove the version dependency from dev field in `setup.py` .

RutvikGupta (approved):
Looks good to merge!

examples/__init__.py
```diff
@@ -8,3 +8,6 @@ def required_to(cls, thing):
                You may not have installed the [train] or [test] dependencies required to run the ray dependent example.
                Install them first using the command `pip install -e .[train, test]` at the source directory to install the package ray[rllib]==1.0.1.post1"""
         )
+
+
+from . import argument_parser as argument_parser
```
Adaickalavan, line 13: Should the import statement be at the top of file?

envision/utils/multiprocessing_queue.py
```diff
@@ -25,6 +25,7 @@
 # Please refer to https://eli.thegreenplace.net/pages/about for more info about the author
 
 import multiprocessing
+import multiprocessing.queues
```
Adaickalavan, line 28: So we need both `import multiprocessing` and `import multiprocessing.queues`?

RutvikGupta: yes, we use both for this implementation.

Adaickalavan (approved):
Minicity scenario is being removed?

Gamenot:
> Minicity scenario is being removed?

No, that snuck in, I just noticed and restored it.

Makefile
Gamenot: We are currently dependent on a specific version of `black` but the intention here is to log that rather than comment it in. The version should only be indicated in the `setup.py` and `requirements.txt` if at all.

envision/utils/multiprocessing_queue.py
Gamenot: `multiprocessing.queues` is used to access `multiprocessing.queues.Queue` specifically.

Gamenot:
I intend to squash and merge this since the commits are noisy.

PR #1121, opened 2021-11-11
Observation space modifier in FrameStack wrapper

Adaickalavan:
1. Included observation space modifier in the frame stack wrapper.
2. If the `observation_space` of the underlying `env` is undefined, i.e., `None`, then the `observation_space` of `FrameStack` wrapper is also `None`.


smarts/env/wrappers/frame_stack.py
```diff
@@ -66,17 +81,29 @@ def step(
...
+        """Resets the environment.
+
+        Returns:
+            Dict[str, List[sensors.Observation]]: Observation upon reset for each agent.
+        """
         env_observations = super(FrameStack, self).reset()
         for agent_id, observation in env_observations.items():
             for _ in range(self._num_stack - 1):
-                self._frames[agent_id].append(observation)
+                self._frames[agent_id].appendleft(observation)
```
Gamenot, line 107: Is this for ordering observations like a queue?

Gamenot (approved):

smarts/env/wrappers/frame_stack.py
Adaickalavan: Yes, newer observation gets added to the front (left side) of the queue, instead of to the back (right side) of the queue.

PR #1124, opened 2021-11-12
Automatically update requirements.txt file

Adaickalavan:
1. Automatically updates the `requirements.txt` file using Github Actions when a pull request is made.
2. Authors do not need to update the `requirements.txt` file anymore.
3. Minimum package versions were specified for packages in `setup.py` which did not have any version specified previously. This helps reduce time spent by pip dependency resolver in [backtracking](https://pip.pypa.io/en/latest/topics/dependency-resolution/).


.github/workflows/ci-formatting-and-dependencies.yml
```diff
@@ -25,16 +25,29 @@ jobs:
...
+          add-apt-repository -y ppa:git-core/ppa && \
+            apt-get update && \
+            apt-get install -y git
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Test pip packages
+      - name: Update requirements
         run: |
           cd $GITHUB_WORKSPACE
-          . ./utils/setup/test_pip_packages.sh
```
saulfield, line 38: It looks like this script is no longer called anywhere. Should it be removed or updated?

.github/workflows/ci-formatting-and-dependencies.yml
```diff
@@ -25,16 +25,29 @@ jobs:
...
+          python3.7 -m venv ${{env.venv_dir}}
+          . ${{env.venv_dir}}/bin/activate
+          pip install --upgrade pip
+          pip install .[test,train,camera-obs]
+          pip freeze | grep -v 'smarts' | grep -v 'pkg-resources==0.0.0' > requirements.txt
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v7
+        with:
+          message: 'Update requirements.txt'
+          add: '*.txt'
```
saulfield, line 52: Is there a reason we can't just add requirements.txt explicitly here? I'm not sure if it could ever cause a problem, but a wildcard in an automated commit seems slightly dangerous :)

saulfield (approved):

.github/workflows/ci-formatting-and-dependencies.yml
```diff
@@ -25,16 +25,29 @@ jobs:
...
+  update-requirements-linux:
     runs-on: ubuntu-18.04
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     container: huaweinoah/smarts:v0.4.18-minimal
     steps:
+      - name: Install git
+        run: |
+          add-apt-repository -y ppa:git-core/ppa && \
+            apt-get update && \
+            apt-get install -y git
```
RutvikGupta, line 37: Why is git being installed?

RutvikGupta (approved):

.github/workflows/ci-formatting-and-dependencies.yml
Adaickalavan: Restricted automated commit to `git add requirements.txt`.

.github/workflows/ci-formatting-and-dependencies.yml
Adaickalavan: To automatically push changes, we need to git add, git commit, and git push in the ci job. Hence, we need to install git, as our smarts minimal container does not have git pre-installed.

Gamenot:
What is blocking this PR?

Gamenot:
We talked and found that the automatic updates should only occur in develop.

.github/workflows/ci-formatting-and-dependencies.yml
Adaickalavan: The `/utils/setup/test_pip_packages.sh` file is removed.

Gamenot:
I wonder if the `license/cla` test will even trigger on a commit outside of a merge request.

.github/workflows/ci-dependencies-mac.yml
```diff
@@ -0,0 +1,40 @@
+name: SMARTS CI Dependencies Mac
+
+on:
+  schedule:
+    - cron: '0 23 * * 1'
+      # Runs at 11.00pm, UTC, every Monday
```
Gamenot, line 6: I think this could get into a race condition with the `SMARTS CI Base Tests Mac` action. This action will either need to trigger the `SMARTS CI Base Tests Mac` action or run at some time comfortably before.

Gamenot (changes requested):

.github/workflows/ci-dependencies-mac.yml
Adaickalavan: Modified workflows such that completion of Mac requirements update will trigger Mac base tests.

Gamenot (approved):

Issue #945, opened 2021-06-16
Standard Gym Environment


Gamenot:
**Is your feature request related to a problem? Please describe.**
The current SMARTS environments do not completely conform to Gym API.

**Describe the solution you'd like**
The intention here is to provide a Gym standard environment for single agent evaluation. In this case the `action_space` and `observation_space` should contained within the environment and required at initialisation. And `action` provided to `step` should be provided as a single action as the agent should only be a single agent.

**Additional context**
https://github.com/openai/gym/blob/ee5ee3a4a5b9d09219ff4c932a45c4a661778cd7/gym/core.py#L8

PR #1127, opened 2021-11-19
Single-agent env wrapper

Adaickalavan:
1. Converts a single-agent SMARTS environment's `step` and `reset` output to be compliant with gym spaces.
2. Helps users to easily train or evaluate SMARTS with existing single-agent reinforcement-learning algorithms.

### DIFF
--- a/smarts/env/wrappers/single_agent.py
+++ b/smarts/env/wrappers/single_agent.py
@@ -1,0 +1,72 @@
+# MIT License
+#
+# Copyright (C) 2021. Huawei Technologies Co., Ltd. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


smarts/env/wrappers/single_agent.py
```diff
@@ -0,0 +1,72 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from typing import Any, Tuple
+
+import gym
+
+
+class SingleAgent(gym.Wrapper):
```
Gamenot, line 28: OK, I think we will want to use this explicitly in the [single_agent.py](https://github.com/huawei-noah/SMARTS/blob/wrapper-single-agent/examples/single_agent.py) example.

Gamenot (commented):
Also this is related to #945.

smarts/env/wrappers/single_agent.py
Adaickalavan: Currently, `single_agent.py` illustrates the multi-agent interface of SMARTS. The `episode.record_step()` on line 67 works on dictionary inputs for multi-agents. 

I attempted to change `single_agent.py` to use `SingleAgent` wrapper. Using `SingleAgent` wrapper makes the code look like normal single-agent gym-compliant code. But line 67 will then be written as:
episode.record_step({AGENT_ID:obs}, {AGENT_ID:reward}, {AGENT_ID:done, “__all__”:done}, {AGENT_ID:info}) 

1) So should we proceed with using `SingleAgent` wrapper in `single_agent.py` example? 
2) If we are using `SingleAgent` wrapper in `single_agent.py`, shall we just remove the `episode.record_step()` functionality in order to streamline the example?

Gamenot: > 1. Should we proceed with using SingleAgent wrapper in single_agent.py example?

Yes, I think we should still do so.

> 2. If we are using `SingleAgent` wrapper in `single_agent.py`, shall we just remove the` episode.record_step()` functionality in order to streamline the example?

The use of record step is just for visual feedback when running the example. I am not sure if we can get rid of it, I could suggest we can make an overload or simplified forwarding version of `record_step()`.

Adaickalavan: Introduced `SingleAgent` wrapper into `single_agent.py` example.

Gamenot (approved):

sah-huawei (approved):

PR #1128, opened 2021-11-19
RGB-image env wrapper

Adaickalavan:
1. Filters SMARTS environment observation and returns only top-down RGB image as observation.
2. If the base env's observation space is frame stacked using the `FrameStack` wrapper, then the returned observation will be a numpy array of stacked images with shape (image_width, image_height, 3*`num_stack`). Here, `num_stack` refers to the number of frames stacked in the base env's observation.
3. Helps users to easily transform SMARTS observation space to gym-compliant numpy array of stacked images.


smarts/env/wrappers/rgb_image.py
```diff
@@ -0,0 +1,88 @@
...
+    def observation(self, obs: Dict[str, Any]) -> Dict[str, np.ndarray]:
+        wrapped_obs = {}
+        for agent_id, agent_obs in obs.items():
+            if self._num_stack == 1:
+                agent_obs = [agent_obs]
+
+            images = []
+            for agent_ob in agent_obs:
+                image = agent_ob.top_down_rgb.data
+                images.append(image.astype(np.uint8))
```
Gamenot, line 83: As far as I can tell, `num_stack` is intended cap the number of used frames from the frame stack. I think this limit should be applied here.

Gamenot (changes requested):

smarts/env/wrappers/rgb_image.py
Adaickalavan: A user could have stacked the wrappers in either of the following two approaches.

Approach1:
`HiwayEnv (obs_space: None) -> FrameStack (obs_space: None) -> RGBImage (obs_space: {agent_id: Box(w,h,3*num_stack)})`

Approach2:
`HiwayEnv (obs_space: None) -> RGBImage (obs_space: {agent_id: Box(w,h,3*1)})`

Previously in Approach1, it was difficult for `RGBImage` wrapper to know the underlying env’s `obs_space` since it is `None`. Hence, the need to feed in `num_stack` argument explicitly to `RGBImage` wrapper so that `RGBImage` wrapper outputs the proper `obs_space`.

In Approach2, `num_stack` argument to `RGBImage` would simply be fed as 1 by the user.

Going forward, I have changed `HiWayEnv`’s observation space. Additionally, I have changed the `FrameStack` wrapper’s observation space to `obs_space: {agent_id: Tuple([gym.Space]*num_stack)}`. With this change, the correct `num_stack` parameter would be auto deduced by `RGBImage` wrapper and the users do not need to feed this value to `RGBImage` wrapper anymore.

The changes for `HiWayEnv` and `Framestack` are in pr #1121. The necessary changes for this `RGBImage` wrapper is not made yet as it is dependent on the pr #1121 being merged first.

Gamenot: Sure, yes, I was commenting on the mismatch between the `FrameStack` and `RGBImage` like with the following case:

```python
fs=FrameStack(env, num_stack=7)
RGBImage(fs, num_stack=3)
```

I think it should fail as early and clearly as possible which is why I think it should be checked or capped here (behaviour documented) to make sure the output matches the observation space rather than taking as many frames as the `FrameStack` will give.

smarts/env/wrappers/rgb_image.py
```diff
@@ -0,0 +1,88 @@
...
+                )
+                for agent_id in agent_specs.keys()
+            }
+        )
+
+    def observation(self, obs: Dict[str, Any]) -> Dict[str, np.ndarray]:
+        wrapped_obs = {}
+        for agent_id, agent_obs in obs.items():
+            if self._num_stack == 1:
+                agent_obs = [agent_obs]
```
Gamenot, line 78: My one other concern might be that `FrameStack` with `num_stack=1` can cause an edge case.

Adaickalavan: `FrameStack` wrapper asserts to disallow `num_stack == 1`. 
https://github.com/huawei-noah/SMARTS/blob/c3b3ccc65879d329216128dc42483e6c2df799b2/smarts/env/wrappers/frame_stack.py#L40

smarts/env/wrappers/rgb_image.py
Adaickalavan: Ensured user supplied `num_stack` matches true num stack value.

Gamenot (approved):

PR #1129, opened 2021-11-19
Minor changes

Adaickalavan:
Minor changes
+ format code
+ improve install script
+ update outdated url
+ remove unnecessary imports


utils/setup/install_deps.sh
```diff
@@ -27,18 +27,18 @@ function set_SUMO_HOME {
...
          libspatialindex-dev \
          sumo sumo-tools sumo-doc \
          build-essential cmake
 
     #only a problem for linux
     if ! check_python_version_gte_3_7; then
          echo "A >=3.7 python version not found"
-         if  [[ "$1" = "-y" ]]; then
-              install_python_3_7
+         if  [[ $1 = "-y" ]]; then
```
Gamenot, line 40: This is fine to do although it is still fine to use `"$1"` as habit so there is not an error when empty in single square bracket comparison.

Gamenot (approved):

utils/setup/install_deps.sh
RutvikGupta: The quotation is what bash script recommends but both work.

RutvikGupta (approved):

utils/setup/install_deps.sh
Adaickalavan: Reverted to using double quotes for comparing variable to string.

PR #1132, opened 2021-11-21
Misc. Optimizations

sah-huawei:
(These came out of profiling.)

- Don't create a new `Pose` object on every pybullet substep.
- Cache the results of pybullet API calls during each step.
- Cache conversions of our `Point` objects to Shapely `Point` objects (which are expensive to create).
- Avoid unnecessary work when there are no vehicles served by the `MotionPlannerProvider`.


sah-huawei:
Although I can't reproduce it locally, the failing CI test seems to be a "real failure".  So I will have another push soon to fix that.  In the meantime, you can hold off on reviewing this.  Thanks.

sah-huawei:
> Although I can't reproduce it locally, the failing CI test seems to be a "real failure". So I will have another push soon to fix that. In the meantime, you can hold off on reviewing this. Thanks.

OK.  It should be good to go now.

smarts/core/chassis.py
```diff
@@ -629,20 +651,30 @@ def control(self, throttle=0, brake=0, steering=0):
...
             self._lat_forces, self._lon_forces = self._tire_model.apply_tire_forces(
                 self,
                 self.bullet_client,
                 [(1 / self._max_torque) * np.array(throttle_list), brake, steering],
             )
+            self._clear_step_cache()
             return
         self._apply_throttle(throttle_list)
         self._apply_brake(brake)
+        self._clear_step_cache()
```
Gamenot, line 677: Is it necessary to invalidate the cache when applying immediate forces? These forces do not have effect until after a physics step.

smarts/core/coordinates.py
```diff
@@ -65,11 +66,30 @@ def equal_if_defined(self, length: float, width: float, height: float) -> bool:
...
+    def as_shapely(self) -> SPoint:
+        # Shapley Point construction is expensive!
+        # Note that before python3.8, @cached_property was not thread safe,
+        # nor can it be used in a NamedTuple (which doesn't have a __dict__).
+        # (Points can be used by multi-threaded client code, even when
+        # SMARTS is still single-threaded, so we want to be safe here.)
+        # So we use the private global _shapely_points as a cache instead.
+        # Here we are relying on CPython's implementation of dict and setdefault
+        # to be thread-safe, which is apparently the case right now.
+        return _shapely_points.setdefault(self, SPoint((self.x, self.y, self.z)))
```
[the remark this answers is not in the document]
sah-huawei: doh!  you're right, good catch!
This shows the insidious nature of optimizations.  I had done it -- and profiled it in with the agents -- originally with an if/else block where the `SPoint` was only constructed on one of the paths, but then when I was adding the comment about dicts being threadsafe, I thought about how ugly it looked and that `setdefault` would look so much nicer, so I just changed it without actually testing it again.  I am ashamed!  

Anyway, I will change it back to the (uglier-but-faste) if-else-block.  Thanks!

smarts/core/chassis.py
sah-huawei: Yes, I think so.  I followed the principle here of invalidating the cache whenever a mutating pybullet API call is made (or in this case, doing it only once after the 3 were made above to steering, throttle and brake).   The reason is because at this point, it's not always clear from the pybullet documentation whether subsequent API calls to pybullet will return the same value or not -- i.e., pybullet's documentation doesn't make it clear _which_ of its API calls are only affected by stepping and which ones "take immediate effect".  In this uncertainty, it seems necessary to err on the side of caution.

The cache is still useful though for "downstream" accesses to things like `pose` that happen frequently on a single step from multiple points in SMARTS and shouldn't require repeated pybullet API calls.

smarts/core/coordinates.py
```diff
@@ -208,6 +228,20 @@ def __eq__(self, other) -> bool:
...
+            # which can be a problem if it was initialized with ints and
+            # now we are assigning it floats, so we just cast it...
+            self.position = np.float64(self.position)
+        self.position[:] = position
+        if "point" in self.__dict__:
+            # clear the cached_property
+            del self.__dict__["point"]
+        if heading != self.heading_:
+            self.orientation = fast_quaternion_from_angle(heading)
+            self.heading_ = heading
```
Gamenot, line 243: Is the postfix `_` a convention? I am not quite familiar with it.

sah-huawei: I've seen it for some codebases, but it's less common.
I don't know why it was used here (`self.heading_` was like this from long ago, I believe -- I definitely didn't add it myself).

smarts/core/chassis.py
sah-huawei: lol, I don't know what the eyes emoji means!  (feeling old)

smarts/core/smarts.py
```diff
@@ -735,7 +736,6 @@ def _nondynamic_provider_step(
 
             vehicle = self._vehicle_index.vehicle_by_id(vehicle_id)
             assert isinstance(vehicle.chassis, BoxChassis)
-            vehicle.step(self._elapsed_sim_time)
```
Gamenot, line 738: I am not entirely convinced it is reasonable to move `vehicle.step()` to after generating the `VehicleState`. It would appear to me that the most recent state should rely on a step having been performed.

smarts/core/chassis.py
Gamenot: I am not sure either, I could not find an 👌 so I just used eyes.

smarts/core/smarts.py
sah-huawei: The reason I moved it is because it needs to happen whenever `pybullet` is stepped and having it happen in different locations runs the risk of it being missed.  Calling it is now a stronger "requirement" because the "step cache" in `chassis.py` needs to be cleared after each step, and this is the way to trigger that.

One effect of moving it into the `_step_pybullet()` function is to make it happen slightly earlier than it would have before.  Before, we waited to call it until after collisions were processed, but this does not seem to be problematic.  (Note that I think that, even before, it has always been called before generating the `VehicleState` object.)

However, another effect of moving it is that it is now called for _every_ vehicle, as opposed to just those with agents controlling them (due to moving it to before the `continue` in `_pybullet_provider_step()`).  I looked into this when I did it and it didn't seem problematic  to me, but I definitely think this decision should be reviewed.  (The `_has_stepped` state in `vehicle.py` just seems to be for controlling a warning message.)

smarts/core/chassis.py
```diff
@@ -440,26 +441,46 @@ def __init__(
...
-        return pose
+        if not self._pose:
+            self._pose = Pose.from_explicit_offset(
+                [0, 0, 0],
+                np.array(pos),
+                heading,
+                local_heading=Heading(0),
+            )
+        else:
+            self._pose.reset_with(pos, heading)
```
sah-huawei, line 472: Of all of the changes in this PR, this one was probably the most risky.
Instead of creating a new `Pose` object on every (sub)step of pybullet (indeed, on every single call to the `pose` property!), we now reuse the same object over and over.  

This is a huge win in terms of efficiency and performance, but it carries that risk that anywhere that a `Pose` object reference was stored and kept for later use might now be dealing with a changed object!

I did a bit of an audit of this, and I think things are ok (all of the tests now pass), but it was the reason for the ~2 test failures that I encountered along the way.  This is why it was important to actually _copy_ the `Pose` and/or its `position` field before returning it in the `Observations` class.

The change made to also do this before emitting states to envision was technically un-necessary, but was made necessary by the way the data replay test was written (specifically the way the websocket was mocked), but I made it anyway, just to be consistent.

smarts/core/chassis.py
```diff
@@ -629,20 +651,30 @@ def control(self, throttle=0, brake=0, steering=0):
         assert 0 <= brake <= 1, f"brake ({brake}) must be in [0, 1]"
         assert -1 <= steering <= 1, f"steering ({steering}) must be in [-1, 1]"
 
+        # If we apply brake at low speed using reverse torque
+        # the vehicle starts to roll back. we need to apply a condition
+        # on brake such that, the reverse torque is only applied after
+        # a threshold is passed for vehicle velocity.
+        # Thus, brake is applied if: vehicle speed > 1/36 (m/s)
+        if brake > 0 and self.longitudinal_lateral_speed[0] < 1 / 36:
```
sah-huawei, line 659: Interestingly, probably the most important optimization in this entire PR was the addition of the `brake > 0` check above. 

This prevents the call to `longitudinal_lateral_speed` (which makes a pybullet API call) -- as well as the subsequent call to `pose` -- in the majority of cases.  (Vehicles rarely hit the brakes, lol!)  Because `control()` is called for every pybullet _substep_, this is a _huge_ win.

Gamenot (approved):

smarts/core/chassis.py
Gamenot: The external mutability of the pose object is a bit concerning but I do agree that it is a good win for efficiency.

sah-huawei:
@saulfield :  it'd be great if you could look at this one soon.

smarts/core/coordinates.py
```diff
@@ -65,11 +66,35 @@ def equal_if_defined(self, length: float, width: float, height: float) -> bool:
...
+    @property
+    def as_shapely(self) -> SPoint:
+        # Shapley Point construction is expensive!
+        # Note that before python3.8, @cached_property was not thread safe,
+        # nor can it be used in a NamedTuple (which doesn't have a __dict__).
+        # (Points can be used by multi-threaded client code, even when
+        # SMARTS is still single-threaded, so we want to be safe here.)
+        # So we use the private global _shapely_points as a cache instead.
+        # Here we are relying on CPython's implementation of dict
+        # to be thread-safe.
```
saulfield, line 86: Just checking -- are we sure we don't need to worry about this for our other uses of `@cached_property`?

saulfield (approved):

smarts/core/coordinates.py
sah-huawei: > Just checking -- are we sure we don't need to worry about this for our other uses of `@cached_property`?

lol, no!

I am fairly confident (above 95%) that the vast majority of our uses of `@cached_property` are ok, as SMARTS is only supposed to be run with a single thread anyway.  However,  internal uses that come via the ROS wrapper can possibly made on other threads, so I did audit those uses (as well as uses in an alternative `RoadMap` implementation that has background processes -- all uses there were converted to `@threaded_cached_property`).  

This is partly why I added the comment above, in order to remind myself and others every time I look at this code, in case we encounter some weirdness down the road that might be related to an unsafe use of `@cached_property`.  

But while at this point, I am still not 100% confident that we have managed this appropriately, I do thin the above (including the comment) is an improvement over what we had originally, so I'm willing to go ahead and merge the PR now.

Gamenot: I say go ahead when you have the chance.

Gamenot:
It looks like we may need to regenerate the `test_data_replay` generated file.

Issue #803, opened 2021-04-29
Connection closed by SUMO when doing imitation learning


zbzhu99:
## **BUG REPORT**
**High Level Description**
SMARTS will raise the "Connection closed by SUMO" error when doing imitation learning experiments.

**SMARTS version**
0.4.15

**Steps to reproduce the bug**
When running imitation learning experiments, we need to replay the same history data multiple times. But from my test, it will cause the connection error of SUMO when we do the data replay more than one time.

Add a loop to run the following code twice:

https://github.com/huawei-noah/SMARTS/blob/6cb8ffda74e235c25d63b74313623b2e03e402f7/examples/history_vehicles_replacement_for_imitation_learning.py#L28-L60

And the error will be raised on the (about) 335th agent at the second loop time.

**System information**
 
- Operating system & Environment [Please note if using Docker]: **outside** Docker
- Number of system cores available
- Python version 3.7.10
- SUMO version 1.8.0
 
**Impact [If known]**

sah-huawei:
Since this seems to be a crash within Sumo (or Traci), and since Sumo isn't needed when replaying traffic histories, we are going to try simply bypassing Sumo traffic for now.

(Of course, we should eventually circle back and try to understand why Sumo is crashing.)

Issue #295, opened 2020-12-10
connection closed by SUMO


BBDrive:
When I run multiple instances with `ray`, it gives an error.
```(pid=26763) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26758) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26765) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26767) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26759) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26750) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26763) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26753) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26766) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26752) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26759) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26758) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26765) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26750) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26767) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26752) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26766) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26753) ERROR:Zoo Worker:Failure while handling connection EOFError()
(pid=26760) ERROR:RemoteAgentBuffer:Waiting for local zoo worker to start up, retrying 0 / 3
(pid=26760) ERROR:Zoo Worker:Failure while handling connection EOFError()
```
But it can still work for a few rounds. After running for a while, it crashed.
```
(pid=26763) ERROR:SMARTS:Simulation crashed with exception. Attempting to cleanly shutdown.
[...]

Issue #158, opened 2020-11-13
Running the maddpg benchmark for a long time results in a traci error.


rsuwa:
## Reproduction

If you run the task for a long time (1-2 hours), there is a high probability that the following error will occur.

## Command

`python run.py scenarios/intersections/4lane -f agents/maddpg/baseline-lane-control.yaml`

## Full logs

```bash
Failure # 1 (occurred at 2020-11-13_14-32-54)
Traceback (most recent call last):
  File "/home/ryota/src/github.com/huawei-noah/SMARTS/benchmark/.venv/lib/python3.7/site-packages/ray/tune/trial_runner.py", line 726, in _process_trial
    result = self.trial_executor.fetch_result(trial)
  File "/home/ryota/src/github.com/huawei-noah/SMARTS/benchmark/.venv/lib/python3.7/site-packages/ray/tune/ray_trial_executor.py", line 489, in fetch_result
    result = ray.get(trial_future[0], timeout=DEFAULT_GET_TIMEOUT)
  File "/home/ryota/src/github.com/huawei-noah/SMARTS/benchmark/.venv/lib/python3.7/site-packages/ray/worker.py", line 1452, in get
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(FatalTraCIError): [36mray::MADDPG2.train()[39m (pid=8802, ip=192.168.10.106)
  File "python/ray/_raylet.pyx", line 482, in ray._raylet.execute_task
  File "python/ray/_raylet.pyx", line 436, in ray._raylet.execute_task.function_executor
  File "/home/ryota/src/github.com/huawei-noah/SMARTS/benchmark/.venv/lib/python3.7/site-packages/ray/rllib/agents/trainer.py", line 517, in train
    raise e
  File "/home/ryota/src/github.com/huawei-noah/SMARTS/benchmark/.venv/lib/python3.7/site-packages/ray/rllib/agents/trainer.py", line 506, in train
    result = Trainable.train(self)
  File "/home/ryota/src/github.com/huawei-noah/SMARTS/benchmark/.venv/lib/python3.7/site-packages/ray/tune/trainable.py", line 336, in train
    result = self.step()
  File "/home/ryota/src/github.com/huawei-noah/SMARTS/benchmark/.venv/lib/python3.7/site-packages/ray/rllib/agents/trainer_template.py", line 147, in step
    res = next(self.train_exec_impl)
[...]

PR #1138, opened 2021-11-25
Graceful handle of TraCI connection errors

Gamenot:
This is an attempt to gracefully handle TraCI connection errors until the root issue is fixed.

See(for example):
https://github.com/huawei-noah/SMARTS/issues/158
https://github.com/huawei-noah/SMARTS/issues/295
https://github.com/huawei-noah/SMARTS/issues/803

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ test: build-all-scenarios
 		--doctest-modules \
 		--forked \
 		--dist=loadscope \
-		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
+		-n `nproc --ignore 2` \
 		--nb-exec-timeout 65536 \
 		./examples/tests ./smarts/env ./envision ./smarts/contrib ./smarts/core ./smarts/sstudio ./tests \
 		--ignore=./smarts/core/tests/test_smarts_memory_growth.py \
@@ -26,7 +26,7 @@ sanity-test: build-all-scenarios
 		--forked \
 		--dist=loadscope \
 		--junitxml="sanity_test_result.xml" \
-		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
+		-n `nproc --ignore 2` \
 		./smarts/core/tests/test_python_version.py::test_python_version \
 		./smarts/core/tests/test_sumo_version.py::test_sumo_version \
 		./smarts/core/tests/test_dynamics_backend.py::test_set_pose \
@@ -45,7 +45,7 @@ test-memory-growth: build-all-scenarios
 		--cov=smarts \
 		--forked \
 		--dist=loadscope \
-		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
+		-n `nproc --ignore 1` \
 		./smarts/core/tests/test_smarts_memory_growth.py
 	rm -f .coverage.*
 	rm -f .coverage*
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ test: build-all-scenarios
 		--dist=loadscope \
 		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
 		--nb-exec-timeout 65536 \
-		./examples/tests ./envision ./smarts/contrib ./smarts/core ./smarts/env ./smarts/sstudio ./tests \
+		./examples/tests ./smarts/env ./envision ./smarts/contrib ./smarts/core ./smarts/sstudio ./tests \
 		--ignore=./smarts/core/tests/test_smarts_memory_growth.py \
 		--ignore=./smarts/core/tests/test_env_frame_rate.py \
 		--ignore=./smarts/env/tests/test_benchmark.py \
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,9 @@ test: build-all-scenarios
 		--doctest-modules \
 		--forked \
 		--dist=loadscope \
-		-n `nproc --ignore 2` \
-		--nb-exec-timeout -1 \
-		./envision ./smarts/contrib ./smarts/core ./smarts/env ./smarts/sstudio ./tests ./examples/tests \
+		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
+		--nb-exec-timeout 65536 \
+		./examples/tests ./envision ./smarts/contrib ./smarts/core ./smarts/env ./smarts/sstudio ./tests \
 		--ignore=./smarts/core/tests/test_smarts_memory_growth.py \
 		--ignore=./smarts/core/tests/test_env_frame_rate.py \
 		--ignore=./smarts/env/tests/test_benchmark.py \
@@ -26,7 +26,7 @@ sanity-test: build-all-scenarios
 		--forked \
 		--dist=loadscope \
 		--junitxml="sanity_test_result.xml" \
-		-n `nproc --ignore 2` \
+		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
 		./smarts/core/tests/test_python_version.py::test_python_version \
 		./smarts/core/tests/test_sumo_version.py::test_sumo_version \
 		./smarts/core/tests/test_dynamics_backend.py::test_set_pose \
@@ -45,7 +45,7 @@ test-memory-growth: build-all-scenarios
 		--cov=smarts \
 		--forked \
 		--dist=loadscope \
-		-n `nproc --ignore 1` \
+		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
 		./smarts/core/tests/test_smarts_memory_growth.py
 	rm -f .coverage.*
 	rm -f .coverage*
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -214,10 +214,7 @@ class AgentManager:
                 scores[agent_id] = vehicle.trip_meter_sensor()
 
         if sim.should_reset:
-            dones = {
-                agent_id: True
-                for agent_id in self.agent_ids
-            }
+            dones = {agent_id: True for agent_id in self.agent_ids}
             dones["__sim__"] = True
 
         return observations, rewards, scores, dones
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -218,6 +218,7 @@ class AgentManager:
                 agent_id: True
                 for agent_id in self.agent_ids
             }
+            dones["__sim__"] = True
 
         return observations, rewards, scores, dones
 
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -213,6 +213,12 @@ class AgentManager:
                 rewards[agent_id] = vehicle.trip_meter_sensor(increment=True)
                 scores[agent_id] = vehicle.trip_meter_sensor()
 
+        if sim.should_reset:
+            dones = {
+                agent_id: True
+                for agent_id in self.agent_ids
+            }
+
         return observations, rewards, scores, dones
 
     def _vehicle_reward(self, vehicle_id, sim):
--- a/smarts/core/agent_manager.py
+++ b/smarts/core/agent_manager.py
@@ -508,6 +508,7 @@ class AgentManager:
         for agent_id in ids_:
             self._agent_interfaces.pop(agent_id, None)
 
+        self._pending_agent_ids = self._pending_agent_ids - ids_
         return ids_
 
     def reset_agents(self, observations):
--- a/smarts/core/bubble_manager.py
+++ b/smarts/core/bubble_manager.py
@@ -30,7 +30,7 @@ from shapely.affinity import rotate, translate
 from shapely.geometry import CAP_STYLE, JOIN_STYLE, Point, Polygon
 
 from smarts.core.data_model import SocialAgent
-from smarts.core.plan import Mission, Plan, PositionalGoal, Start
+from smarts.core.plan import EndlessGoal, Mission, Plan, PositionalGoal, Start
 from smarts.core.road_map import RoadMap
 from smarts.core.utils.id import SocialAgentId
 from smarts.core.utils.string import truncate
@@ -550,9 +550,13 @@ class BubbleManager:
         # Setup mission (also used for observations)
         # XXX: this is not quite right.  route may not be what the agent wants to take.
         route = sim.traffic_sim.vehicle_route(vehicle_id=vehicle.id)
+        if len(route) > 1:
+            goal = PositionalGoal.from_road(route[-1], sim.scenario.road_map)
+        else:
+            goal = EndlessGoal()
         mission = Mission(
             start=Start(vehicle.position[:2], vehicle.heading),
-            goal=PositionalGoal.from_road(route[-1], sim.scenario.road_map),
+            goal=goal
         )
         plan.create_route(mission)
 
--- a/smarts/core/bubble_manager.py
+++ b/smarts/core/bubble_manager.py
@@ -554,10 +554,7 @@ class BubbleManager:
             goal = PositionalGoal.from_road(route[-1], sim.scenario.road_map)
         else:
             goal = EndlessGoal()
-        mission = Mission(
-            start=Start(vehicle.position[:2], vehicle.heading),
-            goal=goal
-        )
+        mission = Mission(start=Start(vehicle.position[:2], vehicle.heading), goal=goal)
         plan.create_route(mission)
 
     def _start_social_agent(
--- a/smarts/core/bubble_manager.py
+++ b/smarts/core/bubble_manager.py
@@ -550,7 +550,7 @@ class BubbleManager:
         # Setup mission (also used for observations)
         # XXX: this is not quite right.  route may not be what the agent wants to take.
         route = sim.traffic_sim.vehicle_route(vehicle_id=vehicle.id)
-        if len(route) > 1:
+        if len(route) > 0:
             goal = PositionalGoal.from_road(route[-1], sim.scenario.road_map)
         else:
             goal = EndlessGoal()
--- a/smarts/core/provider.py
+++ b/smarts/core/provider.py
@@ -26,7 +26,7 @@ from .scenario import Scenario
 from .vehicle import VehicleState
 
 
-class ProviderRecoveryOptions(IntFlag):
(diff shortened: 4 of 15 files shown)


smarts/core/bubble_manager.py
```diff
@@ -563,9 +563,13 @@ def _prepare_sensors_for_agent_control(
         # Setup mission (also used for observations)
         # XXX: this is not quite right.  route may not be what the agent wants to take.
         route = sim.traffic_sim.vehicle_route(vehicle_id=vehicle.id)
+        if len(route) > 1:
+            goal = PositionalGoal.from_road(route[-1], sim.scenario.road_map)
+        else:
+            goal = EndlessGoal()
```
RutvikGupta, line 569: What's the relevance of this with TraCI connection issue?

smarts/core/sumo_traffic_simulation.py
```diff
@@ -260,6 +265,7 @@ def setup(self, next_scenario) -> ProviderState:
         # restart sumo process only when map file changes
         restart_sumo = (
             not self._scenario
+            or self.connected == False
```
RutvikGupta, line 268: where is `self.connected` being set?

smarts/core/bubble_manager.py
Gamenot: The route is returned from the traffic simulation and can now default to `[]`. Thanks for pointing it out because I realise now that the length check for the first case should be `len(route)>0`.

Adaickalavan:
What is the desired course of action when TraCI error happens (e.g., terminate, restart connection)?

Gamenot:
> What is the desired course of action when TraCI error happens (e.g., terminate, restart connection)?

@Adaickalavan From conversation, the intention would be to gracefully end the episode.

smarts/core/sumo_traffic_simulation.py
```diff
@@ -331,6 +354,17 @@ def teardown(self):
...
+    @property
+    def connected(self):
+        return self._traci_conn != None
+
+    @property
+    def recovery_options(self) -> ProviderRecoveryOptions:
+        return (
+            ProviderRecoveryOptions.EPISODE_REQUIRED
+            | ProviderRecoveryOptions.ATTEMPT_RECOVERY
+        )
```
RutvikGupta, line 366: This means if `SUMO TraCI` connection fails, we need to rerun that episode or try to attempt to reconnect to `TraCI`?`

smarts/core/sumo_traffic_simulation.py
```diff
@@ -692,13 +748,15 @@ def _reroute_vehicles(self, vehicle_states):
             new_route_edges = route_edges[-1:] + route_edges
             self._traci_conn.vehicle.setRoute(vehicle_id, new_route_edges)
```
RutvikGupta, line 750: Should the error handling about traci connection be done here and for any reference call to `self._traci_conn` in this file.
```python
if not self.connected:
            return []
        try:
            self._traci_conn.vehicle.setRoute(vehicle_id, new_route_edges)
        except self._traci_exceptions as e:
````

File: smarts/core/sumo_traffic_simulation.py [hunk 5]
```diff
@@ -491,7 +542,12 @@ def _move_vehicle(self, vehicle_id, position, heading, speed):
         self._traci_conn.vehicle.setSpeed(vehicle_id, speed)
```
RutvikGupta, line 543: Connection error handling here as well?

smarts/core/sumo_traffic_simulation.py
Gamenot: It will attempt  `SumoTrafficSimulation.recover()` for now which just currently disconnects or re-throws the exception from the provider.

smarts/core/sumo_traffic_simulation.py
Gamenot: I wrapped these calls at a higher level. I only do exception handling and connection checks at the public method body level to reduce the amount of checks needed.

RutvikGupta (approved):

RutvikGupta (commented):
I think this should be now rebased on `develop` since its an important change.

smarts/core/sumo_traffic_simulation.py
```diff
@@ -260,6 +270,7 @@ def setup(self, next_scenario) -> ProviderState:
         # restart sumo process only when map file changes
         restart_sumo = (
             not self._scenario
+            or self.connected == False
```
sah-huawei, line 273: just `or not self.connected` would be better

smarts/core/sumo_traffic_simulation.py
```diff
@@ -331,6 +354,17 @@ def teardown(self):
         self._to_be_teleported = dict()
         self._reserved_areas = dict()
 
+    @property
+    def connected(self):
+        return self._traci_conn != None
```
sah-huawei, line 359: Prefer:  `return self._traci_conn is not None`

smarts/core/provider.py
```diff
@@ -76,3 +88,50 @@ def reset(self):
...
+        """
+        return True
+
+    @property
+    def recovery_options(self) -> ProviderRecoveryOptions:
+        """If this provider must respond each step for an episode to continue."""
+        return ProviderRecoveryOptions.EXPERIMENT_REQUIRED
+
+
+class EmptyProvider(Provider):
```
sah-huawei, line 108: Another option is to add some default implementations to the base class, `Provider`, and then call those using `super` as needed in the derived provider classes.  I think I might prefer that, although I don't feel that strongly about it.

smarts/core/provider.py
```diff
@@ -76,3 +88,50 @@ def reset(self):
...
+    def connected(self):
+        """Determine if the provider is still responsive. (e.g. the case that the provider is
+        sending provider state over the internet and has stopped responding)
+        """
+        return True
+
+    @property
+    def recovery_options(self) -> ProviderRecoveryOptions:
+        """If this provider must respond each step for an episode to continue."""
+        return ProviderRecoveryOptions.EXPERIMENT_REQUIRED
```
sah-huawei, line 105: I don't think it's possible (or desirable) for the _provider itself_ to know whether or not it's "necessary" (and how) in the context of the overall simulation.  That may depend on things outside of its scope and control.

Hence, I don't think this is in the right place.

sah-huawei (changes requested):

Makefile
```diff
@@ -45,7 +45,7 @@ test-memory-growth: build-all-scenarios
 		--cov=smarts \
 		--forked \
 		--dist=loadscope \
-		-n `nproc --ignore 1` \
+		-n $$(( `nproc`<40 ? `nproc`/2 : 20 )) \
```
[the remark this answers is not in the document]
Gamenot: I will revert the changes in the `Makefile`. This was intended to lessen the clog-up of CPUs when doing testing. MARL commits been removed.

smarts/core/provider.py
Gamenot: Right, it would be more reasonable for that information to be applied at the time when the provider is added.

smarts/core/provider.py
```diff
@@ -76,3 +88,50 @@ def reset(self):
 
     def teardown(self):
         raise NotImplementedError
+
+    def recover(self, scenario, elapsed_sim_time: float, error: Exception) -> bool:
```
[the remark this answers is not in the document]
Gamenot: I have added a docstring for it.

smarts/core/provider.py
Gamenot: Yes, I considered that, I have some reluctance since that would mean that a provider would not be forced to overload the critical methods which could easily lead to unintended results.

smarts/core/provider.py
Gamenot: I have addressed it from SMARTS level

smarts/core/provider.py
```diff
@@ -76,3 +88,50 @@ def reset(self):
...
+
+    def create_vehicle(self, provider_vehicle: VehicleState):
+        return
+
+    def reset(self):
+        return
+
+    def teardown(self):
+        return
+
```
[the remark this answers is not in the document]
Gamenot: I am unsure if I want to keep this class around.

Makefile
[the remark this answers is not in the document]
Gamenot: Reverted except for the notebook test changes.

smarts/core/provider.py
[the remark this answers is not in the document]
Gamenot: I am not sure if I want to keep this class around.

Gamenot:
Odd that the notebook test failed since it worked locally. I know the github workers seem to be having issues.

smarts/core/agent_manager.py
```diff
@@ -213,6 +213,10 @@ def observe(self, sim):
                 rewards[agent_id] = vehicle.trip_meter_sensor(increment=True)
                 scores[agent_id] = vehicle.trip_meter_sensor()
 
+        if sim.should_reset:
+            dones = {agent_id: True for agent_id in self.agent_ids}
+            dones["__sim__"] = True
```
Gamenot, line 218: The idea here is that we could treat the sim as also having the possibility to be done.

Gamenot: It can come through the `dones` from `SMARTS`; however, at the gym level it would need to be translated.

smarts/core/bubble_manager.py
```diff
@@ -563,10 +563,11 @@ def _prepare_sensors_for_agent_control(
...
         # XXX: this is not quite right.  route may not be what the agent wants to take.
         route = sim.traffic_sim.vehicle_route(vehicle_id=vehicle.id)
-        mission = Mission(
-            start=Start(vehicle.position[:2], vehicle.heading),
-            goal=PositionalGoal.from_road(route[-1], sim.scenario.road_map),
-        )
+        if len(route) > 0:
+            goal = PositionalGoal.from_road(route[-1], sim.scenario.road_map)
+        else:
+            goal = EndlessGoal()
```
Gamenot, line 569: I believe this was a bug with a 0 length route being considered a positional goal when there is no end edge.

smarts/core/smarts.py
```diff
@@ -108,16 +109,28 @@ def __init__(
...
             self._trajectory_interpolation_provider,
-        ]
+        )
         if self._traffic_sim:
-            self._providers.insert(0, self._traffic_sim)
+            self._insert_provider(
+                0,
+                self._traffic_sim,
+                recovery_flags=ProviderRecoveryFlags.EPISODE_REQUIRED
+                | ProviderRecoveryFlags.ATTEMPT_RECOVERY,
```
Gamenot, line 129: It is possible to choose if the provider will attempt to recover from an error by specifying flags when adding the provider to the simulation.

smarts/core/smarts.py
```diff
@@ -373,9 +387,24 @@ def setup(self, scenario: Scenario):
...
 
-    def add_provider(self, provider):
+    def add_provider(
+        self,
+        provider: Provider,
+        recovery_flags: ProviderRecoveryFlags = ProviderRecoveryFlags.EXPERIMENT_REQUIRED,
+    ):
         assert isinstance(provider, Provider)
         self._providers.append(provider)
+        self._provider_recovery_flags[provider] = recovery_flags
```
Gamenot, line 397: The simulation keeps track of if the provider should attempt recovery.

smarts/core/smarts.py
```diff
@@ -790,6 +822,24 @@ def _reset_providers(self):
...
             provider.reset()
 
+    def _handle_provider(self, provider: Provider, provider_error):
+        provider_problem = bool(provider_error or not provider.connected)
+        if not provider_problem:
+            return
+
+        recovery_flags = self._provider_recovery_flags.get(
+            provider, ProviderRecoveryFlags.EXPERIMENT_REQUIRED
+        )
```
Gamenot, line 832: The provider will default to being required in the simulation.

smarts/core/smarts.py
```diff
@@ -790,6 +822,24 @@ def _reset_providers(self):
...
+        recovered = False
+        if recovery_flags & ProviderRecoveryFlags.ATTEMPT_RECOVERY:
+            recovered = provider.recover(self._scenario, self.elapsed_sim_time)
+
+        if not recovered:
+            if recovery_flags & ProviderRecoveryFlags.EPISODE_REQUIRED:
+                self._reset_required = True
+            elif recovery_flags & ProviderRecoveryFlags.EXPERIMENT_REQUIRED:
+                raise provider_error
```
Gamenot, line 841: The meaning of required is that it will re-raise the error it had if it cannot (or will not recover) from the error.

smarts/env/hiway_env.py
```diff
@@ -203,12 +201,12 @@ def step(self, agent_actions):
             observations[agent_id] = agent_spec.observation_adapter(observation)
             infos[agent_id] = agent_spec.info_adapter(observation, reward, info)
 
-        for done in agent_dones.values():
+        for done in dones.values():
```
Gamenot, line 204: If I am placing `"__sim__"` in `dones` I am unsure how to translate that through the gym interface. I can just remove it since all the agents are set done, I wonder if it should be signalled through `infos` as well.

smarts/core/provider.py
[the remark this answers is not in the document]
Gamenot: I have removed it since it was not needed.

smarts/core/smarts.py
```diff
@@ -312,7 +326,26 @@ def done_agent_ids(dones):
...
+            try:
+                self._resetting = True
+                return self._reset(scenario)
+            except Exception as e:
+                if not first_exception:
+                    first_exception = e
+            finally:
+                self._resetting = False
+        self._log.error(f"Failed to successfully reset after {tries} times.")
+        raise first_exception
```
Gamenot, line 341: I am somewhat iffy about this change intended to give a slight leeway in case an edge case occurs that breaks `reset`. Essentially, because we do not return `dones` or `infos` from this `reset` it is not easily possible to communicate to the user that `reset` failed for some reason. 

There are two directions we can take with this: 
- The `reset` has the responsibility to give a working simulation state. (currently going with this)
- The user has to determine if `reset` worked.

smarts/core/smarts.py
```diff
@@ -312,7 +326,26 @@ def done_agent_ids(dones):
         self._teardown_vehicles(vehicles_to_teardown)
 
     def reset(self, scenario: Scenario):
-        if scenario == self._scenario and self._reset_agents_only:
+        tries = 2
+        first_exception = None
```
RutvikGupta, line 330: Whats the reason behind 2 tries?

RutvikGupta:
You should probably also resolve the merge conflicts to see if things are working on the current branch as expected.

smarts/core/provider.py
```diff
@@ -18,13 +18,25 @@
...
 
+class ProviderRecoveryFlags(IntFlag):
+    NOT_REQUIRED = 0x00000000
+    """Not needed for the current step. Error causes skip."""
+    EPISODE_REQUIRED = 0x00000010
+    """Needed for the current episode. Results in episode ending."""
+    EXPERIMENT_REQUIRED = 0x00000100
+    """Needed for the experiment. Results in exception if an error is thrown."""
+    ATTEMPT_RECOVERY = 0x00001000
+    """Provider should attempt to recover from the exception or disconnection."""
```
sah-huawei, line 37: This seems to be orthogonal to the others.  For example, we might consider a Provider to be required for an episode _and_ want to attempt recovery, only ending the episode if recovery fails.

I'm guessing that's why you used non-overlapping bits for each of these enums.

I think it would be good to also indicate this with more of a comment on `ATTEMPT_RECOVERY` too.  Something like "Note: this can be combined with any of the above by creating a bitmask out of the enum values using `|`."

smarts/core/smarts.py
sah-huawei: I think you're right to go with the first one.

smarts/core/smarts.py
```diff
@@ -337,6 +370,7 @@ def reset(self, scenario: Scenario):
         self._total_sim_time += self._elapsed_sim_time
         self._elapsed_sim_time = 0
         self._step_count = 0
+        self._reset_required = False
```
sah-huawei, line 373: I might move this to the very end of this function, just in case something between here and then raises.

smarts/core/smarts.py
```diff
@@ -373,9 +407,24 @@ def setup(self, scenario: Scenario):
...
         self._is_setup = True
 
-    def add_provider(self, provider):
+    def add_provider(
+        self,
+        provider: Provider,
+        recovery_flags: ProviderRecoveryFlags = ProviderRecoveryFlags.EXPERIMENT_REQUIRED,
+    ):
         assert isinstance(provider, Provider)
         self._providers.append(provider)
```
sah-huawei, line 416: I think it's cleaner to use `_insert_provider()` here.  (That way, if anything is ever changed in or added to that function, all paths go through there.)

smarts/core/provider.py
```diff
@@ -76,3 +88,28 @@ def reset(self):
 
     def teardown(self):
         raise NotImplementedError
+
+    def recover(
+        self, scenario, elapsed_sim_time: float, error: Optional[Exception] = None
+    ) -> bool:
```
sah-huawei, line 94: I'm wondering if this shouldn't also return a new `ProviderState`, just in case recovery results in something different than the original attempt.

smarts/env/hiway_env.py
sah-huawei: Yeah, I think setting all agents to done should be sufficient for gym.

sah-huawei (approved):
None of my remaining comments/suggestions are blockers, so approving this.

smarts/core/smarts.py
Gamenot: Situational problems occur within the reset sometimes, I considered more than 1 retry but then there is no particular number that will work if the engine is failing. My reasoning here is that reset should not fail for an edge case on `reset` but I would say 
`reset` failure twice in a row indicates it is not an edge case.

smarts/core/smarts.py
Gamenot: My first instinct was to do that but it needs to be set `False` before calling `step` or gathering observations. If there is a raise again it will be set true again anyway and forces another `reset`. I had to make a specific case for where there is an exception in `reset` as you saw with retries since it is the responsibility of `reset` to provide a working simulation state.

PR #1142, opened 2021-11-30
Fix excessive map rescaling

Gamenot:
Comparison of z was causing a problem with the conditions for re-scaling the map. This was causing a lot of extra slowdown on the physics simulation.

### DIFF
--- a/smarts/core/utils/bullet.py
+++ b/smarts/core/utils/bullet.py
@@ -95,10 +95,10 @@ class BulletBoxShape:
     def __init__(self, bbox, bullet_client):
         self._client = bullet_client
 
-        width, length, height = bbox
+        length, width, height = bbox
         collision_box = bullet_client.createCollisionShape(
             shapeType=pybullet.GEOM_BOX,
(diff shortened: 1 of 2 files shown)


smarts/core/utils/bullet.py
```diff
@@ -95,10 +95,10 @@ class BulletBoxShape:
     def __init__(self, bbox, bullet_client):
         self._client = bullet_client
 
-        width, length, height = bbox
+        length, width, height = bbox
         collision_box = bullet_client.createCollisionShape(
             shapeType=pybullet.GEOM_BOX,
-            halfExtents=(length * 0.5, width * 0.5, height * 0.5),
+            halfExtents=(width * 0.5, length * 0.5, height * 0.5),
```
Gamenot, line 101: This is intentional from how I have looked into the `pybullet` gui.

sah-huawei: Yeah, I think it has to do with us having 0 degrees be North in SMARTS, unlike pybullet.

sah-huawei (approved):

RutvikGupta (approved):

Issue #1137, opened 2021-11-25
Collision detection fails in NGSIM (history replacement)


zbzhu99:
## **BUG REPORT**
**High Level Description**
When I run the imitation learning experiment in NGSIM environment, the off-road detection works fine but the collision detection (with other vehicles) does not work. 

**SMARTS version**
current master branch

**Steps to reproduce the bug**
Replace a history vehicle as a controlled ego vehicle in NGSIM dataset using similar code in examples/history_vehicles_replacement_for_imitation_learning.py and assign random action to that vehicle.

**Resulting and expected behaviour**
The controlled ego vehicle will not terminate even if it collides with other background vehicles.

**Error logs and screenshots**
<img width="655" alt="image" src="https://user-images.githubusercontent.com/32090492/143374098-f57ac9df-f57e-46b3-8e0f-f551a65e8455.png">

**Notes**
I uploaded the NGSIM scenario files I used here.
[NGSIM.zip](https://github.com/huawei-noah/SMARTS/files/7599962/NGSIM.zip)

sah-huawei:
@zbzhu99:  Sorry I never found this problem for Issue #969. 

For what it's worth, I had three hypotheses at the time:

1. there might be a problem with the logic related to `dones` in `history_vehicles_replacement_for_imitation_learning.py` such that if there is more than one agent to be replaced in the sample (i.e., `k>1`), then the simulation will continue after the first agent collides because it thinks there are more to do.
2.  the problem might be related to the `pybullet` physics engine not using the vehicle chassis sizes specified in the history database.
3. the problem might be related to using a `BoxChassis` for the agent vehicle (which is deliberate here, but historically our agents have used `AckermannChassis`.

I had investigated (2) and (3) somewhat and, while I hadn't yet ruled them out, I was satisfied that if either of these was the problem, it wasn't straightforward.  I.e., they seemed on the surface to be doing the correct thing.

It was at this point that I got distracted by other work and never came back to this.  Sorry!

From what you wrote above, it appears we can probably rule out (1) as well because it seems you are only replacing one vehicle in this test.  (If that is not the case, please let us know though.)

I have written all this here so if someone else wants to investigate this too they might have a head start.

CyberSY:
@sah-huawei :
I and @zbzhu99 have tested with single agent replacement and found the collision detection problem still existed.

zbzhu99:
The problem has been resolved with PR #1143.

PR #1143, opened 2021-12-01
Pybullet Step Fix

sah-huawei:
Fixed a simple, but annoying logic bug causing `pybullet` to not be stepped.  This should fix issues #969 and #1137.

Also added a unit test for ensuring that collisions between two `BoxChassis` vehicles are detected too.

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -712,11 +712,11 @@ class SMARTS:
         return provider_state
 
     def _nondynamic_provider_step(
-        self, agent_actions, step_pybullet: bool
+        self, agent_actions, stepped_pybullet: bool
     ) -> ProviderState:
         self._perform_agent_actions(agent_actions)
 
-        if step_pybullet:
+        if not stepped_pybullet:
             self._step_pybullet()
 
         self._process_collisions()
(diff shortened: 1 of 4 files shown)


smarts/core/smarts.py
```diff
@@ -712,11 +712,11 @@ def _pybullet_provider_step(self, agent_actions) -> ProviderState:
         return provider_state
 
     def _nondynamic_provider_step(
-        self, agent_actions, step_pybullet: bool
+        self, agent_actions, stepped_pybullet: bool
     ) -> ProviderState:
         self._perform_agent_actions(agent_actions)
```
Gamenot, line 717: It seems odd to have the `_perform_agent_actions()` occur potentially before or after `_step_pybullet()`

Gamenot (approved):

RutvikGupta (approved):

smarts/core/smarts.py
sah-huawei: It seems to me it always happens before `_step_pybullet()`.  Where do you see it happening before?

sah-huawei: Nevermind, I understand what you mean now.

sah-huawei: I'll try a mild refactor to avoid this...

sah-huawei: Good call.  That would have been an issue in the future.  Just pushed a refactor to attempt to address it.  Please review again!

Gamenot (approved):

RutvikGupta (approved):

PR #1144, opened 2021-12-01
relative import fix

sah-huawei:
Fix all relative imports of `argument_parser`, which cause problems when running our example scripts from the command line.
(These were made relative in PR #1120 due to a dependency having an `examples` module.)


sah-huawei:
I'm going to go ahead and merge this without reviews since it's currently broken / non-functional and there are people who are trying to test it right now.  But this should still be reviewed after the fact.

PR #1146, opened 2021-12-02
Specific package version and tensorflow update

Adaickalavan:
1. Package versions were specified for packages in `setup.py` which did not have any version specified previously. 
1. Updated tensorflow version.
1. Avoids unexpected code breaking due to changes in dependencies.
1. Helps reduce time spent by pip dependency resolver in [backtracking](https://pip.pypa.io/en/latest/topics/dependency-resolution/).


setup.py
```diff
@@ -22,57 +22,57 @@
         #   tensorboard needs >=41
         #   50.0 is broken: https://github.com/pypa/setupatools/issues/2353
         "setuptools>=41.0.0,!=50.0",
-        "cached-property",
-        "click",  # used in scl
+        "cached-property==1.5.2",
+        "click==8.0.3",  # used in scl
```
Gamenot, line 26: This hard lock of the dependencies seems overly harsh. This does improve the stability of smarts but it also locks us out of improvements provided by newer packages. I think we should go with `>=` rather than `==`.

setup.py
```diff
@@ -22,57 +22,57 @@
...
+        "psutil==5.8.0",
         "pybullet==3.0.6",
-        "pynput",  # Used by HumanKeyboardAgent
-        "rich",
-        "rtree",  # Used by sumolib
-        "sh",
-        "shapely",
-        "sklearn",  # KDTree from sklearn is used by sumo lanepoints
-        "tableprint",
+        "pynput==1.7.4",  # Used by HumanKeyboardAgent
```
Gamenot, line 32: I believe particular dependency is very specific to the `human_in_the_loop.py` example. We should be able to take it out of the main dependencies.

setup.py
```diff
@@ -22,57 +22,57 @@
...
+        "rich==10.13.0",
+        "rtree==0.9.7",  # Used by sumolib
+        "sh==1.14.2",
+        "shapely==1.8.0",
+        "sklearn==0.0",  # KDTree from sklearn is used by sumo lanepoints
```
Gamenot, line 37: This version seemed odd to me so I looked into it. It looks like the `PyPI` entry recommends to use `scikit-learn` instead.

requirements.txt
```diff
@@ -1,13 +1,13 @@
 absl-py==0.15.0
-aiohttp==3.8.0
+aiohttp==3.8.1
```
Gamenot, line 2: Would `mac_requirements.txt` also change?

setup.py
Adaickalavan: Changed `sklearn` to `scikit-learn`.

setup.py
Adaickalavan: Changed `==` to `>=` for packages where versions were newly specified.

setup.py
Gamenot: It is going to be removed for a different reason: https://github.com/huawei-noah/SMARTS/issues/1152

requirements.txt
Gamenot: @RutvikGupta would you be able to comment on this?

RutvikGupta: I believe it should be updated as well.

Gamenot:
This will need to go through formatting before merge.

Gamenot (approved):

requirements.txt
Adaickalavan: Automated update of `mac_requirements.txt` for mac is included in pr #1124.

Adaickalavan:
Closed in favour of pr #1124 .

PR #1148, opened 2021-12-03
Flexible ParallelEnv wrapper

Adaickalavan:
1) Modified env constructor type from `Callable[[str], gym.Env]` to `Callable[[], gym.Env]`.
2) Improved parallel environment example in order to be more appealing to the user.

### DIFF
--- a/examples/parallel_environment.py
+++ b/examples/parallel_environment.py
@@ -59,8 +59,8 @@ def main(
         for agent_id in agent_ids
     }
 
-    # Create a callable env constructor. Here, for illustration purposes, each 
-    # environment is wrapped with a FrameStack wrapper which returns stacked 
+    # Create a callable env constructor. Here, for illustration purposes, each
+    # environment is wrapped with a FrameStack wrapper which returns stacked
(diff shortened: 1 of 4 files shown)


examples/parallel_environment.py
```diff
@@ -267,7 +237,7 @@ def parallel_env_sync(
         sim_name=args.sim_name,
         headless=args.headless,
         seed=args.seed,
-        num_agents=args.num_agents,
+        num_agents=args.num_agents + 1,
```
Gamenot, line 240: Is this intended to always have multi-agent?

Gamenot (approved):

examples/parallel_environment.py
Adaickalavan: When a user runs the example, I wanted to make the transition in the envision display between `asynchronous` and `synchronous` approaches obvious. A simple way would be to change the number of agents in each approach. Let me know if you have a better way to clearly show the transition in the envision display.

liu-allan (approved):

Issue #1140, opened 2021-11-29
Include Tire Model in Package


Gamenot:
Tire model is missing from the package, this causes the package to be unable to run MPC.

AC:
- [ ] Add tire model to package (move to package resources from examples)
- [ ] Evaluate other resource files in examples

Source: https://github.com/huawei-noah/SMARTS/issues/1136
```
Hello @rshn1994 , we are currently looking into those previously mentioned files to give a better reply.

> Hello @Gamenot Thanks a lot for your reply, I wanted to add one more thing, these parameters are called from the source code after installation. So it would be better if you could include this point in your new release.
>
> Basically the stiffness parameters are not there in src after installation, so it will raise a ValueError: MPC requires providing tire stiffnesses in tire model

Yes, we realise now that this should have been included in the package resources. It should be included in the next release.

> Also, could you also please give me the reference article for the aforementioned tire model and also other tire models. That would be very helpful

I will need to get in contact with @iman512003 to be able to provide an article.

_Originally posted by @Gamenot in https://github.com/huawei-noah/SMARTS/issues/1136#issuecomment-982035616_
```

PR #1149, opened 2021-12-03
move tire model from example to smarts model

cz0204:
Include tire model suggested in issue #1140. Move the tire_parameters.yaml from example folder to ~/smarts/core/models


Gamenot:
I believe there is an out-of date hard-coded reference to this file in `./examples/tools/pybullet_linear_tire_example.py` that should be updated. Please also make a note of the move of the file in the `CHANGELOG`.

Gamenot (changes requested):
Please also make a note of the move of the file in the `CHANGELOG`.

Issue #1139, opened 2021-11-29
To generate myself scenario


zyk516:
**High Level Description**
[To generate myself scenario]

**Desired SMARTS version**
[e.g. 0.4.18]

**Operating System**
[ubuntu 18.04]

**Problems**
I  generated 10 social agents using follow codes:
```
car_nums = 10
car_index_list = [i for i in range(car_nums)]
laner_actors = []
laner_missions = []
for car_index in car_index_list:
    laner_actors.append(t.SocialAgentActor(name=str(car_index), agent_locator="zoo.policies:keep-lane-agent-v0"))
    laner_missions.append(t.Mission(t.Route(begin=("445633932", car_index, 'random'), end=("445633932", 0, 'max'))))

missions = [
    t.Mission(t.Route(begin=("445633932", 2, 'random'), end=("445633932", 0, 'max')))
]
gen_scenario(
    t.Scenario(
        # traffic={'cars':traffic},
        ego_missions=missions,
        social_agent_missions={"all": (laner_actors, laner_missions)}

    ),
    output_dir=Path(__file__).parent,
)
```
But when I visualize this scenario with envision, I can only see one social agent.It seems that one social agent randomly chose from 10 social agents I generate will appear  in one episode.I want to know how the 10 social agents appear at the same time.

Gamenot:
@zyk516 Hello sorry for the late reply, I have looked at the problem. The way you are attempting to create the social vehicles should be correct. I was surprised by this issue, it looks like at some point we tried to share the mission generation code with the ego missions and it ended with the behavior you note.

I am putting up a solution for it.

Gamenot:
Solution in review with PR #1151.

saulfield:
Oh I just realized the PR was abandoned. I will reopen.

PR #1151, opened 2021-12-04
Bugfix single social vehicle only

Gamenot:
As noted here: https://github.com/huawei-noah/SMARTS/issues/1139 scenario generation only applied a single social vehicle to the episode scene.


smarts/core/scenario.py
```diff
@@ -301,7 +291,7 @@ def discover_friction_map(scenario_root):
     @lru_cache(maxsize=16)
     def _discover_social_agents_info(
         scenario,
-    ) -> Sequence[Dict[str, SocialAgent]]:
+    ) -> Sequence[Tuple[Sequence[Tuple[str, SocialAgent]], Sequence[Mission]]]:
```
Gamenot, line 294: This type-hint was completely incorrect before. It is now accurate.

smarts/core/scenario.py
```diff
@@ -195,22 +187,20 @@ def variations_for_all_scenario_roots(
             ) in product(
                 np.roll(routes, roll_routes, 0),
                 np.roll(agent_missions, roll_agent_missions, 0),
-                np.roll(social_agents, roll_social_agents, 0),
+                np.roll(social_agent_infos_and_missions, roll_social_agents, 0),
                 np.roll(traffic_histories, roll_traffic_histories, 0),
             ):
+                if concrete_social_agents is None or len(concrete_social_agents) == 0:
```
Gamenot, line 193:
```suggestion
                if concrete_social_agents is None:
```

smarts/core/scenario.py
```diff
@@ -195,22 +187,20 @@ def variations_for_all_scenario_roots(
...
                 np.roll(traffic_histories, roll_traffic_histories, 0),
             ):
+                if concrete_social_agents is None:
+                    concrete_social_agents = []
                 concrete_social_agent_missions = {
                     agent_id: mission
-                    for agent_id, (_, (_, mission)) in (
-                        concrete_social_agents or {}
-                    ).items()
+                    for (agent_id, _), mission in (zip(*concrete_social_agents))
```
Gamenot, line 197:
```suggestion
                    for (agent_id, _), mission in zip(*concrete_social_agents)
```

smarts/core/scenario.py
```diff
@@ -195,22 +187,20 @@ def variations_for_all_scenario_roots(
...
                 }
 
                 # Filter out mission
                 concrete_social_agents = {
-                    agent_id: (_agent_spec, social_agent)
-                    for agent_id, (_agent_spec, (social_agent, _)) in (
-                        concrete_social_agents or {}
-                    ).items()
+                    agent_id: (social_agent.to_agent_spec(), social_agent)
+                    for (agent_id, social_agent), _ in (zip(*concrete_social_agents))
```
Gamenot, line 203:
```suggestion
                    for (agent_id, social_agent), _ in zip(*concrete_social_agents)
```

Gamenot:
The solution works. I will address the social agent variation test and CHANGELOG requirements tomorrow.

smarts/core/scenario.py
```diff
@@ -195,22 +187,20 @@ def variations_for_all_scenario_roots(
             ) in product(
                 np.roll(routes, roll_routes, 0),
                 np.roll(agent_missions, roll_agent_missions, 0),
-                np.roll(social_agents, roll_social_agents, 0),
+                np.roll(social_agent_infos_and_missions, roll_social_agents, 0),
                 np.roll(traffic_histories, roll_traffic_histories, 0),
             ):
+                if concrete_social_agents is None:
```
liu-allan, line 193: referenced before assignment?

smarts/core/scenario.py
```diff
@@ -317,61 +304,36 @@ def _discover_social_agents_info(
...
+                        agent_locator=actor.agent_locator,
+                        policy_kwargs=actor.policy_kwargs,
+                        initial_speed=actor.initial_speed,
                     )
+                    for actor in actors
                 )
-                count += 1
+            ]
+            social_agent_missions = [
+                Scenario._extract_mission(mission, road_map) for mission in missions
```
liu-allan, line 331: will the missions in the for loop here `for mission in missions` correspond to the actor in `for actor in actors`? Does it matter if they correspond?

smarts/core/scenario.py
```diff
@@ -317,61 +304,36 @@ def _discover_social_agents_info(
...
-                        extracted_mission,
+                actors, missions = pickle.load(missions_file)
+            namespace = os.path.basename(missions_file_path)
+            namespace = os.path.splitext(namespace)[0]
+
+            social_agent_infos = [
+                (agent.id, agent)
+                for agent in (
+                    SocialAgent(
+                        id=SocialAgentId.new(actor.name, group=namespace),
```
RutvikGupta, line 319: Is there a need to use a for loop here? Can't you just assign the `agent` variable the newly created `SocialAgent` instance?

smarts/core/scenario.py
```diff
@@ -180,13 +172,13 @@ def variations_for_all_scenario_roots(
...
-
             if shuffle_scenarios:
-                roll_routes = random.randint(0, len(routes))
-                roll_agent_missions = random.randint(0, len(agent_missions))
-                roll_social_agents = random.randint(0, len(social_agents))
+                roll_routes = random.randint(1, len(routes))
+                roll_agent_missions = random.randint(1, len(agent_missions))
+                roll_social_agents = random.randint(
+                    1, len(social_agent_infos_and_missions)
+                )
```
RutvikGupta, line 180: why was the roll needed to be at least 1?

smarts/core/scenario.py
```diff
@@ -317,61 +304,36 @@ def _discover_social_agents_info(
...
+            social_agent_missions = [
+                Scenario._extract_mission(mission, road_map) for mission in missions
+            ]
 
-        social_agents_info = []
-        for l in agent_bucketer:
-            social_agents_info.append(
-                {agent.id: (agent, mission) for agent, mission in l}
-            )
+            items.append((social_agent_infos, social_agent_missions))
```
RutvikGupta, line 334: The idea here is that for every element in `items` list, you take every element `agent` in `social_agent_infos` and match it to every `social_agent_missions` giving you a cross product between the two array right?. I asked this since the old return type ensured this but the new one doesn't. This also means the cross product is happening at other level since i don't see that change in this pr.

smarts/core/scenario.py
Gamenot: This is just to prevent overlap, `random.randint` gives an inclusive range `[m:n]`.  The beginning and end of the range then overlaps when using cycle.

```python
>>> rs = [random.randint(0, 3) for _ in range(10000)]
>>> min(rs) == 0
True
>>> max(rs) == 3
True
>>> list(np.roll([4,3,2], 3))
[4,3,2]
```

smarts/core/scenario.py
Gamenot: No, after discussion with @sah-huawei we made the decision to make the agents and missions explicit for each generated group rather than working with permutations. This throws the configuration explicitly to the scenario designer.

smarts/core/scenario.py
Gamenot: These missions should explicitly map to the actors. I am going to do a change to the interface here to do an explicit mapping of mission to agent for scenario generation.

smarts/core/scenario.py
Gamenot: It is going to see a change to an explicit Agent->Mission mapping.

smarts/core/scenario.py
Gamenot: No it is created from the `for` iteration.

smarts/core/tests/test_scenario.py
```diff
@@ -98,6 +102,6 @@ def test_scenario_variations_of_social_agents(scenario_root):
         for group, speed in itertools.product(groups, speeds)
     }
 
-    assert (
-        len(all_social_agent_ids - expected_social_agent_ids) == 0
+    assert (len(all_social_agent_ids & expected_social_agent_ids)) == len(
+        expected_social_agent_ids
```
Gamenot, line 106: I made this length check explicit since the old check was subtracting a superset from the resulting set.

Gamenot:
@liu-allan @RutvikGupta @sah-huawei I need another round of review so this can go in.

qianyi-sun (approved):

RutvikGupta:
Do you know why the Rendering tests are failing on this branch?

Gamenot:
> Do you know why the Rendering tests are failing on this branch?

@RutvikGupta I believe it is related to this issue: https://github.com/huawei-noah/SMARTS/issues/1175. It appears that there may be some bad interaction between threads.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - Added `smarts.core.utils.import_utils` to help with the dynamic import of modules.
 - Added `single_agent` env wrapper and unit test. The wrapper converts a single-agent SMARTS environment's step and reset output to be compliant with gym spaces.
 - Added `rgb_image` env wrapper and unit test. The wrapper filters SMARTS environment observation and returns only top-down RGB image as observation.
+- Added `smarts.sstudio.gen_social_agents(...)` to enforce actor to agent paring.
```
sah-huawei, line 29: "paring" -> "pairing"

smarts/sstudio/genscenario.py
```diff
@@ -153,7 +210,7 @@ def gen_social_agent_missions(
     seed: int = 42,
     overwrite: bool = False,
 ):
```
RutvikGupta, line 212: Shouldn't this be removed if it's deprecated? Or do we want to warn the user to use `gen_social_agents` instead?

scenarios/zoo_intersection/scenario.py
```diff
@@ -2,6 +2,7 @@
 import random
 
 from smarts.sstudio import gen_missions, gen_social_agent_missions, gen_traffic
```
RutvikGupta, line 4: import to `gen_social_agent_missions` should be removed if not used.

smarts/sstudio/genscenario.py
```diff
@@ -145,6 +141,67 @@ def gen_traffic(
...
+    """
+
+    # For backwards compatibility we support both a single value and a sequence
+    social_actor_mission_pairs = social_actor_mission_pairs
+    if not isinstance(social_actor_mission_pairs, collections.abc.Sequence):
+        raise ValueError(f"{list(social_actor_mission_pairs)}")
+
+    ActorMissionPairs = Sequence[Tuple[types.SocialAgentActor, types.Mission]]
+    actors: ActorMissionPairs = [a for a, _ in social_actor_mission_pairs]
+    missions: Sequence[types.Mission] = [m for _, m in social_actor_mission_pairs]
```
RutvikGupta, line 176: They can also be extracted in a single for loop.

smarts/sstudio/genscenario.py
```diff
@@ -145,6 +141,67 @@ def gen_traffic(
...
+        raise ValueError(
+            f"{gen_social_agents.__name__}(...) can't be called with BoidAgentActor, got:"
+            f"{social_actor_mission_pairs}"
+        )
+
+    actor_names = [a.name for a in actors]
+    if len(actor_names) != len(set(actor_names)):
+        raise ValueError(f"Actor names={actor_names} must not contain duplicates")
+    if len(missions) != len(set(missions)):
+        raise ValueError(f"Supplied Missions must be unique.")
```
RutvikGupta, line 188: Why does `(actor, mission)` pair needs to be unique? Why can't the same actor have different missions or vice-versa?

RutvikGupta: Doesn't this seems to contradict the `(actor, mission)` we create in `4_lane_t` map where the two actors have the same `reused_mission`?

smarts/core/scenario.py
sah-huawei: lol, now I'm second guessing this!  maybe it's worth having one more (quick) discussion about this today?

smarts/core/scenario.py
```diff
@@ -195,22 +187,20 @@ def variations_for_all_scenario_roots(
...
             ):
+                if concrete_social_agents is None:
+                    concrete_social_agents = []
                 concrete_social_agent_missions = {
                     agent_id: mission
-                    for agent_id, (_, (_, mission)) in (
-                        concrete_social_agents or {}
-                    ).items()
+                    for (agent_id, _), mission in zip(*concrete_social_agents)
                 }
```
Gamenot, line 198: I need to go back an make this pairing to be explicit.

smarts/sstudio/types.py
```diff
@@ -714,21 +717,15 @@ class RoadSurfacePatch:
...
     # e.g. { "turning_agents": ([actors], [missions]), ... }
-    social_agent_missions: Optional[
-        Dict[str, Tuple[Sequence[SocialAgentActor], Sequence[Mission]]]
+    social_agents: Optional[
+        Dict[str, Sequence[Tuple[SocialAgentActor, Mission]]]
     ] = None
-    """Every dictionary item {group: (actors, missions)} gets run simultaneously. If
+    """Values in a dictionary item {group: actor_mission_pairs} get run simultaneously. If
     actors > 1 and missions = 0 or actors = 1 and missions > 0 we cycle through them
     every episode. Otherwise actors must be the same length as missions.
```
sah-huawei, line 730: I don't see where this comment is "enforced".  In `_gen_missions()` there is no check on the lengths of missions relative to actors, nor does there appear to be in `scenario.py`.

sah-huawei (commented):
My main concern is making sure the semantics are clear for SStudio users in the new "non-permutation" version.

smarts/sstudio/genscenario.py
Gamenot: I do not want to immediately remove this method. I want to give a chance for users to move to the `gen_social_agents(...)` method first.

smarts/sstudio/genscenario.py
Gamenot: Good catch, it looks like this is not getting triggered properly.

Gamenot: Actually, it looks like this is triggered properly but the terminology was not clear. I have updated `Scenario.social_agents` -> `Scenario.social_agent_groups` to remove the confusion.

Gamenot:
This is too far out of date.

Issue #1159, opened 2021-12-08
`scl scenario build` throws FILENOTFOUNDERROR


Gamenot:
Calling `scl scenario build` with an empty directory (subdirectories within a directory) will cause an exception.

See #1153:
>@qianyi-sun Can u check if you have a `scenarios/behaviour_model` folder in your cloned SMARTS repository? 
Removing that folder will resolve this issue for you. @sah-huawei removed the `map.net.xml` from all behavior model scenarios but since we generate a shifted normalized `map.net.xml` for `SUMO` maps, which are git ignored, they exist as local files on cloned repositories before this change. 
|
Since build-all command in `cli/studio.py` looks for `.net.xml` references in all scenarios folder, it picks up the behavior models but does not find the `map.net.xml` reference and hence throws the `FILENOTFOUNDERROR`.
|
P.S.: I had the same issue. Removing the folder from my local repo resolved the errors..

_Originally posted by @RutvikGupta in https://github.com/huawei-noah/SMARTS/issues/1153#issuecomment-987146360_

sah-huawei:
Looks like `scl scenario build-all missing` doesn't raise, but `scl scenario build missing` does (in both the case where "missing" is an empty folder and a non-existent folder).
I personally think this is "as it should be", with the caveat that the exception that is raised in the latter case should just report that the folder is empty. 
That is,  I don't think we should assume or require that the scenario folder should contain a _SUMO_ map file, per se, as we now support scenarios with alternative map formats as well.

Issue #1153, opened 2021-12-06
No such file or directory 'scenarios/behavior_models/.../map.net.xml'


qianyi-sun:
## **BUG REPORT**
**High Level Description**
During `make sanity-test`, An error occurs 'FileNotFoundError: [Errno 2] No such file or directory: 'scenarios/behavior_models/uturn/4lane/map.net.xml'

**SMARTS version**
0.4.18

**Previous associated issues**
N/A

**Steps to reproduce the bug**
Follow the setup steps in README.md and run `make sanity-test`

**Resulting and expected behaviour**
[e.g. Should print out "Hello world" but crashes with "Error: stack overflow"]

**Error logs and screenshots**
```
(.venv) sqy201x@Patrick-Desktop:~/SMARTS$ make sanity-test
scl scenario build-all scenarios
build-scenario scenarios/loop
build-scenario scenarios/minicity
build-scenario scenarios/minicity
build-scenario scenarios/tests/same_location
build-scenario scenarios/cloverleaf
build-scenario scenarios/tests/same_location
build-scenario scenarios/zoo_intersection
build-scenario scenarios/intersections/2lane
build-scenario scenarios/zoo_intersection
build-scenario scenarios/intersections/2lane
build-scenario scenarios/intersections/2lane_circle
build-scenario scenarios/intersections/4lane
build-scenario scenarios/intersections/4lane
build-scenario scenarios/intersections/6lane
build-scenario scenarios/intersections/6lane
build-scenario scenarios/intersections/4lane_t
build-scenario scenarios/figure_eight
build-scenario scenarios/straight
build-scenario scenarios/behavior_models/uturn/4lane
Exception in thread Thread-19:
Traceback (most recent call last):
  File "/usr/share/sumo/tools/sumolib/net/__init__.py", line 825, in readNet
build-scenario scenarios/behavior_models/uturn/1lane
    parse(gzip.open(filename), netreader)
  File "/usr/lib/python3.8/gzip.py", line 58, in open
Exception in thread Thread-20:
Traceback (most recent call last):
    binary_file = GzipFile(filename, gz_mode, compresslevel)
  File "/usr/lib/python3.8/gzip.py", line 173, in __init__
  File "/usr/share/sumo/tools/sumolib/net/__init__.py", line 825, in readNet
    parse(gzip.open(filename), netreader)
  File "/usr/lib/python3.8/gzip.py", line 58, in open
[...]

PR #1161, opened 2021-12-08
Raise warning message for building scenarios without map.net.xml file.

RutvikGupta:
See Issue #1159 and #1153. This also prevents building scenarios with shifted map offset network file but not the main `map.net.xml` file.

### DIFF
--- a/cli/studio.py
+++ b/cli/studio.py
@@ -65,6 +65,13 @@ def _build_single_scenario(clean, allow_offset_map, scenario):
 
     scenario_root = Path(scenario)
     map_net = str(scenario_root / "map.net.xml")
+    if not os.path.isfile(map_net):
+        click.echo(
+            "FILENOTFOUND: no reference to map.net.xml was found in {}.  "
+            "Please make sure the path passed is a valid SUMO scenario with SUMO network file (map.net.xml) required "
+            "for scenario building.".format(str(scenario_root))
+        )
+        return
     if not allow_offset_map or scenario.traffic_histories:
         SumoRoadNetwork.from_file(map_net, shift_to_origin=True)
     elif os.path.isfile(SumoRoadNetwork.shifted_net_file_path(map_net)):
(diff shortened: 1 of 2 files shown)


cli/studio.py
```diff
@@ -65,6 +65,13 @@ def _build_single_scenario(clean, allow_offset_map, scenario):
 
     scenario_root = Path(scenario)
     map_net = str(scenario_root / "map.net.xml")
+    if not os.path.isfile(map_net):
+        click.echo(
+            "FILENOTFOUND: no reference to map.net.xml was found in {}.  "
+            "Please make sure the path passed is a valid SUMO scenario with SUMO network file (map.net.xml) required "
+            "for scenario building.".format(str(scenario_root))
+        )
+        return
```
sah-huawei, line 74: The above seems ok in the very short term ("for now") but we should start thinking about generalizing this to support scenarios with other map formats (including scenarios where the map is from a server instead of a file) very soon.

cli/studio.py
```diff
@@ -141,7 +148,9 @@ def build_all_scenarios(clean, allow_offset_maps, scenarios):
     builder_threads = {}
     for scenarios_path in scenarios:
         path = Path(scenarios_path)
-        for p in path.rglob("*.net.xml"):
+        # We search along directories having reference to file map.net.xml and not .net.xml to prevent calling build
+        # command on files having the offset-map network file but not the original file
+        for p in path.rglob("*map.net.xml"):
```
sah-huawei, line 153: (same idea here)

sah-huawei (approved):

cli/studio.py
RutvikGupta: I have generalized this to support new map formats like `.xodr` for `OpenDRIVE `in the `opendrive-road-network` branch.

Gamenot:
Please log this change in the `CHANGELOG`.

cli/studio.py
Gamenot: I think this might drive the need for a cleaner way of switch on different types of scenarios maps. It seems reasonable if it comes in through the `OpenDRIVE` branch.

Gamenot:
Odd, it looks like the CI server had a connection failure: https://github.com/huawei-noah/SMARTS/runs/4461512766?check_suite_focus=true

Gamenot:
Interesting, I have not seen this issue before. Is it an issue with the github servers or us? https://github.com/huawei-noah/SMARTS/runs/4462035138?check_suite_focus=true

It would appear to me that the system was unable to build a thread.

RutvikGupta:
Ok, seems to be some issue with Github server. Its back to normal again.

Gamenot (approved):

PR #1166, opened 2021-12-09
Trip meter sensor

Adaickalavan:
Previously, no score was computed when scenario had ego agents with mission.
Because we were comparing `str` with `obj`.


Gamenot (approved):
Please add an entry in the CHANGELOG.

RutvikGupta (approved):

Issue #1171, opened 2021-12-10
RuntimeWarning: invalid value encountered in double_scalars 


Adaickalavan:
## **BUG REPORT**
**High Level Description**
While running the scenarios, sometimes the following error is thrown
```
/SMARTS/smarts/core/sumo_road_network.py:331: RuntimeWarning: invalid value encountered in double_scalars
lane_angle = np.dot(my_vect, lv) / (my_norm * np.linalg.norm(lv))
```
This is because the lane vector is zero, i.e., `lv = lane.vector_at_offset(offset) = [0 0 0]` , at line `SMARTS/smarts/core/sumo_road_network.py:331`. A divide by zero occurs, producing a nan and the error message.

**SMARTS version**
Latest `develop` branch

**Previous associated issues**
None

**Steps to reproduce the bug**
```
$ git checkout scenario-error
$ scl scenario build-all ./scenarios/intersections/2lane_test/
$ python3.7 ./examples/single_agent.py ./scenarios/intersections/2lane_test/
```

**Resulting and expected behaviour**
Proper lane vectors should not be zero.

**Error logs and screenshots**
![image](https://user-images.githubusercontent.com/24203326/145624457-eb1c815a-8e3c-4026-87f7-aff755d78453.png)
 
**System information**
 
- Operating system & Environment: Ubuntu 18.04
- Python version: 3.7
- SUMO version: 1.10.0
- Date when bug occurred: 10 Dec 2021

sah-huawei:
@Adaickalavan Could you please fill in the "steps to reproduce this bug" section too?  Thanks.

Adaickalavan:
@sah-huawei 
I have included the steps to reproduce the error message in the issue description.

PR #1173, opened 2021-12-10
Bug Fix lane vector for lane offset >= lane's length

RutvikGupta:
See Issue #1171. 
We set the offset in `vector_at_offset` method to lane's length if it is greater than road's length.

### DIFF
--- a/smarts/core/road_map.py
+++ b/smarts/core/road_map.py
@@ -301,6 +301,7 @@ class RoadMap:
             else:
                 s_offset = start_offset
                 end_offset = start_offset + 1  # a little further down the lane
+            s_offset = max(s_offset, 0)
             p1 = self.from_lane_coord(RefLinePoint(s=s_offset))
             p2 = self.from_lane_coord(RefLinePoint(s=end_offset))
             return np.array(p2) - np.array(p1)
--- a/smarts/core/road_map.py
+++ b/smarts/core/road_map.py
@@ -60,7 +60,7 @@ class RoadMap:
         return 1.0
 
     def to_glb(self, at_path):
-        """ build a glb file for camera rendering and envision """
+        """build a glb file for camera rendering and envision"""
         raise NotImplementedError()
 
     def surface_by_id(self, surface_id: str) -> RoadMap.Surface:
@@ -93,7 +93,7 @@ class RoadMap:
         via: Sequence[RoadMap.Road] = None,
         max_to_gen: int = 1,
     ) -> List[RoadMap.Route]:
-        """ Routes will be returned in order of increasing length """
+        """Routes will be returned in order of increasing length"""
         raise NotImplementedError()
 
     def random_route(self, max_route_len: int = 10) -> RoadMap.Route:
@@ -290,14 +290,18 @@ class RoadMap:
         def edges_at_point(self, point: Point) -> Tuple[Point, Point]:
             offset = self.offset_along_lane(point)
             width = self.width_at_offset(offset)
-            left_edge = RefLanePoint(s=offset, t=width / 2)
-            right_edge = RefLanePoint(s=offset, t=-width / 2)
-            return (self.from_lane_coord(left_edge), self.from_lane_coord(right_edge))
+            left_edge = RefLinePoint(s=offset, t=width / 2)
+            right_edge = RefLinePoint(s=offset, t=-width / 2)
+            return self.from_lane_coord(left_edge), self.from_lane_coord(right_edge)
 
         def vector_at_offset(self, start_offset: float) -> np.ndarray:
-            add_offset = 1  # a little further down the lane
-            end_offset = start_offset + add_offset
-            p1 = self.from_lane_coord(RefLinePoint(s=start_offset))
+            if start_offset >= self.length:
+                s_offset = self.length - 1
+                end_offset = self.length
+            else:
+                s_offset = start_offset
+                end_offset = start_offset + 1  # a little further down the lane
+            p1 = self.from_lane_coord(RefLinePoint(s=s_offset))
             p2 = self.from_lane_coord(RefLinePoint(s=end_offset))
             return np.array(p2) - np.array(p1)
 
--- a/smarts/core/sumo_road_network.py
+++ b/smarts/core/sumo_road_network.py
@@ -324,16 +324,19 @@ class SumoRoadNetwork(RoadMap):
             if not nearby_lanes:
                 return result
             my_vect = self.vector_at_offset(offset)
-            my_norm = max(np.linalg.norm(my_vect), 1e-6)
+            my_norm = np.linalg.norm(my_vect)
+            if my_norm == 0:
+                return result
             threshold = -0.995562  # cos(175*pi/180)
             for lane, _ in nearby_lanes:
                 if lane == self:
                     continue
                 lane_refline_pt = lane.to_lane_coord(pt)
                 lv = lane.vector_at_offset(lane_refline_pt.s)
-                if np.array_equal(lv, np.array([0.0, 0.0, 0.0])):
+                lv_norm = np.linalg.norm(lv)
+                if lv_norm == 0:
                     continue
-                lane_angle = np.dot(my_vect, lv) / (my_norm * np.linalg.norm(lv))
+                lane_angle = np.dot(my_vect, lv) / (my_norm * lv_norm)
                 if lane_angle < threshold:
                     result.append(lane)
             return result
--- a/smarts/core/sumo_road_network.py
+++ b/smarts/core/sumo_road_network.py
@@ -178,7 +178,7 @@ class SumoRoadNetwork(RoadMap):
 
     @property
     def source(self) -> str:
-        """ This is the net.xml file that corresponds with our possibly-offset coordinates. """
+        """This is the net.xml file that corresponds with our possibly-offset coordinates."""
         return self._net_file
 
     @cached_property
@@ -195,7 +195,7 @@ class SumoRoadNetwork(RoadMap):
         return self._default_lane_width / SumoRoadNetwork.DEFAULT_LANE_WIDTH
 
     def to_glb(self, at_path):
-        """ build a glb file for camera rendering and envision """
+        """build a glb file for camera rendering and envision"""
         polys = self._compute_road_polygons()
         glb = self._make_glb_from_polys(polys)
         glb.write_glb(at_path)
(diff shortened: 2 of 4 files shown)


smarts/core/road_map.py
```diff
@@ -290,14 +290,18 @@ def center_at_point(self, point: Point) -> Point:
...
+            left_edge = RefLinePoint(s=offset, t=width / 2)
+            right_edge = RefLinePoint(s=offset, t=-width / 2)
+            return self.from_lane_coord(left_edge), self.from_lane_coord(right_edge)
 
         def vector_at_offset(self, start_offset: float) -> np.ndarray:
-            add_offset = 1  # a little further down the lane
-            end_offset = start_offset + add_offset
-            p1 = self.from_lane_coord(RefLinePoint(s=start_offset))
+            if start_offset >= self.length:
+                s_offset = self.length - 1
```
sah-huawei, line 299: This is still problematic if `self.length` is less than 1.  Probably best would be to replace the entire `if`/`else` block with:
```
before_offset = max(start_offset - 0.5, 0)
after_offset = min(start_offset + 0.5, self.length)
if before_offset == after_offset:
    return np.array((0,0,0))
p1 = self.from_lane_coord(RefLinePoint(s=before_offset))
p2 = self.from_lane_coord(RefLinePoint(s=after_offset))
return np.array(p2) - np.array(p1)
```

(Note that for most map formats, this "compare to the same point offset by a little" technique is un-necessary because the vector is given by the direction of the nearest segment.)

sah-huawei: ... and because this can still return (0,0,0) in rare occasions (for example, where a lane has 0 length), we should still also add a check downstream to prevent the divide-by-zero exception.

smarts/core/tests/test_map.py
```diff
@@ -94,6 +96,10 @@ def test_sumo_map(scenario):
     assert "edge-north-NS_0" in foe_set  # entering from north
     assert ":junction-intersection_5_0" in foe_set  # crossing from east-to-west
 
+    # Test the lane vector for a refline point outside lane
+    lane_heading_at_offset = lane.vector_at_offset(55.7)
+    assert np.array_equal(lane_heading_at_offset, np.array([0.0, -1.0, 0.0]))
```
sah-huawei, line 101: Good!

sah-huawei (commented):
On the right track...
(and good catch on the `RefLanePoint` fixes!)

smarts/core/road_map.py
sah-huawei: On 

> This is still problematic if `self.length` is less than 1. Probably best would be to replace the entire `if`/`else` block with:

On second thought, given that we don't want to change the behavior for existing code/scenarios, I guess we should only do the above when `start_offset` is greater than or equal to `self.length`.  

So I withdraw my suggested code (but you will still need to add a `max` to prevent `s_offset` from being less than 0).

RutvikGupta: ok, sure!

smarts/core/sumo_road_network.py
```diff
@@ -324,10 +324,15 @@ def oncoming_lanes_at_offset(self, offset: float) -> List[RoadMap.Lane]:
             if not nearby_lanes:
                 return result
             my_vect = self.vector_at_offset(offset)
-            my_norm = np.linalg.norm(my_vect)
+            my_norm = max(np.linalg.norm(my_vect), 1e-6)
             threshold = -0.995562  # cos(175*pi/180)
```
Gamenot, line 328: This seems unclear, what is the reason for this threshold and the specific value?

smarts/core/tests/test_map.py
```diff
@@ -45,6 +46,7 @@ def test_sumo_map(scenario):
     assert lane.road.contains_point(point)
     assert lane.is_drivable
     assert len(lane.shape()) >= 2
+    assert lane.length == 55.6
```
Gamenot, line 49: Can we make this `55.6` a constant? I mention this only because it is used elsewhere.

smarts/core/sumo_road_network.py
RutvikGupta: for `oncoming_lanes_at_offset` we are checking for lanes that are in opposite direction of travel to the current lane. We find the heading of all the nearby lanes to this offset and check whether the dot product of the current lane's vector and nearby lane vector is `cos(175*pi/180)` which is the above threshold to find if they are in opposite direction to the given lane or not.

smarts/core/tests/test_map.py
RutvikGupta: No, its the length of the lane `edge-north-NS_0` which is being referred by `lane` variable here. Its specific to it so we can't create a constant for it (although it won't matter as well if we do).

Gamenot: OK, I mention that because it looks like this length value is used again on line 100.

smarts/core/sumo_road_network.py
```diff
@@ -324,10 +324,15 @@ def oncoming_lanes_at_offset(self, offset: float) -> List[RoadMap.Lane]:
...
             for lane, _ in nearby_lanes:
-                lv = lane.vector_at_offset(offset)
+                if lane == self:
```
Adaickalavan, line 330: Why does `nearby_lanes` have `self` in it? It feels counterintuitive.

smarts/core/sumo_road_network.py
```diff
@@ -324,10 +324,15 @@ def oncoming_lanes_at_offset(self, offset: float) -> List[RoadMap.Lane]:
             if not nearby_lanes:
                 return result
             my_vect = self.vector_at_offset(offset)
```
Adaickalavan, line 326: When does this `my_vect = self.vector_at_offset(offset)` evaluate to [0,0,0]?

smarts/core/road_map.py
```diff
@@ -290,14 +290,19 @@ def center_at_point(self, point: Point) -> Point:
...
-            return (self.from_lane_coord(left_edge), self.from_lane_coord(right_edge))
+            left_edge = RefLinePoint(s=offset, t=width / 2)
+            right_edge = RefLinePoint(s=offset, t=-width / 2)
+            return self.from_lane_coord(left_edge), self.from_lane_coord(right_edge)
 
         def vector_at_offset(self, start_offset: float) -> np.ndarray:
-            add_offset = 1  # a little further down the lane
-            end_offset = start_offset + add_offset
-            p1 = self.from_lane_coord(RefLinePoint(s=start_offset))
+            if start_offset >= self.length:
```
Adaickalavan, line 298: When does the `start_offset >= self.length` ?

smarts/core/tests/test_map.py
RutvikGupta: Yes we want to test what happens to the vector if offset passed is greater than the length of lane.

smarts/core/sumo_road_network.py
RutvikGupta:
```python
 nearby_lanes = self._map.nearest_lanes(pt, radius=radius)
```
nearby_lanes is obtained from `nearest_lanes` method which is a method of parent `RoadMap` class. So it doesn't know about the current lane and hence can return the current lane `self` as one of the nearby lanes to this point.

smarts/core/road_map.py
RutvikGupta: if the offset is being computed for a point at the end of the lane lying outside the lane.

smarts/core/sumo_road_network.py
```diff
@@ -324,10 +324,15 @@ def oncoming_lanes_at_offset(self, offset: float) -> List[RoadMap.Lane]:
...
-            my_norm = np.linalg.norm(my_vect)
+            my_norm = max(np.linalg.norm(my_vect), 1e-6)
```
Adaickalavan, line 327: Do we use `1e-6` here to avoid divide by zero in line `lane_angle = np.dot(my_vect, lv) / (my_norm * np.linalg.norm(lv))` below?

If so, instead of doing `max(..., 1e-6)` here, can we just return empty `result=[]` if `np.linalg.norm(my_vect)=[0,0,0]` ?

smarts/core/sumo_road_network.py
```diff
@@ -324,10 +324,15 @@ def oncoming_lanes_at_offset(self, offset: float) -> List[RoadMap.Lane]:
...
             threshold = -0.995562  # cos(175*pi/180)
             for lane, _ in nearby_lanes:
-                lv = lane.vector_at_offset(offset)
+                if lane == self:
+                    continue
+                lane_refline_pt = lane.to_lane_coord(pt)
```
sah-huawei, line 332: This was a good bug catch and fix!  Nice.

smarts/core/sumo_road_network.py
sah-huawei: > Do we use `1e-6` here to avoid divide by zero in line `lane_angle = np.dot(my_vect, lv) / (my_norm * np.linalg.norm(lv))` below?

Yes, I think that was the intent there.

 
> If so, instead of doing `max(..., 1e-6)` here, can we just return empty `result=[]` if `np.linalg.norm(my_vect)=[0,0,0]` ?

Yes, I think this is an option -- at least when `my_norm == 0`, since when this happens, the lane at this offset doesn't really have a direction (probably because its length is (very near to) 0, then determining lanes that go in the "opposite" direction doesn't really make sense.  I would support either returning `[]` or `None` here as opposed to doing the `max()` in this case.

Between `[]` and `None`, I would probably prefer `None` in order to denote that this is an anomalous case, although that's a bit more painful for callers of this method, because then they'd have to check for it (which I doubt any do at this point).  Because of this, `[]` would also be ok, just because this event is hopefully going to be rare anyway.

sah-huawei (approved):
I think all of my concerns have been addressed here.

smarts/core/sumo_road_network.py
```diff
@@ -324,10 +324,15 @@ def oncoming_lanes_at_offset(self, offset: float) -> List[RoadMap.Lane]:
...
+                lv = lane.vector_at_offset(lane_refline_pt.s)
+                if np.array_equal(lv, np.array([0.0, 0.0, 0.0])):
```
sah-huawei, line 334: Perhaps a slightly nicer option here would be to first compute the norm and then compare it to 0, like:
```python
...
lv_norm = np.linalg.norm(lv)
if lv_norm == 0:
    continue
lane_angle = np.dot(my_vect, lv) / (my_norm * lv_norm)
...
````

smarts/core/sumo_road_network.py
RutvikGupta: Ok i will go with returning `[ ]`.

Adaickalavan (approved):

Issue #1152, opened 2021-12-06
pynput causes exception on WSL2


Gamenot:
## **BUG REPORT**
**High Level Description**
Running pip install for SMARTS on WSL2 will cause `evdev` compilation (under pynput) to fail to build.

**SMARTS version**
0.4.18

**Previous associated issues**
N/A

**Steps to reproduce the bug**
```
# WSL2
$ cd $SMARTS_REPO
$ pip install -e .
```

**Resulting and expected behaviour**
Should install SMARTS fully but instead crashes on building `evdev`.

**Error logs and screenshots**
![003f73d0714ec300ab1be15636a31c9](https://user-images.githubusercontent.com/59240680/144888915-df8e46df-b559-4653-89c5-ee67156b8164.png)
 
**System information**
 
- Operating system & Environment [Please note if using Docker]:
WSL2 
- Number of system cores available
4
- Python version [Please note if using Jupyter]
python 3.8
- Python packages [e.g. from `pip freeze`]
- SUMO version
- Date when bug occurred  dd_mm_yyyy
06_12_2021
 
**Impact [If known]**
Unable to install SMARTS on WSL2.

qianyi-sun:
Here's the screenshot:

![003f73d0714ec300ab1be15636a31c9](https://user-images.githubusercontent.com/59240680/144888915-df8e46df-b559-4653-89c5-ee67156b8164.png)

RutvikGupta:
I installed `sudo apt-get install python3.8-dev` the developer version of `python3.8` (same for `python3.7`) for my WSL2 setup and didn't have this issue.

Gamenot:
> I installed `sudo apt-get install python3.8-dev` the developer version of `python3.8` (same for `python3.7`) for my WSL2 setup and didn't have this issue. More about this here: https://stackoverflow.com/questions/26053982/setup-script-exited-with-error-command-x86-64-linux-gnu-gcc-failed-with-exit

I think that is useful but I would prefer that we do not require the developer version of python to run SMARTS.

PR #1178, opened 2021-12-14
Move pynput dep out of main dependencies and into extras package

qianyi-sun:
This PR is related to #1152.

### DIFF
--- a/examples/human_in_the_loop.py
+++ b/examples/human_in_the_loop.py
@@ -7,10 +7,11 @@ NOTE: You will need to install [extras] to run this example. `pip install -e .[e
 import logging
 
 import gym
+
 try:
     from pynput.keyboard import Key, Listener
 except ImportError:
-    raise ImportError('pynput dependency is missing, please pip install -e .[extras]')
+    raise ImportError("pynput dependency is missing, please pip install -e .[extras]")
 
 from smarts.core.agent import Agent, AgentSpec
 from smarts.core.agent_interface import AgentInterface, AgentType
@@ -91,13 +92,7 @@ class HumanKeyboardAgent(Agent):
         return self.action
 
 
-def main(
-    scenarios,
-    sim_name,
-    headless,
-    num_episodes,
-    seed,
-):
+def main(scenarios, sim_name, headless, num_episodes, seed):
     agent_spec = AgentSpec(
         interface=AgentInterface.from_type(
             AgentType.StandardWithAbsoluteSteering, max_episode_steps=3000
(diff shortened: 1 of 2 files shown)


examples/human_in_the_loop.py
```diff
@@ -82,13 +92,7 @@ def act(self, obs):
...
 
 
-def main(
-    scenarios,
-    sim_name,
-    headless,
-    num_episodes,
-    seed,
-):
+def main(scenarios, sim_name, headless, num_episodes, seed):
```
Gamenot, line 95: What version of black did you use?

setup.py
```diff
@@ -84,17 +83,10 @@
...
-            "catkin_pkg",
-            "rospkg",
-        ],
-        "waymo": [
-            "waymo-open-dataset-tf-2-2-0",
-        ],
+        "camera-obs": ["Panda3D==1.10.9", "panda3d-gltf==0.13"],
+        "ros": ["catkin_pkg", "rospkg"],
+        "waymo": ["waymo-open-dataset-tf-2-2-0"],
+        "extras": ["pynput>=1.7.4"],  # Used by HumanKeyboardAgent
```
Gamenot, line 89: Seeing this now, I wonder if it would be best to remove the `human_in_the_loop` example and move the agent to the zoo. Then this requirement can sit in the agent's requirements.

examples/human_in_the_loop.py
```diff
@@ -1,7 +1,17 @@
+"""
+This examples runs the human-keyboard Agent, which allows you to control and monitor input devices.
+
+NOTE: You will need to install [extras] to run this example. `pip install -e .[extras]`
+"""
+
```
Gamenot, line 6: Seeing this now, I wonder if it would be best to remove the `human_in_the_loop` example and move the agent to the zoo. Then this requirement can sit in the agent's requirements.

I do not think we need a completely separate example to demonstrate an agent that would generally be used for debugging.

Gamenot: I think this is fine for now but please add an issue to move the `HumanKeyboardAgent` to the agent zoo.

Gamenot (approved):

examples/human_in_the_loop.py
qianyi-sun: `black --version` -> `black, version 19.3b0`

examples/human_in_the_loop.py
RutvikGupta: @qianyi-sun Can you please follow up on these changes?

RutvikGupta (approved):

Issue #1175, opened 2021-12-13
Test


Gamenot:
## **BUG REPORT**
**High Level Description**
The test `smarts/core/tests/test_renderers.py::test_multiple_renderers` causes a segmentation fault when run.

**SMARTS version**
0.4.18

**Previous associated issues**
[Link any associated issues here]

**Steps to reproduce the bug**
```
$ ##Run
$ pip install -e .[test,camera-obs]
$ pytest smarts/core/tests/test_python_version.py::test_python_version
```

**Resulting and expected behaviour**
A "Segment fault" is thrown. Test should pass without issue.

**Error logs and screenshots**
See error log: https://github.com/huawei-noah/SMARTS/runs/4490851081?check_suite_focus=true#step:5:253
 
**System information**
 
- Operating system & Environment [Please note if using Docker]:
github worker smarts 0.4.18 docker image.
- Number of system cores available
2
- Python version [Please note if using Jupyter]
python 3.7
- Python packages [e.g. from `pip freeze`]
n/a
- SUMO version
1.08
- Date when bug occurred
 
**Impact [If known]**
It appears as if it may not be possible to use multiple `Renderer`s on the same process.

sah-huawei:
fwiw, I can't reproduce this locally.  also, I wanted to double-check about which test shows this.  In the "High Level Description" section it says `test_renderers.py` but in the "Steps to Reproduce" it says `test_python_version.py`.  (Both test files currently pass for me locally.)

Gamenot:
Sorry, that was my mistake the reproduction should be in `test_renderers.py`. I have a more intensive test running and it does not seem to be catching anything. As always it could be an issue with the Github workers.

sah-huawei:
If there's a way to do this on the Github worker, it might be worth checking:
```
from panda3d.core import Thread
Thread.isThreadingSupported()
```
This should be `True` when Panda has been compiled with threading support.  (I checked this locally and mine has.)

Gamenot:
Another test has failed the same way: https://github.com/huawei-noah/SMARTS/runs/4523507673?check_suite_focus=true#step:5:253

PR #1179, opened 2021-12-14
Investigate failing rendering threads

Gamenot:
The rendering test appears to be failing consistently now and causing the tests to fail.

See: #1175

### DIFF
--- a/smarts/core/tests/test_renderers.py
+++ b/smarts/core/tests/test_renderers.py
@@ -23,6 +23,7 @@ import math
 import threading
 
 import pytest
+from panda3d.core import Thread as p3dThread
 
 from smarts.core.agent_interface import (
     ActionSpaceType,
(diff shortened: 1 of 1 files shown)


smarts/core/tests/test_renderers.py
```diff
@@ -104,6 +105,7 @@ def test_renderer(self):
 
 
 def test_multiple_renderers(scenario):
+    assert p3dThread.isThreadingSupported()
```
Gamenot, line 108: This was mentioned by @sah-huawei as something we should check. At minimum it should help narrow down the cause of the test failure.

sah-huawei: Darn.  I was hoping it would be as simple as that, but it looks like not (given that the tests for this commit pass).

I would still opt for leaving this assert in (i.e., I think we can go ahead and merge this PR even though it didn't find the problem in Issue #1175) because it could be useful for other users who've already installed panda3d on their own system.

sah-huawei (approved):

Issue #1168, opened 2021-12-10
Avoid creating road map multiple times


saulfield:
**Is your feature request related to a problem? Please describe.**
It appears that `smarts.core.default_map_factory.create_road_map(...)` is called 3 times per episode, which can be expensive for large maps.

**Describe the solution you'd like**
Some way of caching maps that have already been loaded, and potentially a refactor of the uses of `create_road_map` in `smarts.core.scenario`.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1180, opened 2021-12-15
Map Creation Refactor

sah-huawei:
This PR addresses Issue #1168 by adding a simple cache (that only keeps at most 1 map in memory at a time) to the default `RoadMap` builder.

Along the way, I changed from a "factory" pattern to a "builder" pattern for creating `RoadMap` objects (to facilitate downstream projects defining their own builder for other map formats and passing it to SMARTS without having to modify or add anything to the SMARTS codebase).

This also necessitated finally adding the `MapSpec` type to our SStudio DSL, although it remains optional to specify it that way.  (This is also another small stepping stone towards #830.)


smarts/sstudio/types.py
```diff
@@ -240,6 +241,33 @@ class BoidAgentActor(SocialAgentActor):
...
+    """A path or URL or name uniquely designating the map source."""
+    lanepoint_spacing: Optional[float] = None
+    """If specified, the default distance between pre-generated Lane Points (Waypoints)."""
+    default_lane_width: Optional[float] = None
+    """If specified, the default width (in meters) of lanes on this map."""
+    builder_fn: Optional[MapBuilder] = get_road_map
+    """If specified, this should return an object derived from the RoadMap base class
+    and a hash that uniquely identifies it (changes to the hash should signify
+    that the map is different enough that map-related caches should be reloaded.
+    The parameter is this Map object itself.
```
saulfield, line 266:
```suggestion
    The parameter is this MapSpec object itself.
```

saulfield (commented):
This looks great! How about a test that uses the new `MapSpec` type? :)

sah-huawei:
> This looks great! How about a test that uses the new `MapSpec` type? :)

Good call.  Will add!

sah-huawei:
> > This looks great! How about a test that uses the new `MapSpec` type? :)
> 
> Good call. Will add!

Done.

saulfield (approved):

RutvikGupta (approved):

RutvikGupta:
@sah-huawei Can u merge this soon? We want to extend the refactoring here to support `OpenDRIVE` maps before we open the PR.

smarts/sstudio/genscenario.py
```diff
@@ -121,13 +133,22 @@ def gen_scenario(
         )
 
 
+def gen_map(scenario: str, map_spec: types.MapSpec, output_dir: str = None):
+    output_path = os.path.join(output_dir or scenario, "map_spec.pkl")
+    with open(output_path, "wb") as f:
+        # we use cloudpickle here instead of pickle because the
+        # map_spec object may contain a reference to a map_builder callable
+        cloudpickle.dump(map_spec, f)
```
Gamenot, line 141: I think this might be overly unsafe to load a method from a file that could be from an unknown source without any checks.

The first warning `cloudpickle` mentions in documentation is to not load data from un-trusted sources.

> **Security notice**: one should only load pickle data from trusted sources as otherwise pickle.load can lead to arbitrary code execution resulting in a critical security vulnerability.
https://github.com/cloudpipe/cloudpickle#cloudpickle

`cloudpickle` also mentions that `pickle` handles references. Are we doing something dynamic that is causes `pickle` to fail?

> An important difference between cloudpickle and pickle is that cloudpickle can serialize a function or class by value, whereas pickle can only serialize it by reference. Serialization by reference treats functions and classes as attributes of modules, and pickles them through instructions that trigger the import of their module at load time.
https://github.com/cloudpipe/cloudpickle#overriding-pickles-serialization-mechanism-for-importable-constructs

smarts/core/sumo_road_network.py
```diff
@@ -181,6 +183,21 @@ def source(self) -> str:
         """This is the net.xml file that corresponds with our possibly-offset coordinates."""
         return self._net_file
 
+    def is_same_map(self, map_spec: MapSpec) -> bool:
+        dlw = (
+            default_lane_width
```
Gamenot, line 188: I do not see where this is defined. Perhaps:

```suggestion
            map_spec.default_lane_width
```

Gamenot (changes requested):

smarts/core/sumo_road_network.py
sah-huawei: > I do not see where this is defined. Perhaps:

Good catch.  Fixed.

smarts/sstudio/genscenario.py
sah-huawei: As discussed earlier, I've added a comment to warn about this as well as filing Issue #1188 to track this in our backlog.  (Please feel free to update the issue as you see fit.)

Gamenot (approved):
We had some discussion and allowed this until we need to look at #1188.

Issue #1181, opened 2021-12-15
Rewrite the ray_multi_instance example in test


cz0204:
The ray_multi_instance example in the set up phase while executing 'make sanity-test' is not designed as what it is intended to. The objective is to show that SMARTS can run in parallel but it is now only a few scenarios run in sequence. Also, there is no backpropagation in the training for this example.

PR #1182, opened 2021-12-15
delete example ray multi-instance

cz0204:
The multi-instance example is deleted while run 'make sanity-test' in the set up phase. Another related issue is created for #1181


Gamenot (approved):
I was going to say that we should fix it but removing it is probably correct.

cz0204:
yes. We should finally fix the example as what we intended to do. So I start another issue #1181 for fixing the example later.

PR #1185, opened 2021-12-16
Improvements on rendering tests

Gamenot:
Along debugging issues with rendering I found that the tests were not catching issues with rendering. I have updated the rendering test and added some configuration and documentation.

This will allow rendering without need for X11.


smarts/core/smarts.py
```diff
@@ -596,15 +612,20 @@ def observe_from(self, vehicle_ids):
...
-                    self._renderer.setup(self._scenario)
-                    self._vehicle_index.begin_rendering_vehicles(self._renderer)
+                self._renderer = get_renderer(self._sim_id, self._config)
+            except ImportError as e:
+                self._log.warning("Import failed: " + repr(e))
+                raise e
             except Exception as e:
                 self._log.warning("unable to create Renderer:  " + repr(e))
                 self._renderer = None
+                raise e
```
Gamenot, line 624: I have elevated these to re-raise assertions so that people do not scratch their head.

A note here that I may not need the `repr()` if I am re-raising.

smarts/core/renderer.py
```diff
@@ -47,25 +48,59 @@
...
+    """Get the appropriate renderer given the simulation id and config.
+    Args:
+        sim_id (str): The id of a SMARTS simulation the renderer should be related to.
+        config (SmartsConfig): A configuration file determining extra configuration.
+
+    Returns:
+        Renderer: A new renderer based on the given configuration.
+    """
+    _renderer = Renderer(sim_id, config)
+    return _renderer
```
sah-huawei, line 64: Given that it's just a local variable, I'm not sure why the underscore prefix was used for `_renderer` although it's of course fine.

smarts/core/renderer.py
```diff
@@ -76,6 +111,14 @@ def __new__(cls):
             it.init()
         return it
 
+    @classmethod
+    def set_rendering_mode(cls, debug_mode: DEBUG_MODE):
+        cls._debug_mode = debug_mode
```
sah-huawei, line 116: This will have no effect after the `_ShowBaseInstance` object has been created, right?
If not, maybe add an assert here if called too late.

smarts/core/renderer.py
```diff
@@ -76,6 +111,14 @@ def __new__(cls):
...
+
+    @classmethod
+    def set_rendering_backend(cls, backend: str):
+        cls._rendering_backend = backend
```
sah-huawei, line 120: Same as comment on `set_rendering_mode()`.

smarts/core/smarts.py
```diff
@@ -77,26 +78,41 @@ class SMARTSNotSetupError(Exception):
 
 
 class SMARTS:
+    """The SMARTS simulator. This is the direct interface to all parts of the simulation.
+    Args:
+        agent_interfaces: The interfaces providing SMARTS with the understanding of what features an agent needs.
```
sah-huawei, line 83: I suggest changing "an" to "each".

smarts/core/smarts.py
```diff
@@ -596,15 +612,22 @@ def observe_from(self, vehicle_ids):
...
 
-                self._renderer = Renderer(self._sim_id)
-                if self._scenario:
-                    self._renderer.setup(self._scenario)
-                    self._vehicle_index.begin_rendering_vehicles(self._renderer)
+                self._renderer = get_renderer(self._sim_id, self._config)
+            except ImportError as e:
+                self._log.warning(
+                    "Renderer import failed. Please install `[camera-obs]`"
+                )
```
sah-huawei, line 621: They could also just remove the camera sensors from their agents.  You might add this as an alternative in the warning message.

smarts/core/tests/test_renderers.py
```diff
@@ -35,22 +36,51 @@
 from smarts.core.scenario import Scenario
 from smarts.core.smarts import SMARTS
 from smarts.core.sumo_traffic_simulation import SumoTrafficSimulation
+from smarts.core.utils.config import SmartsConfig
 from smarts.core.vehicle import RendererException
 
 
+def _smarts_with_agent(request, agent) -> SMARTS:
```
sah-huawei, line 43: I don't see where the `request` parameter gets used here (or anywhere else below).

smarts/core/tests/test_renderers.py
```diff
@@ -103,25 +140,38 @@ def test_renderer(self):
...
-def test_optional_renderer(smarts, scenario):
-    smarts.reset(scenario)
+def test_optional_renderer(smarts: SMARTS, scenario):
     assert not smarts.is_rendering
-    for _ in range(10):
-        smarts.step({})
-
     renderer = smarts.renderer
     if not renderer:
         raise RendererException.required_to("run test_renderer.py")
```
sah-huawei, line 157: This seems odd.  If the `assert` passes, then it seems like it will throw here.

smarts/core/tests/test_renderers.py
```diff
@@ -103,25 +140,38 @@ def test_renderer(self):
...
 
+    smarts._renderer = None
```
sah-huawei, line 159: I think we also need to check that `smarts._renderer` defaults to `None` if no camera sensors are attached.

smarts/core/smarts.py
```diff
@@ -77,26 +78,41 @@ class SMARTSNotSetupError(Exception):
...
+        agent_interfaces: Dict[str, AgentInterface],
         traffic_sim,  # SumoTrafficSimulation
         envision: EnvisionClient = None,
         visdom: VisdomClient = None,
         fixed_timestep_sec: float = 0.1,
         reset_agents_only: bool = False,
-        zoo_addrs=None,
+        zoo_addrs: Optional[Tuple[str, int]] = None,
         external_provider: bool = False,
+        config=SmartsConfig(),
```
sah-huawei, line 104: I'm not sure about adding a custom `SmartsConfig` object just for renderer configuration, especially if there are (currently) only two values that might get set there.  I'm wondering if this can be done a different way.

smarts/core/tests/test_renderers.py
Gamenot: This part of the test requires SMARTS to create a `Renderer`. It should never throw here if rendering is working correctly.

smarts/core/tests/test_renderers.py
Gamenot: That is done in `test_no_renderer(..)`.

smarts/core/tests/test_renderers.py
Gamenot: I gets added through the command line from the local `conftest.py`.

smarts/core/smarts.py
Gamenot: I removed the config and have a direct configuration to the renderer.

smarts/core/renderer.py
Gamenot: I have set it to affect at any point.

smarts/core/renderer.py
Gamenot: Removed this utility for now.

smarts/core/utils/config.py
```diff
@@ -0,0 +1,70 @@
...
+        "p3headlessgl",
+        "pandagl",
+        "pandadx9",
+        "pandagles",
+        "pandagles2",
+        "p3tinydisplay",
+    },
+    # See https://docs.panda3d.org/1.10/python/programming/configuration/configuring-panda3d
+}
+
```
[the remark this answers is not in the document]
Gamenot: It is only needed when rendering. I have set defaults for now.

sah-huawei (approved):
Just some minor comments you can address when you want, but should be good to merge whenever.

smarts/core/renderer.py
```diff
@@ -42,30 +45,57 @@
...
 from .colors import SceneColors
 from .coordinates import Pose
 from .masks import RenderMasks
 from .scenario import Scenario
 
 
+class DEBUG_MODE(IntEnum):
+    """The rendering debug information level."""
+
+    SPAM = 1
```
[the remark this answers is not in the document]
Gamenot: It is probably more accurate, at that level you tend to lose most useful information. :)

smarts/core/renderer.py
```diff
@@ -139,24 +175,39 @@ def render_node(self, sim_root: NodePath):
...
+    @property
+    def log(self) -> logging.Logger:
+        """The rendering logger."""
+        return self._log
+
+    def remove_buffer(self, buffer):
+        """Remove the rendering buffer."""
+        self._showbase_instance.graphicsEngine.removeWindow(buffer)
+
+    def setup(self, scenario: Scenario) -> None:
```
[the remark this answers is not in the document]
Gamenot: It did not have to be specified, I can remove it, I added it while I was somewhat frustrated with satisfying the types problems.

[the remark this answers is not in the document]
Gamenot: It has been removed.

smarts/core/tests/test_renderers.py
```diff
@@ -65,21 +89,25 @@ def scenario():
...
-        except Exception as e:
+            from smarts.core.renderer import Renderer, DEBUG_MODE as RENDERER_DEBUG_MODE
+
+            self._rdr = Renderer(
+                self._rid, RENDERER_DEBUG_MODE[renderer_debug_mode.upper()]
+            )
+        except ImportError as e:
+            raise e
+        except Exception:
             raise RendererException.required_to("run test_renderer.py")
```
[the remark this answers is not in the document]
Gamenot: You are right. I think I was concerned about unrelated errors and got the cases flipped around.

`ImportError` should throw the `RendererException` and `Exception` should just re-throw.

saulfield (approved):

PR #1190, opened 2021-12-18
Fix SMARTS.__del__() throwing at program exit

### LINKED ISSUE #1189: AttributeError: 'AgentManager' object has no attribute 'log'
**High Level Description**
When a SMARTS instance is created along with a scenarios_iterator, it will raise an AttributeError when tearing down agent manager.

**Desired SMARTS version**
develop branch @ commit 46c34971066c7177335897ac72a8620e9054734f

**Steps to reproduce**
```python


smarts = SMARTS(
    agent_interfaces={},
    traffic_sim=None,
    envision=None,
)

scenarios_iterator = Scenario.scenario_variations(["./ngsim"], [])
```

**Results**

```
Exception ignored in: <function SMARTS.__del__ at 0x7f989db9f710>
  File "/NAS2020/Workspaces/DRLGroup/zbzhu/MADPO/SMARTS_Imitation/SMARTS/smarts/core/smarts.py", line 505, in __del__
  File "/NAS2020/Workspaces/DRLGroup/zbzhu/MADPO/SMARTS_Imitation/SMARTS/smarts/core/smarts.py", line 483, in destroy
  File "/NAS2020/Workspaces/DRLGroup/zbzhu/MADPO/SMARTS_Imitation/SMARTS/smarts/core/smarts.py", line 460, in teardown
  File "/NAS2020/Workspaces/DRLGroup/zbzhu/MADPO/SMARTS_Imitation/SMARTS/smarts/core/agent_manager.py", line 72, in teardown
  File "/home/zbzhu/miniconda3/envs/smarts/lib/python3.7/logging/__init__.py", line 1365, in debug
  File "/home/zbzhu/miniconda3/envs/smarts/lib/python3.7/logging/__init__.py", line 1621, in isEnabledFor
TypeError: 'NoneType' object is not callable
Exception ignored in: <function BulletClient.__del__ at 0x7f98c3ed3950>
  File "/home/zbzhu/miniconda3/envs/smarts/lib/python3.7/site-packages/pybullet_utils/bullet_client.py", line 38, in __del__
AttributeError: 'NoneType' object has no attribute 'error'
```

Gamenot:
See `CHANGELOG`.

Fixes #1189

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -650,11 +650,15 @@ class SMARTS:
             )
 
     def __del__(self):
-        if not self._is_destroyed:
+        try:
+            self.destroy()
+        except (TypeError, AttributeError) as e:
             # This is a print statement because the logging module may be deleted at program exit.
-            print(
+            raise SMARTSDestroyedError(
                 "ERROR: A SMARTS instance may have been deleted by gc before a call to destroy."
-                " Please call `SMARTS.destroy()` to make this error go away."
+                " Please explicitly call `del obj` or `SMARTS.destroy()` to make this error"
+                " go away.",
+                e,
             )
 
     def _teardown_vehicles(self, vehicle_ids):
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -76,6 +76,10 @@ class SMARTSNotSetupError(Exception):
     pass
 
 
+class SMARTSDestroyedError(Exception):
+    pass
+
+
 class SMARTS:
     def __init__(
         self,
@@ -91,6 +95,7 @@ class SMARTS:
         self._log = logging.getLogger(self.__class__.__name__)
         self._sim_id = Id.new("smarts")
         self._is_setup = False
+        self._is_destroyed = False
         self._scenario: Optional[Scenario] = None
         self._renderer = None
         self._envision: Optional[EnvisionClient] = envision
@@ -171,6 +176,7 @@ class SMARTS:
         """Note the time_delta_since_last_step param is in (nominal) seconds."""
         if not self._is_setup:
             raise SMARTSNotSetupError("Must call reset() or setup() before stepping.")
+        self._check_valid()
         assert not (
             self._fixed_timestep_sec and time_delta_since_last_step
         ), "cannot switch from fixed- to variable-time steps mid-simulation"
@@ -342,6 +348,7 @@ class SMARTS:
         raise first_exception
 
     def _reset(self, scenario: Scenario):
+        self._check_valid()
         if (
             scenario == self._scenario
             and self._reset_agents_only
@@ -390,6 +397,7 @@ class SMARTS:
         return observations_for_ego
 
     def setup(self, scenario: Scenario):
+        self._check_valid()
         self._scenario = scenario
 
         self._bubble_manager = BubbleManager(scenario.bubbles, scenario.road_map)
@@ -414,6 +422,7 @@ class SMARTS:
         provider: Provider,
         recovery_flags: ProviderRecoveryFlags = ProviderRecoveryFlags.EXPERIMENT_REQUIRED,
     ):
+        self._check_valid()
         assert isinstance(provider, Provider)
         self._insert_provider(len(self._providers), provider, recovery_flags)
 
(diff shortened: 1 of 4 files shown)


smarts/core/smarts.py
```diff
@@ -555,7 +555,7 @@ def teardown(self):
 
     def destroy(self):
         self.teardown()
-
+        self._vehicle_index = None
```
Gamenot, line 558: This line is needed to make sure that `SMARTS` does not try to call the vehicle index cleanup code again.

Adaickalavan: Is the line `self._vehicle_index=None` included here so that when `destroy()` is called again, the `self._vehicle_index.teardown()` inside `teardown()` will not be called again?

Why and how is the `destroy()` function being called multiple times within a code?

Gamenot: > Is the line self._vehicle_index=None included here so that when destroy() is called again, the self._vehicle_index.teardown() inside teardown() will not be called again?

Yes, although, I think there is the potential alternative that instead the `_vehicle_index` could be set to `None` in the `teardown()` method and then no special action is required in the `destroy()` method.

> Why and how is the destroy() function being called multiple times within a code?

The problem case is if `destroy()` is called and then the `SMARTS` instance reaches the end of the program and is cleaned up by uncontrolled garbage collection where some modules might not be available by the time the instance `__del__()` is called.

However, I am not sure if we should remove the `destroy()` call from `__del__()` because putting `destroy()` in `__del__()` allows `del` to also properly clean up a `SMARTS` instance, and provides a shorthand (using the exception catch) for figuring out if `SMARTS` has already been properly cleaned up.

Adaickalavan: > The problem case is if destroy() is called and then the SMARTS instance reaches the end of the program and is cleaned up by uncontrolled garbage collection where some modules might not be available by the time the instance __del__() is called.

Gamenot: > This line is needed to make sure that SMARTS does not try to call the vehicle index cleanup code again.

Sorry, I think my original comment was misleading, even though this is the effect of the change. The current version of `destroy()` and `teardown()` can be called any number of times without problem. The crashing issue with `destroy()` or `teardown()` only occurs at the end of program garbage collection. I can note the following cases:

1) No Error: `del smarts` is called which calls `__del__()` removing the SMARTS object before it reaches end of program.
2) Fixed error: `destroy()` is called which then will cause an error at end of program because the attribute `_vehicle_index` was not set to `None` and it will access modules that may already be deleted.
3) Error: Neither are called and `destroy()` throws an error  caused by end of program garbage collection which is caught by `__del__()` and then asks the user to call `destroy()` explicitly.

To do exactly what you are suggesting I would need to repurpose `__del__()` to only check if `destroy()` was previously called before `__del__()` and then disallow multiple calls to `destroy()`. This might be reasonable, but I am on the fence about it because it removes the `del smarts_obj` syntax.

sah-huawei: > However, I am not sure if we should remove the `destroy()` call from `__del__()` because putting `destroy()` in `__del__()` allows `del` to also properly clean up a `SMARTS` instance, and provides a shorthand (using the exception catch) for figuring out if `SMARTS` has already been properly cleaned up.

What about the SMARTS object needs to be "properly cleaned up" (as opposed to just cleaned up with normal garbage collection)?  I.e., I'd like to understand why normal garbage collection is not enough for us (either in the normal shutdown case or in the exception case).

And whatever this is should be clearly documented in a comment in `__del__()` (and/or `destroy()`).

Gamenot: I believe that the main sources of problem requiring cleanup are port, file, and subprocess related: TraCI and grpc. I believe the `SumoTrafficSimuation` and `AgentManager` are the core parts that would want guaranteed shutdown.

I think to get the exact specifics would need some additional information. As it stands, I am now leaning towards removing the cleanup from `__del__()` and requiring `destroy()` to be called.

Gamenot:
I have made the requirement to call `destroy()` explicit.

smarts/core/smarts.py
sah-huawei: OK, yeah, I think you're right about Sumo Traffic and the Remote (zoo) Agents, both of which spawn sub-processes that need to be killed.  It might be worth adding a comment about that in either `__del__()` or `destroy()`.

I think we should make both `destroy()` and `__del__()`  as idempotent as possible.  I think your changes here (in particular, setting `self._vehicle_index` to `None`) achieve that, or at least improve on the situation, so I'm inclined to approve this "as is" (with the addition of the comment I requested).

The nice part of keeping the cleanup (call to `destroy()`) in `__del__()` too is that it allows for multiple SMARTS objects to be created (without having to require the user to explicitly call `destroy()` in between, which is an oddity in general) without leaking processes.  But I would support either approach if you guys feel strongly one way or another.

sah-huawei (approved):
As I wrote in the comments, I approve this "as is", although I would also support it if you changed it to require explicitly calling `destroy()`.

Adaickalavan (approved):

PR #1201, opened 2021-12-21
Map Creation Fixes

sah-huawei:
As pointed out by @RutvikGupta , `is_same_map()` in `SumoRoadNetwork` objects did not always return `True` when it should (e.g, in cases where the `net_file` name was changed if it had been shifted, or when the `net_file`filename was filled in with the default).

This fixes that by keeping track of the `MapSpec` object used to create the map.

### DIFF
--- a/smarts/core/default_map_builder.py
+++ b/smarts/core/default_map_builder.py
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+from dataclasses import replace
 import os
 from typing import NamedTuple, Tuple
 
@@ -58,23 +59,19 @@ def get_road_map(map_spec) -> Tuple[RoadMap, str]:
         _existing_map = None
         gc.collect()
 
-    map_path = map_spec.source
-    if not os.path.isfile(map_path):
+    if not os.path.isfile(map_spec.source):
         map_path = os.path.join(map_spec.source, "map.net.xml")
         if not os.path.exists(map_path):
             raise FileNotFoundError(
                 f"Unable to find map in map_source={map_spec.source}."
             )
+        map_spec = replace(map_spec, source=map_path)
 
     # Keep this a conditional import so Sumo does not have to be
     # imported if not necessary:
     from smarts.core.sumo_road_network import SumoRoadNetwork
 
-    road_map = SumoRoadNetwork.from_file(
-        map_path,
-        default_lane_width=map_spec.default_lane_width,
-        lanepoint_spacing=map_spec.lanepoint_spacing,
-    )
+    road_map = SumoRoadNetwork.from_spec(map_spec)
 
     road_map_hash = file_md5_hash(road_map.source)
 
--- a/smarts/core/sumo_road_network.py
+++ b/smarts/core/sumo_road_network.py
@@ -139,7 +139,7 @@ class SumoRoadNetwork(RoadMap):
 
     @classmethod
     def from_spec(cls, map_spec: MapSpec, shift_to_origin: bool = False):
-        net_file = map_spec.source
+        net_file = SumoRoadNetwork._map_path(map_spec)
 
         # Connections to internal lanes are implicit. If `withInternal=True` is
         # set internal junctions and the connections from internal lanes are
@@ -177,18 +177,17 @@ class SumoRoadNetwork(RoadMap):
             else SumoRoadNetwork.DEFAULT_LANE_WIDTH
         )
 
+    @staticmethod
+    def _map_path(map_spec: MapSpec) -> str:
+        if os.path.isdir(map_spec.source):
+            # map.net.xml is the default Sumo map name; try that:
+            return os.path.join(map_spec.source, "map.net.xml")
+        return map_spec.source
+
     def is_same_map(self, map_spec: MapSpec) -> bool:
         if map_spec.source != self._map_spec.source:
-            cur_source = (
-                map_spec.source
-                if os.path.isfile(map_spec.source)
-                else os.path.join(map_spec.source, "map.net.xml")
-            )
-            orig_source = (
-                self._map_spec.source
-                if os.path.isfile(self._map_spec.source)
-                else os.path.join(self._map_spec.source, "map.net.xml")
-            )
+            cur_source = SumoRoadNetwork._map_path(map_spec)
+            orig_source = SumoRoadNetwork._map_path(self._map_spec)
             if cur_source != orig_source:
                 return False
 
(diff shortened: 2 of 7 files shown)


smarts/core/default_map_builder.py
```diff
@@ -58,23 +59,19 @@ def get_road_map(map_spec) -> Tuple[RoadMap, str]:
...
 
-    map_path = map_spec.source
-    if not os.path.isfile(map_path):
+    if not os.path.isfile(map_spec.source):
         map_path = os.path.join(map_spec.source, "map.net.xml")
         if not os.path.exists(map_path):
             raise FileNotFoundError(
                 f"Unable to find map in map_source={map_spec.source}."
             )
+        map_spec = replace(map_spec, source=map_path)
```
RutvikGupta, line 68: Why is there a need to use `dataclasses.replace` here? Why can't you use `map_spec.source = map_path`?

smarts/core/default_map_builder.py
```diff
@@ -58,23 +59,19 @@ def get_road_map(map_spec) -> Tuple[RoadMap, str]:
         _existing_map = None
         gc.collect()
 
-    map_path = map_spec.source
-    if not os.path.isfile(map_path):
+    if not os.path.isfile(map_spec.source):
         map_path = os.path.join(map_spec.source, "map.net.xml")
```
RutvikGupta, line 63: Isn't there a need to check if the offset were normalized and the shifted network file needs to be used here since we create the road map class using this file?

smarts/core/sumo_road_network.py
```diff
@@ -143,13 +138,9 @@ def _shift_coordinates(cls, net_file_path, shifted_path):
...
     @classmethod
-    def from_file(
-        cls,
-        net_file,
-        shift_to_origin=False,
-        default_lane_width=None,
-        lanepoint_spacing=None,
-    ):
+    def from_spec(cls, map_spec: MapSpec, shift_to_origin: bool = False):
+        net_file = map_spec.source
```
RutvikGupta, line 142: Can we safely assume here that `map_spec.source` can be a path to the network file now as opposed to a path to scenario directory? I thought you told me that a user can set `MapSpec` when creating the Scenario?

smarts/core/sumo_road_network.py
```diff
@@ -171,32 +162,47 @@ def from_file(
...
             )
-            and dlw == self._default_lane_width
-        )
+            orig_source = (
+                self._map_spec.source
+                if os.path.isfile(self._map_spec.source)
+                else os.path.join(self._map_spec.source, "map.net.xml")
+            )
+            if cur_source != orig_source:
+                return False
```
RutvikGupta, line 193: Do we need to consider the `shifted_AUTOGEN_map.net.xml` file here for both current and original sources?

smarts/core/sumo_road_network.py
```diff
@@ -171,32 +162,47 @@ def from_file(
...
+
+        if map_spec.lanepoint_spacing != self._map_spec.lanepoint_spacing:
+            return False
+
+        if (
+            map_spec.default_lane_width != self._map_spec.default_lane_width
+            and SumoRoadNetwork._spec_lane_width(map_spec)
+            != SumoRoadNetwork._spec_lane_width(self._map_spec)
```
RutvikGupta, line 201: Isn't this check redundant since it is being checked by the first part of this and statement?

smarts/sstudio/genscenario.py
```diff
@@ -434,7 +434,8 @@ def gen_traffic_histories(scenario: str, histories_datasets, overwrite: bool):
     if os.path.exists(road_network_path):
         from smarts.core.sumo_road_network import SumoRoadNetwork
 
-        road_network = SumoRoadNetwork.from_file(road_network_path)
+        map_spec = types.MapSpec(road_network_path)
+        road_network = SumoRoadNetwork.from_spec(map_spec)
```
RutvikGupta, line 438: Why is there a need to re-create `sumo_road_network`

smarts/core/default_map_builder.py
sah-huawei: Not here.  The shifting of a the map is just a SUMO-specific thing... we try to encapsulate it and the shifted map's name entirely in the `SumoRoadNetwork` class.  However here, in `default_map_builder.py`, we are trying to determine _which_ type of map the scenario uses, in general (SUMO or OpenDrive), so `map_spec.source` is just a folder, but there is presence of a `map.net.xml` file, which is the default Sumo road network filename, this is indicative that we are using Sumo and hence we can proceed.  

Note that since this hasn't yet been merged with the OpenDrive PR, all the current implementation is doing is just trying to determine if this scenario indeed uses a Sumo road network or not.  Even if we ultimately shift the map in `sumo_road_network.py`, there sill should always be a `map.net.xml` file in the folder if there wasn't a network file named explicitly in the `map_spec`.

smarts/core/sumo_road_network.py
sah-huawei: Yeah, I suppose we should add `map.net.xml` to it here (notwithstanding what's done in `default_map_builder.py`) if it's still just a folder, just in case we arrive here having not gone through the default builder.  Good point!  Will do.

smarts/core/sumo_road_network.py
sah-huawei: No, because that happens "downstream" of the `map_spec`.  That is, even if it gets shifted in `from_spec()`, the original value is still maintained in `self._map_spec`.

smarts/core/sumo_road_network.py
sah-huawei: The second part here is to catch the case where the original `map_spec` specified `None` whereas the new one explicitly specifies `3.2` (which is what `DEFAULT_LANE_WIDTH` is).  In that case, we'd still consider these the "same map" even though their `map_specs` are not identical.

smarts/sstudio/genscenario.py
sah-huawei: Good point.  This is tricky because `gen_traffic_histories()` can be called independently or from `gen_scenario()`, and it might have been created in `gen_scenario()`.   I'll make it a parameter with a default value to solve this.

smarts/core/default_map_builder.py
sah-huawei: We could probably allow this given our current implementations, but it's usually considered "bad practice" to change a data object being passed to you b/c it changes it in the caller too, which the caller might not expect (unless it's documented that way).  So that's why I'm avoiding it here.

smarts/core/sumo_road_network.py
sah-huawei: Done.

smarts/sstudio/genscenario.py
sah-huawei: Done.

smarts/core/default_map_builder.py
RutvikGupta: Yup, this makes it easy for OpenDRIVE extension since we only have one network file here.

RutvikGupta (approved):

Gamenot (approved):

Issue #1176, opened 2021-12-13
Update to `setup.cfg`


Gamenot:
**Is your feature request related to a problem? Please describe.**
`setup.py` is an overly verbose older style way of listing python dependencies.

**Describe the solution you'd like**
Switch to using `setup.cfg` as the default. Keep a simple `setup.py` that calls `setuptools.setup()`.

**Describe alternatives you've considered**
Leave as is.

**Additional context**
See the following for context: https://stackoverflow.com/questions/39484863/whats-the-difference-between-setup-py-and-setup-cfg-in-python-projects/60885212#60885212

Gamenot:
A new consideration is to use `pyproject.toml` and `Pipfile` with `Pipfile.lock`.

PR #1203, opened 2021-12-21
Upgrade to setup.cfg

qianyi-sun:
#1176


setup.py
```diff
@@ -1,94 +1,4 @@
-from os import path
-from setuptools import setup, find_packages
+from setuptools import setup
```
Gamenot, line 1: Also, it looks like if we use [the newest version of `pip`](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#editable-installation) we will not need the `setup.py` workaround. We may still want to keep it though since some users will not be using the most recent version of `pip`.

Gamenot, line 1: I can also note that this `setup.py` is probably why the CI is failing because it expects `setuptools` but that is not required for the build.

`pip` documentation mentions that the following is the default build system:
```
[build-system]
requires = ["setuptools>=40.8.0", "wheel"]
build-backend = "setuptools.build_meta:__legacy__"
```

Gamenot (commented):
I think this deserves more conversation since it looks like `setuptools` does not yet support `pyproject.toml` which is a bit painful. 

So it means choice between the following:
- Drop `pyproject.toml` and use `setup.cfg`
- Use `pyproject.toml` with `poetry` back-end and require a `>=31` version of `pip`
- Use `poetry` for both front-end and back-end
- Use `pipenv`

Gamenot:
I have updated my suggestions.

PR #1205, opened 2021-12-22
Misc. backfill

sah-huawei:
A collection of miscellaneous changes that were made in a SMARTS fork for behavioral model development.


smarts/core/scenario.py
```diff
@@ -551,7 +551,7 @@ def _extract_mission(mission, road_map):
 
         def resolve_offset(offset, lane_length):
             # epsilon to ensure we are within this edge and not the subsequent one
-            epsilon = 1e-6
+            epsilon = 2 if lane_length > 2 else 1e-6
```
Gamenot, line 554: I think there is a bigger issue here, the problem this tried to address is that injecting a vehicle so close to the beginning of the lane would potentially inject on top of another vehicle but this is not the way to fix that issue.

Gamenot (approved):

smarts/core/scenario.py
sah-huawei: Do you want to file an issue about this?

Gamenot: https://github.com/huawei-noah/SMARTS/issues/1217

smarts/core/controllers/trajectory_tracking_controller.py
```diff
@@ -208,14 +208,39 @@ def perform_trajectory_tracking_PD(
         lateral_gain = lerp(3, final_lateral_gain, normalized_speed)
         heading_gain = lerp(0.03, final_heading_gain, normalized_speed)
         steering_filter_constant = lerp(
-            2, final_steering_filter_constant, normalized_speed
+            12, final_steering_filter_constant, normalized_speed
         )
-        heading_error_derivative_gain = lerp(
+        heading_error_derivative_gain = 0 * lerp(
             1.5, final_heading_error_derivative_gain, normalized_speed
         )
```
Gamenot, line 215: This seems to be invalidating the configuration. That is not necessarily wrong because the controller can interpret the configuration in any way it chooses but it is definitely controller configuration.

These changes to the controller appear to be overly specific.

sah-huawei: Yes, I agree.  That's how it was in the working (I assume) solution I saw in another repo, so that's why I brought that in here.  But I don't like it as is.  I think I'll use an if-block instead of multiplying by `0`, to make it clearer that the config values are not being used.

sah-huawei: ok, I tried to make the current situation a bit clearer (without changing the overall result).

smarts/core/controllers/trajectory_tracking_controller.py
```diff
@@ -208,14 +208,43 @@ def perform_trajectory_tracking_PD(
         lateral_gain = lerp(3, final_lateral_gain, normalized_speed)
         heading_gain = lerp(0.03, final_heading_gain, normalized_speed)
```
liu-allan, line 209: `lateral_gain` and `heading_gain` are being reassigned on line 225 and 226 without being used

https://github.com/huawei-noah/SMARTS/pull/1205/files#diff-4c1da4781052a940dd05e04f2ebf416b41ddf16f511b16e87fe1c41fa2dd6946R225-R226
```suggestion

```

smarts/core/controllers/trajectory_tracking_controller.py
```diff
@@ -208,14 +208,43 @@ def perform_trajectory_tracking_PD(
...
+            abs(
+                TrajectoryTrackingController.curvature_calculation(
+                    trajectory, 0, num_points=3
+                )
+            )
+            < 150
+        ):
+            heading_gain = 0.05
+            # lateral_gain = 0.71
+            # velocity_gain *= 0.5
```
liu-allan, line 241:
```suggestion

```

smarts/core/controllers/trajectory_tracking_controller.py
sah-huawei: Ok, I did a bit more reorg to clean this up some more.  See what you think now.

liu-allan (approved):

Issue #1207, opened 2021-12-23
Add type annotation test


Gamenot:
**Is your feature request related to a problem? Please describe.**
The is based on a need to detect issues with code such as cases where the return type is not as listed or where types are not determined properly.

**Describe the solution you'd like**
Add a `mypy` test to ensure that auto code documentation is consistent with the behaviour of the code. 

**Describe alternatives you've considered**
-Needs investigation

**Additional context**
See: https://mypy.readthedocs.io/en/stable/introduction.html

Adaickalavan:
An alternative type analyzer to consider would be [Pytype](https://google.github.io/pytype/) .

Gamenot:
We should go with Pytype for this test.

PR #1208, opened 2021-12-24
Type annotation

Adaickalavan:
This is an incremental change towards the type annotation issue #1207.


Gamenot (approved):
This looks like it can go straight in.

smarts/core/motion_planner_provider.py
```diff
@@ -29,6 +29,8 @@
 
 
 class MotionPlannerProvider(Provider):
+    _poses: np.ndarray
```
sah-huawei, line 32: What was the intent here?  (`MotionPlannerProvider._poses` is not the same thing as `self._poses` below.  Is this used anywhere?)

smarts/core/scenario.py
```diff
@@ -77,13 +75,13 @@ class Scenario:
...
         scenario_root: str,
-        route: str = None,
-        missions: Dict[str, Mission] = None,
-        social_agents: Dict[str, SocialAgent] = None,
-        log_dir: str = None,
+        route: Optional[str] = None,
+        missions: Optional[Dict[str, Mission]] = None,
+        social_agents: Optional[Dict[str, SocialAgent]] = None,
+        log_dir: Optional[str] = None,
         surface_patches: list = None,
```
sah-huawei, line 82: Why not `Optional` for `surface_patches` too?  (Maybe it could even be `Optional[Sequence[Dict[str, Any]]]`.)

smarts/sstudio/types.py
```diff
@@ -259,7 +259,7 @@ class MapSpec:
     """If specified, the default distance between pre-generated Lane Points (Waypoints)."""
     default_lane_width: Optional[float] = None
     """If specified, the default width (in meters) of lanes on this map."""
-    builder_fn: Optional[MapBuilder] = get_road_map
+    builder_fn: MapBuilder = get_road_map
```
sah-huawei, line 262: I'm not sure about this one since it's got a default value that isn't `None`.

smarts/core/scenario.py
Adaickalavan: The pytype throws hundreds of type errors. Not all type errors have been fixed yet. This pull request is an incremental fix.

I have included your suggested type for `surface_patches` now.

smarts/sstudio/types.py
Adaickalavan: Since `builder_fn` has a default value which is not `None`, the `Optional[ ]` type was removed.

smarts/core/motion_planner_provider.py
Adaickalavan: pytype throws attribute error on `self._poses`. See [here](https://google.github.io/pytype/errors.html#attribute-error) for explanation on attribute error and suggested solution.

sah-huawei: Ah ok, I guess that makes sense.  Thanks for the link.  

Maybe a short comment here might prevent someone else thinking this is extraneous and removing it in the future.

(Personally, I'm not a big fan of cluttering the code just to for `pytype`, but I guess this is just a matter of opinion.)

sah-huawei (approved):

smarts/sstudio/types.py
sah-huawei: oops, thanks.  I read that backwards originally.

PR #1215, opened 2021-12-29
HiWayEnv Gym Environment to support OpenDRIVE Maps

RutvikGupta:
Modified `HiWay` env to support `OpenDRIVE` so that `SMARTS` object is instantiated without the `SUMO` traffic provider and social agents. All existing examples not dependent on `SUMO` traffic provider (like imitation examples) can be simulated on `OpenDRIVE` scenarios. 
P.S.: This needs to be merged after `OpenDRIVE` support PR is merged.


examples/opendrive_agent.py
```diff
@@ -0,0 +1,89 @@
+import logging
```
Gamenot, line 1: I do not think we need this example. Let's just use the other example.

smarts/env/opendrive_env.py
```diff
@@ -0,0 +1,207 @@
...
+        if visdom:
+            visdom_client = VisdomClient()
+
+        self._smarts = SMARTS(
+            agent_interfaces=agent_interfaces,
+            traffic_sim=None,
+            envision=envision_client,
+            visdom=visdom_client,
+            fixed_timestep_sec=fixed_timestep_sec,
+        )
```
Gamenot, line 122: Looking at it, the only major change I see between this env and `HiWayEnv` is the traffic simulator. I could suggest we detect the version of scenario we are using instead and decide if to add the traffic.

Gamenot (changes requested):

smarts/env/opendrive_env.py
RutvikGupta: I don't think it will be a good idea to do so. `opendrive_road_network.py` does not support `SumoTrafficSimulation` but we still be importing it creating a dependency. If we do conditional imports based on scenario version there's still the case that we have a mix of OpenDRIVE and SUMO scenarios. I can add the change you suggested above, but it will look a bit messy.

smarts/env/hiway_env.py
```diff
@@ -129,21 +129,61 @@ def __init__(
...
+                    num_external_sumo_clients=num_external_sumo_clients,
+                    sumo_port=sumo_port,
+                    auto_start=sumo_auto_start,
+                    endless_traffic=endless_traffic,
+                ),
+                envision=envision_client,
+                visdom=visdom_client,
+                fixed_timestep_sec=fixed_timestep_sec,
+                zoo_addrs=zoo_addrs,
+            )
```
Gamenot, line 160: I think this also should work:
```suggestion
        traffic_sim = None
        if is_opendrive:
            # We currently don't support the Native SUMO Traffic Provider or Social Agents for OpenDRIVE maps
            pass
        else:
            from smarts.core.sumo_traffic_simulation import SumoTrafficSimulation
            traffic_sim = SumoTrafficSimulation(
                headless=sumo_headless,
                time_resolution=fixed_timestep_sec,
                num_external_sumo_clients=num_external_sumo_clients,
                sumo_port=sumo_port,
                auto_start=sumo_auto_start,
                endless_traffic=endless_traffic,
            )

        self._smarts = SMARTS(
            agent_interfaces=agent_interfaces,
            traffic_sim=traffic_sim,
            envision=envision_client,
            visdom=visdom_client,
            fixed_timestep_sec=fixed_timestep_sec,
            zoo_addrs=zoo_addrs,
        )
```

smarts/env/hiway_env.py
```diff
@@ -129,22 +129,59 @@ def __init__(
...
+            traffic_sim = SumoTrafficSimulation(
                 headless=sumo_headless,
                 time_resolution=fixed_timestep_sec,
                 num_external_sumo_clients=num_external_sumo_clients,
                 sumo_port=sumo_port,
                 auto_start=sumo_auto_start,
                 endless_traffic=endless_traffic,
-            ),
+            )
+            zoo_addrs = zoo_addrs
```
Gamenot, line 148: Sorry, I was still in the middle of editing.

```suggestion
```

smarts/env/hiway_env.py
```diff
@@ -129,22 +129,59 @@ def __init__(
...
 
-        self._smarts = SMARTS(
-            agent_interfaces=agent_interfaces,
-            traffic_sim=SumoTrafficSimulation(
+        is_opendrive = HiWayEnv.check_scenario_versions(scenarios)
+        traffic_sim = None
+        zoo_addrs = None
+        if is_opendrive:
+            # We currently don't support the Native SUMO Traffic Provider or Social Agents for OpenDRIVE maps
+            pass
```
Gamenot, line 137:
```suggestion
            # We currently don't support the Native SUMO Traffic Provider or Social Agents for OpenDRIVE maps
            if zoo_addrs:
                warnings.warn("`zoo_addrs` should not be used with OpenDRIVE scenarios")
                zoo_addrs = None
            pass
```

smarts/env/hiway_env.py
RutvikGupta: oh sorry, about that! But i have included the above change.

smarts/env/hiway_env.py
```diff
@@ -173,7 +210,7 @@ def scenario_log(self):
             "mission_hash": str(hash(frozenset(scenario.missions.items()))),
         }
 
-    def seed(self, seed: int) -> int:
+    def seed(self, seed=None) -> int:
```
Gamenot, line 213: What is the reason for this change?

RutvikGupta: It didn't match the `OpenAI` gym's `seed` method header and was giving me a warning in my IDE. But i will revert it back for now.

smarts/env/hiway_env.py
```diff
@@ -129,22 +129,62 @@ def __init__(
...
+                )
+
+        try:
+            assert (num_sumo == 0 and num_opendrive == len(scenarios)) or (
+                num_sumo == len(scenarios) and num_opendrive == 0
+            )
+        except AssertionError:
+            raise AssertionError(
+                "All scenarios passed as a parameters for the gym environment need to be of the same version (OpenDRIVE or SUMO)."
+            )
```
sah-huawei, line 184: I guess this may be moot if you make changes like I suggested above, but here I'm wondering, why not just:
```python
if (num_sumo > 0 and num_opendrive > 0) or (num_sumo + num_opendrive != len(scenarios)):
    raise AssertionError( "All scenarios passed as a parameters for the gym environment need to be of the same version (OpenDRIVE or SUMO).")
```

smarts/env/hiway_env.py
```diff
@@ -129,22 +129,62 @@ def __init__(
...
+        scenario_list = Scenario.get_scenario_list(scenarios)
+        for scenario_root in scenario_list:
+            if os.path.exists(os.path.join(scenario_root, "map.net.xml")):
+                num_sumo += 1
+            elif os.path.exists(os.path.join(scenario_root, "map.xodr")):
+                num_opendrive += 1
+            else:
+                raise FileNotFoundError(
+                    f"Unable to find map in map_source={scenario_root}."
+                )
```
sah-huawei, line 175: Note that our dsl allows for `map_spec` to specify _any_ file name for the map source in a scenario.  The point there is to support _custom_ map types that are extensions to the `RoadMap` class, not _just_ Sumo and OpenDrive.  Users can use their own custom map builder functions.

This loop should be rewritten to reflect that.  Instead of relying on filenames, the map builder fn should be actually be used to construct the map object and then its type should be checked.  I suggest changing `_build_map()` to `build_map()` in `scenario.py` for this purpose.

Also, since the only type of map for which we currently support traffic sim is Sumo, that's the only type we really care about here.  That is, we can only support traffic if all of the maps are `SumoRoadNetworks`, but it doesn't really matter (and we don't need to count) what type they are instead.

sah-huawei (changes requested):

smarts/env/hiway_env.py
```diff
@@ -173,7 +213,7 @@ def scenario_log(self):
             "mission_hash": str(hash(frozenset(scenario.missions.items()))),
         }
 
-    def seed(self, seed: int) -> int:
+    def seed(self, seed: int = None) -> int:
```
Adaickalavan, line 216: Why is the default value `None` ? Seed value cannot be `None`.

RutvikGupta: I added that since otherwise it was not following the  OpenAI gym's seed definition. But i will drop it for now.

smarts/env/hiway_env.py
RutvikGupta: Done!

Gamenot (approved):
I have no further comments :ship:

PR #1216, opened 2021-12-29
type hint fixes

sah-huawei:
After seeing Issue #1207 and PR #1208, I was curious about the `pytype` tool, so I just installed it and tried it out.

I was not intending to do a thorough "audit" of our type hint use at this time (although I think I heard that Adai might be doing that already) -- I only tried it on `smarts/core/smarts.py`.  This PR just contains fixes for some of the "easier" problems that `pytype` reported when I did this (including a couple return-type inconsistencies).

I figured I might as well do a PR for these now, notwithstanding that audit that might be currently underway, just in case it isn't yet.

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -23,7 +23,7 @@ import math
 import os
 import warnings
 from collections import defaultdict
-from typing import List, Optional, Set
+from typing import Iterable, List, Optional
 
 import numpy as np
 
@@ -652,7 +652,7 @@ class SMARTS:
     def version(self) -> str:
         return VERSION
 
-    def teardown_agents_without_vehicles(self, agent_ids: Set[str]):
+    def teardown_agents_without_vehicles(self, agent_ids: Iterable[str]):
         """
         Teardown agents in the given list that have no vehicles registered as
         controlled-by or shadowed-by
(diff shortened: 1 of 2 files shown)


RutvikGupta (approved):
For how minor the changes are this should be good to go. But I think if Adai is going to continue working on this in his free time, we should not be creating a separate PR for every file fixed by `pytype` lol.

sah-huawei:
> For how minor the changes are this should be good to go. But I think if Adai is going to continue working on this in his free time, we should not be creating a separate PR for every file fixed by `pytype` lol.

Yep, I completely agree!  (I had to stop myself from doing more files as it's kind of "fun" -- lol -- to do these fixes.  But yeah, I don't think more than one of us should be working on them at a time.)

In fact, I'm ok with not merging this PR if it's already subsumed by whatever Adai is doing.  I just made the push in case it wasn't yet.

Adaickalavan:
> For how minor the changes are this should be good to go. But I think if Adai is going to continue working on this in his free time, we should not be creating a separate PR for every file fixed by `pytype` lol.

Given there are numerous type errors, I would not be able to fix all of them on my own. Everyone is encouraged to check their types using pytype for all new code or new pull requests. I guess, we will eventually include a pytype test in the CI workflow. I believe our code structure and maintenance improves with proper typing.

Adaickalavan (approved):

Issue #1214, opened 2021-12-29
The stationary vehicle keeps rotating


JamesZhutheThird:
**High Level Description**
The envision platform seems to use the direction of velocity as the heading of a vehicle. And when the vehicle is stationary, it may look like rotating. Wondering if that's the problem of the dataset (NGSIM) or the platform😵‍💫

![Envision — Firefox Nightly 2021-12-29 16-17-06~1](https://user-images.githubusercontent.com/35760593/147642387-cb13457c-490b-494a-b28a-b1492de9eb19.gif)

sah-huawei:
Hello @JamesZhutheThird , thanks for the report.

Yes, we are aware of this issue.  The problem is due to the way headings are inferred from the NGSIM dataset by our `genhistories` script -- i.e., the problem is not in the NGSIM dataset itself (NGSIM does not include headings).  

We will try to improve on this in future versions.  In a past version, our inference algorithm produced smoother, more natural results, but this meant that the position at time `t+1` was no longer predictable using basic kinematics from the position and velocity at time `t`, so we reverted that in order to support a specific type of model training.  In the future, we may add an option to allow users such as yourself to control how much smoothing is done when inferring headings.

Gamenot:
Hi @sah-huawei , I am wondering why we cannot generate the orientation inference and then keep it separate from the position and velocity of the current `t`.

sah-huawei:
> Hi @sah-huawei , I am wondering if we should generate the orientation inference and then keep it separate from the position and velocity of the current `t`.

Well that's kind of what we do already.  That is, while position over two successive time-steps is used in order to infer heading (orientation), the computed headings have no impact on the positions we ultimately import.

(Note that at the request of zbzhu99 last summer, we currently also recompute velocities based on position deltas, as opposed to importing the _instantaneous_ velocities in the dataset, in order to ensure that the speed and heading at each time-step `t` will precisely predict the position at time `t+1`.   **Put another way, we are taking the _positions_ in the NGSIM dataset as accurate / definitive.**)

[...]

PR #1219, opened 2022-01-03
Smooth headings

sah-huawei:
This attempts to address issue #1214 by adding three different options to the dataset spec yaml file:
1. `max_angular_velocity` - if specified, near-instantaneous heading changes can be avoided;
2. `heading_inference_window` -  heading inference is calculated using a rolling window (moving average) across this many position changes;
3. `heading_inference_min_speed` - a speed threshold below which headings are assumed not to change.

Each of these addresses noise and imprecision issues in the NGSIM dataset in different ways, while allowing for the request that triggered PR #896 to be honored as well.

For example, I added it to my NGSIM yaml like this:
```yaml
trajectory_dataset:                                                                                                                                             
    name: i80_0400-0415
    source: NGSIM
    input_path: ./original/xy-trajectories/i80/trajectories-0400-0415.txt
 
    # 55mph is roughly 25m/s.  100km/h is roughly 28m/s.
    speed_limit_mps: 28
 
    flip_y: True
    swap_xy: True
 
    x_margin_px: 60
 
    # When inferring headings from positions, each vehicle's angular velocity
    # will be limited to be at most this amount (in rad/sec) to prevent lateral-coordinate
    # noise in the dataset from causing near-instantaneous heading changes.
    max_angular_velocity: 4

    # When inferring headings from positions, a sliding window (moving average)
    # of this size will be used to smooth inferred headings and reduce their dependency
    # on any individual position changes.  Defaults to 2 if not specified.
    heading_inference_window: 5 

    # Speed threshold below which a vehicle's heading is assumed not to change.
    # This is useful to prevent abnormal heading changes that may arise from 
    # noise in position estimates in a trajectory dataset dominating real position
    # changes in situations where the real position changes are very small. 
    # Defaults to .22 if not specified.
    heading_inference_min_speed:  .22

    map_net:
        # These are dimensions of the Sumo network in map.net.xml.
        # We map the coordinates in the input dataset onto this space.
        # max_y is required since flip_y is True.
        max_y: 25.02
 
        # The map width is used to filter position_x values
        # to ensure they're in range.
        width: 310.92
```

I'll update the Issue with instructions (and caveats) once this PR has been approved.

### DIFF
--- a/smarts/sstudio/genhistories.py
+++ b/smarts/sstudio/genhistories.py
@@ -35,6 +35,8 @@ import yaml
 from numpy.lib.stride_tricks import as_strided as stride
 from numpy.lib.stride_tricks import sliding_window_view
 
+from smarts.core.utils.math import vec_to_radians
+
 try:
     from waymo_open_dataset.protos import scenario_pb2
 except ImportError:
@@ -221,7 +223,7 @@ class Interaction(_TrajectoryDataset):
         assert not self._flip_y
         self._max_angular_velocity = dataset_spec.get("max_angular_velocity", None)
         self._heading_min_speed = dataset_spec.get("heading_inference_min_speed", 0.22)
-        self._prev_heading = 3 * math.pi / 2
+        self._prev_heading = None
         self._next_row = None
         # See: https://interaction-dataset.com/details-and-format
         # position and length/width are in meters.
@@ -294,9 +296,8 @@ class Interaction(_TrajectoryDataset):
                 dy = float(self._next_row["y"]) - float(row["y"])
                 dm = np.linalg.norm((dx, dy))
                 if dm != 0.0 and dm > self._heading_min_speed:
-                    r = math.atan2(dy / dm, dx / dm)
-                    new_heading = (r - math.pi / 2) % (2 * math.pi)
-                    if self._max_angular_velocity:
+                    new_heading = vec_to_radians((dx, dy))
+                    if self._max_angular_velocity and self._prev_heading is not None:
                         # XXX: could try to divide by sim_time delta here instead of assuming .1s
                         angular_velocity = (new_heading - self._prev_heading) / 0.1
                         if abs(angular_velocity) > self._max_angular_velocity:
@@ -328,11 +329,11 @@ class NGSIM(_TrajectoryDataset):
         super().check_dataset_spec(dataset_spec)
         hiw = dataset_spec.get("heading_inference_window", 2)
         # 11 is a semi-arbitrary max just to keep things "sane".
-        if hiw < 2 or hiw > 11:
+        if not 2 <= hiw <= 11:
             raise ValueError("heading_inference_window must be between 2 and 11")
 
-    def _cal_heading(self, window) -> float:
-        window = window[0]
+    def _cal_heading(self, window_param) -> float:
+        window = window_param[0]
         new_heading = 0
         prev_heading = None
         den = 0
@@ -355,8 +356,7 @@ class NGSIM(_TrajectoryDataset):
                     den += 1
                 continue
             vhat = (n - c) / s
-            r = math.atan2(vhat[1], vhat[0])
-            inst_heading = (r - math.pi / 2) % (2 * math.pi)
+            inst_heading = vec_to_radians(vhat)
             if prev_heading is not None:
                 if inst_heading == 0 and prev_heading > math.pi:
                     inst_heading = 2 * math.pi
@@ -508,6 +508,16 @@ class OldJSON(_TrajectoryDataset):
     """This exists because SMARTS used to use JSON files for traffic histories.
     We provide this to help people convert these previously-created .json
     history files to the new .shf format."""
+
+    def __init__(self, dataset_spec: Dict[str, Any], output: str):
+        from warnings import warn
+
+        warn(
+            f"The {self.__class__.__name__} class has been deprecated.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        super().__init__(dataset_spec, output)
 
     @property
     def rows(self) -> Generator[Tuple, None, None]:
--- a/smarts/sstudio/genhistories.py
+++ b/smarts/sstudio/genhistories.py
@@ -33,6 +33,7 @@ import numpy as np
 import pandas as pd
 import yaml
 from numpy.lib.stride_tricks import as_strided as stride
+from numpy.lib.stride_tricks import sliding_window_view
 
 try:
     from waymo_open_dataset.protos import scenario_pb2
@@ -219,7 +220,8 @@ class Interaction(_TrajectoryDataset):
         super().__init__(dataset_spec, output)
         assert not self._flip_y
         self._max_angular_velocity = dataset_spec.get("max_angular_velocity", None)
-        self._prev_heading = -math.pi / 2
+        self._heading_min_speed = dataset_spec.get("heading_inference_min_speed", 0.22)
+        self._prev_heading = 3 * math.pi / 2
         self._next_row = None
         # See: https://interaction-dataset.com/details-and-format
         # position and length/width are in meters.
@@ -230,6 +232,14 @@ class Interaction(_TrajectoryDataset):
             "position_x": "y" if self._swap_xy else "x",
             "position_y": "x" if self._swap_xy else "y",
         }
+
+    def check_dataset_spec(self, dataset_spec: Dict[str, Any]):
+        super().check_dataset_spec(dataset_spec)
+        hiw = dataset_spec.get("heading_inference_window", 2)
+        if hiw != 2:
+            raise ValueError(
+                "heading_inference_window not yet supported for Interaction datasets."
+            )
 
     @property
     def rows(self) -> Generator[Dict, None, None]:
@@ -280,7 +290,9 @@ class Interaction(_TrajectoryDataset):
                 dx = float(self._next_row["x"]) - float(row["x"])
                 dy = float(self._next_row["y"]) - float(row["y"])
                 dm = np.linalg.norm((dx, dy))
-                if dm > 0.0:
+                if dm != 0.0 and (
+                    self._heading_min_speed is None or dm > self._heading_min_speed
+                ):
                     r = math.atan2(dy / dm, dx / dm)
                     new_heading = (r - math.pi / 2) % (2 * math.pi)
                     if self._max_angular_velocity:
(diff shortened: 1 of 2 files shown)


smarts/sstudio/genhistories.py
```diff
@@ -301,10 +318,19 @@ def _cal_heading(self, window):
...
         vhat = (n - c) / s
         r = math.atan2(vhat[1], vhat[0])
-        self._prev_heading = (r - math.pi / 2) % (2 * math.pi)
+        new_heading = (r - math.pi / 2) % (2 * math.pi)
+        if self._max_angular_velocity:
+            # XXX: could try to divide by sim_time delta here instead of assuming .1s
+            angular_velocity = (new_heading - self._prev_heading) / 0.1
+            if abs(angular_velocity) > self._max_angular_velocity:
+                new_heading = (
+                    self._prev_heading
```
RutvikGupta, line 327: Maybe factor this out to a method to avoid code duplication.

smarts/sstudio/genhistories.py
```diff
@@ -301,10 +318,19 @@ def _cal_heading(self, window):
...
+                    + np.sign(angular_velocity) * self._max_angular_velocity * 0.1
+                )
+        self._prev_heading = new_heading
```
RutvikGupta, line 330: Maybe factor this out to a method to avoid code duplication.

smarts/sstudio/genhistories.py
```diff
@@ -279,19 +282,33 @@ def column_val_in_row(self, row, col_name: str):
...
+                    if self._max_angular_velocity:
+                        # XXX: could try to divide by sim_time delta here instead of assuming .1s
+                        angular_velocity = (new_heading - self._prev_heading) / 0.1
+                        if abs(angular_velocity) > self._max_angular_velocity:
+                            new_heading = (
+                                self._prev_heading
+                                + np.sign(angular_velocity)
+                                * self._max_angular_velocity
+                                * 0.1
+                            )
```
RutvikGupta, line 295: `if abs(angular_velocity) > self._max_angular_velocity`, shouldn't the `self._max_angular_velocity` be updated to the new `angular_velocity` ?

smarts/sstudio/genhistories.py
sah-huawei: Good point, will do!

smarts/sstudio/genhistories.py
sah-huawei: No, b/c this is a parameter set by the user that should remain fixed.  I.e., we're not trying to _find_ the maximum here, rather, this is the maximum _allowable_.  This is a way of filtering out "noise" in the dataset positions.

RutvikGupta: I see. Makes sense!

Gamenot (approved):

RutvikGupta (approved):

sah-huawei:
Note: I've updated the description of the PR to describe the new features I've added.

smarts/sstudio/genhistories.py
```diff
@@ -229,8 +233,16 @@ def __init__(self, dataset_spec, output):
             "position_y": "x" if self._swap_xy else "y",
         }
 
+    def check_dataset_spec(self, dataset_spec: Dict[str, Any]):
+        super().check_dataset_spec(dataset_spec)
+        hiw = dataset_spec.get("heading_inference_window", 2)
+        if hiw != 2:
+            raise ValueError(
+                "heading_inference_window not yet supported for Interaction datasets."
+            )
```
Gamenot, line 242: Is this to reserve this argument?

sah-huawei: The reason I put this there is because headings are inferred for both the NGSIM and Interaction, so one might assume that the 3 heading-inference-related YAML options that I added here would work for both of these. 

However, since (a) it's unclear if the nature of the noise in Interaction is the same as in NGSIM and (b) this particular option (`heading_inference_window`) would require non-trivial changes to the way the `Interaction` subclass works, I instead added this exception so that if someone tries to add `heading_inference_window` to their yaml file for an Interaction dataset (like they have in, say, their NGSIM yaml files) they can be made aware that "yes, that option _could_ be added here in the future if you think you need it, but we haven't done it yet."

The reason 2 is the default value there is because that's basically the "no window" case.

sah-huawei: Just added a comment about this.

smarts/sstudio/genhistories.py
```diff
@@ -178,10 +179,10 @@ def create_output(self, time_precision=3):
...
                     time_precision,
                 ),
-                float(self.column_val_in_row(row, "position_x") + x_offset)
-                * self.scale,
-                float(self.column_val_in_row(row, "position_y") + y_offset)
-                * self.scale,
+                float(self.column_val_in_row(row, "position_x"))
+                + x_offset * self.scale,
+                float(self.column_val_in_row(row, "position_y"))
+                + y_offset * self.scale,
```
RutvikGupta, line 185: Why was there a need to do this?

smarts/sstudio/genhistories.py
```diff
@@ -229,8 +233,19 @@ def __init__(self, dataset_spec, output):
             "position_y": "x" if self._swap_xy else "y",
         }
 
+    def check_dataset_spec(self, dataset_spec: Dict[str, Any]):
+        super().check_dataset_spec(dataset_spec)
+        hiw = dataset_spec.get("heading_inference_window", 2)
+        if hiw != 2:
+            # Adding support fot this would require changing the rows() generator
```
RutvikGupta, line 240: T
```suggestion
            # Adding support for this would require changing the rows() generator
```

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +293,96 @@ def column_val_in_row(self, row, col_name: str):
                 dx = float(self._next_row["x"]) - float(row["x"])
                 dy = float(self._next_row["y"]) - float(row["y"])
                 dm = np.linalg.norm((dx, dy))
-                if dm > 0.0:
+                if dm != 0.0 and (
+                    self._heading_min_speed is None or dm > self._heading_min_speed
```
RutvikGupta, line 297: How can `self._heading_min_speed` be `None` if its default value is 0.22?

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +293,96 @@ def column_val_in_row(self, row, col_name: str):
...
-            return self._prev_heading
-        vhat = (n - c) / s
-        r = math.atan2(vhat[1], vhat[0])
-        self._prev_heading = (r - math.pi / 2) % (2 * math.pi)
-        return self._prev_heading
+        self._prev_heading = 3 * math.pi / 2
+        self._max_angular_velocity = dataset_spec.get("max_angular_velocity", None)
+        self._heading_window = dataset_spec.get("heading_inference_window", 2)
+        # .22 corresponds to roughly 5mph.
+        self._heading_min_speed = dataset_spec.get("heading_inference_min_speed", 0.22)
```
RutvikGupta, line 327: Where these default values computed through Trial and Error?

smarts/sstudio/genhistories.py
```diff
@@ -229,8 +233,19 @@ def __init__(self, dataset_spec, output):
...
+    def check_dataset_spec(self, dataset_spec: Dict[str, Any]):
+        super().check_dataset_spec(dataset_spec)
+        hiw = dataset_spec.get("heading_inference_window", 2)
+        if hiw != 2:
+            # Adding support fot this would require changing the rows() generator
+            # (since we're not using Pandas here like we are for NGSIM).
+            # So wait until if/when users request it...
+            raise ValueError(
+                "heading_inference_window not yet supported for Interaction datasets."
+            )
```
RutvikGupta, line 245: Do we want to raise an Error here? Shouldn't we just warn the user this is not supported for now?

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +293,96 @@ def column_val_in_row(self, row, col_name: str):
...
+        self._heading_window = dataset_spec.get("heading_inference_window", 2)
+        # .22 corresponds to roughly 5mph.
+        self._heading_min_speed = dataset_spec.get("heading_inference_min_speed", 0.22)
+
+    def check_dataset_spec(self, dataset_spec: Dict[str, Any]):
+        super().check_dataset_spec(dataset_spec)
+        hiw = dataset_spec.get("heading_inference_window", 2)
+        # 11 is a semi-arbitrary max just to keep things "sane".
+        if hiw < 2 or hiw > 11:
+            raise ValueError("heading_inference_window must be between 2 and 11")
```
RutvikGupta, line 334: Should we default to 2, if the user passes value outside this range instead?

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +293,96 @@ def column_val_in_row(self, row, col_name: str):
...
+                    if abs(angular_velocity) > self._max_angular_velocity:
+                        inst_heading = (
+                            prev_heading
+                            + np.sign(angular_velocity)
+                            * self._max_angular_velocity
+                            * 0.1
+                        )
+            den += 1
+            new_heading += inst_heading
+            prev_heading = inst_heading
```
RutvikGupta, line 377: Shouldn't `prev_heading` be equal to old value of `new_heading`  here?

RutvikGupta (commented):
I marked some doubts but overall looks great! Especially the Sliding Window Algorithm :)

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +290,95 @@ def column_val_in_row(self, row, col_name: str):
...
+        self._max_angular_velocity = dataset_spec.get("max_angular_velocity", None)
+        self._heading_window = dataset_spec.get("heading_inference_window", 2)
+        # .22 corresponds to roughly 5mph.
+        self._heading_min_speed = dataset_spec.get("heading_inference_min_speed", 0.22)
+
+    def check_dataset_spec(self, dataset_spec: Dict[str, Any]):
+        super().check_dataset_spec(dataset_spec)
+        hiw = dataset_spec.get("heading_inference_window", 2)
+        if hiw < 2 or hiw > 11:
+            raise ValueError("heading_inference_window must be between 2 and 11")
```
Gamenot, line 332: The `NGSIM` dataset seems to have a different requirement on this compared to the interaction dataset.

smarts/sstudio/genhistories.py
sah-huawei: > How can `self._heading_min_speed` be `None` if its default value is 0.22?

Good point.  I originally hadn't given it a default value, but changed my mind when I realized they could just set it to `-1` if they didn't want to use it.  I'll update this...

smarts/sstudio/genhistories.py
sah-huawei: Yes.  

fyi, in the old version from last spring, I'd just hardcoded this (.22) as a constant, but now I want to allow them to set it because it will depend on the dataset noise.

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +290,95 @@ def column_val_in_row(self, row, col_name: str):
...
+
+    def _cal_heading(self, window) -> float:
+        window = window[0]
```
Gamenot, line 335: I would suggest assigning to a different variable or defining the type of the `window` argument if the variable needs to be replaced in this way. I think `_cal_heading(..)` is unclear about what it needs.

smarts/sstudio/genhistories.py
sah-huawei: Good catch!!
A change was required there for when `self.column_val_in_row()` returns a string (which happens for Interaction datasets).  This can't be "added" to `x_offset` or `y_offset` until it is cast to a float first.  However, I screwed up and didn't keep the outer parentheses by accident.

I'm glad you caught this!!

smarts/sstudio/genhistories.py
sah-huawei: No because this one (unlike `self._prev_heading`) is used to keep track of the previous value within the window (i.e.,  at just `w`), meanwhile `new_heading` is _accumulating_ _all_ values through the window (in order to compute the average after the loop).

smarts/sstudio/genhistories.py
sah-huawei: Maybe?  Only if we don't consider it an error (i.e., don't raise) in the Interaction dataset like you suggested in your other comment.  I'm considering this for a few minutes...

sah-huawei: I think I'm going to keep it as a fatal error for now.  In general, if the user explicitly asks for something that we can't do, I think it's important not to let them think it worked.  I realize that they _should_ notice a warning message, but I have too many times seen people ignore these or fail to see them b/c they're dumping the output to a log file.

smarts/sstudio/genhistories.py
sah-huawei: I think I'm going to keep it a fatal error for now (for the reason I wrote in your other comment).

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +290,95 @@ def column_val_in_row(self, row, col_name: str):
...
+                self._heading_min_speed is not None
+                and (s < self._heading_min_speed or ispeed < self._heading_min_speed)
+            ):
+                if prev_heading is not None:
+                    new_heading += prev_heading
+                    den += 1
+                continue
+            vhat = (n - c) / s
+            r = math.atan2(vhat[1], vhat[0])
+            inst_heading = (r - math.pi / 2) % (2 * math.pi)
```
Gamenot, line 359: I believe our `math` should have this conversion method but it does not. (I believe it is being added in the OpenDRIVE pr)

smarts/sstudio/genhistories.py
```diff
@@ -277,34 +290,95 @@ def column_val_in_row(self, row, col_name: str):
...
+            if prev_heading is not None:
+                if inst_heading == 0 and prev_heading > math.pi:
+                    inst_heading = 2 * math.pi
```
Gamenot, line 362: Is this intended to constrain the heading in some way? I cannot quite grasp how it interacts with the averaging for `new_heading`.

smarts/sstudio/genhistories.py
```diff
@@ -254,7 +267,7 @@ def _lookup_agent_type(self, agent_type: str) -> int:
         self._log.warning(f"unknown agent_type:  {agent_type}.")
         return 0
 
-    def column_val_in_row(self, row, col_name: str):
+    def column_val_in_row(self, row, col_name: str) -> Any:
```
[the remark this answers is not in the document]
sah-huawei: Although not quite.  The first return in this function (around line 276) is one that could be something other than a `float`.  For example, in the bug related to the `float()` cast and parentheses that Rutvik flagged, the issue that I was trying to address there (incorrectly, originally) was that this sometimes returns a string.

smarts/sstudio/genhistories.py
```diff
@@ -215,9 +216,12 @@ def create_output(self, time_precision=3):
 
 
 class Interaction(_TrajectoryDataset):
-    def __init__(self, dataset_spec, output):
+    def __init__(self, dataset_spec: Dict[str, Any], output: str):
         super().__init__(dataset_spec, output)
         assert not self._flip_y
+        self._max_angular_velocity = dataset_spec.get("max_angular_velocity", None)
+        self._heading_min_speed = dataset_spec.get("heading_inference_min_speed", 0.22)
+        self._prev_heading = 3 * math.pi / 2
```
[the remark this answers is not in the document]
sah-huawei: Ah, very good point!  Yes, it could. 
I'll try to do something better here...

smarts/sstudio/genhistories.py
sah-huawei: Yes, `heading_inference_window` is only supported for NGSIM right now, but I put the place-holder into the Interaction class as well for the reasons I wrote about in one of your other comments.

For Interaction datasets, it must always be 2 (which is basically no window) at this point.

smarts/sstudio/genhistories.py
sah-huawei: agreed, will fix.

smarts/sstudio/genhistories.py
```diff
@@ -430,7 +510,7 @@ class OldJSON(_TrajectoryDataset):
     history files to the new .shf format."""
 
     @property
-    def rows(self):
+    def rows(self) -> Generator[Tuple, None, None]:
```
Gamenot, line 513: I think we will want to deprecate `OldJSON` unless we are supporting migrations. It is additional work to keep servicing it.

smarts/sstudio/genhistories.py
sah-huawei: This was a nasty thing I discovered in my debugging related to averaging around a circle.

Imagine you want to average 2 headings, one of which might either be 0 or `2*pi`.  If you use 0, then it will bring the average down, but if you use `2*pi` it will bring the average up.  Which do you use?  

The heuristic here is to use the one that's the smallest angle away from the other heading(s) you're averaging with (under the assumption that this is more likely).  So if the other heading was, say `5*pi/3`, then it's likely that you want to use `2*pi` to get  `11*pi/12` for an average, rather than using `0` to get `5*pi/6`.

smarts/sstudio/genhistories.py
sah-huawei: Good idea.  I'll add a comment to that now.

smarts/sstudio/genhistories.py
sah-huawei: Yeah, actually I think I can just use `vec_to_radians()`.  It probably should have been that all along!

Issue #1206, opened 2021-12-23
Add `pre-commit`


Gamenot:
**Is your feature request related to a problem? Please describe.**
Manual steps can lead to mistakes which increase review time.

**Describe the solution you'd like**
Add in the `pre-commit` `git hook` to handle the manual linting, formatting, and basic tests stages of pushing a commit.

**Describe alternatives you've considered**
Automate formatting step on the CI side.

**Additional context**
See https://pre-commit.com/ for details.

PR #1223, opened 2022-01-06
Add pre commit hooks for formatting

liu-allan:
Made to address #1206, black and isort are run on the staged files

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,11 +5,12 @@ repos:
         name: black
         entry: black
         language: system
-        exclude_types: [yaml]
+        types: [python]
     -   id: isort
         name: isort
         entry: isort
         language: system
+        types: [python]
         args: ["--profile", "black"]
     -   id: pylint
         name: pylint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,8 +8,7 @@ repos:
     rev: 5.5.4
     hooks:
     -   id: isort
-        args:
-          - --profile black
+        args: ["--profile", "black"]
 
 -   repo: local
     hooks:
@@ -20,6 +19,7 @@ repos:
         types: [python]
         args:
             [
+            "--errors-only",
             "-rn", # Only display messages
             "-sn", # Don't display the score
             ]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,3 @@
-# repos:
-# -   repo: https://github.com/psf/black
-#     rev: 20.8b1
-#     hooks:
-#     -   id: black
-
-# -   repo: https://github.com/pycqa/isort
-#     rev: 5.5.4
-#     hooks:
-#     -   id: isort
-#         args: ["--profile", "black"]
-
 repos:
 -   repo: local
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,17 +1,28 @@
+# repos:
+# -   repo: https://github.com/psf/black
+#     rev: 20.8b1
+#     hooks:
+#     -   id: black
+
+# -   repo: https://github.com/pycqa/isort
+#     rev: 5.5.4
+#     hooks:
+#     -   id: isort
+#         args: ["--profile", "black"]
+
 repos:
--   repo: https://github.com/psf/black
-    rev: 20.8b1
-    hooks:
-    -   id: black
-
--   repo: https://github.com/pycqa/isort
-    rev: 5.5.4
-    hooks:
-    -   id: isort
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -111,6 +111,7 @@ pexpect==4.8.0
 pickleshare==0.7.5
 Pillow==8.2.0
 pluggy==1.0.0
+pre-commit==2.16.0
```
Adaickalavan, line 114: The `requirements.txt` file will be overwritten and updated automatically, on commit to the develop branch, by `ci-dependencies-linux.yml` in the CI.

Currently, we only include [test,train,camera-obs] options from `setup.py` inside `requirements.txt` file.

If needed, perhaps `pre-commit==2.16.0` and `pylint==2.12.2` can be included in the [dev] option in `setup.py` file.

liu-allan: I have made this change, is there anything else?

setup.py
```diff
@@ -80,6 +81,8 @@
             "black==20.8b1",
             "grpcio-tools==1.32.0",
             "isort==5.7.0",
+            "pre-commit==2.16.0",
+            "pylint==2.12.2",
```
Gamenot, line 85: Is there some description here of what we would need to do to get `pre-commit` set up?

Gamenot (changes requested):

setup.py
liu-allan: user just has to run `pre-commit install`, where should i include this instruction?

Gamenot: I think add instructions for setup in in the `CONTRIBUTING.md`.

Gamenot: You will need to take a read-through and update the file.

CONTRIBUTING.md
```diff
@@ -18,10 +18,11 @@ We encourage all forms of contributions to SMARTS, not limited to:
 
 ## Setting up your dev environment
 
-In addition to following the setup steps in the README, you'll need to install the [dev] dependencies.
+In addition to following the setup steps in the README, you'll need to install the [dev] dependencies and install the pre-commit hooks.
 
 ```bash
 pip install -e .[dev]
+pre-commit install
```
Gamenot, line 25: This should not be needed because it will be installed with `[dev]`.

liu-allan: ah with that i meant the `pre-commit install` command

CONTRIBUTING.md
```diff
@@ -90,13 +90,9 @@ Tag the issue as a feature request using `enhancement` and if it takes more than
 
 ### Python(Format)
 
-1. Always run `make format` before committing.
-
```
Gamenot, line 94: nit: Perhaps reference the need to have `pre-commit` set up for the repository.

CONTRIBUTING.md
```diff
@@ -90,13 +90,9 @@ Tag the issue as a feature request using `enhancement` and if it takes more than
...
 The project follows a strict format requirement for python code. We made a decision early on in the project to use [Black](https://github.com/psf/black). This makes formatting consistent while eliminating [bike shedding](http://bikeshed.com/).
 If you do not already have it please install it via `pip install black`.
 
-Formatting guarantees that your code will pass the CI formatting test case.
-
```
Gamenot, line 99: I do not think these lines need to be deleted.

CONTRIBUTING.md
liu-allan: it is, this is to setup our pre-commit hooks for use not the actual install

Gamenot: Oh, sorry, yes!

Gamenot (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
+repos:
+-   repo: https://github.com/psf/black
+    rev: 19.3b0
```
Adaickalavan, line 3: We are using `black==20.8b1` in `setup.py`

https://github.com/huawei-noah/SMARTS/blob/ce1061bba9e7e9f00510443b2ce09a07d9f6ed64/setup.py#L81

Should we use the same black version here?

Adaickalavan (approved):

CONTRIBUTING.md
```diff
@@ -42,8 +43,8 @@ Please take care in using good commit messages as they're useful for debugging,
 
 ## Pre-Push Checklist
 
-1. Do your best to see that your code compiles locally.
-2. Run `make format`. See [Formatting](#Formatting).
+1. Make sure the pre-commit hooks are installed `pre-commit install`
```
Adaickalavan, line 46:
```suggestion
1. Make sure the pre-commit hooks are installed via `pre-commit install`.
```

Gamenot:
It is not good that we are seeing determinism issues.

Issue #1228, opened 2022-01-07
Remove `route_waypoints` from `RoadWaypoints`


Gamenot:
**Is your feature request related to a problem? Please describe.**
Road waypoints should not have any relation to the route of a vehicle, just waypoints in proximity to a vehicle.

**Describe the solution you'd like**
Remove the `route_waypoints` part of the road waypoints observation.

**Additional context**
This functionality seems to have been added to support a "`Task`" feature we already removed.

PR #1229, opened 2022-01-07
Remove route_waypoints from RoadWaypoints

Adaickalavan:
+ Removes `route_waypoints` attribute from `observation.RoadWaypoints`
+ Addresses #1228 .

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -74,7 +74,6 @@ class EgoVehicleObservation(NamedTuple):
 
 class RoadWaypoints(NamedTuple):
     lanes: Dict[str, List[List[Waypoint]]]
-    route_waypoints: List[List[Waypoint]]
 
 
 class GridMapMetadata(NamedTuple):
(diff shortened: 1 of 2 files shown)


Gamenot (approved):

RutvikGupta (approved):

smarts/core/sensors.py
```diff
@@ -901,11 +900,11 @@ def __init__(self, vehicle, sim, plan, horizon=32):
         self._plan = plan
         self._horizon = horizon
 
-    def __call__(self):
+    def __call__(self) -> RoadWaypoints:
         veh_pt = self._vehicle.pose.point
         lane = self._road_map.nearest_lane(veh_pt)
         if not lane:
-            return RoadWaypoints(lanes={}, route_waypoints=[])
+            return RoadWaypoints(lanes={})
```
sah-huawei, line 907: Since this affects the interface (and so is a "breaking change"), could you please add a line about this in CHANGELOG.md?

smarts/core/sensors.py
```diff
@@ -901,11 +900,11 @@ def __init__(self, vehicle, sim, plan, horizon=32):
         self._plan = plan
         self._horizon = horizon
 
-    def __call__(self):
+    def __call__(self) -> RoadWaypoints:
         veh_pt = self._vehicle.pose.point
         lane = self._road_map.nearest_lane(veh_pt)
         if not lane:
-            return RoadWaypoints(lanes={}, route_waypoints=[])
```
sah-huawei, line 908: I suspect the reason for the redundancy here was so that users would only ever have to add one type of waypoints sensor, either `RoadWaypointsSensor` or `WaypointsSensor`, but not both.  I.e., I think this was just to simplify things for users.

sah-huawei (approved):
I think this can go in, but since it's a breaking change it either has to go into v0.5 or v0.6 (not for example, v0.5.1).

smarts/core/sensors.py
Adaickalavan: I have added the change into changelog.md.

PR #1234, opened 2022-01-10
map building fixes

sah-huawei:
- Changed so that if more than one qualifying map file exists in a the `map_spec.source` folder, `get_road_map()` in `default_map_builder.py` will prefer to return the default files (`map.net.xml` or `map.xodr`) if they exist.
- Cleaned up map building in scenario studio, including:
    - don't assume map filenames (use the `MapSpec` `builder_fn` instead) and honor custom map building from `scenario.py`
    - correctly handle the `shift_to_origin` parameter, which was broken before

### DIFF
--- a/smarts/core/default_map_builder.py
+++ b/smarts/core/default_map_builder.py
@@ -72,7 +72,7 @@ def _find_mapfile_in_dir(map_dir: str) -> Tuple[int, str]:
         if f.endswith(".net.xml"):
             map_type = _SUMO_MAP
             map_path = os.path.join(map_dir, f)
-        if f.endswith(".xodr"):
+        elif f.endswith(".xodr"):
             map_type = _OPENDRIVE_MAP
             map_path = os.path.join(map_dir, f)
     return map_type, map_path
@@ -110,24 +110,24 @@ def get_road_map(map_spec) -> Tuple[RoadMap, str]:
     if map_type == _SUMO_MAP:
         from smarts.core.sumo_road_network import SumoRoadNetwork
 
-        clazz = SumoRoadNetwork
+        map_class = SumoRoadNetwork
 
     elif map_type == _OPENDRIVE_MAP:
         from smarts.core.opendrive_road_network import OpenDriveRoadNetwork
 
-        clazz = OpenDriveRoadNetwork
+        map_class = OpenDriveRoadNetwork
 
     else:
         return None, None
 
     if _existing_map:
-        if isinstance(_existing_map.obj, clazz) and _existing_map.obj.is_same_map(
+        if isinstance(_existing_map.obj, map_class) and _existing_map.obj.is_same_map(
             map_spec
         ):
             return _existing_map.obj, _existing_map.map_hash
         _clear_cache()
 
-    road_map = clazz.from_spec(map_spec)
+    road_map = map_class.from_spec(map_spec)
     if os.path.isfile(road_map.source):
         road_map_hash = file_md5_hash(road_map.source)
     else:
--- a/smarts/core/default_map_builder.py
+++ b/smarts/core/default_map_builder.py
@@ -23,7 +23,7 @@ from dataclasses import replace
 from typing import NamedTuple, Tuple
 from pathlib import Path
 from smarts.core.road_map import RoadMap
-from smarts.core.utils.file import file_md5_hash
+from smarts.core.utils.file import file_md5_hash, path2hash
 
 
 _existing_map = None
@@ -113,7 +113,7 @@ def get_road_map(map_spec) -> Tuple[RoadMap, str]:
         clazz = SumoRoadNetwork
 
     elif map_type == _OPENDRIVE_MAP:
-        from smarts.core.sumo_road_network import OpenDriveRoadNetwork
+        from smarts.core.opendrive_road_network import OpenDriveRoadNetwork
 
         clazz = OpenDriveRoadNetwork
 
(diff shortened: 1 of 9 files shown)


smarts/core/default_map_builder.py
```diff
@@ -54,63 +97,41 @@ def get_road_map(map_spec) -> Tuple[RoadMap, str]:
...
+        if map_source.endswith(".net.xml"):
+            map_type = _SUMO_MAP
+        elif map_source.endswith(".xodr"):
+            map_type = _OPENDRIVE_MAP
+
+    if map_type == _SUMO_MAP:
         from smarts.core.sumo_road_network import SumoRoadNetwork
 
-        road_map = SumoRoadNetwork.from_spec(map_spec)
+        clazz = SumoRoadNetwork
```
RutvikGupta, line 113: `clazz` lol ? Maybe you can use `map_class`.

RutvikGupta (commented):
I would recommend doing a similar change in `cli/studio.py` where we build the scenarios.

smarts/core/default_map_builder.py
sah-huawei: ok, done.

sah-huawei:
> I would recommend doing a similar change in `cli/studio.py` where we build the scenarios.

OK, good call.  Done.  (Cleanup there was urgently needed.)
Although now this PR is considerably bigger!  ;)
(... and I guess I'd better change the name and description too!)

smarts/sstudio/types.py
```diff
@@ -277,6 +277,9 @@ class MapSpec:
     """If specified, the default distance between pre-generated Lane Points (Waypoints)."""
     default_lane_width: Optional[float] = None
     """If specified, the default width (in meters) of lanes on this map."""
+    shift_to_origin: bool = False
+    """If True, a map whose bounding-box does not intersect with the origin point (0,0)
+    whill be shifted such that it does upon creation."""
```
saulfield, line 282:
```suggestion
    will be shifted such that it does upon creation."""
```

smarts/core/default_map_builder.py
```diff
@@ -23,11 +23,61 @@
...
+
+def _clear_cache():
+    global _existing_map
+    if _existing_map:
+        import gc
+
+        # Try to only keep one map around at a time...
+        del _existing_map
+        _existing_map = None
+        gc.collect
```
saulfield, line 52:
```suggestion
        gc.collect()
```

smarts/core/default_map_builder.py
```diff
@@ -23,11 +23,61 @@
...
+    for f in os.listdir(map_dir):
+        cand_map_type = map_filename_type.get(f)
+        if cand_map_type is not None:
+            return cand_map_type, os.path.join(map_dir, f)
+        if f.endswith(".net.xml"):
+            map_type = _SUMO_MAP
+            map_path = os.path.join(map_dir, f)
+        elif f.endswith(".xodr"):
+            map_type = _OPENDRIVE_MAP
+            map_path = os.path.join(map_dir, f)
```
saulfield, line 77: Very minor, but maybe we want to return here since this could continue looping needlessly if we've already found a map file.

saulfield (changes requested):

smarts/core/default_map_builder.py
sah-huawei: Unfortunately we can't do that b/c we need to make sure we don't have a file like `map.net.xml` later in the directory list.   (If we do, we want to return that one instead.  Actually that was the main point of the original change.)

smarts/core/default_map_builder.py
sah-huawei: good catch.  I typed that in instead of copy/pasting like I should have done.  will fix.

saulfield (approved):

smarts/core/default_map_builder.py
```diff
@@ -23,11 +23,61 @@
...
```
RutvikGupta, line 77: If we have two files ending with `.net.xml`, wouldn't it add to the path creating a invalid path. I can see it happening for a SUMO scenario having the shifted map offset and the original network file.

smarts/core/opendrive_road_network.py
```diff
@@ -226,6 +226,11 @@ def from_spec(
         cls,
         map_spec: MapSpec,
     ):
+        if map_spec.shift_to_origin:
+            logger = logging.getLogger(cls.__name__)
+            logger.warning(
+                "OpenDrive road networks do not yet support the 'shift_to_origin' option."
+            )
```
RutvikGupta, line 233: Unlike `SUMO`, `OpenDRIVE` doesn't require shifting of origin, since offset computation are fairly more accurate in `OpenDRIVE`.

RutvikGupta (approved):
Some minor concerns, but looks good to merge.

smarts/core/default_map_builder.py
sah-huawei: I'm not sure I understand this one correctly, but `map_path` is set "fresh" each time (not concatenated) from `map_dir`, so it doesn't "add to the path".

If there are both `map.net.xml` and `shifted_map-AUTOGEN.net.xml` in the same folder, the first one that is encountered (which should be `map.net.xml` if they are listed alphabetically, but it doesn't matter either way) will be returned on line 71 and the `map_type` will be SUMO.  If there is _also_ some other file in the folder, say `abc.net.xml`, it won't be returned, it will just cause the local `map_path` to be set and then discarded upon line 71.

smarts/core/opendrive_road_network.py
sah-huawei: understood.  But I think the warning is still needed here just in case some user tries to use `shift_to_origin` for some other reason of their own.

PR #1235, opened 2022-01-11
[Bugfix] Secondary TraCI error

Gamenot:
A secondary error has showed up in which calling `traci_conn.close()` will throw an exception because `TraCI` automatically closes and invalidates the underlying connection object upon the initial exception.

### DIFF
--- a/smarts/core/sumo_traffic_simulation.py
+++ b/smarts/core/sumo_traffic_simulation.py
@@ -295,14 +295,24 @@ class SumoTrafficSimulation(Provider):
         return self._compute_provider_state()
 
     def _close_traci_and_pipes(self):
+        """We should expect this method to always work without throwing"""
+
+        def __safe_close(conn):
+            try:
+                conn.close()
+            except:
+                pass
+
         if self._sumo_proc:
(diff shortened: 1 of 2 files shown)


qianyi-sun (approved):

smarts/core/sumo_traffic_simulation.py
```diff
@@ -295,14 +295,24 @@ def setup(self, next_scenario) -> ProviderState:
...
+            except:
+                pass
+
         if self._sumo_proc:
-            self._sumo_proc.stdin.close()
-            self._sumo_proc.stdout.close()
-            self._sumo_proc.stderr.close()
+            __safe_close(self._sumo_proc.stdin)
+            __safe_close(self._sumo_proc.stdout)
+            __safe_close(self._sumo_proc.stderr)
```
RutvikGupta, line 309: why is this being tried 3 times?

Gamenot: There are 3 different streams here that need to be closed: `stdin`, `stdout`, `stderr`.

RutvikGupta (approved):

PR #1242, opened 2022-01-14
Add introduction in readme.md for zsh user

qianyi-sun:
The current `pip install -e .[*]` does not work with zsh console. Instead, quotation marks are necessary outside the `.[*]`. Therefore additional introduction has been added in the readme.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,18 +73,16 @@ source .venv/bin/activate
 # upgrade pip, a recent version of pip is needed for the version of tensorflow we depend on
 pip install --upgrade pip
 
-# for the following `pip install -e .[*]` commands, if you are a zsh user,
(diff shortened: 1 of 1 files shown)


qianyi-sun:
@Gamenot There are basically two options to do this:
1) Add additional intro in the comment for the `pip install -e [*]` commands to let the `zsh` users know that they need quotation marks in the command. (current way)
2) Keep the intro to remain unchanged and change the  `pip install -e [*]` command into the quotation mark version, since with or without them, the command works as usual in regular bash console.

Gamenot:
I think let's just use the quoted version since it covers a broader base. Then we can mention the escape '\' needed to install using a `fish` shell.

sah-huawei (approved):
Yes, glad you fixed this.  I hit this when I first joined too!

Gamenot (approved):

Issue #1192, opened 2021-12-21
Conform to Gym Spaces at Environment Level


Gamenot:
Gym environments should provide the following:

AC:
- [x] The exact `gym.Space` for the observations it returns (`observation_space`)
- [x] The exact `gym.Space` for the actions it accepts (`action_space`)
- [x] Environment `action_space`
- [ ] A way to isolate per-agent `action_space`s

Adaickalavan:
1. A SMARTS environment will have gym-compatible action space attribute when wrapped with the `FormatAction` env wrapper from `SMARTS/smarts/env/wrappers/format_action.py`.

2. What is needed to statisfy the requirement _"A way to isolate per-agent action_spaces"_ ?

Adaickalavan:
Related pull requests:

Adaickalavan:
1. SMARTS does not inherently provide gym compliant or RL-friendly interface.
1. Related pull requests in making inherent changes to SMARTS interface:
    + #691 
    + #712 
1. Alternatively, a SMARTS environment can be made gym-compatible by wrapping it with the `FormatObs` and `FormatAction` env wrappers.
1. What is needed to statisfy the requirement _"A way to isolate per-agent action_spaces"_ ?

Adaickalavan:
Closed by #1793

PR #1248, opened 2022-01-19
Standard obs

### LINKED ISSUE #548: Make observation gym friendly

### LINKED ISSUE #945: Standard Gym Environment

Adaickalavan:
1. Added `StandardObs` wrapper which converts SMARTS observations to standardized gym-compliant RL-friendly observations and returns `StdObs`.

1. Example usage:
    ```
    env = StandardObs(env=env)
    ```

1. Addresses #1192

1. Closes #548
1. Closes #945


smarts/env/wrappers/standard_obs.py
```diff
@@ -0,0 +1,582 @@
...
+    AgentInterface attributes.
+    """
+
+    def __init__(self, env: gym.Env):
+        """
+        Args:
+            env (gym.Env): SMARTS environment to be wrapped.
+
+        Raises:
+            AssertionError: If all agents do not have the same AgentInterface.
```
Gamenot, line 209: It occurs to me also that the agent interface adaptors are also potentially incompatible with this wrapper.

examples/single_agent.py
```diff
@@ -78,14 +67,24 @@ def main(scenarios, sim_name, headless, num_episodes, seed, max_episode_steps=No
     env.close()
 
 
+def _build_scenario(scenario):
+    build_scenario = f"scl scenario build-all --clean {scenario}"
+    os.system(build_scenario)
```
RutvikGupta, line 72: `subprocess.Popen()` is more flexible than `os.system` and i suggest we should use that instead.

smarts/env/wrappers/standard_obs.py
```diff
@@ -0,0 +1,582 @@
...
+        waypoint. shape=(4,20). dtype=np.float32.
+    pos:
+        Coordinate of a waypoint. Defaults to np.array([0,0,0]). 
+        shape=(4,20,3). dtype=np.float32.
+    speed_limit:
+        Lane speed limit at a waypoint in m/s. shape=(4,20). dtype=np.float32.
+    """
+
+
+class StandardObs(gym.ObservationWrapper):
```
sah-huawei, line 194: Given the possible confusion between `StandardObs` and `StdObs` noted above, I would suggest renaming this class to `StdObsWrapper`.

smarts/env/wrappers/standard_obs.py
```diff
@@ -0,0 +1,582 @@
...
+        "angular_acceleration": val.angular_acceleration.astype(np.float32),
+        "angular_jerk": val.angular_jerk.astype(np.float32),
+        "angular_velocity": val.angular_acceleration.astype(np.float32),
+        "box": np.array(val.bounding_box.as_lwh).astype(np.float32),
+        "heading": np.float32(val.heading),
+        "lane_index": np.int8(val.lane_index),
+        "linear_acceleration": val.linear_acceleration.astype(np.float32),
+        "linear_jerk": val.linear_jerk.astype(np.float32),
+        "linear_velocity": val.linear_velocity.astype(np.float32),
+        "pos": val.position.astype(np.float32),
```
sah-huawei, line 409: I think our positions are `np.float64` in the core.  Is that supported in spaces?  If so, we probably shouldn't lose precision here.

smarts/core/sensors.py
```diff
@@ -661,8 +654,6 @@ def __call__(self) -> OccupancyGridMap:
         grid = np.frombuffer(mem_view, np.uint8)
         grid.shape = (self._camera.tex.getYSize(), self._camera.tex.getXSize(), 1)
         grid = np.flipud(grid)
-        grid = grid.clip(min=0, max=1).astype(np.int8)
-        grid *= 100  # full confidence on known cells
```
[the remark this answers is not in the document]
Adaickalavan: This is intended removal of unnecessary lines of code.

examples/single_agent.py
```diff
@@ -1,68 +1,57 @@
...
 
 from smarts.core.agent import Agent, AgentSpec
 from smarts.core.agent_interface import AgentInterface, AgentType
-from smarts.core.sensors import Observation
 from smarts.core.utils.episodes import episodes
 from smarts.env.wrappers.single_agent import SingleAgent
+from smarts.env.wrappers.standard_obs import StandardObs, StdObs
 
-# The following ugliness was made necessary because the `aiohttp` #
-# dependency has an "examples" module too.  (See PR #1120.)
```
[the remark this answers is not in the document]
Adaickalavan: We can create an issue and link this ugliness inside the issue rather than adding ugliness to an especially user impression creating code.

examples/single_agent.py
Adaickalavan: Changed.

examples/egoless.py
```diff
@@ -14,17 +14,13 @@
...
     env = gym.make(
         "smarts.env:hiway-v0",
         scenarios=scenarios,
         agent_specs={},
-        sim_name=sim_name,
         headless=headless,
         sumo_headless=True,
-        visdom=False,
-        seed=seed,
-        fixed_timestep_sec=0.1,
```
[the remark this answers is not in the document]
Adaickalavan: I had removed them to keep our examples more focused and lean, which helps create a better user impression.

Is the `sim_name` really necessary here in these examples? 

If `sim_name` is not provided, they will simply default to `None`.

smarts/core/events.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020. Huawei Technologies Co., Ltd. All rights reserved.
+# Copyright (C) 2022. Huawei Technologies Co., Ltd. All rights reserved.
```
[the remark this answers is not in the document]
Adaickalavan: Ok.

examples/single_agent.py
```diff
@@ -1,68 +1,57 @@
...
 
 import gym
+import numpy as np
 
 from smarts.core.agent import Agent, AgentSpec
 from smarts.core.agent_interface import AgentInterface, AgentType
-from smarts.core.sensors import Observation
 from smarts.core.utils.episodes import episodes
 from smarts.env.wrappers.single_agent import SingleAgent
+from smarts.env.wrappers.standard_obs import StandardObs, StdObs
```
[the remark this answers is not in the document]
Adaickalavan: Changes:
`StandardObs` -> `FormatObs`
`StdObs` -> `StdObs`

Relation:
`FormatObs` formats the observations to produce gym-compliant vectorized observation `StdObs`.

smarts/env/wrappers/standard_obs.py
Adaickalavan: Answered above.

examples/single_agent.py
```diff
@@ -1,68 +1,57 @@
...
 
-    # Wrap a single-agent env with SingleAgent wrapper to make `step` and `reset`
-    # output compliant with gym spaces.
-    env = SingleAgent(env)
+    # Convert SMARTS observations to standardized gym-compliant observations.
+    env = StandardObs(env=env)
+
+    # Convert `env.step()` and `env.reset()` from multi-agent interface to
+    # single-agent interface.
+    env = SingleAgent(env=env)
```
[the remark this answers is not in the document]
Adaickalavan: Upcoming standard environments, for example `intersection_env` which is under development, will be wrapped with `FormatObs` inside the gym registration itself. 

I simply intended to show the usage of `StandardObs` here, but it might be unnecessary, hence it is removed from `single_agent.py` example.

smarts/core/events.py
```diff
@@ -17,16 +17,23 @@
...
+    not_moving: bool
     off_road: bool
     off_route: bool
     on_shoulder: bool
-    wrong_way: bool
-    not_moving: bool
     reached_goal: bool
     reached_max_episode_steps: bool
-    agents_alive_done: bool
+    wrong_way: bool
```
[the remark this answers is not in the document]
Adaickalavan: Only the type annotation of `collisions: Sequence  # Sequence[Collision]` was corrected to become ` collisions: List[Optional[Collision]]`, whereas the actual data returned for `Events.collisions` remains unchanged.

The order of args is reverted.

smarts/core/sensors.py
```diff
@@ -124,34 +124,27 @@ class Vias:
...
     occupancy_grid_map: OccupancyGridMap
+    road_waypoints: RoadWaypoints
     top_down_rgb: TopDownRGB
-    road_waypoints: RoadWaypoints = None
-    via_data: Vias = None
-
-
-@dataclass
-class Collision:
-    collidee_id: str
```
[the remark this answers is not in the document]
Adaickalavan: The `class Collision` was moved due to circular dependency and poor code structure.

smarts/env/wrappers/standard_obs.py
Adaickalavan: `FormatObs` wraps the original smarts observations controlled by the `AgentInterface`. Hence, the base environment fed to `FormatObs` should not use `AgentSpec.observation_adapter`. However, `AgentSpec.observation_adapter` can be used in subsequent environments derived and built upon the output environment of the `FormatObs` wrapper.

Following from https://github.com/huawei-noah/SMARTS/pull/1248#discussion_r791060268 and https://github.com/huawei-noah/SMARTS/pull/1248#discussion_r794074825, `FormatObs` is primarily going to be applied only inside gym registration itself. Hence, we can apply the `AgentSpec.observation_adapter`, if any is provided, after `FormatObs` wrapper inside the gym registration. Moreover, once the gym register returns an environment, the users are free to use `AgentSpec.observation_adapter` to modify the observations in an agent-specific manner.

With the above being said, I think observation modifier should be inside the agent and not inside the environment.

smarts/env/wrappers/standard_obs.py
Adaickalavan: Granted.

smarts/core/sensors.py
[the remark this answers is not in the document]
sah-huawei: Sure, I have no problem with moving it -- that makes sense.  What I was saying was that if some user who has written code that might parse collisions has something like this:
```python
from smarts.core.sensors import Collisions
```
They will have to change their code to:
```python
from smarts.core.events import Collisions
```

In general, changes in SMARTS that require our users to change their own code constitute "interface changes" should (at least)  be noted in the CHANGELOG.

examples/single_agent.py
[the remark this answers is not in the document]
sah-huawei: OK, point taken about the comment potentially adding to the existing ugliness.  

The reference to the _reason_ for the ugliness is what I don't want to lose (otherwise we'll be stuck with this ugliness forever, as opposed to hopefully cleaning it up once `aiohttp` is fixed.)...  I will create the issue to track this.

(fwiw, the reason I had the comment in the code instead of just in an issue was because this is in our public-facing examples area, and I didn't want potential users to "judge us" based upon this, when it wasn't really our fault that we had to make it so convoluted.)

smarts/env/hiway_env.py
```diff
@@ -226,7 +259,13 @@ def step(self, agent_actions):
 
         return observations, rewards, dones, infos
 
-    def reset(self):
+    def reset(self) -> Dict[str, Observation]:
+        """Resets the environment.
+
+        Returns:
+            Dict[str, Observations]: Agents' observation.
```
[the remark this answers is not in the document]
sah-huawei: Actually, I meant that the comment here says "Observations" instead of "Observation".

smarts/env/wrappers/standard_obs.py
sah-huawei: Cool.  Now I'm wondering about some of the other `np.float32` fields as well.  Did you audit these to see if any of them should also be `np.float64`?

.github/workflows/ci-format.yml
```diff
@@ -36,4 +36,35 @@ jobs:
...
             --ignore marl_benchmark,examples,scenarios,docs,manager_pb2_grpc.py,worker_pb2_grpc.py \
             --msg-template='{path}: line {line}: {msg_id}: {msg}' \
-            ./smarts ./envision
\ No newline at end of file
+            ./smarts ./envision
+
+  test-types:
+    runs-on: ubuntu-18.04
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    container: huaweinoah/smarts:v0.4.18-minimal
```
sah-huawei, line 44: can we just use "latest" for the version here?

.github/workflows/ci-format.yml
```diff
@@ -36,4 +36,35 @@ jobs:
...
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v13.2  
+        with:
+          base_sha: origin/develop     
+      - name: Check types
+        if: contains(steps.changed-files.outputs.all_changed_files, '.py')
+        run: |
+          . ${{env.venv_dir}}/bin/activate
+          pytype --disable=pyi-error ${{steps.changed-files.outputs.all_changed_files}}
```
sah-huawei, line 70: So does this mean that if I make a change to, say, `smarts.py`, then `pytype` will be run on that file and I will be forced to fix any existing or lingering type issues there (even if they don't relate to the thing I just changed)?

Or is it only run on _new_ files that I'm adding to the repo?

If it's the former, I'm not sure I would support this.

examples/multi_agent.py
```diff
@@ -1,16 +1,13 @@
...
 from smarts.core.agent_interface import AgentInterface, AgentType
 from smarts.core.utils.episodes import episodes
 
-# The following ugliness was made necessary because the `aiohttp` #
-# dependency has an "examples" module too.  (See PR #1120.)
-if __name__ == "__main__":
-    from argument_parser import default_argument_parser
-else:
-    from .argument_parser import default_argument_parser
-
```
RutvikGupta, line 13: Is this issue fixed?

examples/ray_multi_instance.py
```diff
@@ -17,18 +17,11 @@
     raise RayException.required_to("ray_multi_instance.py")
 
 
+from examples.argument_parser import default_argument_parser
```
sah-huawei, line 20: I take it the problem with `aiohttp` that meant we had to use relative imports no longer exists?  Is this the case?

examples/multi_agent.py
```diff
@@ -65,10 +61,15 @@ def main(scenarios, sim_name, headless, num_episodes, seed, max_episode_steps=No
     parser = default_argument_parser("multi-agent-example")
     args = parser.parse_args()
 
+    if not args.scenarios:
+        args.scenarios = [
+            str(pathlib.Path(__file__).absolute().parents[1] / "scenarios" / "loop")
+        ]
+
+    build_scenario(args.scenarios)
```
RutvikGupta, line 69:
```suggestion
    if not args.scenarios:
        args.scenarios = [
            str(pathlib.Path(__file__).absolute().parents[1] / "scenarios" / "loop")
        ]

        build_scenario(args.scenarios)
```
I don't think its a good idea to rebuild the scenario if it was passed as an argument. I can see that the user can forget to build the scenario before running an example on it, but we have check for those cases down the stream.

smarts/core/sensors.py
```diff
@@ -799,7 +796,7 @@ def __init__(
         self,
         vehicle,
         bullet_client,
-        sensor_params: SensorParams = None,
+        sensor_params: Optional[SensorParams] = None,
         lidar_offset=(0, 0, 1),
```
RutvikGupta, line 800: Why are there no type defs for other three attributes?

smarts/core/sensors.py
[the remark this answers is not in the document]
Adaickalavan: Since this code change is not integral to this PR, this code change has already been undone previously.

.github/workflows/ci-format.yml
Adaickalavan: The container `huaweinoah/smarts:v0.4.18-minimal` only has the necessary apt packages and sumo installed. All other smarts python files are installed dynamically during CI. No change appears to be needed here for now.

Whereas the container `huaweinoah/smarts:latest` contains apt packages, sumo, and pre-installed smarts files. This is not suited for our CI jobs.

See the difference between the two docker files here: https://github.com/huawei-noah/SMARTS/tree/standard-obs/utils/docker .

examples/single_agent.py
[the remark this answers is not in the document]
Adaickalavan: The latest version of `aiohttp` package does not cause issues with using the `example` module. The `argument_parser` imports have already been fixed.

.github/workflows/ci-format.yml
Adaickalavan: Currently all added, copied, modified, and renamed files, with respect to the `develop` branch will be checked by pytype.

Poor unchecked coding and typing is precisely how we ended up at https://github.com/huawei-noah/SMARTS/pull/1248#discussion_r791071308 and at https://github.com/huawei-noah/SMARTS/pull/1248#discussion_r791097068, which continually bugs the users in a never ending cycle.

pytype is very lenient in type checking. Please give it a try first.

examples/ray_multi_instance.py
Adaickalavan: The latest version of `aiohttp` package does not cause issues with using the `examples` module. The `argument_parser` imports have thus been fixed.

smarts/core/sensors.py
Adaickalavan: Here, I had simply fixed an erroneous type caught by pytype.

smarts/env/wrappers/standard_obs.py
Adaickalavan: I think the other fields are okay to use `np.float32`.

examples/egoless.py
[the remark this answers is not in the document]
Gamenot: It is useful for viewing in envision, especially if there are multiple examples going or a playback running. I don't have a heavy preference toward keeping it.

smarts/env/hiway_env.py
```diff
@@ -164,27 +174,23 @@ def __init__(
 
     @property
     def agent_specs(self) -> Dict[str, AgentSpec]:
-        """Agent specs currently in use for this simulation.
+        """Agents' specifications used in this simulation.
 
         Returns:
-            A list of AgentSpec.
+            (Dict[str, AgentSpec]): Agents' specifications.
```
Gamenot, line 180: I wonder if the type annotation is needed in the docstring when it is provided in the type hint. It ends up looking redundant. Although, I know it is a requirement of this comment format.

.github/workflows/ci-format.yml
sah-huawei: I like `pytype` and have been using it now for a few weeks (including fixing old code in unrelated PRs, such as #1219!)  No need to just "give it a try" -- I am using it regularly, including using it to fix old code (see #1208 for example).

My objection is not to pytype itself, but to forcing people making changes/fixes to existing code to have to _also_ change other, completely unrelated code elsewhere in the file that potentially has nothing to do with the Issue on which they're working (and which they may not understand adequately enough to make changes at that point, especially if they are not part of the core smarts team) just to meet the demands of our now-automated checking.  

Note that I'm _not_ saying such fixes shouldn't ever be made:  rather I think they should be made in their own, dedicated PRs, rather than "along the way" in other PRs.  For example, I would argue that even though all of the type hint changes I made in, say, PR #1219 were needed (and I'm glad I did them), they were just a distraction for the actual fixes in that particular PR and made it much harder to review.

Regarding the comments you linked about the `Collisions` class, while I agree with moving it into Events (and said so there), the issue there had _very_ _little_ to do with type hints (which were added retroactively well after both classes were written), and someone might even debate with you that there was anything whatsoever wrong with the way it was.  (I'm not doing that though.)

Where something like `Collision` is defined should _not_ be _solely_ a matter of what uses it within SMARTS.   You may consider it "poor coding" that it was originally defined or remained in `sensors.py` -- I'm not really arguing that one way or another. (I think the circumstances of it remaining  there while `Events` was moved to its own file predate both of us on this project... it's hard for either of us to evaluate that decision without context.)

While I do believe that it _currently_ makes sense to define the `Collision` class in `events.py` in this case, I also believe that it should not be _required_ to be just so by _pytype_ -- that's a decision that has to do with more than _just_ the dependency tree!  

And indeed, I believe the fact that you felt the need to move it -- a breaking change -- in a PR about something entirely different _is a case in point_ about the _misuse_ of the `pytype` tool and/or a misunderstanding of the PR process.

Aside:  In my opinion, this dispute is partly caused by forgetting that type hints are just that -- _hints_.  Because of the fact that the _python import model isn't currently designed for strong typing_, using type hints religiously brings about many spurious issues with circular dependencies that are not real issues, but rather artifacts of python's inability to support forward declarations like C++ does.  In my opinion,  trying to bend over backwards (i.e., make the code more convoluted/complex) in order to accommodate an (imperfect) third-party tool detracts from the value of type hints.  I'd rather leave them off completely in such cases (while using them where they don't interfere).

.github/workflows/ci-format.yml
sah-huawei: ok

examples/ray_multi_instance.py
sah-huawei: Good!

.github/workflows/ci-format.yml
sah-huawei: The above is my only outstanding issue with this PR.  I suspect we aren't going to quickly come to an agreement on this point.  So in the interest of getting this PR merged, I suggest that we move the `pytype` CI change to a separate PR on its own (if such a PR doesn't already exist) for further discussion on these points and out of this PR.  If so, I can approve this.

smarts/env/hiway_env.py
Adaickalavan: I was simply using the automated Python Docstring Generator extension in my VSCode which had docstring format options of docblockr, google, sphinx, and numpy. Mine was set to use google. 

We can decide on a suitable docstring format, and ensure all docstring conforms to it.

.github/workflows/ci-format.yml
sah-huawei: > ... I suggest that we move the `pytype` CI change to a separate PR on its own (if such a PR doesn't already exist) ...

Oh, I see I missed reviewing the `pytype` CI PR and that this change has _already_ made it into the `develop` branch.  In that case, I can approve this PR now "as is".

I will bring up my objections to the `pytype` CI change in a meeting and/or Issue instead.

sah-huawei (approved):
Despite my objections to the `pytype` CI change here, I am going to approve this because it seems the CI change has already made it into `develop` anyway, independently of this.

examples/multi_agent.py
Adaickalavan: I desired the code to run seamlessly even if the user forgets to pass a scenario argument or forgets to build the scenario. Catching unbuilt scenario later down stream and throwing an error, kind of spoils the seamless experience. 

I have removed the `--clean` option from the `scl scenario build-all` command, which should reduce some of the things regenerated. 

Please do suggest, if you have other better code ideas.

examples/single_agent.py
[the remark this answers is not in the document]
Gamenot: OK, that is reasonable.

Gamenot (approved):

smarts/env/hiway_env.py
Gamenot: I think it would be best to get everyone on-board with the same `docstring` format. https://github.com/huawei-noah/SMARTS/issues/1263

.github/workflows/ci-format.yml
sah-huawei: Note:  What just happened in PR #1266 is exactly the type of thing I was worried about above.  It was a fairly simple PR about not setting `dones` during reset, but because of the pytype CI reports, Allan felt the need to _also_ attempt to fix the (currently broken) vehicle typing scheme as part of that PR, even though he didn't fully understand the reasons for its being broken or the complexity of the change needed.

PR #1249, opened 2022-01-21
Add continuous closed loop flow [closes #1239]

### LINKED ISSUE #1239: Generating continuous loop scenario
**High Level Description**
I am new to SMARTS and Sumo. I  am trying to generate a continuous loop scenario. The scenario that I wish to deploy has already been implemented in the Project Flow ([Video for the scenario](https://www.youtube.com/watch?time_continue=1&v=vFJvjPC8eNo&feature=emb_title))

Scenario Generated

**Desired SMARTS version**
0.4.18

**Operating System**
Arch Linux

**Problems**
I read the documentation and the example scenarios. However still have problem to tackle the following challenges:

1. The vehicles are generated one by one instead of all together as shown the image below.
2. It seems that the vehicles are vanished after they reach the end of the road/edge. I can see in the [Sumo documentation ](https://sumo.dlr.de/docs/Tutorials/Driving_in_Circles.html) that we need to add rerouters. But,  I am not sure how can we do this using Scenario Studio.
3. How can I add/generate car follow model?

Scenario Generated

I created the map and my code for generating the scenario is as follows

```from pathlib import Path

from smarts.sstudio import gen_scenario
from smarts.sstudio import types as t
from smarts.sstudio import gen_traffic


traffic = t.Traffic(
    flows=[
        t.Flow(
            route=t.Route(begin=("top", 0, 'random'), end=("bottom", 0, 'max')),
            # route=t.Route(begin=("445633932", 0, 'random'), end=("445633931", 0, 'max')),
            begin=0, end=1 * 60 * 60, rate=0.1,
            actors={t.TrafficActor(name="car"): 1},
        )
        for i in range(20)
    ]
)

# open_agent_actor = t.SocialAgentActor(
#     name="open-agent", agent_locator="zoo.policies:keep-lane-agent-v0"
# )

# laner_actor = t.SocialAgentActor(
#     name="keep-lane-agent",
#     agent_locator="zoo.policies:keep-lane-agent-v0",
# )

gen_scenario(
    t.Scenario(
        traffic={"basic": traffic},
    ),
    output_dir=Path(__file__).parent,
)
```

I would be grateful if you could help me with the above challenges.

Thank you

Gamenot:
Addresses the need for a continuous closed loop traffic flow.

### DIFF
--- a/smarts/sstudio/types.py
+++ b/smarts/sstudio/types.py
@@ -392,6 +392,36 @@ class Flow:
     def __eq__(self, other):
         return self.__class__ == other.__class__ and hash(self) == hash(other)
 
+@dataclass
+class SingleVehicle:
+    """A single vehicle that can be interchanged with flow."""
+    route: Union[RandomRoute, Route]
+    """The route that the vehicle will use."""
+
+    actors: Dict[TrafficActor, float] = field(default_factory=dict)
+    """An actor to weight mapping associated as\\: { actor: weight }
+    actor:
+        The traffic actors that are provided.
+    weight:
+        The chance of this actor appearing as a ratio over total weight.
+    """
+
+@dataclass
+class Rerouter:
+    """A route changing pad that will change the end route of an actor."""
+    begin: Tuple[str, int, Any]
+    """The rerouter location."""
+    ends: Dict[Tuple[str, int, Any], float] = field(default_factory=dict)
+    """The end locations that actors will be rerouted to."""
+    time_threshold: float = 0
+    """The time offset from the start of the simulation until this rerouter becomes active."""
+    probability: float = 1
+    """The probability [0:1] that this rerouter will reroute an actor."""
+    vehicle_types: Sequence[str] = field(default=set)
+    """The actor types that this rerouter affects. Empty definition affects everything."""
+    off: bool = False
+    """Deactivate."""
+
 
 @dataclass(frozen=True)
 class JunctionEdgeIDResolver:
--- a/smarts/sstudio/types.py
+++ b/smarts/sstudio/types.py
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 import collections.abc as collections_abc
+from functools import cached_property
 import hashlib
 import logging
 import pickle
@@ -393,7 +394,7 @@ class Flow:
         return self.__class__ == other.__class__ and hash(self) == hash(other)
 
 
-@dataclass
+@dataclass(frozen=True)
 class SingleVehicle:
     """A single vehicle that can be interchanged with flow."""
 
@@ -407,6 +408,23 @@ class SingleVehicle:
     weight:
         The chance of this actor appearing as a ratio over total weight.
     """
+
+    depart: float = 0
+
+    @property
+    def id(self) -> str:
+        return "single-vehicle-{}-{}-".format(
+            self.route.id,
+            str(_pickle_hash(sorted(self.actors.items(), key=lambda a: a[0].name))),
+        )
+
+    def __hash__(self):
+        # Custom hash since self.actors is not hashable, here we first convert to a
+        # frozenset.
+        return _pickle_hash((self.route, self.rate, frozenset(self.actors.items())))
+
+    def __eq__(self, other):
+        return self.__class__ == other.__class__ and hash(self) == hash(other)
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


smarts/sstudio/types.py
```diff
@@ -434,6 +486,31 @@ class Traffic:
...
+    """Single instance vehicles"""
+
+    @cached_property
+    def actors(self) -> Sequence[TrafficActor]:
+        actors = set()
+        for a in [a for flow in self.flows for a in flow.actors]:
+            actors.add(a)
+
+        for a in [a for sv in self.single_vehicles for a in sv.actors]:
+            actors.add(a)
```
sah-huawei, line 499: It seems this can be done more efficiently with set comprehensions:
```python
    return {a for flow in self.flows for a in flow.actors} | {a for sv in self.single_vehicles for a in sv.actors}
```

smarts/sstudio/types.py
```diff
@@ -434,6 +486,31 @@ class Traffic:
...
+    @cached_property
+    def routes(self) -> Sequence[Route]:
+        routes = set()
+        for r in (f.route for f in self.flows):
+            routes.add(r)
+
+        for r in (sv.route for sv in self.single_vehicles):
+            routes.add(r)
+
+        return routes
```
sah-huawei, line 512: Same here:
```python
return {f.route for f in self.flows} | {sv.route for sv in self.single_vehicles}
```

smarts/sstudio/types.py
```diff
@@ -393,6 +394,57 @@ def __eq__(self, other):
...
+
+
+@dataclass
+class Rerouter:
+    """A route changing pad that will change the end route of an actor."""
+
+    begin: Tuple[str, int, Any]
+    """The rerouter location."""
+    ends: Dict[Tuple[str, int, Any], float] = field(default_factory=dict)
+    """The end locations that actors will be rerouted to."""
```
sah-huawei, line 437: The comment here should indicate that the dict key is (I assume) `(road, lane_index, offset)`. But what is the `float` value in this dict for?

smarts/sstudio/types.py
```diff
@@ -393,6 +394,57 @@ def __eq__(self, other):
...
+    def __eq__(self, other):
+        return self.__class__ == other.__class__ and hash(self) == hash(other)
+
+
+@dataclass
+class Rerouter:
+    """A route changing pad that will change the end route of an actor."""
+
+    begin: Tuple[str, int, Any]
+    """The rerouter location."""
```
sah-huawei, line 435: I'm assuming this is `(road, lane_index, offset)` right?  (If so, why are we using `Any` for offset instead of `float`?)  I think you should indicate the semantics of the tuple in the comment.

smarts/sstudio/types.py
```diff
@@ -393,6 +394,57 @@ def __eq__(self, other):
...
+    def id(self) -> str:
+        return "single-vehicle-{}-{}-".format(
+            self.route.id,
+            str(_pickle_hash(sorted(self.actors.items(), key=lambda a: a[0].name))),
+        )
+
+    def __hash__(self):
+        # Custom hash since self.actors is not hashable, here we first convert to a
+        # frozenset.
+        return _pickle_hash((self.route, self.rate, frozenset(self.actors.items())))
```
sah-huawei, line 424: What is `self.rate` here?

smarts/sstudio/types.py
```diff
@@ -393,6 +394,57 @@ def __eq__(self, other):
...
+    depart: float = 0
+
+    @property
+    def id(self) -> str:
+        return "single-vehicle-{}-{}-".format(
+            self.route.id,
+            str(_pickle_hash(sorted(self.actors.items(), key=lambda a: a[0].name))),
+        )
+
+    def __hash__(self):
```
sah-huawei, line 421: I think this can be a `@cached_property`.

smarts/sstudio/types.py
```diff
@@ -393,6 +394,57 @@ def __eq__(self, other):
...
+
+    actors: Dict[TrafficActor, float] = field(default_factory=dict)
+    """An actor to weight mapping associated as\\: { actor: weight }
+    actor:
+        The traffic actors that are provided.
+    weight:
+        The chance of this actor appearing as a ratio over total weight.
+    """
+
+    depart: float = 0
```
sah-huawei, line 412: This needs a docstring explaining what it is.

smarts/sstudio/types.py
```diff
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 import collections.abc as collections_abc
+from functools import cached_property
```
sah-huawei, line 21: I think this should be:
```python
from cached_property import cached_property
```

PR #1250, opened 2022-01-21
Driving in traffic - dependency version and type annotation

Adaickalavan:
1. Changed dependency version, namely `smarts[camera-obs]==0.5.0`
2. Added type annotations
3. Made minor improvements to `logdir` handling


examples/driving_in_traffic/run.py
```diff
@@ -111,7 +90,24 @@ def _build_scenario():
...
+def _setup_gpu():
+    gpus = tf.config.list_physical_devices("GPU")
+    if gpus:
+        try:
+            for gpu in gpus:
+                tf.config.experimental.set_memory_growth(gpu, True)
+            logical_gpus = tf.config.list_logical_devices("GPU")
+            print(len(gpus), "Physical GPUs,", len(logical_gpus), "Logical GPUs")
+        except RuntimeError as e:
+            print(e)
```
RutvikGupta, line 102: We don't need to raise the error?

examples/driving_in_traffic/run.py
```diff
@@ -57,46 +58,24 @@ def main(args):
...
-                "eval_eps": 1e8,
-            }
-        )
+        logdir = args.logdir
+        config_dv2 = config_dv2.update({"eval_eps": 1e8})
     else:
         raise KeyError(
             f'Expected \'train\' or \'evaluate\', but got {config_env["mode"]}.'
         )
+    config_dv2 = config_dv2.update({"logdir": logdir})
```
RutvikGupta, line 78: Why is the `logdir` path stored in the configuration file if it is being passed by user and not being used by the config file anywhere?

RutvikGupta (approved):

examples/driving_in_traffic/run.py
Adaickalavan: It is being used later in the `run` function.
https://github.com/huawei-noah/SMARTS/blob/a3a32ddb96e0d4cd41d69e7de7c8d559f2dab585/examples/driving_in_traffic/run.py#L126

smarts/core/utils/logging.py
```diff
@@ -37,7 +37,7 @@ def timeit(name: str, logger):
 def isnotebook():
     try:
         shell = get_ipython().__class__.__name__
-        if shell == "ZMQInteractiveShell":
+        if shell == "ZMQInteractiveShell" or "google.colab" in sys.modules:
```
Gamenot, line 40: Does is this check for "google.colab" need to be after `ipython`.

Gamenot (approved):

smarts/core/utils/logging.py
Adaickalavan: `"google.colab" in sys.modules` is to detect whether the code is being executed in Colab.

examples/driving_in_traffic/run.py
Adaickalavan: In this code, we are trying to set the memory growth property of the GPU. However, depending on the user’s GPU setup, we might fail in setting this property. Failure in setting this property does not mean the code is unable to run. Even without the call to `_setup_gpu()`, the code should be able to run using the available GPU or CPU. Therefore, printing a message on the failure to set GPU memory property and allowing the code to continue running appears to be the better step in order not to unnecessarily frustrate the user.

For more reference, see: 
+ https://www.tensorflow.org/guide/gpu#limiting_gpu_memory_growth
+ https://www.tensorflow.org/api_docs/python/tf/config/experimental/set_memory_growth

PR #1252, opened 2022-01-24
Auto format and commit

Adaickalavan:
Automatically format (`isort`, `black`, `prettier`) the code and commit the changes using CI job in `ci-auto-commit-linux.yml` workflow.


Gamenot (approved):
I can bring up the new issue that merge requests will require branches based on another branch to coordinate merges so as to not come into conflict with the auto-formatting.

.github/workflows/ci-auto-commit-linux.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          . ${{env.venv_dir}}/bin/activate
+          pip install .[dev]
+          isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/marl_benchmark/ ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./ultra ./zoo
+          black .
+          npx prettier --write ./envision/web/src
+      - name: Commit format changes
+        uses: EndBug/add-and-commit@v7
+        with:
+          add: '*.py'
```
RutvikGupta, line 58: Shouldn't the files formatted by `npx prettier --write` like the `.html` files also need to be added?

Adaickalavan: Good catch. I have changed the git add command to `add: '["*.py", "*.html", "*.js"]'` .

Issue #1167, opened 2021-12-10
Example agent that follows a mission


saulfield:
**Is your feature request related to a problem? Please describe.**
While testing the OpenDrive implementation of waypoints, we noticed there doesn't seem to be an agent in the examples that follows waypoints for a mission. This would be a good example to have for users and for testing.

**Describe the solution you'd like**
An audit of the example agents, and a new agent that fits this role if we don't already have a similar one.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

liu-allan:
Is there an example scenario in mind to test? By waypoints do you mean something similar to vias?

PR #1254, opened 2022-01-24
Waypoint following agent

liu-allan:
#1167 

The agent follows waypoint trajectories and changes lanes to follow the mission.
image

In testing, I found that in this location on `od_merge` there are multiple waypoint paths that have the same trajectory. Here there are 6 waypoint paths and they don't change lanes like in the image above

image

I think this agent is only useful for testing OpenDrive maps because it doesn't seem like sumo supports waypoints that change lanes like in the first image. In general for sumo maps, the `single_agent` example will follow the mission.


examples/waypoint_tracking_agent.py
```diff
@@ -0,0 +1,158 @@
...
+    def act(self, obs):
+
+        current_lane = obs.ego_vehicle_state.lane_index
+        # Desired speed is in m/s
+        desired_speed = 10
+        wp_index = 0
+
+        num_lanes = 0
+        for path in obs.waypoint_paths:
+            num_lanes = max(num_lanes, path[0].lane_index + 1)
```
RutvikGupta, line 37: Are we trying to find here the total lanes that this road has or number of different lanes over which all the waypoints paths start? I ask this since these two values can differ.

examples/waypoint_tracking_agent.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # This is when there are waypoint paths that change lanes
+        if (
+            len(obs.waypoint_paths) > num_lanes
+            and len(obs.waypoint_paths) <= num_lanes ** 2
```
RutvikGupta, line 42:
```suggestion
            num_lanes < len(obs.waypoint_paths) <= num_lanes ** 2
```

RutvikGupta, line 42: Can we say this can accurately find whether we have a waypoint path that has a lane change?

examples/waypoint_tracking_agent.py
```diff
@@ -0,0 +1,158 @@
...
+
+        else:
+            for i in range(len(obs.waypoint_paths)):
+                if obs.waypoint_paths[i][0].lane_index == current_lane:
+                    wp_index = i
+                    break
+
+        if self.waypoint_path:
+
+            num_trajectory_points = min([5, len(self.waypoint_path)])
```
RutvikGupta, line 72: why do we take at minimum 5 trajectory points?

examples/waypoint_tracking_agent.py
```diff
@@ -0,0 +1,158 @@
...
+                [
+                    obs.waypoint_paths[wp_index][i].pos[1]
+                    for i in range(num_trajectory_points)
+                ],
+                [
+                    obs.waypoint_paths[wp_index][i].heading
+                    for i in range(num_trajectory_points)
+                ],
+                [desired_speed for i in range(num_trajectory_points)],
+            ]
```
RutvikGupta, line 98: This if and else block can be simplified by storing the `waypoint_path` chosen into a temporary variable and and then computing the trajectory using that temp variable.

examples/waypoint_tracking_agent.py
liu-allan: This was completely arbitrary, I think I will change it back to 10 which was used in the `trajectory_tracking_agent` example.

examples/waypoint_tracking_agent.py
liu-allan: Total lanes of the road, however that is a good idea to check the number of different lanes over which they start/end

examples/waypoint_tracking_agent.py
liu-allan: I am not sure, my reasoning was that by default there must be an equal number of waypoint paths as lanes, and then any lane changing path would be in addition to the number of lanes. And then the maximum would be the number of lanes squared unless there are multiple waypoint paths that start and end in the same lane.

examples/waypoint_tracking_agent.py
```diff
@@ -0,0 +1,140 @@
...
+from smarts.core.sensors import Observation
+from smarts.core.utils.episodes import episodes
+from smarts.env.wrappers.single_agent import SingleAgent
+
+# The following ugliness was made necessary because the `aiohttp` #
+# dependency has an "examples" module too.  (See PR #1120.)
+if __name__ == "__main__":
+    from argument_parser import default_argument_parser
+else:
+    from .argument_parser import default_argument_parser
```
sah-huawei, line 17: Now that PR #1248 has been merged, I think you can simplify this back to just `from examples.argument_parser import default_argument_parser`.

examples/waypoint_tracking_agent.py
sah-huawei: For arbitrary magic numbers like this, I would suggest a constant with a symbolic name.  (Or even a class field, depending on how configurable you want to make things.)

RutvikGupta (approved):

Gamenot (approved):

Issue #1218, opened 2022-01-02
Vehicle starts on road shoulder


Adaickalavan:
## **BUG REPORT**
Let `done_criteria` to include `on_shoulder=True`.

https://github.com/huawei-noah/SMARTS/blob/1258f7e671f2e4bc2daefecb99fed816b1b46ccd/smarts/core/smarts.py#L350-L356

The smarts `reset` code is shown above. Here, when the vehicle appears and returns `observations_for_ego`, it is sometimes already `done=True` as the `observations_for_ego.events.on_shoulder=True`. Simply put, the vehicle appeared on road shoulder. Hence, the subsequent first call to `step` after `reset` will throw error as the agent is already `done`. 

**SMARTS version**
0.4.18.

**Steps to reproduce the bug**
Set `done_criteria` to include `on_shoulder=True`.

**Resulting and expected behaviour**
Vehicle should preferably not start on road shoulder.

**Impact [If known]**
If this is not fixed, then `done_criteria` cannot include `on_shoulder` condition.

Gamenot:
This appears to be an issue with translating to `Gym` reset. I am not quite sure how to deal with this issue. I think `Gym` requires a valid initial state to be available after reset so the requirement in order to satisfy `Gym` might be to force the vehicle into a valid initial state.

PR #1258, opened 2022-01-26
Agents being done during reset

liu-allan:
#1218 

Currently looping until the agent(s) are not done on the reset step.
Also addresses the problem with `step_count` where it could be greater than `max_episode_steps` before the episode begins.

### DIFF
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -414,7 +414,6 @@ class SMARTS:
 
         self._total_sim_time += self._elapsed_sim_time
         self._elapsed_sim_time = 0
-        self._step_count = 0
         self._reset_required = False
 
         self._vehicle_states = [v.state for v in self._vehicle_index.vehicles]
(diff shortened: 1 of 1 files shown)


sah-huawei (commented):
Do we yet understand how/why vehicles are starting on the shoulder to begin with in Issue #1218?  It seems that this is the real root of the problem (even before Gym gets involved).  If so, changing `reset()` may just _mask whatever root problem(s) may be occurring here_.  Thus, I'm hesitant to approve anything until we know the answers to some of the questions below.

Assuming the Agent is getting assigned a vehicle via a Trap here, there are roughly two ways that the vehicle's initial position will be determined:  from the position of an existing background vehicle that is captured within the trap, or if the trap's patience expires, from the starting point of its mission.   (Note that the trap is created at the mission starting point, so this is doubly implicated.)

So to evaluate what sort of fix to make for this, I think we need a few more pieces of information:  
- first, is the Agent in question being given a random, endless mission, or a user-defined mission in #1218.  (I'm assuming the former here, otherwise, the user should just adjust their mission start point.)
- second, is the problem happening due to a _capture_ of a vehicle within a Trap, or due the trap expiring and a new vehicle being created?
    - if it's due to _capture_, then is Sumo being used to create the background traffic?  assuming so, why/how are Sumo vehicles on the shoulder at any point?  the most likely (only?) explanation I can think of for that would be some map issue (since Sumo vehicles try to remain in the center of their lane, always), like that roads are too curved relative to vehicle lengths.   So what scenario is being used here?
    - if it's not due to capture, then why are new vehicles being created on the shoulder?  the initial position is a function of the Agent's mission Start, which in the case of random, endless missions is chosen at the center point of the lane.  If so, this _also_ indicates a potential problem with the scenario and/or map, and something the user potentially needs to fix.
 
In general, I'm not necessarily opposed to the changes proposed to the `reset()` function here, but I think we need a better understanding of the "bug(s)" before adding them, as they will mask what could be existing bugs in our Trap/Mission/Scenario -related code and/or one or more of our scenario maps.

liu-allan:
@Adaickalavan

Issue #1264, opened 2022-01-31
Customizable time length for not_moving done criteria


liu-allan:
**Is your feature request related to a problem? Please describe.**

The `not_moving` done criteria currently checks if the agent has not moved more than 1 meter in 60s

**Describe the solution you'd like**
A way for the user to determine how long to check before activating the `not_moving` done criteria

PR #1265, opened 2022-02-01
Allow user to specify `not_moving` check duration

liu-allan:
#1264 

Allows the user to specify how long to check the vehicle has not moved for the `not_moving` done criteria


smarts/core/agent_interface.py
```diff
@@ -188,7 +188,7 @@ class DoneCriteria:
     """End the episode when the agent drives in the wrong direction, even though it
     may be driving on the mission route.
     """
-    not_moving: bool = False
+    not_moving: Tuple = (False, 60)
```
Gamenot, line 191: I think we should use a `dataclass` here that names the two values.

smarts/core/agent_interface.py
```diff
@@ -188,7 +188,7 @@ class DoneCriteria:
...
     """End the episode when the agent is not moving for 60 seconds or more. To account
     for controller noise not moving means <= 1 meter of displacement within 60 seconds.
     """
```
Gamenot, line 194: Since we are making this change then the description will need to be updated.

smarts/core/agent_interface.py
Gamenot, line 191: I think we have run into a semantic issue here. This is no longer simply `DoneCriteria`. This is full event configuration, these events still fire even if they do not cause done so changing the time it takes for the `not_moving` event to fire is independent from if it is flagged as a done criteria.

One of the following must be done here:
- change this class into an `EventConfiguration`
- have `EventConfiguration` separate from the `DoneCriteria`.

smarts/core/sensors.py
```diff
@@ -448,7 +448,7 @@ def _is_done_with_events(cls, sim, agent_id, vehicle, sensor_state):
         collided = sim.vehicle_did_collide(vehicle.id)
         is_off_road = cls._vehicle_is_off_road(sim, vehicle)
         is_on_shoulder = cls._vehicle_is_on_shoulder(sim, vehicle)
-        is_not_moving = cls._vehicle_is_not_moving(sim, vehicle)
+        is_not_moving = cls._vehicle_is_not_moving(sim, vehicle, done_criteria.not_moving[1])
```
Gamenot, line 451: My point here (see my previous comment). This result is has implications on the `Events` object as well as `done`.

Gamenot (changes requested):

smarts/core/agent_interface.py
```diff
@@ -172,6 +172,12 @@ class AgentsAliveDoneCriteria:
...
 
 
+@dataclass(frozen=True)
+class EventConfiguration:
+    not_moving: int = 60
+    """How long to check the agent is not moving"""
+
```
liu-allan, line 179: Currently `not_moving` is the only event I really see that has a configurable aspect to it

liu-allan:
@Gamenot please take another look!

smarts/core/agent_interface.py
Gamenot: `not_moving` also has the 1 meter value that could potentially be configurable.

I also wonder if `wrong_way` could be time configured as well to implicitly account for overtaking. But that is not for this PR.

smarts/core/agent_interface.py
```diff
@@ -172,6 +172,12 @@ class AgentsAliveDoneCriteria:
     """
 
 
+@dataclass(frozen=True)
+class EventConfiguration:
+    not_moving: int = 60
+    """How long to check the agent is not moving"""
```
Gamenot, line 178:
```suggestion
    """How long in seconds to check the agent after which it is considered not moving"""
```

Gamenot:
> @Gamenot please take another look!

Yes, this seems to be a proper direction to take the configuration.

Gamenot:
@liu-allan Please update the `CHANGELOG`.

smarts/core/agent_interface.py
```diff
@@ -172,6 +172,12 @@ class AgentsAliveDoneCriteria:
     """
 
 
+@dataclass(frozen=True)
+class EventConfiguration:
```
Gamenot, line 176: `docstring` is needed here.

Gamenot (changes requested):

smarts/core/agent_interface.py
```diff
@@ -172,6 +172,16 @@ class AgentsAliveDoneCriteria:
...
+    """Configure the conditions in which an Event is triggered."""
+
+    not_moving_time: int = 60
+    """How long in seconds to check the agent after which it is considered not moving"""
+    not_moving_distance: int = 1
+    """How many meters to check for the agent to move after which it is considered not moving"""
```
Gamenot, line 182: I believe these values should be `float`.

```suggestion
    not_moving_time: float = 60
    """How long in seconds to check the agent after which it is considered not moving"""
    not_moving_distance: float = 1
    """How many meters the agent's actor has to move under which the agent is considered not moving"""
```

Gamenot (changes requested):

RutvikGupta (approved):

Issue #1230, opened 2022-01-07
CI tests have non-deterministic results


sah-huawei:
## **BUG REPORT**
**High Level Description**
This week I've noticed our CI tests failing more often.  However, when they tests are re-run, they then pass.  We need to track down the source of this non-determinism.

**SMARTS version**
latest (~0.5) on the `develop` branch

This happened to me on PRs #1222, #1219, and #1205.  Each failed on one of the last commits before merging, but then passed after the tests were re-run.

At least some of the failures seem to have to do with gRPC timeouts.

Gamenot:
It happened again. https://github.com/huawei-noah/SMARTS/runs/4742297574?check_suite_focus=true#step:5:995

I can note that the error is simulator state based so the retry of `reset()` does not help.

sah-huawei:
After much trial-and-error, I've finally made some progress in diagnosing this.  

It looks like the problem causing the test to sometimes-but-not-always fail in `test_stack_frame.py` is:
- when it passes, "Agent-002" will capture a vehicle ("car-flow-random-route-17fc695a-7772817341121806087--4-0.0") right at the point when its patience is also expiring for its trap; but
- when it fails, "Agent-002" will not capture the vehicle and instead its patience expires.  when that happens, it is blocked from adding a new vehicle in that location by the presence of "car-flow-random-route-17fc695a-7772817341121806087--4-0.0", so "Agent-002" will not get a vehicle at all on the step and the result of the `reset()` call will only show one vehicle ("Agent-001").

Note that `reset()` triggers multiple steps (about 11) for this use of the `loop` scenario because it doesn't return until there are observations for an ego agent, so until one or both agents capture a vehicle or their trap patience expires.

Although I still don't know the source of the non-determinism, it seems that failure path is more likely to be taken when multiple CI jobs are running simultaneously (which is what I've been doing in order to get any info about it at all), presumably "bogging down" the CPUs on the server instance where the CI runs (?).  I don't yet understand why this would have an effect on the Sumo-controlled vehicles positioning on each step, but at least I now have a clue as to where to look.

sah-huawei:
Also, regarding the failures coming from the tests in `test_parallel_env.py`, I've noticed that the `seed` passed to the `sumo` command when starting it in `SumoTrafficSimulation` is different for one of the two parallel processes than is used for the other process or for the single-env case.  This may be relevant, or a red herring... I still need to investigate this further.

Gamenot:
[...]

PR #1266, opened 2022-02-04
Prevent done during reset

liu-allan:
Disables `dones` from returning true during a reset

#1258


smarts/core/sensors.py
```diff
@@ -458,15 +458,16 @@ def _is_done_with_events(cls, sim, agent_id, vehicle, sensor_state):
         )
 
         done = (
-            (is_off_road and done_criteria.off_road)
+            (not sim._resetting) and
```
sah-huawei, line 461: It would  be better to use a non-private property here, i.e., to promote `_resetting` to `resetting` in the Smarts class.

sah-huawei (commented):
I'm not sure why the CI tests are failing.  It seems to be something different than #1230.  We can probably ignore that here, but it might be good for more than one of us to double-check that before we merge this.

Gamenot:
#1267 will need to be merged for the tests to succeed here.

smarts/core/sensors.py
```diff
@@ -458,15 +458,16 @@ def _is_done_with_events(cls, sim, agent_id, vehicle, sensor_state):
         )
 
         done = (
-            (is_off_road and done_criteria.off_road)
+            (not sim.resetting) and
```
Gamenot, line 461: nit:
```suggestion
            not sim.resetting and 
```

Gamenot (approved):

sah-huawei (approved):

smarts/core/smarts.py
```diff
@@ -954,7 +954,7 @@ def providers(self) -> List[Provider]:
         # TODO: It's inconsistent that pybullet is not here
         return self._providers
 
-    def get_provider_by_type(self, requested_type) -> Provider:
+    def get_provider_by_type(self, requested_type):
```
liu-allan, line 957: @Gamenot Can you look over these changes in `smarts.py`? This was to fix pytype errors when it would return `None` instead of `Provider`

sah-huawei, line 957: I would suggest this instead be:  `-> Optional[Provider]`

smarts/core/smarts.py
```diff
@@ -994,7 +994,7 @@ def _reset_providers(self):
             except Exception as provider_error:
                 self._handle_provider(provider, provider_error)
 
-    def _handle_provider(self, provider: Provider, provider_error) -> ProviderState:
+    def _handle_provider(self, provider: Provider, provider_error):
         provider_problem = bool(provider_error or not provider.connected)
         if not provider_problem:
             return
```
sah-huawei, line 1000: Same with for this one (`-> Optional[ProviderState]`).  But also, I think we actually should specify `return None` on line 1000.

smarts/core/smarts.py
```diff
@@ -1127,6 +1127,11 @@ def should_reset(self):
         """If the simulation requires a reset."""
         return self._reset_required
 
+    @property
+    def resetting(self):
```
Gamenot, line 1131: Nit since it is evident:
```suggestion
    def resetting(self) -> bool:
```

smarts/core/smarts.py
```diff
@@ -1380,7 +1385,7 @@ def _try_emit_envision_state(self, provider_state: ProviderState, obs, scores):
                 )
                 traffic[v.vehicle_id] = envision_types.TrafficActorState(
                     actor_type=envision_types.TrafficActorType.SocialVehicle,
-                    vehicle_type=veh_type,
+                    vehicle_type=envision_types.VehicleType(veh_type),
```
sah-huawei, line 1388: We will probably need to add a function to convert the types here.
However, I would suggest that it should be done in a different PR.

smarts/core/smarts.py
Gamenot: I have the same opinion as @sah-huawei

smarts/core/smarts.py
sah-huawei: (... the reason I suggest a different PR is because we've long needed to clean up our handling of vehicle types anyway and it's a bigger job than just this.)

liu-allan: Is what is here okay for now though?

liu-allan:
> #1267 will need to be merged for the tests to succeed here.

It works, 1267 should be merged before this

smarts/core/smarts.py
liu-allan: @sah-huawei can you take another look at this please?

envision/types.py
```diff
@@ -38,7 +38,7 @@ class VehicleType(str, Enum):
     Coach = "coach"
     Truck = "truck"
     Trailer = "trailer"
-    Car = "car"
+    Passenger = "passenger"
```
sah-huawei, line 41: This should be reverted too.  
Note that the list here in `envison/types` will not match the union of the two possible type lists that we currently support (the keys to the `VEHICLE_CONFIG` dict in `vehicle.py` and then the values of the `vehicle_type` fields in the `VehicleConfig` objects in that dict).

To fix this (in another PR), we will either need to add a function to map from the bigger list into the list of 5 things that envision supports, or (probably better, but non-trivial) unify across our vehicle typing schemes.

smarts/core/smarts.py
```diff
@@ -1350,7 +1355,7 @@ def _try_emit_envision_state(self, provider_state: ProviderState, obs, scores):
                 traffic[v.vehicle_id] = envision_types.TrafficActorState(
                     name=self._agent_manager.agent_name(agent_id),
                     actor_type=actor_type,
-                    vehicle_type=envision_types.VehicleType.Car,
+                    vehicle_type=envision_types.VehicleType.Passenger,
```
sah-huawei, line 1358: When the above is reverted, don't forget to revert this too.

smarts/core/smarts.py
sah-huawei: As described above, and as we just chatted about, getting this right is going to take more than this.  Thus, I think it's best to just revert this line here to what it was originally.  I will open another issue about fixing vehicle types in general.

sah-huawei (changes requested):
Everything is fine w.r.t. the original point of this PR, I just ask that you revert the changes related to vehicle types.

PR #1267, opened 2022-02-07
Update package

Adaickalavan:
Some dependency versions were updated to fix the failing CI.

1) Pip dependency resolver took extra long time in backtracking, causing CI to hang. Hence, latest package versions were specified.
2) I am unsure which core dependency requires the opencv package, but the latest opencv package causes the rllib test to fail. Therefore, `opencv-python==4.1.2.30` and `opencv-python-headless==4.1.2.30` were needed to pass the rllib test.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ setup(
             "pytest-xdist>=2.4.0",
         ],
         "train": [
-            "opencv-python==4.1.2.30",
+            "opencv-contrib-python-headless==4.1.2.30",
             "ray[rllib]==1.0.1.post1",
             "tensorflow>=2.4.0",
             "torch==1.4.0",
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ setup(
             "pytest-xdist>=2.4.0",
         ],
         "train": [
-            "opencv-python-headless==4.1.2.30",
+            "opencv-python==4.1.2.30",
             "ray[rllib]==1.0.1.post1",
             "tensorflow>=2.4.0",
             "torch==1.4.0",
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
         "protobuf>=3.19.1",
         "PyYAML>=6.0",
         "twisted>=21.7.0",
-        "opendrive2lanelet",
+        "opendrive2lanelet>=1.2.1",
     ],
     extras_require={
         "camera-obs": ["Panda3D==1.10.9", "panda3d-gltf==0.13"],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -76,17 +76,17 @@
...
             "pytest-cov>=3.0.0",
-            "pytest-notebook>=0.6.1",
+            "pytest-notebook>=0.7.0",
             "pytest-xdist>=2.4.0",
-            "ray[rllib]==1.0.1.post1",  # We use Ray for our multiprocessing needs
-            "tensorflow>=2.4.0",  # For rllib tests
         ],
         "train": [
+            "opencv-python==4.1.2.30",
+            "opencv-python-headless==4.1.2.30",
```
Gamenot, line 89: Do we really need both? Usually these would be exclusive of each other.

Edit:

Yes, it looks like it is exclusive. I would suggest we go with the headless version.

>>> Select the correct package for your environment:
>>>
>>> **There are four different packages (see options 1, 2, 3 and 4 below) and you should SELECT ONLY ONE OF THEM.** Do not install multiple different packages in the same environment. There is no plugin architecture: all the packages use the same namespace (cv2). If you installed multiple different packages in the same environment, uninstall them all with pip uninstall and reinstall only one package.
>>>
>>> a. Packages for standard desktop environments (Windows, macOS, almost any GNU/Linux distribution)
>>>
>>> Option 1 - Main modules package: pip install opencv-python
>>> Option 2 - Full package (contains both main modules and contrib/extra modules): pip install opencv-contrib-python (check contrib/extra modules listing from [OpenCV documentation](https://docs.opencv.org/master/))
b. Packages for server (headless) environments (such as Docker, cloud environments etc.), no GUI library dependencies
>>> 
>>> These packages are smaller than the two other packages above because they do not contain any GUI functionality (not compiled with Qt / other GUI components). This means that the packages avoid a heavy dependency chain to X11 libraries and you will have for example smaller Docker images as a result. You should always use these packages if you do not use cv2.imshow et al. or you are using some other package (such as PyQt) than OpenCV to create your GUI.
>>> 
>>> **Option 3 - Headless main modules package: pip install opencv-python-headless**
>>> Option 4 - Headless full package (contains both main modules and contrib/extra modules): pip install opencv-contrib-python-headless (check contrib/extra modules listing from [OpenCV documentation](https://docs.opencv.org/master/))

Gamenot (changes requested):

setup.py
Adaickalavan: Removed `opencv-python`.

Adaickalavan: Unfortunately, removing `opencv-python` brings back the original error causing rllib test to fail.
```
.venv/lib/python3.7/site-packages/cv2/__init__.py:9: in <module>
    from .cv2 import _registerMatType
E   ImportError: cannot import name '_registerMatType' from 'cv2.cv2' (/__w/SMARTS/SMARTS/.venv/lib/python3.7/site-packages/cv2/cv2.cpython-37m-x86_64-linux-gnu.so)
```

Gamenot: Odd, in theory we should only need one of them and we should not need the full package. It looks like RLlib has different plans.

Gamenot (approved):

liu-allan (approved):

PR #1274, opened 2022-02-11
Intersection-v0

Adaickalavan:
Standard left turn environment `intersection-v0` is added in `smarts.env.intersection_env.intersection_env`. 

A few things to consider:
1) An all-way stop is used for the traffic.
2) Image resolution for `rgb`, `ogm`, and `dagm`, is currently 0.39m/pixel = 100m/256pixel .
3) Endless traffic is set to `false`. Traffic end time is set to 3600s, which is longer than the maximum episode time of 300s.
4) The roads are of length: Left - 105m, North - 105m, Right - 50m, South - 50m.
5) Mission route: agent starts at 55m mark on Left road and ends at 50m mark on North road. Total travel distance should be around 100m.
6) `reward` per step is the distance travelled per step. 
7) `info["score"]` is the total distance travelled.
8)  To test the `intersection-v0`, run the temporarily added `examples.single_agent_test.py` file by executing `python3.7 ./examples/single_agent_test.py` command.
9)  Score calculation in smarts might need to be checked. The score (i.e., cumulative distance travelled) appears to exceed 100 when the agent has merely reached the intersection which is actually way before the expected 100m mark. This can be visualised by running the test file `python3.7 ./examples/single_agent_test.py`. The 100m mark should ideally be reached once the agent turns left and completes driving 50m into the North road.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 ### Added
 - Added `get_vehicle_start_time()` method for scenarios with traffic history data.  See Issue #1210.
 - Added `FormatObs` wrapper which converts SMARTS observations to gym-compliant RL-friendly vectorized observations and returns `StdObs`.
+- Added standard intersection environment, `intersection-v0`, for reinforcement learning where agents have to make a left turn in the presence of traffic.
 ### Changed
 - If more than one qualifying map file exists in a the `map_spec.source` folder, `get_road_map()` in `default_map_builder.py` will prefer to return the default files (`map.net.xml` or `map.xodr`) if they exist.
 - Moved the `smarts_ros` ROS node from the `examples` area into the `smarts.ros` module so that it can be distributed with SMARTS packages.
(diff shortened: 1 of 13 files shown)


liu-allan:
I noticed a bug where the episode ends with a collision 6 steps in. However upon viewing the simulation using sumo/envision, the ego vehicle does not appear to be colliding with anything at all

image

https://user-images.githubusercontent.com/44624435/155161454-c0be0b24-e4a8-4c0b-9e24-9bd0359ef4d4.mp4

Gamenot:
I am not sure why this is happening but it looks like a vehicle stalled: 
image

Gamenot:
It looks like the physics did not manage to catch up to the vehicle positions:

image

Gamenot:
Source of the problem appears to be because forces are not being applied(because no simulation update happens) until an ego takes control of a vehicle in the simulation.

https://github.com/huawei-noah/SMARTS/blob/aed3b2684101b728e08d61a5387fa354ce59f7be/smarts/core/smarts.py#L1049-L1063

Adaickalavan:
This pull request is being closed, as it is superseded by pull request #1427 .

Issue #1275, opened 2022-02-11
`Pose` attributes `position` and `orientation` do not have standard data format


Gamenot:
### Description of the Issue
Currently `Pose` does not constrain `position` and `orientation` to being numpy arrays so `all` is not an attribute of `tuple`.

### Requested solution
Force the `position` and `orientation` attributes to be numpy arrays. This may be possible through `__post_init__`

### More Context
https://github.com/huawei-noah/SMARTS/pull/1273

PR #1276, opened 2022-02-12
Fix pose data format

Gamenot:
- Force `Pose.position` to be a length 3 `np.array`.
- Force `Pose.orientation` to be a length 4 `np.array`.
- Fix various type issues

See #1275

### DIFF
--- a/smarts/core/coordinates.py
+++ b/smarts/core/coordinates.py
@@ -173,7 +173,7 @@ class Heading(float):
         float.__init__(value)
 
     def __new__(
-        self, x: Union[SupportsFloat, SupportsIndex, ellipsis] = ...
+        self, x: Union[SupportsFloat, SupportsIndex, Ellipsis] = ...
     ) -> "Heading":
         """A override to constrain heading to -pi to pi"""
         value = x
--- a/smarts/core/coordinates.py
+++ b/smarts/core/coordinates.py
@@ -173,7 +173,7 @@ class Heading(float):
         float.__init__(value)
 
     def __new__(
-        self, x: Union[SupportsFloat, SupportsIndex, Ellipsis] = ...
+        self, x: Union[SupportsFloat, SupportsIndex, Ellipsis.__class__] = ...
     ) -> "Heading":
         """A override to constrain heading to -pi to pi"""
         value = x
--- a/smarts/core/coordinates.py
+++ b/smarts/core/coordinates.py
@@ -255,7 +255,6 @@ class Heading(float):
 class Pose:
     """A pair of position and orientation values."""
 
-    # TODO: these should be np.ndarray
     position: np.ndarray  # [x, y, z]
     orientation: np.ndarray  # [a, b, c, d] -> a + bi + cj + dk = 0
     heading_: Optional[Heading] = None  # cached heading to avoid recomputing
--- a/smarts/core/coordinates.py
+++ b/smarts/core/coordinates.py
@@ -402,7 +402,7 @@ class Pose:
 
         return self.heading_
 
-    def as_position2(self) -> np.ndarray:
+    def as_position2d(self) -> np.ndarray:
         """Convert to a 2d position array"""
         return self.position[:2]
 
--- a/smarts/core/coordinates.py
+++ b/smarts/core/coordinates.py
@@ -176,7 +176,7 @@ class Heading(float):
 
     @classmethod
     def from_bullet(cls, bullet_heading):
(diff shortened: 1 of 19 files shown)


smarts/core/coordinates.py
```diff
@@ -258,12 +256,20 @@ class Pose:
...
-    orientation: Union[
-        NamedTuple, Tuple, np.ndarray
-    ]  # [a, b, c, d] -> a + bi + cj + dk = 0
+    position: np.ndarray  # [x, y, z]
+    orientation: np.ndarray  # [a, b, c, d] -> a + bi + cj + dk = 0
     heading_: Optional[Heading] = None  # cached heading to avoid recomputing
 
+    def __post_init__(self):
+        if not isinstance(self.position, np.ndarray):
+            self.position = np.array(self.position, dtype=np.float64)
```
RutvikGupta, line 265: You should use `dtype=float` since `np.float64` is deprecated in version of numpy.

smarts/core/utils/math.py
```diff
@@ -307,9 +307,6 @@ def position_at_offset(
     if is_close(dist, offset):
         return p2
 
-    if offset > dist:
-        return None
-
     return p1[0] + (p2[0] - p1[0]) * (offset / dist), p1[1] + (p2[1] - p1[1]) * (
```
RutvikGupta, line 310: Why was this removed?

smarts/core/tire_models.py
```diff
@@ -116,6 +116,10 @@ def _calculate_slip_angles(chassis, client, action):
 
         return slip_angles
 
+    def apply_tire_forces(self, chassis, client, action):
+        """Applies tire forces"""
+        raise NotImplementedError
```
RutvikGupta, line 121: Will this throw an error since it is being used in `smarts/core/chassis.py` but not implemented?

RutvikGupta (approved):

smarts/core/coordinates.py
```diff
@@ -258,12 +256,20 @@ class Pose:
     """A pair of position and orientation values."""
 
     # TODO: these should be np.ndarray
```
Adaickalavan, line 258: Can this `# TODO` be removed now, since change to np.ndarray has been made?

smarts/core/sumo_road_network.py
```diff
@@ -150,15 +150,19 @@ def from_spec(cls, map_spec: MapSpec):
         # Connections to internal lanes are implicit. If `withInternal=True` is
         # set internal junctions and the connections from internal lanes are
         # loaded into the network graph.
+        # pytype: disable=module-attr
         G = sumolib.net.readNet(net_file, withInternal=True)
+        # pytype: enable=module-attr
```
Adaickalavan, line 155:
```suggestion
        G = sumolib.net.readNet(net_file, withInternal=True)  # pytype: disable=module-attr
```
If disabling pytype for a single line, perhaps use a single line pytype disable comment.

smarts/core/sumo_road_network.py
```diff
@@ -150,15 +150,19 @@ def from_spec(cls, map_spec: MapSpec):
...
 
         if not cls._check_net_origin(G.getBoundary()):
             shifted_net_file = cls.shifted_net_file_path(net_file)
             if os.path.isfile(shifted_net_file) or (
                 map_spec.shift_to_origin
                 and cls._shift_coordinates(net_file, shifted_net_file)
             ):
+                # pytype: disable=module-attr
                 G = sumolib.net.readNet(shifted_net_file, withInternal=True)
+                # pytype: enable=module-attr
```
Adaickalavan, line 165: Similar suggestion, as previous, to use a single line pytype disable comment.
```suggestion
                G = sumolib.net.readNet(shifted_net_file, withInternal=True)  # pytype: disable=module-attr
```

Adaickalavan (approved):

smarts/core/sumo_road_network.py
Gamenot: I tried to do that but the `black` formatting caused issues.

smarts/core/utils/math.py
Gamenot: This case never gets hit, I looked through the code and found that if this case would get hit it would cause errors in the code since we always unpack the result without checking checking the truthy-ness.

smarts/core/coordinates.py
Gamenot: `np.float64` does not seem to be deprecated although `np.float` is.

Issue #1269, opened 2022-02-08
Convert string to vehicle types


liu-allan:
**Is your feature request related to a problem? Please describe.**
Currently strings are being used instead of `VehicleType`, so to support proper type usage, we need to have a way to convert types

**Describe the solution you'd like**
Function to convert types

**Describe alternatives you've considered**

**Additional context**

sah-huawei:
see Issue #1270 for a bit more detail.

Issue #1270, opened 2022-02-09
Unify vehicle typing schemes


sah-huawei:
Due to historical reasons, SMARTS currently must support two different vehicle type schemes (as evidenced by both the `veh_type` and  `veh_config_type` fields in the `VehicleState` class).

The "config" types index into the `VEHICLE_CONFIGS` dict in `vehicle.py` historically come from [Sumo vehicle classes](https://sumo.dlr.de/docs/Definition_of_Vehicles%2C_Vehicle_Types%2C_and_Routes.html#abstract_vehicle_class) and include:  passenger, bus, coach, truck, trailer, pedestrian, and motorcycle.  

Meanwhile, SMARTS attempted to support its own vehicle types in the set:  car, bus, coach, trailer, truck.  While overlapping with the Sumo types, these were meant to be independent of them and specific to SMARTS.  These were used in `envision/types.py` in the `VehicleType` enum and in Scenario Studio (`sstudio/types.py`) in the `TrafficActor` class as part of the DSL.

So the idea was to map between SMARTS types, which would be exposed to end users of SMARTS, and Sumo types internally when necessary to create a Sumo vehicle.  However, at some point (probably when the ROS interface was created), the Sumo vehicle ("config") types also got exposed to end users.

This should be rectified. 

Along the way, code like in `smarts.py` that passes vehicle types to Envision should be fixed to ensure that only types supported by Envision are passed to it.  So for example, we should not be passing `veh_type` directly to Envision,when `veh_type` is set as follows:
```python
                veh_type = (
                    v.vehicle_config_type if v.vehicle_config_type else v.vehicle_type
                )
 ```

sah-huawei:
Oops, accidentally duplicated issue #1269.

PR #1278, opened 2022-02-14
Convert vehicle types

liu-allan:
#1269

Added the other sumo vehicle types, but only add them to the traffic if they are currently supported by Envision

### DIFF
--- a/envision/types.py
+++ b/envision/types.py
@@ -39,6 +39,8 @@ class VehicleType(str, Enum):
     Truck = "truck"
     Trailer = "trailer"
     Car = "car"
+    Pedestrian = "pedestrian"
+    Motorcycle = "motorcycle"
 
 
 class TrafficActorState(NamedTuple):
--- a/envision/types.py
+++ b/envision/types.py
@@ -41,7 +41,7 @@ class VehicleType(str, Enum):
     Car = "car"
     Pedestrian = "pedestrian"
     Motorcycle = "motorcycle"
-    Passenger = "car"
+    Passenger = "passenger"
 
 
 class TrafficActorState(NamedTuple):
--- a/envision/types.py
+++ b/envision/types.py
@@ -39,9 +39,6 @@ class VehicleType(str, Enum):
     Truck = "truck"
     Trailer = "trailer"
     Car = "car"
-    Pedestrian = "pedestrian"
-    Motorcycle = "motorcycle"
-    Passenger = "passenger"
 
 
 class TrafficActorState(NamedTuple):
--- a/envision/types.py
+++ b/envision/types.py
@@ -39,6 +39,9 @@ class VehicleType(str, Enum):
     Truck = "truck"
     Trailer = "trailer"
     Car = "car"
+    Pedestrian = "pedestrian"
+    Motorcycle = "motorcycle"
+    Passenger = "car"
 
 
 class TrafficActorState(NamedTuple):
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -1385,13 +1385,17 @@ class SMARTS:
                 veh_type = (
                     v.vehicle_config_type if v.vehicle_config_type else v.vehicle_type
                 )
-                traffic[v.vehicle_id] = envision_types.TrafficActorState(
-                    actor_type=envision_types.TrafficActorType.SocialVehicle,
-                    vehicle_type=envision_types.VehicleType(veh_type),
-                    position=tuple(v.pose.position),
-                    heading=float(v.pose.heading),
-                    speed=v.speed,
-                )
+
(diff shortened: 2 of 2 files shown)


smarts/core/smarts.py
```diff
@@ -1385,13 +1385,17 @@ def _try_emit_envision_state(self, provider_state: ProviderState, obs, scores):
...
-                    actor_type=envision_types.TrafficActorType.SocialVehicle,
-                    vehicle_type=veh_type,
-                    position=tuple(v.pose.position),
-                    heading=float(v.pose.heading),
-                    speed=v.speed,
-                )
+
+                SUPPORTED_VEHICLES = {"car", "truck", "passenger", "bus", "trailer", "coach"}
+
+                if veh_type in SUPPORTED_VEHICLES:
```
Gamenot, line 1391: I am not entirely confident in skipping any default without some kind of logging. The best option would be to pass on the vehicle type even if it is wrong and then log it on envision's side.

liu-allan: how should we prevent the unsupported types from causing issues though? Currently passenger, motorcycle and pedestrian aren't supported

smarts/core/smarts.py
```diff
@@ -1385,13 +1385,21 @@ def _try_emit_envision_state(self, provider_state: ProviderState, obs, scores):
...
                 )
-                traffic[v.vehicle_id] = envision_types.TrafficActorState(
-                    actor_type=envision_types.TrafficActorType.SocialVehicle,
-                    vehicle_type=veh_type,
-                    position=tuple(v.pose.position),
-                    heading=float(v.pose.heading),
-                    speed=v.speed,
-                )
+
+                SUPPORTED_VEHICLES = {"car", "truck", "passenger", "bus", "trailer", "coach"}
```
sah-huawei, line 1389: Since this is about what _envision_ supports, this should be defined in `envision_types` and exported from there.  (The `SMARTS` class should not be declaring what the Envision tool supports.)

However, there is already a `VehicleType` enum that can be imported from `envision_types`.  We should just use that here.  I'm pretty sure you can check for membership in a python enum type, although I don't know how off the top of my head.

envision/types.py
```diff
@@ -39,6 +39,9 @@ class VehicleType(str, Enum):
     Truck = "truck"
     Trailer = "trailer"
     Car = "car"
+    Pedestrian = "pedestrian"
+    Motorcycle = "motorcycle"
+    Passenger = "passenger"
```
sah-huawei, line 44: What will Envision do with these if they're used?   Will it display them properly?

sah-huawei (commented):
This falls short of a more comprehensive and "unified" treatment of vehicle types that I think we need.  While I'm not necessarily opposed to something like this as a temporary solution to the Envision interface, I don't believe it fully satisfies my intent when I filed Issue #1270.  Maybe we need to have a meeting about this.

envision/types.py
liu-allan: No this won't work, this was a mistake on my end about how to go about solving the problem

smarts/core/smarts.py
```diff
@@ -1385,13 +1385,21 @@ def _try_emit_envision_state(self, provider_state: ProviderState, obs, scores):
...
-                    heading=float(v.pose.heading),
-                    speed=v.speed,
-                )
+
+                if veh_type == "passenger":
+                    veh_type = "car"
+
+                SUPPORTED_TYPES = set(item.value for item in envision_types.VehicleType)
+
+                if veh_type in SUPPORTED_TYPES:
```
sah-huawei, line 1394: Maybe we should log a warning message if `veh_type` is not in the `SUPPORTED_TYPES` list?  Something like `"not sending vehicle {v.vehicle_id} to Envision since it is of unsupported type {veh_type}"`...

sah-huawei (approved):
just one suggestion about adding a log message...

smarts/core/smarts.py
sah-huawei: ... but if so, it probably only needs to be logged once per `v.vehicle_id`, so I guess this is optional.

smarts/core/smarts.py
sah-huawei: Actually (although I just approved this PR), I agree with Tucker here.  I think this is best handled on the Envision side.   What I would do there is if type is unsupported (either "motorcycle" or "pedestrian"), I would log a warning (once per vehicle id) and then try to display a simple (small) "box" at the position specified.

liu-allan: After some digging it seems as though pedestrian and motorcycle are supported by envision after all:

https://github.com/huawei-noah/SMARTS/blob/develop/envision/web/src/render_helpers.js

Should I just add the missing types to `envision/types.py`?

sah-huawei: Yeah, I think so.

Gamenot (approved):

Issue #753, opened 2021-04-09
Remove Vehicle Type from `AgentInterface`


Gamenot:
**Is your feature request related to a problem? Please describe.**
It is not the case that the vehicle that the agent will take over is always the same as is provided in the agent interface. The pre-defined type in the agent interface then conflicts with the type of the assumed vehicle on capture.

**Describe the solution you'd like**
The vehicle type should be removed from the agent interface and the default (should there be no pre-existing vehicle to assume) should come through scenario studio. A scenario with no existing mission should just result in the default being `"sedan"`.

**Additional context**
Agent attempts to hijack always result in replacement of the target vehicle with a "sedan".

Gamenot:
This issue has morphed to some extent here.

Some discussion is needed here to get the right approach to this issue:
1. Vehicle acquiring issues:
  - AgentInterface requirements that the agent would need to be able to perform.
  - EntryTactic requirements in the scenario for the kind of vehicle that an agent will acquire.
  - How to negotiate conflicts between the two.
2. Specificity of the agent type:
  - How specific should the agent type be.
3. Incorporation of custom vehicle models 
  - How to supply `.urdf` to override vehicles
  - How to supply `.urdf` for a completely custom vehicle
4. Default selection of a vehicle type
  - How to handle cases where the vehicle type should have no default (e.g. for taking on vehicle from imitation learning.)
  - How to supply a default vehicle type.

PR #1294, opened 2022-02-23
Remove vehicle type from agent interface

liu-allan:
#753

### DIFF
--- a/smarts/core/controllers/__init__.py
+++ b/smarts/core/controllers/__init__.py
@@ -66,7 +66,6 @@ class Controllers:
         controller_state,
         sensor_state,
         action_space,
-        vehicle_type,
     ):
         """Calls control for the given vehicle based on a given action space and action.
         Args:
@@ -84,13 +83,9 @@ class Controllers:
                 The state of a vehicle sensor as relates to vehicle sensors.
             action_space:
                 The action space of the provided action.
-            vehicle_type:
-                Vehicle type information about the given vehicle.
         """
         if action is None:
             return
-        if vehicle_type == "bus":
-            assert action_space == ActionSpaceType.Trajectory
         if action_space == ActionSpaceType.Continuous:
             vehicle.control(
                 throttle=np.clip(action[0], 0.0, 1.0),
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -1239,7 +1239,6 @@ class SMARTS:
                     controller_state,
                     sensor_state,
                     agent_interface.action_space,
-                    agent_interface.vehicle_type,
                 )
 
     def _sync_vehicles_to_renderer(self):
--- a/smarts/core/vehicle.py
+++ b/smarts/core/vehicle.py
@@ -394,9 +394,14 @@ class Vehicle:
             SceneColors.Agent.value if trainable else SceneColors.SocialAgent.value
         )
 
-        if agent_interface.vehicle_type == "sedan":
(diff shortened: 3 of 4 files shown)


smarts/core/controllers/__init__.py
```diff
@@ -84,13 +83,9 @@ def perform_action(
                 The state of a vehicle sensor as relates to vehicle sensors.
             action_space:
                 The action space of the provided action.
-            vehicle_type:
-                Vehicle type information about the given vehicle.
         """
         if action is None:
             return
-        if vehicle_type == "bus":
```
liu-allan, line 92: Is this necessary? This was the only use of `vehicle_type` in this function so would a bus still be able to work properly?

smarts/core/vehicle.py
```diff
@@ -394,9 +394,14 @@ def build_agent_vehicle(
             SceneColors.Agent.value if trainable else SceneColors.SocialAgent.value
         )
 
-        if agent_interface.vehicle_type == "sedan":
+        if mission.vehicle_spec == None:
+            vehicle_type = "sedan"
+        else:
+            vehicle_type = mission.vehicle_spec.veh_config_type
```
liu-allan, line 400: is this the right place to get the vehicle type?

Gamenot: Yes, although I am slightly confused as to why the `vehicle_spec` is inside the mission rather than in the plan.

smarts/core/controllers/__init__.py
Gamenot: I doubt this necessary. I would also prefer to have fewer cases whenever possible.

smarts/core/smarts.py
```diff
@@ -1354,10 +1353,15 @@ def _try_emit_envision_state(self, provider_state: ProviderState, obs, scores):
...
+
+                veh_type = (
+                    v.vehicle_config_type if v.vehicle_config_type else v.vehicle_type
+                )
+                
                 traffic[v.vehicle_id] = envision_types.TrafficActorState(
                     name=self._agent_manager.agent_name(agent_id),
                     actor_type=actor_type,
-                    vehicle_type=envision_types.VehicleType.Car,
+                    vehicle_type=veh_type,
```
liu-allan, line 1364: This change seems to cause the ego vehicle to be the wrong color (grey instead of red) on SUMO maps

smarts/core/vehicle.py
liu-allan: where should `vehicle_spec` be specified then?

smarts/sstudio/types.py
```diff
@@ -464,6 +464,14 @@ class TrapEntryTactic(EntryTactic):
     """The speed that the vehicle starts at when the hijack limit expiry emits a new vehicle"""
 
 
+@dataclass(frozen=True)
+class VehicleSpec:
+    """Vehicle specifications"""
+
+    veh_id: str
+    veh_config_type: str
```
sah-huawei, line 472: This is problematic because we have been and are still distinguishing in our (core) code between `veh_type` and `veh_config_type`, where the latter should eventually  be fully encapsulated to just the Sumo types (per Issue #1270).  I do not think we should be exposing the "config" (Sumo) types in Scenario Studio, but to just change the variable name here to `veh_type` would also be a misnomer without further unification in the core.  Thus, I suggest we deal with this as part of the more comprehensive solution that issue #1270 requires.

PR #1295, opened 2022-02-23
Ghost collision

Adaickalavan:
This pull request addresses the ghost collision noted at 
+ https://github.com/huawei-noah/SMARTS/pull/1274#issuecomment-1047897642
+ https://github.com/huawei-noah/SMARTS/pull/1274#issuecomment-1048436471

### DIFF
--- a/smarts/core/lidar.py
+++ b/smarts/core/lidar.py
@@ -69,7 +69,7 @@ class Lidar:
                     self._sensor_params.noise_mu, self._sensor_params.noise_sigma
                 )
             )
-        return np.array(static_lidar_noise, dtype=np.float)
+        return np.array(static_lidar_noise, dtype=np.float64)
 
     def compute_point_cloud(
         self,
--- a/smarts/core/smarts.py
+++ b/smarts/core/smarts.py
@@ -898,15 +898,16 @@ class SMARTS:
                         vehicle_id=vehicle_id,
                         vehicle_config_type=vehicle.vehicle_config_type,
                     )
-                # Update the social vehicle avatar to match the vehicle state
+
+                # Update social vehicle pose when no active agents are present
+                if not self._vehicle_index.agent_vehicle_ids():
+                    social_vehicle.chassis.state_override(
+                        dt=dt, force_pose=vehicle.pose
+                    )
+
                 if not vehicle.updated:
-                    if self._resetting:
-                        social_vehicle.chassis.state_override(
-                            dt=dt, force_pose=vehicle.pose
-                        )
-                    else:
-                        # Note: update_state() happens *after* pybullet has been stepped.
-                        social_vehicle.update_state(vehicle, dt=dt)
+                    # Note: update_state() happens *after* pybullet has been stepped.
+                    social_vehicle.update_state(vehicle, dt=dt)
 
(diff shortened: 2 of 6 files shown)


Gamenot (approved):
@Adaickalavan and I looked into this together. :ship:

smarts/core/smarts.py
```diff
@@ -900,8 +900,13 @@ def _pybullet_provider_sync(self, provider_state: ProviderState):
                     )
                 # Update the social vehicle avatar to match the vehicle state
                 if not vehicle.updated:
-                    # Note:  update_state() happens *after* pybullet has been stepped.
-                    social_vehicle.update_state(vehicle, dt=dt)
+                    if self._resetting:
```
Gamenot, line 903: Actually, I now realise that there is still a potential problem here because the lack of physics update is caused by having no active agents. Just fixing this by updating during reset leaves open the possibility of it happening in the multi-agent case where agent become active in a way that leaves a gap in the scenario where there are no agent.

We might want this case to be based on there not being an active agent.

Adaickalavan: Changed `if self._resetting:` to `if not self._vehicle_index.agent_vehicle_ids():` .

Gamenot (approved):

smarts/core/smarts.py
```diff
@@ -898,9 +898,15 @@ def _pybullet_provider_sync(self, provider_state: ProviderState):
                         vehicle_id=vehicle_id,
                         vehicle_config_type=vehicle.vehicle_config_type,
                     )
-                # Update the social vehicle avatar to match the vehicle state
+
+                # Update social vehicle pose when no active agents are present
+                if not self._vehicle_index.agent_vehicle_ids():
```
Gamenot, line 903: I could suggest calling `agent_vehicle_ids()` once outside the loop since there is a significant cost to the method.

Gamenot (changes requested):

smarts/core/smarts.py
Adaickalavan: Moved `agent_vehicle_ids` to outside `for` loop.

Gamenot (approved):

smarts/core/lidar.py
```diff
@@ -69,7 +69,7 @@ def _compute_static_lidar_noise(self):
                     self._sensor_params.noise_mu, self._sensor_params.noise_sigma
                 )
             )
-        return np.array(static_lidar_noise, dtype=np.float)
+        return np.array(static_lidar_noise, dtype=np.float64)
```
RutvikGupta, line 72: I believe `np.float64` is deprecated in the newer versions of `numpy`. You can just use `float` here.

RutvikGupta (approved):

smarts/core/lidar.py
Adaickalavan: It appears that `np.float` is deprecated, and may be replaced with either `float` or the numpy scalar type name `np.float64`. Here, I simply preferred the explicit bit-width sized alias of `np.float64` .

Deprecation notes: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations .

Issue #1310, opened 2022-03-02
Restore Panda3D to optional dependency regression


sah-huawei:
## **BUG REPORT**
**High Level Description**
In previous versions, we made the import of the Panda3D dependency via `renderer.py` optional (only if camera sensors and the `[camera-obs]` extension are used).

However in a recent release, we accidentally added an unconditional import or `renderer` to `vehicle.py` which now imports Panda3D again, even if cameras are not being used.  (This was done in the PR related to docstrings.)  We need to fix this such that Panda3D is no longer required. 

**SMARTS version**
0.5.1

**Steps to reproduce the bug**
Don't install smarts with the `[camera-obs]` extensions and you will get an exception from the `renderer` import in `vehicle.py`.

PR #1318, opened 2022-03-04
Removed the Renderer import from vehicle.py

RutvikGupta:
See Issue #1310. Also restricted the `pytype` version to make it less aggressive.

### DIFF
--- a/smarts/core/vehicle.py
+++ b/smarts/core/vehicle.py
@@ -57,17 +57,17 @@ class VehicleState:
     vehicle_id: str
     pose: Pose
     dimensions: Dimensions
-    vehicle_type: str = None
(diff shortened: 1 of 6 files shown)


smarts/core/vehicle.py
```diff
@@ -249,8 +238,9 @@ def sensors(self) -> dict:
         return self._sensors
 
     @property
-    def renderer(self) -> Optional[Renderer]:
+    def renderer(self):  # type: ignore
```
Gamenot, line 241: Ideally here we would pull out a generalised `Renderer` interface and use that instead.

Gamenot (approved):
Only a comment, otherwise should be fine to go in.

Issue #1316, opened 2022-03-04
genhistories should account for vehicle angle when adjusting the center point


qianyi-sun:
## **BUG REPORT**
**High Level Description**
When the map is not aligned along the x-axis, the center-point correction for the vehicle length is incorrect.
It needs to account for the vehicle angle.

**SMARTS version**
0.5.1

sah-huawei:
Fixed with #1319

PR #1319, opened 2022-03-05
history position fix

sah-huawei:
This PR has two things in it:

1) A fix for a bug in `genhistories.py` with how it shifts vehicle positions from the front of the car to the center (Issue #1316).  This depends on heading, but the original code just assumed the car was oriented in the x-direction.  (This worked originally b/c the original i-80 map was oriented along the x-axis.)

2) Increasing the minimum version of `numpy` that's required in `setup.py` because one of the APIs we use in `genhistories.py` (`sliding_window_view()`) wasn't introduced until version `1.20`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
         "click>=8.0.3",  # used in scl
         "eclipse-sumo==1.10.0",  # sumo
         "gym==0.19.0",
-        "numpy>=1.20.0",
+        "numpy>=1.19.5",  # required for tf 2.4 below
         "pandas>=1.3.4",
         "psutil>=5.8.0",
         "pybullet==3.0.6",
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
         "click>=8.0.3",  # used in scl
         "eclipse-sumo==1.10.0",  # sumo
         "gym==0.19.0",
-        "numpy>=1.19.5",
+        "numpy>=1.20.0",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -27,7 +27,7 @@
         "click>=8.0.3",  # used in scl
         "eclipse-sumo==1.10.0",  # sumo
         "gym==0.19.0",
-        "numpy>=1.19.5",
+        "numpy>=1.20.0",
```
Gamenot, line 30: Why is this needed?

Gamenot (approved):

setup.py
sah-huawei: It's the second thing I wrote about that in the PR description:  we've been using `sliding_window_view()` in `genhistories.py` for a while and `numpy` has been in `setup.py` for a while without a minimum version.  But recently, we added a minimum version of `numpy` there which is too low such that, if you reinstall and get `1.19.5` (our stated minimum version) then it won't support `sliding_window_view()`.  I encountered this when trying to test the fix above for Patrick.

qianyi-sun (approved):

RutvikGupta (approved):
With `numpy` version set to 1.19.5 again, it looks ready to merge.

setup.py
sah-huawei: ok, due to issue #1321, I've reverted the `numpy` version to `1.19.5`.

Issue #1312, opened 2022-03-02
Add I-80 NGSIM Map


Gamenot:
**Describe the Problem**
Support NGSIM's I-80 map.

**Proposed Solution**
Add the I-80 map as a formal scenario resource.

- [x] Add map
- [x] Test dataset on map

@qianyi-sun This map already exists, part of this could be a first step towards #1309.

PR #1322, opened 2022-03-08
Add i-80 sumo map for corresponding NGSIM dataset

qianyi-sun:
linked issue: #1312

### DIFF
--- a/scenarios/NGSIM/README.md
+++ b/scenarios/NGSIM/README.md
@@ -9,7 +9,7 @@ SMARTS assumes that the data will be in tab-separated files.
 Its import scripts were tested on and work with the data downloaded from [here](http://bit.ly/PPUU-data),
 which is the same as was used by this [PPUU project](https://github.com/Atcold/pytorch-PPUU).
 
-After the datasets been downloaded, you will need to update each of the `.yaml` files
+After the dataset has been downloaded, you will need to update each of the `.yaml` files
 in the scenario folders to point the `input_path` field to its location on your filesystem.
 
 Once that is done, you should be able to build your scenarios in the normal way, for example:
--- a/scenarios/NGSIM/README.md
+++ b/scenarios/NGSIM/README.md
@@ -1,0 +1,31 @@
+# NGSIM traffic history scenarios
+
+The scenarios in the sub-folders here are based on the Next Generation Simulation (NGSIM) 
+dataset created by the US Department of Transportation (DOT) and described
+[here](https://ops.fhwa.dot.gov/trafficanalysistools/ngsim.htm).
+
+To use these scenarios, you must first download the dataset in an appropriate format.
+SMARTS assumes that the data will be in tab-separated files.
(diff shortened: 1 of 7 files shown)


scenarios/NGSIM/README.md
```diff
@@ -0,0 +1,31 @@
...
+The scenarios in the sub-folders here are based on the Next Generation Simulation (NGSIM) 
+dataset created by the US Department of Transportation (DOT) and described
+[here](https://ops.fhwa.dot.gov/trafficanalysistools/ngsim.htm).
+
+To use these scenarios, you must first download the dataset in an appropriate format.
+SMARTS assumes that the data will be in tab-separated files.
+Its import scripts were tested on and work with the data downloaded from [here](http://bit.ly/PPUU-data),
+which is the same as was used by this [PPUU project](https://github.com/Atcold/pytorch-PPUU).
+
+After the datas been downloaded, you will need to update each of the `.yaml` files
```
Gamenot, line 12:
```suggestion
After the datasets been downloaded, you will need to update each of the `.yaml` files
```

smarts/sstudio/genhistories.py
```diff
@@ -803,6 +803,10 @@ def _check_args(args) -> bool:
         with open(args.dataset, "r") as yf:
             dataset_spec = yaml.safe_load(yf)["trajectory_dataset"]
 
+    if dataset_spec.get("input_path") == "<PATH TO FILE GOES HERE>":
```
Gamenot, line 806: I would suggest a default here as well.
```suggestion
    default_input_path = "<PATH TO FILE GOES HERE>"
    if dataset_spec.get("input_path", default_input_path) == default_input_path:
```

smarts/sstudio/genhistories.py
```diff
@@ -803,6 +803,10 @@ def _check_args(args) -> bool:
...
+        print(f"skipping placeholder dataset spec at {args.dataset}.")
+        sys.exit(0)
```
Gamenot, line 808: I believe we should be more aggressive than this. The dataset is not useable without "input_path".

sah-huawei: This was added to effectively skip these scenarios in our CI, since we won't be pushing the datasets in our repo.  (These scenarios can only be built by a user who has downloaded the dataset to their local filesystem and then set their `input_path` accordingly.)  So it has to exit with a status of 0, otherwise CI will think that `scl scenario build-all` failed.  

What did you have in mind here?

sah-huawei (approved):

RutvikGupta (approved):

smarts/sstudio/genhistories.py
Gamenot: Ah, I see. I think my preference would be to exclude these scenarios from the `build-all`.

It is more of a nitpick than a requirement though.

Gamenot (approved):

Issue #1305, opened 2022-03-02
Make `scikit` Optional


Gamenot:
**Description of the Problem**
`scipy` packages `scikit-learn` and `scikit-image` are large packages.

**Proposed solution**
Remove the need for these dependencies.

RutvikGupta:
Something I came across which can be a nice replacement for the `sklearn.neighbors.KDTree` usage in `smarts/core/lanepoints.py`.: https://pypi.org/project/kdtree/.
Its a very light package and has no dependencies:
![image](https://user-images.githubusercontent.com/47334253/156819586-ce0b514a-a932-4a14-b006-dbf7cf7dc614.png)

Much better than writing our own KD Tree. Its implementation: https://github.com/stefankoegl/kdtree/blob/master/kdtree.py

PR #1325, opened 2022-03-08
Make `scikit` Optional

liu-allan:
#1305


ultra/ultra/utils/common.py
```diff
@@ -145,7 +144,7 @@ def copy_source_code(srcs, dst):
 
 
 def resize_im(im, size):
```
liu-allan, line 146: What is the type for `size` here? This is the only location of `resize_im` in all of SMARTS

ultra/ultra/utils/common.py
```diff
@@ -145,7 +144,7 @@ def copy_source_code(srcs, dst):
...
-    im = (resize(im, size, anti_aliasing=True, mode="constant") * 255.0).astype(
+    im = (cv2.resize(im, size) * 255.0).astype(
```
liu-allan, line 147: How important is anti-aliasing? Should I add a blur to mimic this here?

smarts/core/lanepoints.py
```diff
@@ -527,9 +519,10 @@ def _closest_linked_lp_in_kd_tree_batched(
         points, linked_lps, tree: KDTree, k: int = 1
     ):
         p2ds = np.array([vec_2d(p) for p in points])
-        closest_indices = tree.query(
-            p2ds, k=min(k, len(linked_lps)), return_distance=False, sort_results=True
+        _, closest_indices = tree.query(
```
liu-allan, line 522: Removing `scipy` entirely may be difficult. https://github.com/stefankoegl/kdtree This implementation can only find the nearest node instead of nearest k

ultra/ultra/utils/common.py
Gamenot: `cv2.resize` has anti-aliasing integrated. `resize(..,interpolation=..)` works with the following:
- 0 INTER_NEAREST
- 1 INTER_LINEAR
- 2 INTER_AREA
- 3 INTER_CUBIC
- 4 INTER_LANCZOS4

smarts/core/lanepoints.py
RutvikGupta, line 522: image

The distance return by this method is deprecated and is removed in the latest version of `scipy`. We should be careful here whether we want to upgrade to latest version of scipy to avoid the distance return or lock the version if we have a dependency requirement.

Gamenot: We could also conditionally unpack. Although that might be un-optimal. It might also be possible to determine that condition once by wrapping the query method.

liu-allan: Only the condition where k=None which is deprecated. This doesn't affect our use so I don't think we need to worry too much about locking the dependency version. See https://docs.scipy.org/doc/scipy-1.5.4/reference/generated/scipy.spatial.KDTree.query.html for an older version

ultra/ultra/utils/common.py
Gamenot: I am honestly unsure what the implications are.

Gamenot: I am honestly unsure what the implications are.

RutvikGupta (approved):

Gamenot (approved):

PR #1327, opened 2022-03-08
support for extra columns in NGSIM data files

sah-huawei:
There appear to be two NGSIM datafile schemas:  one used by `i80` and `us101` that has 18 columns and one used by `peach` and `lanker` which has 6 extra columns (24 total).  All of the other columns seem to be the same.  Annoyingly though, the 6 extra columns are not just appended to the end of the row, but rather inserted into the middle of it!

If you try to read the latter into a data frame, it screws things up and columns are misaligned.  So this adds an "NGSIM2" type to the dataspec yaml for the `peach` and `lanker` cases to read these files with the extra columns.

### DIFF
--- a/smarts/sstudio/genhistories.py
+++ b/smarts/sstudio/genhistories.py
@@ -85,7 +85,7 @@ class _TrajectoryDataset:
         if "input_path" not in dataset_spec:
             errmsg = "'input_path' field is required in dataset yaml."
         elif dataset_spec.get("flip_y"):
-            if dataset_spec.get("source") != "NGSIM":
+            if not dataset_spec.get("source").startswith("NGSIM"):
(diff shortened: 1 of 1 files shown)


qianyi-sun (approved):

sah-huawei:
> ... Maybe this should be documented somewhere so that users set the source type to be NGSIM2 for other two datasets.

In general I'd agree with you, but in this case we are adding `.yaml` files (and maps) to our repo for at least 3 of the 4 NGSIM datasets, and possibly even `lanker` as well (although it's lower priority for us).  So there should be no need  for our users to have to create NGSIM `.yaml` files themselves, thus they shouldn't need to know about NGSIM2.  (I.e., it's enough that Patrick does.)

Issue #1335, opened 2022-03-09
Switch `pytype` test to added


Gamenot:
**Is your feature request related to a problem? Please describe.**
`pytype` tests are aggressively failing for changed files.

**Describe the solution you'd like**
Switch to track added rather than changed files for now.

PR #1337, opened 2022-03-10
PyType on branch only

sah-huawei:
Per our discussions from earlier Wednesday, this change constrains the files on which `pytype` will be run during CI to those that have been changed on the branch being merged since it was branched.

(So, for example, files changed and independently merged into `develop` in the meantime will not be checked if they were not changed on this branch.)

May address Issue #1335.

### DIFF
--- a/.github/workflows/ci-format.yml
+++ b/.github/workflows/ci-format.yml
@@ -52,11 +52,13 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Install pytest
+      - name: Install SMARTS
         run: |
           python3.7 -m venv ${{env.venv_dir}}
           . ${{env.venv_dir}}/bin/activate
           pip install --upgrade pip wheel
+          pip install .[camera-obs,dev,extras,opendrive,ros,test,train]
+          # install the specific version we want for pytest, just in case...
           pip install pytype==2022.1.13
       - name: Get changed files on branch since branching
         id: changed-files
--- a/.github/workflows/ci-format.yml
+++ b/.github/workflows/ci-format.yml
@@ -52,19 +52,22 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Install SMARTS
+      - name: Install pytest
         run: |
           python3.7 -m venv ${{env.venv_dir}}
           . ${{env.venv_dir}}/bin/activate
           pip install --upgrade pip wheel
-          pip install .[camera-obs,dev,extras,opendrive,ros,test,train]
-      - name: Get changed files
+          pip install pytype==2022.1.13
+      - name: Get changed files on branch since branching
         id: changed-files
-        uses: tj-actions/changed-files@v13.2  
-        with:
-          base_sha: origin/develop     
+        shell: bash
+        run: |
+          cd $GITHUB_WORKSPACE
+          CHANGED=$(git diff --diff-filter="AM" --name-only --ignore-submodules=all origin/develop... | awk -v d=" " '{s=(NR==1?s:s d)$0}END{print s}')
+          echo "::set-output name=changed_files::$CHANGED"
       - name: Check types
-        if: contains(steps.changed-files.outputs.all_changed_files, '.py')
+        if: contains(steps.changed-files.outputs.changed_files, '.py')
         run: |
           . ${{env.venv_dir}}/bin/activate
-          pytype --disable=pyi-error ${{steps.changed-files.outputs.all_changed_files}}
+          pytype --disable=pyi-error ${{steps.changed-files.outputs.changed_files}}
+
--- a/.github/workflows/ci-format.yml
+++ b/.github/workflows/ci-format.yml
@@ -57,9 +57,7 @@ jobs:
           python3.7 -m venv ${{env.venv_dir}}
           . ${{env.venv_dir}}/bin/activate
           pip install --upgrade pip wheel
-          pip install .[camera-obs,dev,extras,opendrive,ros,test,train]
-          # install the specific version we want for pytest, just in case...
-          pip install pytype==2022.1.13
+          pip install .[dev]
       - name: Get changed files on branch since branching
         id: changed-files
         shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-format.yml
```diff
@@ -52,19 +52,22 @@ jobs:
...
           fetch-depth: 0
-      - name: Install SMARTS
+      - name: Install pytest
         run: |
           python3.7 -m venv ${{env.venv_dir}}
           . ${{env.venv_dir}}/bin/activate
           pip install --upgrade pip wheel
-          pip install .[camera-obs,dev,extras,opendrive,ros,test,train]
-      - name: Get changed files
+          pip install pytype==2022.1.13
```
RutvikGupta, line 60: Why wont not installing the packages on this runner throw import errors? I remember seeing error for `waymo` packages since we were not installing its dependencies here and was forced to ignore pytype check on files where it was imported.

.github/workflows/ci-format.yml
```diff
@@ -52,19 +52,22 @@ jobs:
...
+      - name: Get changed files on branch since branching
         id: changed-files
-        uses: tj-actions/changed-files@v13.2  
-        with:
-          base_sha: origin/develop     
+        shell: bash
+        run: |
+          cd $GITHUB_WORKSPACE
+          CHANGED=$(git diff --diff-filter="AM" --name-only --ignore-submodules=all origin/develop... | awk -v d=" " '{s=(NR==1?s:s d)$0}END{print s}')
```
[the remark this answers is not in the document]
sah-huawei: It's basically replacing newlines with spaces.  (There are a lot of ways of doing this in bash, but this seems to be the preferred way for other people writing github workflow rules.)  The `git diff` commend outputs the names of the files on separate lines but we need them in a single string (separated by spaces) to assign them to the `changed_files` output variable and to use them in the next step on a command line with `pytype`.

.github/workflows/ci-format.yml
sah-huawei: I don't think so, but I'm not 100% sure.  I think what matters most is being in the right "root folder" (in this case, `$GITHUB_WORKSPACE`) such that the `smarts` module references resolve correctly in python.  That is, being in the folder where the repo has been cloned (in the checkout step).  

I tried to test this locally using a tool called `act` before pushing, so I'm fairly confident it won't, but I'm not at 100%.  However, if it does fail on github because of that, we can add that back in.

Having to install all of our dependencies for all of our extensions like `[ros]` and `[waymo]` was exactly what I was trying to avoid.  I'd rather have `pytype` skip checking these (using the "ignore" comments that Adai demonstrated yesterday) than deal with having to diagnose issues that perhaps exist in their interfaces.  However, I don't feel to strongly about that, so I'm willing to add the `pip install .[...]` back into the "install" step if you guys feel it necessary.  

I guess like this, we are probably going to need to add more "ignore" comments to some of our import statements (which I would prefer, longer term, but don't necessarily want to force with this PR), so now I'm thinking I should probably add it back in, at least for the time being.  I guess I'll do that now.

Thanks for asking about this!

RutvikGupta (approved):

.github/workflows/ci-format.yml
```diff
@@ -58,13 +58,18 @@ jobs:
           . ${{env.venv_dir}}/bin/activate
           pip install --upgrade pip wheel
           pip install .[camera-obs,dev,extras,opendrive,ros,test,train]
-      - name: Get changed files
+          # install the specific version we want for pytest, just in case...
+          pip install pytype==2022.1.13
```
Gamenot, line 62: nit:I would prefer to not have this specific install since it disconnects the test from what is listed in `setup.py`.

```suggestion
```

.github/workflows/ci-format.yml
```diff
@@ -58,13 +58,18 @@ jobs:
...
+      - name: Get changed files on branch since branching
         id: changed-files
-        uses: tj-actions/changed-files@v13.2  
-        with:
-          base_sha: origin/develop     
+        shell: bash
+        run: |
+          cd $GITHUB_WORKSPACE
+          CHANGED=$(git diff --diff-filter="AM" --name-only --ignore-submodules=all origin/develop... | awk -v d=" " '{s=(NR==1?s:s d)$0}END{print s}')
```
Gamenot, line 68: Interesting, I would guess the ellipsis after a branch means the direct parent node?

Gamenot (approved):

.github/workflows/ci-format.yml
```diff
@@ -58,13 +58,18 @@ jobs:
...
+          echo "::set-output name=changed_files::$CHANGED"
       - name: Check types
-        if: contains(steps.changed-files.outputs.all_changed_files, '.py')
+        if: contains(steps.changed-files.outputs.changed_files, '.py')
         run: |
           . ${{env.venv_dir}}/bin/activate
-          pytype --disable=pyi-error ${{steps.changed-files.outputs.all_changed_files}}
\ No newline at end of file
+          pytype --disable=pyi-error ${{steps.changed-files.outputs.changed_files}}
```
Adaickalavan, line 74:
```suggestion
          pytype -d pyi-error import-error ${{steps.changed-files.outputs.changed_files}}
```

With this change:
+ All third party imported libraries will be ignored.
+ Third party libraries do not need to be ignored individually in each file. 
+ Additionally, the time cosuming `pip install .[camera-obs,dev,extras,opendrive,ros,test,train]` command above can be changed to the simpler `pip install .[dev]` . Option `dev` was kept so that we use the same pytype version from setup.py file.

Not sure if there are any undesirable side-effects which arises from completely ignoring all import-errors.

.github/workflows/ci-format.yml
sah-huawei: The "unfinished ellipsis" is just short for `...HEAD`, i.e., comparing to the head of the current branch.

The key reason for using the triple-dot ellipsis here is so we compare the PR branch to `develop` "at the branch point".  I.e., we are ignoring anything that's happened on the `develop` branch since we branched, so the only files that should show up in this diff are things that were changed in the PR, _not_ things that were changed in the meantime on `develop`.

.github/workflows/ci-format.yml
sah-huawei: My thought here was to allow us to use a newer, possibly-more-restrictive version of `pytype` in our development and local test environments if we want, while keeping the version in CI pinned to something we know is "sane".

However,  I think you're right this might be confusing to people if, at some point, newer versions of `pytype` become _less_ restrictive for whatever reason.  Then they will see things passing locally, but not in CI, which might confuse them.  So I'll take your suggestion.

.github/workflows/ci-format.yml
sah-huawei: Nice!  Thanks for the suggestion.  I'll make this change now.

PR #1340, opened 2022-03-11
Type fix in zoo

Adaickalavan:
+ Minor type fix in zoo.
+ Update dependency version to reduce pip backtracking.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,10 +31,10 @@ setup(
         "pandas>=1.3.4",
         "psutil>=5.8.0",
         "pybullet==3.0.6",
-        "rich>=10.13.0",
-        "rtree>=0.9.7",  # Used by sumolib
+        "rich>=11.2.0",
+        "Rtree>=0.9.7",  # Used by sumolib
         "sh>=1.14.2",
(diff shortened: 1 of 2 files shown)


RutvikGupta (approved):

setup.py
```diff
@@ -31,10 +31,10 @@
         "pandas>=1.3.4",
         "psutil>=5.8.0",
         "pybullet==3.0.6",
-        "rich>=10.13.0",
-        "rtree>=0.9.7",  # Used by sumolib
+        "rich>=11.2.0",
+        "Rtree>=0.9.7",  # Used by sumolib
```
Gamenot, line 35: Does sumolib need this anymore?

Gamenot (approved):

setup.py
Adaickalavan: `Rtree` appears to be used in `opendrive_road_network.py` here:
https://github.com/huawei-noah/SMARTS/blob/873c75e8a71c215f9fae1219795fe64cfdb65a3f/smarts/core/opendrive_road_network.py#L32

PR #1346, opened 2022-03-15
Add sumo vtype override

Gamenot:
This will allow overriding `<vType/>` params in `sumo` traffic generation. Leaves open having other overrides for later uses of `TrafficActor` with other traffic generators.

### DIFF
--- a/smarts/sstudio/types.py
+++ b/smarts/sstudio/types.py
@@ -49,6 +49,7 @@ from smarts.core.utils.math import rotate_around_point
 class _SUMO_PARAMS_MODE(IntEnum):
     TITLE_CASE = 0
     KEEP_SNAKE_CASE = 1
+    SAME_AS_SUMO = 2
 
 
 def _pickle_hash(obj) -> int:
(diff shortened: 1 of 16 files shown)


smarts/sstudio/types.py
```diff
@@ -108,16 +112,31 @@ class LaneChangingModel(_SumoParams):
     """Models how the actor acts with respect to lane changes."""
 
     def __init__(self, **kwargs):
+        warnings.warn(
+            f"The LaneChangingModel has has replaced by {SumoVTypeOverride.__name__}. Instantiating this object type is deprecated.",
```
sah-huawei, line 116:
```suggestion
            f"The LaneChangingModel has been replaced by {SumoVTypeOverride.__name__}. Instantiating this object type is deprecated.",
```

smarts/sstudio/types.py
```diff
@@ -108,16 +112,31 @@ class LaneChangingModel(_SumoParams):
...
+            DeprecationWarning,
+        )
```
sah-huawei, line 118: As you know, I used the `LaneChangingModel` in the `smarts_traffic` branch to allow setting parameters related to cut-ins.  So if you do make the change suggested here, I would rename `SumoVTypeOverride` to just be `VTypeOverride`.  However, I'm not sure even that is general enough because it assumes something called "vtype", which may not always be the case.  (For one thing, note that these apply to traffic _actors_, not just _vehicles_.)

Gamenot:
Closed because we do not need to inject more of SUMO into SMARTS.

Issue #1336, opened 2022-03-09
`pytype` Fails Typecheck on Repository


Gamenot:
**Is your feature request related to a problem? Please describe.**
`pytype` fails when running against the smarts repository.

```sh
# e.g.
pytype --disable=pyi-error smarts envision cli ultra zoo
```

**Describe the solution you'd like**
Run `pytype` across the repository and fix the type issues. Restore the `pytype` test to watch for changed files.

PR #1348, opened 2022-03-18
Fix `pytype` errors

liu-allan:
#1336


saulfield:
An issue came up in the CI run for another PR that might be relevant to this one: https://github.com/huawei-noah/SMARTS/runs/5603527968?check_suite_focus=true

It looks like different modules with the same name cause the pytype build step to fail. There is an open issue for this here: https://github.com/google/pytype/issues/198

In the meantime, we discussed this and decided it makes sense to try to ignore this if we can while a fix is incoming. At first we thought we could ignore the `dupbuild` error it references, but now I realize it's from the build system that pytype uses (ninja). Perhaps we should ignore files with duplicate names when we invoke pytype? (e.g. `scenarios/**/agent_prefabs.py`)

liu-allan:
> An issue came up in the CI run for another PR that might be relevant to this one: https://github.com/huawei-noah/SMARTS/runs/5603527968?check_suite_focus=true
> 
> It looks like different modules with the same name cause the pytype build step to fail. There is an open issue for this here: [google/pytype#198](https://github.com/google/pytype/issues/198)
> 
> In the meantime, we discussed this and decided it makes sense to try to ignore this if we can while a fix is incoming. At first we thought we could ignore the `dupbuild` error it references, but now I realize it's from the build system that pytype uses (ninja). Perhaps we should ignore files with duplicate names when we invoke pytype? (e.g. `scenarios/**/agent_prefabs.py`)

Ignoring files with duplicate names will only work in certain situations. This error also occurs on duplicate filenames that were not modified but instead are inferred during the test. See (https://github.com/huawei-noah/SMARTS/runs/5632808654?check_suite_focus=true) for this pr, where only one `utils.py` is modified on this branch yet the error still triggers due to other `utils.py` being referenced in other modified files.

smarts/core/vehicle.py
```diff
@@ -631,11 +633,13 @@ def teardown(self, exclude_chassis=False):
         for sensor in [
             sensor
             for sensor in [
+                # pytype: disable=attribute-error
                 self._drivable_area_grid_map_sensor,
                 self._ogm_sensor,
                 self._rgb_sensor,
                 self._lidar_sensor,
                 self._driven_path_sensor,
+                # pytype: enable=attribute-error
```
Gamenot, line 642: I think for these we should deal with the dynamic attributes this way:  https://github.com/google/pytype/blob/main/docs/faq.md#im-dynamically-populating-a-class--module-using-setattr-or-by-modifying-locals--globals-now-pytype-complains-about-missing-attributes-or-module-members-how-do-i-fix-this

zoo/evaluation/metrics/diversity/utils.py
```diff
@@ -768,7 +768,9 @@ def eval_diversity(pos_ego, pos_agent, speed_ego, speed_agent, time_ego, time_ag
             pos_ego, pos_agent, speed_ego, speed_agent, time_ego, time_agent
         )
         pos_list1, pos_list2 = count_intersection_area(pos_ego, pos_agent)
-    A1, B1, A2, B2, P1, P2 = count_intersection_segment(pos_list1, pos_list2)
+    # pytype: disable=attribute-error
+    A1, B1, A2, B2, P1, P2 = count_intersection_segment(pos_list1, pos_list2) 
+    # pytype: enable=attribute-error
```
Gamenot, line 773: What is this attribute-error from?

cli/ultra.py
```diff
@@ -69,10 +69,11 @@ def build():
 @click.argument("auth_key", type=str, default=None)
 @click.argument("port", default=7432, type=int)
 def worker(auth_key, port):
-    from smarts.zoo.worker import listen
+    from smarts.zoo.worker import serve
 
     auth_key = auth_key if auth_key else ""
-    listen(port, auth_key)
+    # listen() does not exist anymore, does serve() provide the same function?
+    serve(port)
```
Gamenot, line 76: @Adaickalavan 

Just be careful here since I believe ULTRA is using a different version of SMARTS than the current version. We may want to provide the following:

```suggestion
    auth_key = auth_key if auth_key else ""
    try:
        from smarts.zoo.worker import listen
        listen(port, auth_key)
    except ImportError:
        from smarts.zoo.worker import serve
        serve(port)
```

envision/server.py
```diff
@@ -421,6 +422,7 @@ class ModelFileHandler(FileHandler):
     def initialize(self):
         # We store the resource filenames as values in `path_map` and route them
         # through `importlib.resources` for resolution.
+        # pytype: disable=wrong-arg-types
```
Gamenot, line 425: This should be possible to fix by changing the signature of the base to the following: `FileHandler.initialize(path_map: Dict[str,Union[str, Path]]`.

Gamenot (changes requested):

smarts/core/agent_manager.py
```diff
@@ -233,7 +235,9 @@ def observe(
             dones = {agent_id: True for agent_id in self.agent_ids}
             dones["__sim__"] = True
 
-        return observations, rewards, scores, dones
+        # pytype: disable=bad-return-type
+        return observations, rewards, scores, dones 
+        # pytype: enable=bad-return-type
```
Gamenot, line 240: This is alarming, what is the cause here?

smarts/core/bubble_manager.py
```diff
@@ -475,7 +477,7 @@ def _move_travelling_bubbles(self, sim):
 
     def _airlock_social_vehicle_with_social_agent(
         self, sim, vehicle_id: str, social_agent_actor: SocialAgentActor, bubble: Bubble
-    ) -> str:
+    ) -> None:
```
Gamenot, line 480:
```suggestion
    ):
```

smarts/core/bubble_manager.py
```diff
@@ -508,15 +510,15 @@ def _airlock_social_vehicle_with_social_agent(
...
+            return 
 
         self._start_social_agent(
             sim, agent_id, social_agent, social_agent_actor, bubble
         )
 
     def _hijack_social_vehicle_with_social_agent(
         self, sim, vehicle_id: str, social_agent_actor: SocialAgentActor, bubble: Bubble
-    ) -> str:
+    ) -> None:
```
Gamenot, line 521:
```suggestion
    ):
```

smarts/core/bubble_manager.py
```diff
@@ -618,7 +620,7 @@ def _make_social_agent_id(vehicle_id):
 
     @staticmethod
     def _make_boid_social_agent_id(social_agent_actor):
-        return f"BUBBLE-AGENT-{truncate(social_agent_actor.id, 48)}"
+        return f"BUBBLE-AGENT-{truncate(social_agent_actor.name, 48)}"
```
Gamenot, line 623: What is the reason for changing this? I am concerned it may have a major affect on the agent ids.

smarts/core/external_provider.py
```diff
@@ -52,7 +52,10 @@ def state_update(
 
     @property
     def action_spaces(self) -> Set[ActionSpaceType]:
+        # What should be here?
+        # pytype: disable=bad-return-type
         return {}
+        # pytype: enable=bad-return-type
```
Gamenot, line 58: Ah, it confused me for a second but this is a dictionary playing pretend.

```suggestion
        return set()
```

smarts/core/scenario.py
```diff
@@ -397,8 +397,9 @@ def setdefault(l: Sequence[Any], index: int, default):
             social_agents_info.append(
                 {agent.id: (agent, mission) for agent, mission in l}
             )
-
+        # pytype: disable=bad-return-type
         return social_agents_info
+        # pytype: enable=bad-return-type
```
Gamenot, line 402: I believe this could be fixed by changing the return signature to `Sequence[Dict[str, Tuple[SocialAgent, Mission]]].

Gamenot (changes requested):

smarts/core/sumo_traffic_simulation.py
```diff
@@ -206,8 +206,10 @@ def _initialize_traci_conn(self, num_retries=5):
 
         try:
             # It is mandatory to set order when using multiple clients.
+            # pytype: disable=attribute-error
             self._traci_conn.setOrder(0)
             self._traci_conn.getVersion()
+            # pytype: enable=attribute-error
```
Gamenot, line 212: I think the type of `self._traci_conn` is `Optional[traci.connection.Connection]`.

smarts/core/tests/test_notebook.py
```diff
@@ -26,7 +26,9 @@
 import gym
 import importlib_resources
 import pytest
+# pytype: disable=import-error
 import pytest_notebook.nb_regression as nb
+# pytype: enable=import-error
```
Gamenot, line 31: I think this can be removed by `pip install smarts[test]` in the type test GitHub action.

smarts/core/tests/test_trajectory_interpolation_provider.py
```diff
@@ -300,10 +300,12 @@ def test_trajectory_interpolation_provider_in_smarts(smarts, agent_spec, scenari
         if agent_obs.events.reached_goal:
             reached_goal = True
             break
-
+    
+    # pytype: disable=attribute-error
     curr_position = agent_obs.ego_vehicle_state.position
     curr_heading = agent_obs.ego_vehicle_state.heading
     curr_speed = agent_obs.ego_vehicle_state.speed
+    # pytype: enable=attribute-error
```
Gamenot, line 308: This seems odd to me that we would get an `attribute-error` here, it means that our observation object is not clearly defined.

smarts/core/traffic_history_provider.py
```diff
@@ -87,7 +87,11 @@ def teardown(self):
 
     @property
     def action_spaces(self) -> Set[ActionSpaceType]:
+        # What should be here?
+        # pytype: disable=bad-return-type
         return {}
+        # pytype: enable=bad-return-type
+
```
Gamenot, line 94: Another larping dictionary.

```suggestion
        return set()
```

smarts/core/vehicle.py
```diff
@@ -265,7 +266,9 @@ def state(self) -> VehicleState:
             pose=self.pose,
             dimensions=self._chassis.dimensions,
             speed=self.speed,
+            # pytype: disable=attribute-error
             steering=self._chassis.steering,
+            # pytype: enable=attribute-error
```
Gamenot, line 271: This means that the `Chassis` interface is not well defined. I could imagine steering  should be a property of the `Chassis` abstract class.

smarts/core/tests/test_notebook.py
Adaickalavan: 1. No need for individual `pytype: disable=import-error` commands anymore as we have ignored `import-error` globally during pytest in Github CI.
https://github.com/huawei-noah/SMARTS/blob/3ae9f8479d5ebf81f9a79e52d2eb94332f6c7acd/.github/workflows/ci-format.yml#L72  
2. Thus, I think no need to include `test` option while installing smarts in Github CI.

smarts/core/renderer.py
```diff
@@ -30,6 +30,7 @@
 
 import gltf
 from direct.showbase.ShowBase import ShowBase
+# pytype: disable=import-error
```
Adaickalavan, line 33:
```suggestion
```

smarts/core/renderer.py
```diff
@@ -41,6 +42,7 @@
     WindowProperties,
     loadPrcFileData,
 )
+# pytype: enable=import-error
```
Adaickalavan, line 45:
```suggestion
```

smarts/core/bubble_manager.py
liu-allan: `SocialAgentActor` from smarts/sstudio/types does not have an `id` attribute, only `name`

zoo/policies/open-agent/open_agent/agent.py
```diff
@@ -6,6 +6,7 @@
 from pathlib import Path
 from typing import Sequence
 
+# pytype: disable=import-error
```
Adaickalavan, line 9:
```suggestion
```

smarts/sstudio/genhistories.py
```diff
@@ -38,7 +38,7 @@
 from smarts.core.utils.math import vec_to_radians
 
 try:
-    from waymo_open_dataset.protos import scenario_pb2
+    from waymo_open_dataset.protos import scenario_pb2 # pytype: disable=import-error
```
Adaickalavan, line 41:
```suggestion
    from waymo_open_dataset.protos import scenario_pb2
```

smarts/core/tests/test_trajectory_interpolation_provider.py
liu-allan: It thinks that `agent_obs` is `None`

smarts/core/vehicle.py
liu-allan: `steering` is a property of the `Chassis` abstract class which I am not sure why pytype does not recognize. https://github.com/huawei-noah/SMARTS/blob/652aa23e71bd4e2732e2742140cfcd0ec082a7da/smarts/core/chassis.py#L134

smarts/core/vehicle_index.py
```diff
@@ -462,7 +462,9 @@ def stop_agent_observation(self, vehicle_id):
         vehicle_id = _2id(vehicle_id)
 
         vehicle = self._vehicles[vehicle_id]
-        Vehicle.detach_all_sensors_from_vehicle(vehicle)
+        # pytype: disable=attribute-error
+        Vehicle.detach_all_sensors_from_vehicle(vehicle) 
+        # pytype: enable=attribute-error
```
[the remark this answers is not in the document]
liu-allan: No unfortunately, adding `_HAS_DYNAMIC_ATTRIBUTES = True` did not fix this issue

zoo/policies/open-agent/open_agent/agent.py
liu-allan: `casadi` or `opengen` will fail the CI test because they are not installed in the other packages

smarts/core/agent_manager.py
liu-allan: Ok I have changed it, its just that the return type is quite complicated because of the `boid` condition, this could return a dictionary of observations/rewards... for one agent `{agent_id: Dict[str, Observation]}` or a single observation/reward for one agent `{agent_id: Observation}`

envision/server.py
```diff
@@ -61,6 +61,7 @@ def set_default_headers(self):
         """Setup the default headers.
         In this case they are the minimum required CORS releated headers.
         """
+        # pytype: disable=attribute-error
```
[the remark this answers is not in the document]
liu-allan: I have followed the suggestion for `Mixin` from here https://github.com/google/pytype/blob/main/docs/faq.md#im-dynamically-populating-a-class--module-using-setattr-or-by-modifying-locals--globals-now-pytype-complains-about-missing-attributes-or-module-members-how-do-i-fix-this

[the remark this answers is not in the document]
liu-allan: Nevermind this did not work when running an experiment

ultra/tests/test_episode.py
```diff
@@ -114,9 +114,10 @@ def run_experiment():
             #     # self.assertTrue(
             #     #     abs(result[key] - episode_info["Train"][AGENT_ID].data[key]) <= 0.001
             #     # )
-
-    @unittest.skip
-    def test_episode_counter(self):
+    # pytype: disable=wrong-arg-types
+    @unittest.skip 
+    def test_episode_counter(self): 
+    # pytype: enable=wrong-arg-types
```
Gamenot, line 120:
```suggestion
    @unittest.skip("Experiment test is not necessary at this time.") 
    def test_episode_counter(self):
```

smarts/core/bubble_manager.py
Gamenot: Weird, does this ever even get hit then?

cli/studio.py
```diff
@@ -84,7 +84,7 @@ def _build_single_scenario(clean: bool, allow_offset_map: bool, scenario: str):
 
 
 def _build_single_scenario_proc(
-    clean: bool, allow_offset_map: bool, scenario: str, semaphore: Semaphore
+    clean: bool, allow_offset_map: bool, scenario: str, semaphore: Semaphore # pytype: disable=invalid-annotation
```
Gamenot, line 87: I would assume Semaphore is causing this. It seems kind of curious.

ultra/tests/test_episode.py
```diff
@@ -36,7 +36,7 @@
 task_id = "00"
 task_level = "easy"
 
-
+# pytype: disable=name-error
```
Gamenot, line 39: Where is this error coming from? This is either a bug with `pytest` or a bug with the code here.

Gamenot (approved):

cli/studio.py
liu-allan: Yes it is

smarts/core/vehicle.py
liu-allan: No, that didn't fix it. It is only `chassis.steering` that is not recognized properly for some reason.

zoo/evaluation/metrics/diversity/diversity_evaluation.py
```diff
@@ -73,6 +74,7 @@ def run_evaluation(self):
             df.set_index(["scenario"], inplace=True)
             for agent in agent_list:
                 for scenario_name, result_data in result_dict.items():
+                    # pytype: enable=attribute-error
```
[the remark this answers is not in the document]
liu-allan: Yes it is because of `result_dict`. `diversity_data_analyze` can return `None` which is what I think is causing this error https://github.com/huawei-noah/SMARTS/blob/a7825623fe32a1a2509d3f7ad97dc61eaaa97ff9/zoo/evaluation/metrics/diversity/diversity_evaluation.py#L44

ultra/tests/test_episode.py
liu-allan: It was coming from `episode` being undefined, this is now fixed

smarts/core/bubble_manager.py
liu-allan: I am not sure

smarts/core/agent_manager.py
```diff
@@ -223,25 +223,27 @@ def observe(
...
-                scores[agent_id] = vehicle.trip_meter_sensor()
+                rewards[agent_id] = float(vehicle.trip_meter_sensor(increment=True))
+                scores[agent_id] = float(vehicle.trip_meter_sensor())
 
         if sim.should_reset:
             dones = {agent_id: True for agent_id in self.agent_ids}
             dones["__sim__"] = True
 
-        return observations, rewards, scores, dones
+        return observations, rewards, scores, dones
```
sah-huawei, line 238: extra whitespace

smarts/core/bubble_manager.py
```diff
@@ -184,15 +184,17 @@ def admissibility(
                 )
             )
 
+            # pytype: disable=unsupported-operands
             all_hijacked_vehicle_ids = (
                 current_hijacked_vehicle_ids
-                | vehicle_ids_by_bubble_state[BubbleState.InAirlock][self]
+                | vehicle_ids_by_bubble_state[BubbleState.InAirlock][self]
```
sah-huawei, line 190: whitespace

smarts/core/bubble_manager.py
```diff
@@ -184,15 +184,17 @@ def admissibility(
...
             all_hijacked_vehicle_ids = (
                 current_hijacked_vehicle_ids
-                | vehicle_ids_by_bubble_state[BubbleState.InAirlock][self]
+                | vehicle_ids_by_bubble_state[BubbleState.InAirlock][self] 
             ) - {vehicle_id}
 
             all_shadowed_vehicle_ids = (
                 current_shadowed_vehicle_ids
-                | vehicle_ids_by_bubble_state[BubbleState.InBubble][self]
+                | vehicle_ids_by_bubble_state[BubbleState.InBubble][self]
```
sah-huawei, line 195: whitespace

smarts/core/bubble_manager.py
```diff
@@ -508,15 +510,15 @@ def _airlock_social_vehicle_with_social_agent(
         )
 
         if social_agent is None:
-            return
+            return
```
sah-huawei, line 513: whitespace

smarts/core/tests/test_observations.py
```diff
@@ -24,7 +24,7 @@
 import gym
 import numpy as np
 import pytest
-from panda3d.core import OrthographicLens, Point2, Point3
+from panda3d.core import OrthographicLens, Point2, Point3
```
sah-huawei, line 27: whitespace

smarts/core/tests/test_trajectory_interpolation_provider.py
```diff
@@ -300,7 +300,7 @@ def test_trajectory_interpolation_provider_in_smarts(smarts, agent_spec, scenari
         if agent_obs.events.reached_goal:
             reached_goal = True
             break
-
+
```
sah-huawei, line 303: whitespace

smarts/core/vehicle.py
sah-huawei: Is it because there's no return type specified?  (`Optional[float]`)

smarts/ros/src/smarts_ros/scripts/ros_driver.py
```diff
@@ -28,9 +28,9 @@
 import time
 from collections import deque
 from threading import Lock
-from typing import Any, Dict, List, Sequence, Tuple
+from typing import Any, Dict, List, Sequence, Tuple, Optional
 
-import numpy as np
+import numpy as np
```
sah-huawei, line 33: whitespace

ultra/tests/test_evaluate.py
```diff
@@ -471,7 +472,8 @@ def run_experiment(scenario_info, num_agents, log_dir, headless=True):
 
             active_agent_ids = observations.keys() & next_observations.keys()
             loss_outputs = {
-                agent_id: agents[agent_id].step(
+                agent_id: agents[agent_id].act(
```
sah-huawei, line 475: what was the reason for this change?

sah-huawei (changes requested):
the only real issue I saw was a bunch of places where spurious whitespace was introduced.  I'm surprised this got through `black` (if it wasn't run, it should be of course), but even if so, those changes should be reverted.

.github/workflows/ci-format.yml
```diff
@@ -57,17 +57,21 @@ jobs:
           python3.7 -m venv ${{env.venv_dir}}
           . ${{env.venv_dir}}/bin/activate
           pip install --upgrade pip wheel
-          pip install .[dev]
+          pip install -e .[dev,camera-obs,train,test]
```
Adaickalavan, line 60: Is this `camera-obs,train,test` necessary ?
```suggestion
          pip install -e .[dev] 
```

cli/ultra.py
Adaickalavan, line 76: The latest version of `ultra` code was developed separately in the `ultra-develop` branch. 

This cli command `scl ultra worker` is for manually starting a manager in a distributed computing setting. This version of cli command code appears to use the original implementation from smarts version `v0.4.10` : https://github.com/huawei-noah/SMARTS/blob/v0.4.10/smarts/zoo/worker.py .

I believe the correct cli command is identical to `scl zoo manager`, thus we might simply delete this entire `scl ultra worker` cli command.

Adaickalavan: The latest version of `ultra` code was developed separately in the `ultra-develop` branch. 

This cli command `scl ultra worker` is for manually starting a manager in a distributed computing setting. This version of cli command code appears to use the original implementation from smarts version `v0.4.10` : https://github.com/huawei-noah/SMARTS/blob/v0.4.10/smarts/zoo/worker.py .

I believe the correct cli command is identical to `scl zoo manager`, thus we might simply delete this entire `scl ultra worker` cli command.

envision/server.py
```diff
@@ -17,6 +17,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
+import abc
```
Adaickalavan, line 20: `import abc` appears unused and can be removed.

```suggestion
```

cli/studio.py
Adaickalavan: `multiprocessing.Semaphore` is a `Callable` and thus the `invalid-annotation` error. The semaphore is of type `multiprocessing.synchronize.Semaphore`. 

Hence, either use the appropriate type 
```suggestion
from multiprocessing import Process, Semaphore, synchronize
def _build_single_scenario_proc(
    clean: bool, allow_offset_map: bool, scenario: str, semaphore: synchronize.Semaphore): 
```
or simply leave the type hint empty.
```suggestion
def _build_single_scenario_proc(
    clean: bool, allow_offset_map: bool, scenario: str, semaphore):
```

smarts/core/agent_manager.py
```diff
@@ -223,25 +223,27 @@ def observe(
...
-                        f"Agent `{agent_id}` has raised done with {observations[agent_id].events}",
+                        # pytype: disable=attribute-error
+                        f"Agent `{agent_id}` has raised done with {observations[agent_id].events}", 
+                        # pytype: enable=attribute-error
                     )
 
-                rewards[agent_id] = vehicle.trip_meter_sensor(increment=True)
-                scores[agent_id] = vehicle.trip_meter_sensor()
+                rewards[agent_id] = float(vehicle.trip_meter_sensor(increment=True))
+                scores[agent_id] = float(vehicle.trip_meter_sensor())
```
Adaickalavan, line 232: I suggest we remove such type casting, i.e., `float(...)` . 
```suggestion
                rewards[agent_id] = vehicle.trip_meter_sensor(increment=True)
                scores[agent_id] = vehicle.trip_meter_sensor()
```

If there is any issue with the `vehicle.trip_meter_sensor()` function, then we should correct it at the source at https://github.com/huawei-noah/SMARTS/blob/60239fbba0c341e331867a38277e480e83c0da85/smarts/core/sensors.py#L938-L942

smarts/core/agent_manager.py
```diff
@@ -223,25 +223,27 @@ def observe(
...
+    def _vehicle_reward(self, vehicle_id, sim) -> float:
+        return float(sim.vehicle_index.vehicle_by_id(vehicle_id).trip_meter_sensor(
             increment=True
-        )
+        ))
 
-    def _vehicle_score(self, vehicle_id, sim):
-        return sim.vehicle_index.vehicle_by_id(vehicle_id).trip_meter_sensor()
+    def _vehicle_score(self, vehicle_id, sim) -> float:
+        return float(sim.vehicle_index.vehicle_by_id(vehicle_id).trip_meter_sensor())
```
Adaickalavan, line 246: Avoid ad-hoc type casting, i.e., `float( ... )` . If there is any issue with the vehicle.trip_meter_sensor() function, then we should correct it at the source.

smarts/core/agent_manager.py
Adaickalavan: I believe such whitespaces will be automatically corrected by the `isort, Black, and prettier` step in `ci-auto-commit-linux.yml`. The `ci-auto-commit-linux.yml` file runs and commits automatically upon merging a pull request into the `develop` branch.

smarts/core/agent_manager.py
```diff
@@ -223,25 +223,27 @@ def observe(
                 else:
                     logging.log(
                         logging.DEBUG,
-                        f"Agent `{agent_id}` has raised done with {observations[agent_id].events}",
+                        # pytype: disable=attribute-error
+                        f"Agent `{agent_id}` has raised done with {observations[agent_id].events}", 
+                        # pytype: enable=attribute-error
```
Adaickalavan, line 228: This `attribute-error` is due to the two types of `observations` present. Consider fixing it as follows.
```suggestion
                obs, dones[agent_id] = Sensors.observe(
                    sim, agent_id, sensor_state, vehicle
                )
                observations[agent_id] = obs

                if sim.vehicle_index.vehicle_is_shadowed(vehicle.id):
                    # It is not a shadowing agent's fault if it is done
                    dones[agent_id] = False
                else:
                    logging.log(
                        logging.DEBUG,
                        f"Agent `{agent_id}` has raised done with {obs.events}", 
                    )
```

smarts/core/plan.py
```diff
@@ -57,7 +57,7 @@ def point(self) -> Point:
     def from_pose(cls, pose: Pose):
         """Convert to a starting location from a pose."""
         return cls(
-            position=pose.as_position2d(),
+            position=tuple(pose.as_position2d()),
```
Adaickalavan, line 60: A better fix is as follows:
```
import numpy as np
    
<some code>
    
    @dataclass(frozen=True)
    class Start:  
        position: np.ndarray

        <some code>
    
        @classmethod
        def from_pose(cls, pose: Pose):
            """Convert to a starting location from a pose."""
            return cls(
                position=pose.as_position2d(),
                heading=pose.heading,
                from_front_bumper=False,
            )
```

smarts/core/smarts.py
```diff
@@ -89,7 +89,7 @@ def __init__(
         traffic_sim,  # SumoTrafficSimulation
         envision: Optional[EnvisionClient] = None,
         visdom: Optional[VisdomClient] = None,
-        fixed_timestep_sec: float = 0.1,
+        fixed_timestep_sec: Optional[float] = 0.1,
```
Adaickalavan, line 92: Why use `Optional` here?

`Optional[...]` is a shorthand notation for `Union[..., None]` .

```suggestion
        fixed_timestep_sec: float = 0.1,
```

smarts/core/smarts.py
```diff
@@ -108,7 +108,7 @@ def __init__(
         self._reset_required = False
 
         assert fixed_timestep_sec is None or fixed_timestep_sec > 0
-        self.fixed_timestep_sec: float = fixed_timestep_sec
+        self.fixed_timestep_sec: Optional[float] = fixed_timestep_sec
```
Adaickalavan, line 111: Why use `Optional` here?

`Optional[...]` is a shorthand notation for `Union[..., None]` .

```suggestion
        self.fixed_timestep_sec: float = fixed_timestep_sec
```

smarts/core/smarts.py
```diff
@@ -869,7 +869,9 @@ def _pybullet_provider_sync(self, provider_state: ProviderState):
                     agent_interface = self._agent_manager.agent_interface_for_agent_id(
                         agent_id
                     )
+                    # pytype: disable=attribute-error
                     agent_action_space = agent_interface.action_space
+                    # pytype: enable=attribute-error
```
Adaickalavan, line 874: It looks like `agent_interface_for_agent_id` function is not expected to return `None`, hence a better fix would be to change the following

https://github.com/huawei-noah/SMARTS/blob/60239fbba0c341e331867a38277e480e83c0da85/smarts/core/agent_manager.py#L105-L107

to

```
    def agent_interface_for_agent_id(self, agent_id) -> AgentInterface:
        """Get the agent interface of a specific agent."""
        return self._agent_interfaces[agent_id]
```
Then, delete the pytype disable command here.
```suggestion
                    agent_action_space = agent_interface.action_space
```

smarts/core/agent_manager.py
```diff
@@ -102,7 +102,7 @@ def agent_interfaces(self) -> Dict[str, AgentInterface]:
         """A list of all agent to agent interface mappings."""
         return self._agent_interfaces
 
-    def agent_interface_for_agent_id(self, agent_id) -> AgentInterface:
+    def agent_interface_for_agent_id(self, agent_id) -> Optional[AgentInterface]:
         """Get the agent interface of a specific agent."""
         return self._agent_interfaces.get(agent_id, None)
```
Adaickalavan, line 107: It looks like `agent_interface_for_agent_id` function is not expected to return `None`, hence consider the following.

```suggestion
    def agent_interface_for_agent_id(self, agent_id) -> AgentInterface:
        """Get the agent interface of a specific agent."""
        return self._agent_interfaces[agent_id]
```

smarts/core/smarts.py
```diff
@@ -1039,12 +1041,16 @@ def agent_controls_vehicles(agent_id):
...
+            # pytype: disable=attribute-error
             return interface.action_space in action_spaces
+            # pytype: enable=attribute-error
 
         def matches_no_provider_action_space(agent_id):
             interface = self._agent_manager.agent_interface_for_agent_id(agent_id)
             for provider in self.providers:
+                # pytype: disable=attribute-error
                 if interface.action_space in provider.action_spaces:
+                # pytype: enable=attribute-error
```
Adaickalavan, line 1053: Remove the pytype disable command here.
```suggestion
                if interface.action_space in provider.action_spaces:
```

smarts/core/smarts.py
```diff
@@ -1039,12 +1041,16 @@ def agent_controls_vehicles(agent_id):
 
         def matches_provider_action_spaces(agent_id, action_spaces):
             interface = self._agent_manager.agent_interface_for_agent_id(agent_id)
+            # pytype: disable=attribute-error
             return interface.action_space in action_spaces
+            # pytype: enable=attribute-error
```
Adaickalavan, line 1046: Remove the pytype disable command here.
```suggestion
            return interface.action_space in action_spaces
```

smarts/core/smarts.py
```diff
@@ -1243,8 +1249,10 @@ def _perform_agent_actions(self, agent_actions):
                     vehicle_action,
                     controller_state,
                     sensor_state,
+                    # pytype: disable=attribute-error
                     agent_interface.action_space,
                     agent_interface.vehicle_type,
+                    # pytype: enable=attribute-error
```
Adaickalavan, line 1255: Remove the pytype disable command here.
```suggestion
                    agent_interface.action_space,
                    agent_interface.vehicle_type,
```

smarts/core/vehicle_index.py
```diff
@@ -284,7 +284,7 @@ def vehicles(self):
 
     def vehicleitems(self) -> Tuple[str, Vehicle]:
         """A list of all vehicle IDs paired with their vehicle."""
-        return map(lambda x: (self._2id_to_id[x[0]], x[1]), self._vehicles.items())
+        return tuple(map(lambda x: (self._2id_to_id[x[0]], x[1]), self._vehicles.items()))
```
Adaickalavan, line 287: Returning an iterator is better, hence consider the following change instead:
```
    def vehicleitems(self) -> Iterator[Tuple[str, Vehicle]]:
        """A list of all vehicle IDs paired with their vehicle."""
        return map(lambda x: (self._2id_to_id[x[0]], x[1]), self._vehicles.items())
```

smarts/sstudio/genhistories.py
```diff
@@ -85,7 +85,9 @@ def check_dataset_spec(self, dataset_spec: Dict[str, Any]):
         if "input_path" not in dataset_spec:
             errmsg = "'input_path' field is required in dataset yaml."
         elif dataset_spec.get("flip_y"):
+            # pytype: disable=attribute-error
             if not dataset_spec.get("source").startswith("NGSIM"):
+            # pytype: enable=attribute-error
```
Adaickalavan, line 90: Default value of `get()` is `None`. `None` has no `startswith()` attribute.  Hence, consider the following change.
```suggestion
            if not dataset_spec["source"].startswith("NGSIM"):
```

zoo/policies/cross-rl-agent/cross_rl_agent/train/run_test.py
```diff
@@ -134,7 +134,7 @@ def test(test_scenarios, sim_name, headless, num_episodes, seed):
                 if WITH_SOC_MT:
                     reward = rewards[AGENT_ID]
                 else:
-                    reward = np.sum(reward)
+                    reward = np.sum(rewards.values())
```
Adaickalavan, line 137: What did the original `reward` variable refer to?

smarts/core/agent_manager.py
sah-huawei: for one thing (and this may not be the problem that `pytype` is complaining about), shouldn't the return type hint be `-> float` for `__call__()`?

ultra/tests/test_evaluate.py
Gamenot: This should not be changed, it should be reverted.

Adaickalavan: Reverted.

.github/workflows/ci-format.yml
```diff
@@ -57,17 +57,21 @@ jobs:
...
         shell: bash
         run: |
           cd $GITHUB_WORKSPACE
           CHANGED=$(git diff --diff-filter="AM" --name-only --ignore-submodules=all origin/develop... | awk -v d=" " '{s=(NR==1?s:s d)$0}END{print s}')
-          echo "::set-output name=changed_files::$CHANGED"
+          declare -A SEEN
+          UNIQUE=()
+          for file in ${CHANGED[@]}; do filename=$(basename $file); if [ ! -v "SEEN[$filename]" ]; then SEEN[$filename]+=1; UNIQUE+=($file); fi; done 
+          OUT=$( IFS=$' '; echo "${UNIQUE[*]}")
+          echo "::set-output name=changed_files::$OUT"
```
saulfield, line 71: It looks like this filters out files with duplicate names. Have we confirmed this addresses the issue I brought up in my other comment? i.e. if `scenarios/2_lane/agent_prefabs.py` and `scenarios/4_lane/agent_prefabs.py` are both changed, will this exclude both of them?

saulfield (changes requested):
Looks good to me. No comments on the type hint changes, just want to confirm the issue about duplicate files has been resolved.

.github/workflows/ci-format.yml
liu-allan: It will include the first file seen with a duplicate name

Adaickalavan (approved):

saulfield (approved):

Issue #1279, opened 2022-02-16
Replace `HiWayEnv` with "`StandardGymEnv`"


Gamenot:
## Description of the issue
`HiWayEnv` as is currently implemented breaks with the `gym` interface in a few locations:
- `reset(..)` can throw an exception due to incompatibility between a scenario and an agent interface.
- `step(..)` returns `dones` individual to agent rather than `done` related the end of an episode
- `observation_space` is not implemented (and must be)
- `action_space` is not implemented (and must be)

## Proposed solution
Create a new standard gym environment that fully conforms to the gym interface. Deprecate `HiWayEnv`.

AC:
- [ ] `reset` is guaranteed to re-initialize a working environment by skipping invalid episode configurations see: https://github.com/huawei-noah/SMARTS/issues/1271#issuecomment-1035879435
- [ ] details should be logged if a scenario&agent-interface combination is invalid to be able to start an episod.
- [ ] `dones` should be returned from `step` instead of `done` and is related to the environment episode being `done` rather than individual agents.
- [ ] consider if #1194 can be integrated to automatically evaluate `action_space` from agent interfaces
- [ ] consider if #1193 can be integrated to automatically evaluate `observation_space` from agent interfaces
- [ ] Deprecate `HiWayEnv`

Adaickalavan:
For reference, currently `/smarts/env/wrappers/format_obs.py` can automatically build the appropriate gym `observation_space` from `agent_specs[agent_id].interface`. This can be used as reference when `adapters` are being worked on.

PR #1374, opened 2022-04-08
Replace HiWayEnv with "GymBaseEnv"

liu-allan:
#1279 

Not sure how to check if a combination is invalid here:
- details should be logged if a scenario&agent-interface combination is invalid to be able to start an episode

### DIFF
--- a/smarts/core/utils/episodes.py
+++ b/smarts/core/utils/episodes.py
@@ -74,7 +74,7 @@ class EpisodeLog:
                 observations, rewards, dones, infos
             )
 
-        if dones.get("__all__", False) and infos is not None:
+        if dones and infos is not None:
             for agent, score in infos.items():
                 self.scores[agent] = score["score"]
 
--- a/smarts/env/wrappers/single_agent.py
+++ b/smarts/env/wrappers/single_agent.py
@@ -58,7 +58,7 @@ class SingleAgent(gym.Wrapper):
         return (
             obs[self._agent_id],
             reward[self._agent_id],
-            done[self._agent_id],
+            done,
             info[self._agent_id],
         )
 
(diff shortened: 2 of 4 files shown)


smarts/env/wrappers/single_agent.py
```diff
@@ -58,7 +58,7 @@ def step(self, action: Any) -> Tuple[Any, float, bool, Any]:
         return (
             obs[self._agent_id],
             reward[self._agent_id],
-            done[self._agent_id],
+            done,
```
Adaickalavan, line 61: I think this should not be changed, as it is intended to wrap a multi-agent env to become compatible with a single-agent gym specs.
```suggestion
            done[self._agent_id],
```

smarts/core/utils/episodes.py
```diff
@@ -74,7 +74,7 @@ def record_step(self, observations=None, rewards=None, dones=None, infos=None):
                 observations, rewards, dones, infos
             )
 
-        if dones.get("__all__", False) and infos is not None:
+        if dones and infos is not None:
```
Adaickalavan, line 77: I think `dones` here is a `dict`. Why was this changed?

liu-allan: From the issue #1279 one of the requests was that `done` be related to the environment being done rather than for individual agents. I have changed `done` to represent `dones["__all__"]` instead to reflect the request. See https://github.com/huawei-noah/SMARTS/pull/1374/files#diff-7f6d3d7e26af4d9bdf221b3951430a399520fd204bdf830e475813052c2c1d28R333

smarts/env/wrappers/single_agent.py
liu-allan: See other comment

smarts/core/utils/episodes.py
Gamenot: This is correct for `Gym` specific environments. When we get around to `RLlib` style environments that may be different.

Adaickalavan:
Closed due to becoming obsolete.

PR #1383, opened 2022-04-14
Suppress `stderr` output in notebook test

Gamenot:
Attempt a fix so that the `test_notebook` test does not output an unclosed file to stderr.

### DIFF
--- a/smarts/core/utils/logging.py
+++ b/smarts/core/utils/logging.py
@@ -101,8 +101,17 @@ def _suppress_fileout(stdname):
     except UnsupportedOperation as e:
         if not isnotebook():
             raise e
-        ## This case is notebook which does not have issues with the c_printf
-        return None
+        old_stderr = sys.stderr
+        sys.stderr = open(os.devnull, os.O_WRONLY)
+
+        def cleanup(_):
+            nonlocal old_stderr
+            sys.stderr.flush()
+            os.close(sys.stderr)
+            sys.stderr = old_stderr
+
+        ## This case is notebook
+        return cleanup
 
     dup_std_fno = os.dup(original_std_fno)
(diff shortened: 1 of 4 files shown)


Gamenot:
It looks like this might not fix the issue. The problem stems from TraCI's `Connection.__init__(..)` where `socket.connect()` is being called without calling `socket.close()` when handling `OSError`/`socket.error`.

RutvikGupta (approved):

qianyi-sun (approved):

smarts/core/utils/logging.py
```diff
@@ -101,8 +102,24 @@ def _suppress_fileout(stdname):
...
+        old_std = getattr(sys, stdname)
+        setattr(sys, stdname, file)
+
+        def cleanup(_):
+            nonlocal old_std, stdname
+            new_std = getattr(sys, stdname)
+            new_std.flush()
+            # Dummy attributes because of https://github.com/ipython/ipykernel/issues/867
+            if not hasattr(new_std, "watch_fd_thread"):
+                setattr(new_std, "watch_fd_thread", Mock())
```
saulfield, line 115: This is a clever workaround for the `AttributeError` mentioned in that issue. Does it not matter (at least for our purposes) that it isn't the true attribute? e.g. when `watch_fd_thread.join()` is called, could it cause other problems?

saulfield (approved):

smarts/core/utils/logging.py
Gamenot: I am not sure, I was tempted to just use `Mock()` in place of the file itself, but I was concerned that it would cause issues with manipulating `std` outputs. I do not think this hack matters here, it may break in the future.

On reflection, I believe I should just use `None` here.

PR #1388, opened 2022-04-18
fix us101 map

qianyi-sun:
a quick fix for the us101 map on master/develop branch

### DIFF
--- a/cli/studio.py
+++ b/cli/studio.py
@@ -71,7 +71,7 @@ def _build_single_scenario(clean: bool, allow_offset_map: bool, scenario: str):
     from smarts.core.scenario import Scenario
 
     traffic_histories = Scenario.discover_traffic_histories(scenario_root_str)
-    shift_to_origin = not allow_offset_map or bool(traffic_histories)
+    shift_to_origin = not allow_offset_map
 
     map_spec = Scenario.discover_map(scenario_root_str, shift_to_origin=shift_to_origin)
     road_map, _ = map_spec.builder_fn(map_spec)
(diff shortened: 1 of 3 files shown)


RutvikGupta (approved):

Gamenot (approved):

cli/studio.py
```diff
@@ -71,7 +71,7 @@ def _build_single_scenario(clean: bool, allow_offset_map: bool, scenario: str):
     from smarts.core.scenario import Scenario
 
     traffic_histories = Scenario.discover_traffic_histories(scenario_root_str)
-    shift_to_origin = not allow_offset_map or bool(traffic_histories)
+    shift_to_origin = not allow_offset_map
```
turix2, line 74: I would go a step further here.  I suggest forcing it to not shift the map when there's traffic history data:
```python
    # don't shift maps for scenarios with traffic histories since history data must line up with map
    shift_to_origin = not allow_offset_map and not bool(traffic_histories)
```

Then, where the `--allow-offset-map` option is documented just above that, you should amend it to be something like:
```python
@click.option(
    "--allow-offset-maps",
    is_flag=True,
    default=False,
    help="Allows road networks (maps) to be offset from the origin. If not specified, a new network file is created if necessary.  Defaults to False except when there's Traffic History data associated with the scenario.",
)
```

scenarios/NGSIM/README.md
```diff
@@ -16,6 +16,10 @@ Once that is done, you should be able to build your scenarios in the normal way,
 ```bash
 scl scenario build-all --clean scenarios/NGSIM
 ```
+or the following command to disable generating `shifted_map-AUTOGEN.net.xml`
```
turix2, line 19: I would reword this to:  "or use the `--allow-offset-map` option to prevent auto-shifting the map to the origin, like:  "

(The main reason for the rewording is because the filename `shifted_map-AUTOGEN.net.xml` is encapsulated, i.e., not to be known or cared about by users.)

PR #1399, opened 2022-04-19
Investigate missing envision map

Gamenot:
Debug issues with envision map rendering.

### DIFF
--- a/envision/server.py
+++ b/envision/server.py
@@ -324,7 +324,7 @@ class StateWebSocket(tornado.websocket.WebSocketHandler):
     async def open(self, simulation_id):
         """Open this socket to  listen for webclient playback requests."""
         if simulation_id not in WEB_CLIENT_RUN_LOOPS:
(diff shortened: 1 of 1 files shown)


envision/server.py
```diff
@@ -365,8 +365,13 @@ def initialize(self, path_map: Dict[str, Union[str, Path]] = {}):
 
     async def get(self, id_):
         """Serve a resource requested by id."""
-        if id_ not in self._path_map or not self._path_map[id_].exists():
-            raise tornado.web.HTTPError(404)
+        if id_ not in self._path_map:
+            raise tornado.web.HTTPError(404, f"Map resource {id_} not found")
+
+        if not self._path_map[id_].exists():
```
saulfield, line 371: The type hint for `_path_map` says its values can be `str` or `Path`. It seems that calling `exists()` assumes it is a `Path`.

saulfield (approved):

RutvikGupta (approved):

Issue #1402, opened 2022-04-22
Missing/skipping data during conversion for ngsim us101


qianyi-sun:
## **BUG REPORT**
**High Level Description**
When using SMARTS to convert NGSIM raw data (.txt files) to .shf files, for some `vehicle_id`s, corresponding `sim_time`s get skipped or missed, which leads to vehicles teleport in the visualization. 

Also, there is a total row number difference between the raw data and the `.shf` files.

**SMARTS version**
0.6.1 rc1

**Steps to reproduce the bug**
`scl scenario build --clean --allow-offset-map scenario/NGSIM/us101`

**Resulting and expected behaviour**
The `sim_time` for all vehicles should be continuous and the row number of data before/after the conversion should be the same. 

**Error logs and screenshots**
For example, the `vehicle_id=1936` in `us101_0820-0835.shf`, the `sim_time` jumps from 913.7 to 917.2 (see the line 14,15 in screenshot below).
![image](https://user-images.githubusercontent.com/59240680/164753490-ad564848-f92d-454e-971e-3a551eecad34.png)

Also with same `vehicle_id=1936`, the total row number of it in `us101_0820-0835.shf` is 520, given that in the raw data, the row number of it is 789.

qianyi-sun:
@YuechengLiu We use this issue to record the bug that you reported to us. Please feel free to edit the description for improvement if necessary. Thanks!

turix2:
In general, "gaps" in a vehicle's trajectory can occur if it drives off of the mapped area for some reason.  However, that's not occurring in this case, so this was a bug.  I have a PR that I'm just wrapping up that fixes this.

sah-huawei:
I think this is fixed by PR #1403.  Please reopen this issue if it re-occurs though.

Issue #1354, opened 2022-03-25
Abnormal traffic flow trajectory in us101 map


qianyi-sun:
## **BUG REPORT**
**High Level Description**
In the real map, there are direction changes in different sections but the traffic pattern itself doesn't seem to curve as much as the road.

**SMARTS version**
0.5.1

https://user-images.githubusercontent.com/59240680/160151542-bc0ab5bd-9d93-4b06-b26b-af10e6f045aa.mp4

Here's the x,y coordinates of vehicle_id=2 in 0400-0415.shf:

![image](https://user-images.githubusercontent.com/59240680/160152177-7768878e-90c8-4b41-9a94-5e061dac7554.png)

And in the raw data: (col 1 is the vehicle id, col 5 and 6 are x,y coordinates)

![image](https://user-images.githubusercontent.com/59240680/160152348-892299bd-98e3-4731-8f13-799bfb22db16.png)

Please let me know if more info is needed, thanks!

sah-huawei:
Just to clarify, the data you pasted in above is from the `peach` dataset (`trajectories-0400pm-0415pm.txt`), not `us101`.  But I only have a map for `us101`, not `peach`.

However, I still think I can find a vehicle that does what you described when we talked about this.

qianyi-sun:
> Just to clarify, the data you pasted in above is from the `peach` dataset (`trajectories-0400pm-0415pm.txt`), not `us101`. But I only have a map for `us101`, not `peach`.
> 
> However, I still think I can find a vehicle in the `us101` data that does what you described when we talked about this.

Oh sorry that was a typo, I used Peachtree data for filing this issue, I apologize for the inconvenience due to this typo.

Issue #1321, opened 2022-03-07
version inconsistencies for numpy and tensorflow


sah-huawei:
## **BUG REPORT**
**High Level Description**
We have a version compatibility "deadlock" within our current `setup.py` such that we will get an import error related to `numpy` from `genhistories.py` when we have either our `[train]` or our `[waymo]` extensions installed.

**Details**
Our current `setup.py` specifies a minimum version for the `numpy` package of `1.19.5`.  However, we have been using a function in `genhistories.py` (`sliding_window_view`) for over a year that was only introduced in `numpy` version `1.20.0`.

Yet we cannot currently update our minimum version of `numpy` to be `1.20.0` because our minimum version of `tensorflow` (in the `[train]` extension) is specified as `2.4.0`.  `tensorflow` did not add support for `numpy 1.20.0` until version `2.7`.

Yet we cannot update our minimum version of `tensorflow` to version `2.7` because the `waymo-open-dataset` does not have support beyond version `2.6`.

**SMARTS version**
0.5 and 0.5.1

**Steps to reproduce the bug**
Install smarts with the `[train]` and/or `[waymo]` extension and then do `scl scenario build` on a scenario that contains traffic history data.

**Resulting and expected behaviour**
We should be able to use any and all of our setup extensions together.

**Impact [If known]**
If not fixed, the impact will be felt in future versions (i.e., starting in SMARTS version `0.6`) when a customer tries to use Waymo history data to train a model. So this undermines our addition of Waymo open dataset support as a feature unless we can solve this.

sah-huawei:
Just adding a bit more detail about `sliding_window_view`.  Replacing this with something else seems to be the past of least resistance, but it's non-trivial because at the scale of the NGSIM dataset this is already a bottleneck.  

It takes around an hour on my machine to run `genhistories` on NGSIM because of the sliding windows, and that's _using_ the `numpy` native implementation that is optimized to work "in place" (with the stride tricks).  Doing it with our own code is likely to slow it down by an order of magnitude.

PR #1403, opened 2022-04-22
NGSIM import fixes

sah-huawei:
This PR addresses several Issues related to importing NGSIM datasets, including:
- Issue #1321:  we replace the use of numpy's `sliding_window_view()` with our own custom `_WindowedReader`.  (We now use this for the `INTERACTION` dataset importer too, so we can fully support the heading window there.)
- Issue #1354:  we correct positions from the front-bumper to the vehicle center based on its (inferred) heading.
- Issue #1402:  we fix map boundary checking (along with the fix for #1354) to avoid thinking the vehicles are off of the map.

Along the way, we also fixed several other traffic-history issues, including:
- we now never auto-shift maps for scenarios that have traffic history.
- changed the default value of the `heading_inference_min_speed` parameter to 2.2 m/s (instead of .22 m per .1 sec).
- slightly better heuristics for inferring headings using the position windows.

We're also going to push an example of an INTERACTION dataset scenario (since this was used for testing the #1321 change anyway).  This will probably be useful in the future.

Note that the `peachtree` scenario map that's included here is still incomplete.  We will finish that in another PR to follow.


Gamenot (approved):
Patrick and I looked at fixing it from the CLI side. This should be a useful fix.

qianyi-sun (approved):

smarts/sstudio/genhistories.py
```diff
@@ -54,14 +62,82 @@ def __init__(self, dataset_spec: Dict[str, Any], output: str):
...
+            row_gen: Iterable[Row],
+            transform_fn: Callable[[Row, Deque[Row], Deque[Row]], None],
+            window_before: int = 0,
+            window_after: int = 0,
+            group_col: Optional[str] = None,
+        ):
+            self._row_gen = row_gen
+            self._transform_fn = transform_fn
+            self._before_width = window_before
+            self._after_width = window_after
```
RutvikGupta, line 106: Should there be a default value of these instead of 0 or they vary for different map source?

smarts/sstudio/genhistories.py
```diff
@@ -54,14 +62,82 @@ def __init__(self, dataset_spec: Dict[str, Any], output: str):
...
+        lane_width = self._map_spec.get("lane_width", real_lane_width_m)
         self._scale = lane_width / real_lane_width_m
         self._flip_y = dataset_spec.get("flip_y", False)
         self._swap_xy = dataset_spec.get("swap_xy", False)
+        # most trajectory datasets have .1s time delta (i.e., were collected at 10 Hz)
+        self._dt_sec = 0.1
+
+    class _WindowedReader:
+        """Iterates over the rows in file using a sliding window that keeps track of
+        both a number of rows before the current row and a number of rows after.
```
RutvikGupta, line 76: Maybe this should be moved to `smarts.core.utils`.

RutvikGupta (approved):

smarts/sstudio/genhistories.py
sah-huawei: Yeah, I think it varies by how much "smoothing" the user wants to do within a window.  A default of 0 is basically no smoothing.  (So, for example, because the NGSIM dataset is noisy, we currently smooth its positions with a window size of 15 -- 7 before and 7 after the current row.  But there are circumstances, such as the contest perhaps, where we might just want to use the positions "as is", which would mean 0 before and 0 after the current row.)

smarts/sstudio/genhistories.py
sah-huawei: Yeah, I considered that.  I couldn't think of anywhere else we might use it in the future though.  But maybe I should just do it anyway.

Issue #1342, opened 2022-03-11
specify dataset specifications in scenario.py


sah-huawei:
SMARTS currently supports passing a list of `.yaml` files that specify information about external datasets to `gen_scenario()` in `scenario.py` via the `traffic_histories` property of the `Scenario` object.

It would be cleaner and more consistent to add a type to the DSL's supported types in `smarts.sstudio.types` called something like `TrajectoryDataset` whose possible fields match the fields currently supported in our "dataset spec" `yaml` files as understood by `genhistories.py`.  Then the `trafffic_histories` property of `Scenario` objects could be initialized with a list of `TrajectoryDataset` _objects_, as opposed to strings specifying `.yaml` file paths.

Maybe this could be done in conjunction for the fix for Issue #1321?

If so, the new `scenarios/NGSIM` folder would need to be updated to use the new `TrajectoryDataset` types instead of its `.yaml` files.

PR #1405, opened 2022-04-24
add TrafficHistoryDataset to DSL

sah-huawei:
addresses Issue #1342 

When this gets merged, the Waymo map utility will need to be updated too.


smarts/sstudio/genhistories.py
```diff
@@ -477,22 +482,28 @@ def rows(self) -> Generator[Dict, None, None]:
...
+                    map_bbox.max_pt[0] is not None
+                    and row["position_x"] * self.scale > map_bbox.max_pt[0]
+                ):
+                    self._log.info(
+                        f"skipping row for vehicle {row['vehicle_id']} with x-position ({row['position_x'] * self.scale}) off of map"
+                    )
+                    continue
+                if (
+                    map_bbox.max_pt[1] is not None
+                    and row["position_y"] * self.scale > map_bbox.max_pt[1]
```
Gamenot, line 501:
```suggestion
                    map_bbox.max_pt.y is not None
                    and row["position_y"] * self.scale > map_bbox.max_pt.y
```

smarts/sstudio/genhistories.py
```diff
@@ -692,8 +703,10 @@ def _row_gen(self) -> Generator[_TrajectoryDataset._WindowedReader.Row, None, No
                 if y_margin:
                     row["position_y"] -= y_margin
                 if self._flip_y:
-                    max_y = self._map_spec["max_y"]
-                    row["position_y"] = (max_y / self.scale) - row["position_y"]
+                    map_bb = self._dataset_spec["map_bbox"]
+                    row["position_y"] = (
+                        map_bb.max_pt[1] / self.scale - row["position_y"]
+                    )
```
Gamenot, line 709:
```suggestion
                    row["position_y"] = (
                        map_bb.max_pt.y / self.scale - row["position_y"]
                    )
```

smarts/sstudio/genhistories.py
```diff
@@ -725,22 +738,28 @@ def rows(self) -> Generator[Dict, None, None]:
...
-            if map_height and row["position_y"] * self.scale > map_height:
-                self._log.warning(
-                    f"skipping row for vehicle {row['vehicle_id']} with y-position(s) off of map"
-                )
-                continue
+            if map_bbox:
+                if (
+                    map_bbox.max_pt[0] is not None
+                    and row["position_x"] * self.scale > map_bbox.max_pt[0]
+                ):
```
Gamenot, line 750:
```suggestion
                    map_bbox.max_pt.x is not None
                    and row["position_x"] * self.scale > map_bbox.max_pt.x
                ):
```

smarts/sstudio/genhistories.py
```diff
@@ -725,22 +738,28 @@ def rows(self) -> Generator[Dict, None, None]:
...
+                    self._log.info(
+                        f"skipping row for vehicle {row['vehicle_id']} with x-position ({row['position_x'] * self.scale}) off of map"
+                    )
+                    continue
+                if (
+                    map_bbox.max_pt[1] is not None
+                    and row["position_y"] * self.scale > map_bbox.max_pt[1]
+                ):
```
Gamenot, line 758:
```suggestion
                    map_bbox.max_pt.y is not None
                    and row["position_y"] * self.scale > map_bbox.max_pt.y
                ):
```

smarts/sstudio/genscenario.py
```diff
@@ -439,25 +439,36 @@ def _validate_entry_tactic(mission):
...
-    genhistories_py = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)), "genhistories.py"
-    )
     for hdsr in histories_datasets:
+        if not isinstance(hdsr, str):
+            from . import genhistories
+
+            genhistories.import_dataset(hdsr, scenario, overwrite)
+            continue
+
```
Gamenot, line 460: I would prefer to check the specific type rather than accept any other type.
```suggestion
        if isinstance(hdsr, types.TrafficHistoryDataset):
            from . import genhistories

            genhistories.import_dataset(hdsr, scenario, overwrite)
            continue

        assert isinstance(hdsr, str)
```

smarts/sstudio/genhistories.py
```diff
@@ -990,6 +1009,48 @@ def column_val_in_row(self, row, col_name: str) -> Any:
...
+        if not overwrite:
+            print("file already exists at {output}.  skipping...")
+            return
+        os.remove(output)
+    source = dataset_spec.source_type
+    if source.startswith("NGSIM"):
+        dataset = NGSIM(dataset_spec.__dict__, output)
+    elif source == "INTERACTION":
+        dataset = Interaction(dataset_spec.__dict__, output)
+    elif source == "Waymo":
```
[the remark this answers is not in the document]
sah-huawei: They have different fields in the data files.  The city scenarios (peachtree and lanker) have 6 more columns than the highway scenarios.  (See around lines 546-555 of `genhistories.py`.)

smarts/sstudio/genhistories.py
```diff
@@ -365,8 +368,10 @@ def _row_gen(self) -> Generator[_TrajectoryDataset._WindowedReader.Row, None, No
                 if y_margin:
                     row["position_y"] -= y_margin
                 if self._flip_y:
-                    max_y = self._map_spec["max_y"]
-                    row["position_y"] = (max_y / self.scale) - row["position_y"]
+                    map_bb = self._dataset_spec["map_bbox"]
+                    row["position_y"] = (
+                        map_bb.max_pt[1] / self.scale - row["position_y"]
```
saulfield, line 373:
```suggestion
                        map_bb.max_pt.y / self.scale - row["position_y"]
```

smarts/sstudio/genhistories.py
```diff
@@ -1032,24 +1093,25 @@ def _check_args(args) -> bool:
         os.remove(args.output)
 
     if args.old:
-        dataset_spec = {"source": "OldJSON", "input_path": args.dataset}
+        dataset_spec = {"source_type": "OldJSON", "input_path": args.dataset}
```
saulfield, line 1096: Should we be removing this old format soon? It seems it's been deprecated for a while, and now we have an "old" way of specifying the dataset spec as well as an "old" traffic history format.

smarts/sstudio/types.py
```diff
@@ -795,6 +796,52 @@ class _ActorAndMission:
...
 
 
+@dataclass(frozen=True)
+class TrafficHistoryDataset:
+    """Describes a dataset containing trajectories (time-stamped positions)
+    for a set of vehicles.  Often these have been collected by third parties
+    from real-world observations, hence the name 'history'.  When used
+    with a SMARTS scenario, traffic vehicles will move on the map according
+    to their trajectories as specified in the dataset.  These can be mixed
+    with other types of traffic (such as would be sepcified by an object of
```
saulfield, line 806:
```suggestion
    with other types of traffic (such as would be specified by an object of
```

smarts/sstudio/types.py
```diff
@@ -795,6 +796,52 @@ class _ActorAndMission:
...
+    x_margin_px: float = 0.0
+    """x offset of the map from the data (in pixels)"""
+    y_margin_px: float = 0.0
+    """y offset of the map from the data (in pixels)"""
+    swap_xy: bool = False
+    """if True, the x and y axes the dataset coordinate system will be swapped"""
+    flip_y: bool = False
+    """if True, the dataset will be mirrored around the x-axis; need to specify map_bbox to use"""
+    map_bbox: Optional[BoundingBox] = None
+    """the bounding box of the map, used to filter x- and y-coordinsates to ensure they're in range"""
```
saulfield, line 826:
```suggestion
    """the bounding box of the map, used to filter x- and y-coordinates to ensure they're in range"""
```

smarts/sstudio/types.py
```diff
@@ -795,6 +796,52 @@ class _ActorAndMission:
...
+
+    map_lane_width: float = 3.7
+    """This is used to figure out the map scale, which is map_lane_width / real_lane_width_m.  (So use `real_lane_width_m` here for 1:1 scale - the default.)  It's also used in SMARTS for detecting off-road, etc."""
+    real_lane_width_m: float = 3.7
+    """Averge width in meters of the dataset's lanes in the real world.  US highway lanes are about 12 feet (or ~3.7m, the default) wide."""
```
saulfield, line 831:
```suggestion
    """Average width in meters of the dataset's lanes in the real world.  US highway lanes are about 12 feet (or ~3.7m, the default) wide."""
```

smarts/sstudio/types.py
```diff
@@ -795,6 +796,52 @@ class _ActorAndMission:
...
+    speed_limit_mps: Optional[float] = None
+    """used by SMARTS or the initial speed of new agents being added to the scenario"""
```
saulfield, line 833:
```suggestion
    """used by SMARTS for the initial speed of new agents being added to the scenario"""
```

smarts/sstudio/genhistories.py
```diff
@@ -990,6 +1009,48 @@ def column_val_in_row(self, row, col_name: str) -> Any:
...
+
+    remap(dataset_spec, "source", dataset_spec, "source_type")
+    if "map_net" in dataset_spec:
+        map_spec = dataset_spec["map_net"]
+        remap(map_spec, "lane_width", dataset_spec, "map_lane_width")
+        if "max_y" in map_spec or "width" in map_spec or "height" in map_spec:
+            max_x = map_spec.get("width")
+            max_y = map_spec.get("max_y", map_spec.get("height"))
+            dataset_spec["map_bbox"] = BoundingBox((0, 0), (max_x, max_y))
+
```
RutvikGupta, line 1052: This is to support the deprecated `yaml` files?

smarts/sstudio/genscenario.py
```diff
@@ -439,25 +439,37 @@ def _validate_entry_tactic(mission):
...
+            A sequence of traffic history descriptors.
         overwrite:
             If to forcefully write over the previous existing output file
     """
-    genhistories_py = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)), "genhistories.py"
-    )
     for hdsr in histories_datasets:
+        if isinstance(hdsr, types.TrafficHistoryDataset):
+            from . import genhistories
```
RutvikGupta, line 456: You may need to add the ugly code here to fall to absolute import, if relative import throws an error here.

RutvikGupta (approved):
Some comments but looks good to merge when addressed. Will update the utility files when merged!

scenarios/INTERACTION/merging/scenario.py
```diff
@@ -1,12 +1,21 @@
...
 
 from smarts.sstudio import types as t
 from smarts.sstudio import gen_scenario
 
 
+traffic_histories = [
+    t.TrafficHistoryDataset(
+        name=f"traffic_history_{hd}",
+        source_type="INTERACTION",
+        input_path=None,  # for example: f"./traffic_history_{hd}.txt"
```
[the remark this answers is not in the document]
RutvikGupta: Is this for when users want to build the scenario without overwriting the history dataset?

smarts/sstudio/genhistories.py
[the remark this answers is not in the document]
Gamenot: Interesting, I assume then that peachtree and lanker are newer? 

I suppose my only issue is that this mixes in some extra implementation implications into the name rather than being just a "type".

smarts/sstudio/genhistories.py
```diff
@@ -1032,24 +1093,25 @@ def _check_args(args) -> bool:
...
     if args.y_offset:
         dataset_spec["y_offset"] = args.y_offset
 
-    source = dataset_spec.get("source", "NGSIM")
-    if source == "NGSIM" or source == "NGSIM2":
+    source = dataset_spec.get("source_type", "NGSIM_highway")
+    if source.startswith("NGSIM"):
         dataset = NGSIM(dataset_spec, args.output)
     elif source == "Waymo":
         dataset = Waymo(dataset_spec, args.output)
```
RutvikGupta, line 1117: This and the rest of the lines seem redundant after the addition of `import_dataset` function. Maybe `import_dataset` should be modified to allow passing in both,  `TrafficHistoryDataset` object and user arguments.

smarts/sstudio/genhistories.py
sah-huawei: Yeah, I figure we can remove it at the same time we remove the deprecated yaml-file support (i.e., in our next major version, 0.7).

scenarios/INTERACTION/merging/scenario.py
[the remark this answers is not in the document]
sah-huawei: This is just because we aren't including the dataset itself in our repo, but we do want to include the `scenario.py`.  Our Makefile which is used by our CI does `scl scenario build-all ...`, so we need a way to not have it fail when the dataset isn't present.  So that's why we need the `None` option.

smarts/sstudio/genhistories.py
sah-huawei: Yes.  For the time being...  I think in v0.7, we should stop supporting them entirely and then this can be removed.

smarts/sstudio/genscenario.py
sah-huawei: oh, good point!
although it just worked in my testing, so maybe not... hmm, I don't quite understand why it sometimes works like this!

smarts/sstudio/genhistories.py
sah-huawei: Yeah, it is redundant, I agree.  I mainly left it "as is" to continue to support the deprecated yaml-file approach.  But when we remove that, I had planned to change the `__main__` path to just call `import_dataset()` so that `gehistories.py` can continue to be used either as a script or as a source-code file.

smarts/sstudio/genhistories.py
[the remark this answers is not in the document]
sah-huawei: I agree it's not ideal.  I conceptualized the `_city` and `_highway` as "sub types".    But I might instead try to add something that determines this subtype from the number of fields in the first row of the file (and then just use `NGSIM` as the type).

smarts/sstudio/genscenario.py
RutvikGupta: Some users complained about relative import errors on docker. In my testing i didn't encounter any. Not sure if its based on how your repo is setup.

Gamenot: I doubt that this would cause a problem since it is contained directly within the package directory tree.

smarts/sstudio/genhistories.py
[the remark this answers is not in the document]
sah-huawei: OK, I added code to infer the "subtype" from the first line of the data file, so now we can just use "NGSIM" instead of "NGSIM_city" or "NGSIM_highway".

smarts/sstudio/genscenario.py
sah-huawei: cool, updated

scenarios/NGSIM/i80/scenario.py
```diff
@@ -1,15 +1,27 @@
...
 
+traffic_histories = [
+    t.TrafficHistoryDataset(
+        name=f"i80_{hd}",
+        source_type="NGSIM_highway",
+        input_path=None,  # for example: f"./trajectories-{hd}.txt"
+        x_margin_px=60.0,
+        swap_xy=True,
+        flip_y=True,
+        map_bbox=BoundingBox((0.0, -3.0), (425.00, 25.02)),
```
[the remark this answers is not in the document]
sah-huawei: OK, I've now removed `map_bbox` from the descriptor in favour of `filter_off_map`.

Issue #1407, opened 2022-04-26
API for forwarding ngsim simulation


qianyi-sun:
This is a request from @YuechengLiu, which is about adding an API in SMARTS like the following:

```
class SMARTS:
    def _reset(scenario, start_time=None):
        if start_time:
            self._elapsed_sim_time = start_time
        else:
            self._elapsed_sim_time = 0
```

qianyi-sun:
@YuechengLiu, we've discussed this issue and please leave a comment to correct us if the following understanding is wrong: 

so this request is for allowing the simulation of ngsim data to forward to a specific time step, in order to skip some `sim_time` from the beginning, is this right?

YuechengLiu:
> @YuechengLiu, we've discussed this issue and please leave a comment to correct us if the following understanding is wrong:
> 
> so this request is for allowing the simulation of ngsim data to forward to a specific time step, in order to skip some `sim_time` from the beginning, is this right?

Exactly. And we can discuss about this issue in detail later.

YuechengLiu:
In the case that some of of vehicles are controlled by user and the others are loaded from traffic histories, (just like what has been done in: https://github.com/huawei-noah/SMARTS/blob/master/examples/history_vehicles_replacement_for_imitation_learning.py)
.

I find that the SMARTS._reset() are **not able to get the observations of ego vehicles immediately**, and SMARTS while keep calling SMARTS._step() until ego obeservations are returned. 
https://github.com/huawei-noah/SMARTS/blob/3ef5650b04ac6fb7145cf4e23d5534d73e0929fc/smarts/core/smarts.py#L434
This process will last for 2 steps ~ 20steps。During the procedure, the internal clock (SMARTS._elapsed_sim_time) is still keep going，and the vehicles (which are not controlled by user) provided by the traffic provider is loaded according to the new elapsed_sim_time. However, the state of ego vehicles are still old and **can not match** its initial state from the traffic history under the new internal clock.

Gamenot:
## Understanding

I understand the issues to be the following:
- Simulation time increments during reset.
- Initially specified ego state becomes old during that increment (or are emitted at a location later where they should have captured a vehicle)
- Wasteful time spent simulating in reset.

So your request is to set the jump the simulation time at beginning of reset in an attempt to skip all steps before ego observations start.

## Issue

Then the problem is the following even if we get the `traffic_history_provider` to move to a different time:

```python
>>> ## assuming first agent mission start time is 0
[...]

PR #1412, opened 2022-04-28
Update SMARTS reset with start time and fix dt dependency.

Gamenot:
This provides the ability to move the simulator ahead on reset. And fixes an issue where the Trap manager was using time deltas instead of the current simulation time.

See #1407 for context.


qianyi-sun (approved):

smarts/core/smarts.py
```diff
@@ -362,7 +362,7 @@ def done_agent_ids(dones):
         self._agent_manager.teardown_social_agents(agents_to_teardown)
         self._teardown_vehicles(vehicles_to_teardown)
 
-    def reset(self, scenario: Scenario) -> Dict[str, Observation]:
+    def reset(self, scenario: Scenario, start_time=0) -> Dict[str, Observation]:
```
saulfield, line 365: Perhaps it's worth adding a type annotation for the public interface? I see there are a few other places in this diff that are missing them as well, but I'll leave that up to you.
```suggestion
    def reset(self, scenario: Scenario, start_time: float = 0.0) -> Dict[str, Observation]:
```

saulfield (approved):

RutvikGupta (approved):

YuechengLiu:
Hello @Gamenot , Thanks for your work!

How should I use this new inferface?  
Do I need to set the start_time of `ego_missions` to 0 with the new inferface?

YuechengLiu:
Maybe SMARTS should provide a inferface to discover the mission of a vehicle for any given start_time (not only for the first-seen-time of the vehicle).

Gamenot:
> Hello @Gamenot , Thanks for your work!
> 
> How should I use this new inferface? Do I need to set the start_time of `ego_missions` to 0 with the new inferface?
> 
> Does the interface change have an impact on existing examples? Such as https://github.com/huawei-noah/SMARTS/blob/master/examples/history_vehicles_replacement_for_imitation_learning.py

@YuechengLiu This interface in its current form should have no effect on the current history replacement example. It would only have an effect if you change the `start_time` to something other than 0. The `ego_missions` start time should be set to at or after the `start_time` provided to reset with this changeset or they may be skipped (and a warning given.)

The exact interface implications are the following:

```python
assert all(start_time <= (m.start_time + m.entry_tactic.wait_to_hijack_limit_s) for m in missions)
# Warning occurs with skipped missions that would run out of patience before `start_time` 
smarts.reset(scenario, start_time = start_time) 
```

YuechengLiu:
@Gamenot 

https://github.com/huawei-noah/SMARTS/blob/3ef5650b04ac6fb7145cf4e23d5534d73e0929fc/examples/history_vehicles_replacement_for_imitation_learning.py#L206

According to my understanding, for example, we should remove the line here. Is this right?

Gamenot:
> Maybe SMARTS should provide an interface to discover the mission of a vehicle for any given start_time (not only for the first-seen-time of the vehicle).

A new request seems to be a different issue. I assume the intention is something like the following:

```python
active_vehicle_ids = traffic_history.vehicle_ids_active_between(start_time, end_time)
used_vehicle_ids = custom_filter(active_vehicle_ids)
missions = {
    vehicle_id: mission_between_time(
      traffic_history,
      vehicle_id,
      start_time, # could be different value
      end_time # could be different value
    ) for vehicle_id in used_vehicle_ids
}
assert all((
        m.start_time == start_time 
        && isinstance(m.goal, PositionalGoal)
    ) for m in vehicle_missions.values()
)
```

Gamenot:
I will do some additional tests tomorrow morning before merging this PR.

YuechengLiu:
> > Maybe SMARTS should provide an interface to discover the mission of a vehicle for any given start_time (not only for the first-seen-time of the vehicle).
> 
> A new request seems to be a different issue. I assume the intention is something like the following:
> 
> ```python
> active_vehicle_ids = traffic_history.vehicle_ids_active_between(start_time, end_time)
> used_vehicle_ids = custom_filter(active_vehicle_ids)
> missions = {
>     vehicle_id: mission_between_time(
>       traffic_history,
>       vehicle_id,
>       start_time, # could be different value
>       end_time # could be different value
>     ) for vehicle_id in used_vehicle_ids
> }
> assert all((
>         m.start_time == start_time 
>         && isinstance(m.goal, PositionalGoal)
>     ) for m in vehicle_missions.values()
> )
> ```

This is exactly what I need.

smarts/core/smarts.py
```diff
@@ -388,8 +393,13 @@ def reset(self, scenario: Scenario) -> Dict[str, Observation]:
         self._log.error(f"Failed to successfully reset after {tries} times.")
         raise first_exception
 
-    def _reset(self, scenario: Scenario):
+    def _reset(self, scenario: Scenario, start_time):
```
sah-huawei, line 396: type hint

sah-huawei (commented):
Should we use this in the `reset()` call in `history_vehicles_replacement_for_imitation_learning.py`?  (Currently, it uses the `history_start_time` variable to shift trajectories in time, but it might be cleaner to just start the simulation at the appropriate time.)

Gamenot:
> > > Maybe SMARTS should provide an interface to discover the mission of a vehicle for any given start_time (not only for the first-seen-time of the vehicle).
> > ...
> 
> This is exactly what I need.

I have taken a look into it and I see why it is needed during testing. The exact interface I am going for right now is the following:

```python3
traffic_history = scenario.traffic_history
active_vehicles: Dict[id_str, HistoryVehicleSlice] = traffic_history.vehicle_slice_in_range(start_time, end_time)
# e.g. filter: v.end_time - v.start_time > 6 and v.start_time == start_time
used_vehicle_ids = custom_filter(active_vehicles) 
missions = {
    vehicle_id: scenario.history_mission_for_time(
      vehicle_id,
      active_vehicles[vehicle_id].start_time, # could be different value
      active_vehicles[vehicle_id].end_time, # value > start_time
    ) for vehicle_id in used_vehicle_ids
}
```

PR #1417, opened 2022-05-02
Traffic history mission window

Gamenot:
- Traffic history vehicle time window query and mission conversion
TODO:
- [x] Clean up temporary options in example.
- [x] Remove extra queries.
- [x] Consider better interface.
- [x] Rename slice to window.


smarts/core/traffic_history.py
```diff
@@ -230,6 +285,85 @@ def vehicles_active_between(
...
+                   ) AS D ON V.id = D.vehicle_id
+                   JOIN (
+                    SELECT
+                        vehicle_id,
+                        sim_time,
+                        speed,
+                        position_x,
+                        position_y,
+                        heading_rad
+                    FROM (SELECT * from Trajectory WHERE sim_time < ? ORDER BY sim_time ASC)
```
Gamenot, line 310: Order this inner trajectory by simulation time so that it gets addressed in that order when next processed.

smarts/core/traffic_history.py
```diff
@@ -230,6 +285,85 @@ def vehicles_active_between(
...
+                   ) AS E ON V.id = E.vehicle_id
+                   JOIN (
+                    SELECT
+                        vehicle_id,
+                        sim_time,
+                        speed,
+                        position_x,
+                        position_y,
+                        heading_rad
+                    FROM (SELECT * from Trajectory WHERE sim_time >= ? ORDER BY sim_time DESC)
```
Gamenot, line 321: Same here as before but with the earliest value in the window.

smarts/core/traffic_history.py
```diff
@@ -230,6 +285,85 @@ def vehicles_active_between(
...
+    ):
+        """Find all vehicles active between the given history times."""
+        query = """SELECT V.id, V.type, V.length, V.width, V.height,
+                          S.position_x, S.position_y, S.heading_rad, S.speed, D.avg_speed,
+                          S.sim_time, E.sim_time,
+                          E.position_x, E.position_y, E.heading_rad
+                   FROM Vehicle AS V
+                   INNER JOIN ( 
+                    SELECT vehicle_id, AVG(speed) as "avg_speed"
+                    FROM (SELECT vehicle_id, sim_time, speed from Trajectory WHERE sim_time >= ? AND sim_time < ?)
```
Gamenot, line 299: I figured average speed could be useful.

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -115,7 +119,32 @@ def main(
...
+            nonlocal exists_at_or_after
+            window = 4
+            return (
+                v
+                for v in vehs
+                if v.average_speed > 3
+                and abs(v.start_time) - exists_at_or_after < window
+            )
+
+        used_vehicles = custom_filter(active_vehicles)
```
Gamenot, line 140: I note that this filter is useful. I think I could pass it as an argument `history_mission_for_vehicle_slice(..)` to avoid having to go through `traffic_history.vehicle_slice_in_range(..)` first.

Final interface could be shortened to:
```python
veh_missions = {
    str(veh_history_window.vehicle_id): veh_history_window
    for veh_time_slice in scenario.history_missions_for_window(
        exists_at_or_after, ends_before, minimum_window, custom_filter=speed_and_time
    )
}
```

```

**Comment by Gamenot** [comment]
Follows up #1412.

**Comment by YuechengLiu** [comment]
great work!

File: smarts/core/traffic_history.py [hunk 5]
```diff
@@ -262,12 +399,10 @@ def random_overlapping_sample(
         sample_start_time = vehicle_start_times[choice]
         sample_end_time = self.vehicle_final_exit_time(choice)
         while len(sample) < k:
-            choices = list(
-                self.vehicle_ids_active_between(sample_start_time, sample_end_time)
-            )
+            choices = list(vehicle_start_times.keys())
             if len(choices) <= len(sample):
                 break
-            choice = str(random.choice(choices)[0])
```
Gamenot, line 270: There did not seem to be a reason for a re-query of the vehicles.

qianyi-sun (approved):

YuechengLiu:
While calling `Scenario.history_missions_for_window`, an empty list is returned. Are there any bugs in this inferface?

Gamenot:
> While calling `Scenario.history_missions_for_window`, an empty list is returned. Are there any bugs in this interface?

From my tests it appears to be working assuming there are vehicles in the window between `exists_at_or_after` and `ends_before` that leave the map before `ends_before`.

What window are you calling with? I think perhaps I should not filter the vehicles to those that strictly end before `ends_before`. I can loosen up that to just cap to that value.

YuechengLiu:
> > While calling `Scenario.history_missions_for_window`, an empty list is returned. Are there any bugs in this interface?
> 
> From my tests it appears to be working assuming there are vehicles in the window between `exists_at_or_after` and `ends_before` that leave the map before `ends_before`.
> 
> What window are you calling with? I think perhaps I should not filter the vehicles to those that strictly end before `ends_before`. I can loosen up that to just cap to that value.

I am calling `scenario.history_mission+for_window(exist_at_or_after=242.6, end_before=10000.0, mininum_vehicle_vehicle_window=0.0, filter=lambda x: x)` on ngsim_i80 environment. It is obvious that there should exist some vehicles during this time, and this start_time (242.6) works using the inferface implemented by myself.

Gamenot:
> > > While calling `Scenario.history_missions_for_window`, an empty list is returned. Are there any bugs in this interface?
> > ...
> 
> I am calling `scenario.history_mission_for_window(exist_at_or_after=242.6, end_before=10000.0, mininum_vehicle_vehicle_window=0.0, filter=lambda x: x)` on ngsim_i80 environment. It is obvious that there should exist some vehicles during this time, and this start_time (242.6) works using the inferface implemented by myself.

I will try out `ngsim_i80` if it works for me I will assume that my fear that my previous concern about the SQL query is valid and I will need to filter the data using a different query: https://github.com/huawei-noah/SMARTS/pull/1417#discussion_r862607575

Gamenot:
I was not seeing problems testing with `ngsim_us101` it might be that there is a difference in fields.

Gamenot:
> I am calling `scenario.history_mission_for_window(exist_at_or_after=242.6, end_before=10000.0, mininum_vehicle_vehicle_window=0.0, filter=lambda x: x)` on ngsim_i80 environment. It is obvious that there should exist some vehicles during this time, and this start_time (242.6) works using the inferface implemented by myself.

I realize one potential cause of the issue is that `scenario.history_mission_for_window(..)` returns a generator object that would be consumed if you iterate over it before turning it into a list.

Gamenot:
I am not seeing problems with `ngsim_us101` on my environment, I will check with a different environment.

YuechengLiu:
Thanks for your replay, and I will do more tests on my env.

Gamenot:
For clarity, I have tested `Scenario.history_missions_for_window(exist_at_or_after=242.6, end_before=10000.0, mininum_vehicle_vehicle_window=0.0, ..)` for the `i-80` trajectories with: `trajectories-0400-0415.txt`, `trajectories-0500-0515.txt`, `trajectories-0515-0530.txt`.

Gamenot:
Would you happen to be using `macos`?

YuechengLiu:
I am running on linux (ubuntu 16.04).

Gamenot:
I just finished testing with an Ubuntu 16.04 image on a separate environment, it appeared to work with `i-80`.

If it is not an issue on your end it is likely the problem has to do with the SQL query filtering in a different order due to a different python version or query call order. I will work to break the SQL query up tomorrow.

YuechengLiu:
> I just finished testing with an Ubuntu 16.04 image on a separate environment, it appeared to work with `i-80`.
> 
> If it is not an issue on your end it is likely the problem has to do with the SQL query filtering in a different order due to a different python version or query call order. I will work to break the SQL query up tomorrow.

Thanks for the help, and my python version is 3.7.13.

Gamenot:
@Gamenot Self note: https://stackoverflow.com/a/12608528 approach 2 could work to ensure determinism because `sqlite3` is version `2.6.0` for `python 3.7->3.9`.

Gamenot:
@YuechengLiu I have addressed the possible execution order issue by forcing the order in the SQL query.

Could you try the new version?

smarts/core/traffic_history.py
```diff
@@ -230,6 +297,127 @@ def vehicles_active_between(
...
+                assert r.start_time >= exists_at_or_after
+                assert r.end_time <= ends_before
+                assert r.end_position_x not in seen_pos_x
+                seen_pos_x.add(r.end_position_x)
+                seen.add(r.vehicle_id)
+                rs.append(r)
+            self._log.info(f"Num vehicles suiting time criteria: {len(rs)}")
+            return rs
+
+        thvtss = None
```
saulfield, line 416: I feel like we could come up with a better variable name for this :)

smarts/core/traffic_history.py
```diff
@@ -230,6 +297,127 @@ def vehicles_active_between(
...
+        thvtss = debug(rows)
+
+        return thvtss or (_slice_from_row(row) for row in rows)
```
saulfield, line 419: Is this just to allow easy toggling on/off of the debugging path? It's a bit confusing because one is a list and one is a generator.

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -111,19 +114,36 @@ def main(
...
+            nonlocal exists_at_or_after
+            window = 4
+            return (
+                v
+                for v in vehs
+                if v.average_speed > 3
+                and abs(v.start_time) - exists_at_or_after < window
+            )
+
+        # pytype: disable=attribute-error
```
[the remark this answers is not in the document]
Gamenot: It is frustrating but `pytype` complains about the optional nature of the vehicle spec even though we ensure it exists in the method.

smarts/core/traffic_history.py
Gamenot: Yes, I think it is badly addressed, I will swap to including the debugging in the generator instead of resolving it to a list.

smarts/core/traffic_history.py
Gamenot: Or eliminate it. ;)

examples/history_vehicles_replacement_for_imitation_learning.py
```diff
@@ -111,19 +114,36 @@ def main(
...
+        veh_missions = {
+            mission.vehicle_spec.veh_id: mission
```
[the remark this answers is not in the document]
saulfield: Ah, my mistake. I parsed this as a type hint for some reason.

saulfield (approved):

YuechengLiu:
It still can't work. My python version is `3.7.13` and my sqlite version is `3.38.2`.

Gamenot:
@YuechengLiu I added a method `traffic_history.last_seen_vehicle_time()` does that turn up as `None` if called with that scenario? Does the original `traffic_history.discover_missions_of_traffic_histories()` still work if you revert back to that?

I am perplexed that there is no error associated with the failure.

> It still can't work. My python version is `3.7.13` and my sqlite version is `3.38.2`.

OK, I am going to try testing in those particular versions. I was using `3.7.7` and `3.22.0` respectively.

Gamenot:
Have you rebuilt the traffic history scenario recently? I know there was an update to how it was generated 9 days ago in #1403.

Gamenot:
> It still can't work. My python version is `3.7.13` and my sqlite version is `3.38.2`.

I have now tested these versions. The issue does not seem to show up.

YuechengLiu:
@Gamenot  Thanks for the help!, it works this time. I forget to remove the debug line.
```python
vehicle_windows = self._traffic_history.vehicle_windows_in_range(...)
print(f"{list(vehicle_windows)}")
# after the print line, vehicle_windows becomes empty.
```
I have removed the print line and it works well this time, although I am not quite sure why it happened.

Gamenot:
> @Gamenot Thanks for the help!, it works this time. I forget to remove the debug line.
> 
> ```python
> vehicle_windows = self._traffic_history.vehicle_windows_in_range(...)
> print(f"{list(vehicle_windows)}")
> # after the print line, vehicle_windows becomes empty.
> ```
> 
> I have removed the print line and it works well this time, although I am not quite sure why it happened.

Sorry, yes, the `traffic_history.vehicle_windows_in_range(..)` method returns a generator which is an iterater that consumes each element as it is accessed. I can make the interface return a list instead but I was avoiding it for scaling reasons in the case that histories have hundreds of thousands of vehicles.

YuechengLiu:
Another request. It would be better to provide an interface to get the vehicle_window for **a specific vehicle** (given the vehicle_id). I find that it is very slow to get all vehicle missions first and then select the needed vehicle from them).

Gamenot:
> Another request. It would be better to provide an interface to get the vehicle_window for **a specific vehicle** (given the vehicle_id). I find that it is very slow to get all vehicle missions first and then select the needed one from them).

I have added a utility that will do that which takes the form:

```python3
vehicle_window: TrafficHistoryVehicleWindow | None = traffic_history.vehicle_window_by_id(vehicle_id)
```

YuechengLiu:
@Gamenot Thanks for your work! The request is not urgent to be finished.

It would be better to provide an interface like this,
```python
window = scenario.history_mission_for_window_by_id(vehicle_id, exists_at_or_after, ends_before, mininum_vehicle_window)
``` 

We can simply call `scenario.history_mission_for_window()` and pass a `filter` to filter out the needed one, but it is too time-consuming.

So maybe another SQL Query is needed to get the history information for the specified vehicle

Gamenot:
> @Gamenot Thanks for your work! The request is not urgent to be finished.
> 
> It would be better to provide an interface like this,
> 
> ```python
> window = scenario.history_mission_for_window_by_id(vehicle_id, exists_at_or_after, ends_before, mininum_vehicle_window)
> ```
> 
> We can simply call `scenario.history_mission_for_window()` and pass a `filter` to filter out the needed one, but it is too time-consuming.
> 
> So maybe another SQL Query is needed to get the history information for the specified vehicle

@YuechengLiu I think I may be misunderstanding your intention, if you know the `vehicle_id` why do you need a time window, wouldn't that increase the chances of missing the time that the vehicle is active? 

It would mean that you would need both the `vehicle_id` and time information before accessing the vehicle. 

I have a list of non-exclusive statements here to try and understand single vehicle selection:

- a) `get 10 seconds window for vehicle #4`
- b) `get window of simulation time between 25.4 and 35.4 for vehicle #4`
- c) `get entire traffic history window for vehicle #4`
- d) `get a 10 second window for a random vehicle between time 40.1 and 60.1`
- e) `save ids of vehicles that perform a behaviour for later reference (e.g {'345, '844', '210'} failed lane change)`

YuechengLiu:
@Gamenot 
b) and e) are closer to my needs.

For example, I have filtered out some special cases from the traffic histories or environment interactions (lane change, cut-in, fail cases, etc.), and I want to train a ego vehicle in these special cases. So I just save the start_time and end_time and the corresponding vehicle_id of these special cases and replay them during training (not start from the first-seen-time of the vehicles). 

Only a little of vehicles is ego-controlled (maybe only one in most of scenarios), and the controlled vehicles usually hold different start_time and end_time.


It's ok to keep the current API, since I have realized the function by myself and maybe most of users will not use the function.

Gamenot:
@YuechengLiu I have patched the delayed history problem here.

Issue #1423, opened 2022-05-04
Randomize scenario variations for online RL training


qianyi-sun:
Generally, when we `scl scenario build` the scenarios, the variations such as traffic density, vehicle positions in traffic... would be fixed. It might not be good for RL training if the scenarios are fixed for every episode. 

If possible, maybe we can do something like, for each scenario (left turn in an intersection, merging...) we set 3 to 5 road structure configurations and randomize the road structure as well as the traffic variations, then we would have different traffic difficulties and road structures for each episode.

PR #1418, opened 2022-05-02
add 2 basic cut-off/intersection scenarios for sb3 training 

qianyi-sun:

PR #1424, opened 2022-05-05
support random seed in examples

sah-huawei:
Cf Issue #1423, this hooks up the random seed that might be specified on the command line in various examples.  (It was often ignored before.)

This won't completely address #1423, but it at least allows for Patrick and Soheil to see the types of variation they can get by playing with the `seed`.

### DIFF
--- a/examples/argument_parser.py
+++ b/examples/argument_parser.py
@@ -32,7 +32,12 @@ def default_argument_parser(program: Optional[str] = None):
     parser.add_argument(
         "--headless", help="Run the simulation in headless mode.", action="store_true"
     )
-    parser.add_argument("--seed", type=int, default=42)
+    parser.add_argument(
+        "--seed",
+        help="A seed to use for all random numbers in the simulation.  Using the same seed value should lead to the same results.",
+        type=int,
+        default=42,
+    )
     parser.add_argument(
         "--sim-name",
         help="Simulation name.",
--- a/examples/egoless.py
+++ b/examples/egoless.py
@@ -1,7 +1,7 @@
 import logging
 
 import gym
-from typing import Sequence
+from typing import Optional, Sequence
 
 from examples.argument_parser import default_argument_parser
 from smarts.core.utils.episodes import episodes
@@ -14,7 +14,7 @@ def main(
     headless: bool,
     num_episodes: int,
     seed: int,
-    max_episode_steps: int = None,
+    max_episode_steps: Optional[int] = None,
 ):
     env = gym.make(
         "smarts.env:hiway-v0",
--- a/examples/egoless.py
+++ b/examples/egoless.py
@@ -1,6 +1,7 @@
 import logging
 
 import gym
+from typing import Sequence
 
 from examples.argument_parser import default_argument_parser
 from smarts.core.utils.episodes import episodes
@@ -8,13 +9,20 @@ from smarts.core.utils.episodes import episodes
 logging.basicConfig(level=logging.INFO)
 
 
-def main(scenarios, headless, num_episodes, max_episode_steps=None):
+def main(
+    scenarios: Sequence[str],
+    headless: bool,
+    num_episodes: int,
+    seed: int,
+    max_episode_steps: int = None,
+):
     env = gym.make(
         "smarts.env:hiway-v0",
         scenarios=scenarios,
         agent_specs={},
         headless=headless,
         sumo_headless=True,
+        seed=seed,
     )
 
     if max_episode_steps is None:
@@ -39,4 +47,5 @@ if __name__ == "__main__":
         scenarios=args.scenarios,
         headless=args.headless,
         num_episodes=args.episodes,
+        seed=args.seed,
     )
--- a/examples/multi_agent.py
+++ b/examples/multi_agent.py
@@ -1,7 +1,7 @@
 import pathlib
 
 import gym
-from typing import Sequence
+from typing import Optional, Sequence
 
 from examples.argument_parser import default_argument_parser
 from smarts.core.agent import Agent
@@ -24,7 +24,7 @@ def main(
     headless: bool,
     num_episodes: int,
     seed: int,
-    max_episode_steps: int = None,
+    max_episode_steps: Optional[int] = None,
 ):
     agent_specs = {
         agent_id: AgentSpec(
--- a/examples/multi_agent.py
+++ b/examples/multi_agent.py
@@ -1,6 +1,7 @@
 import pathlib
 
 import gym
+from typing import Sequence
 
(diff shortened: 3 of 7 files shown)


Gamenot (approved):

qianyi-sun (approved):

Adaickalavan:
I think this may not be needed.

a) In order to cycle through different maps upon each episode reset, we could instantiate the smarts env with a list of scenarios, instead of a single scenario.
b) If a given scenario has multiple traffic flows defined in it, the traffic flow is cycled upon each episode reset call.

Variations in map and traffic flow, as yielded by (a) and (b) above, is already achievable without changing the seed.

The general examples in smarts should be lean and easy for users to get started with, and hence the `seed` parameter was previously removed. Re-adding `seed` parameter bloats these examples and users may not necessarily understand what it actually does. It is better for us to explain `seed` usage in the documents for interested users. 

For the various smarts based scenarios being designed for the competition, I believe we should build the appropriate `scenario.py` and its corresponding `env` file together in tandem. For example, see PR #1274 for `intersection-v0` environment. Next, we could have `cut-off-v0` and `merge-v0` task environments. For competition purposes, these task environments could have a configurable seed as input parameter to allow us to test the RL models. 

For general use beyond the competition purposes, these task environments, namely, `intersection-v0`, `cut-off-v0`, and `merge-v0`, should not allow configurable seed in order to ensure that all users will be building and comparing their models on the same standard task environments. 

I have created a `multi_intersection_env` example. Here, two different maps and multiple different traffic flows are cycled through upon each episode reset call. The two maps consists of (i) `1lane_left_turn` and (ii) `1lane_to_2lane_left_turn` scenarios. The `1lane_to_2lane_left_turn` scenario is simply the new scenario proposed in PR #1418. 
```
git checkout sb3-1
python3.7 ./examples/multi_intersection.py 
```

sah-huawei:
> I think this may not be needed.
> 
> a) In order to cycle through different maps upon each episode reset, we could instantiate the smarts env with a list of scenarios, instead of a single scenario. b) If a given scenario has multiple traffic flows defined in it, the traffic flow is cycled upon each episode reset call.
> 
> Variations in map and traffic flow, as yielded by (a) and (b) above, is already achievable without changing the seed.
> 
> The general examples in smarts should be lean and easy for users to get started with, and hence the `seed` parameter was previously removed. Re-adding `seed` parameter bloats these examples and users may not necessarily understand what it actually does. It is better for us to explain `seed` usage in the documents for interested users.

I agree with this point (keeping the examples simple) but did not realize this was why `seed` was (partially) removed before, as it was still left as an option in `argument_parser.py` (which all of the examples share).  Should we just remove it from there too then?  (As it is now, specifying a seed on the command line will not have had any effect for most of the examples.)
 
> For the various smarts based scenarios being designed for the competition, I believe we should build the appropriate `scenario.py` and its corresponding `env` file together in tandem. For example, see PR #1274 for `intersection-v0` environment. Next, we could have `cut-off-v0` and `merge-v0` task environments.
> 
> Situations for variable seed number, include: (i) For competition purposes, these task environments could have a configurable seed as input parameter to allow us to test the RL models. (ii) Different seed number is needed when running parallel environments. But this should be seamlessly taken care by the vectorized env wrapper which is used to run the multiple env in parallel.
> 
> I have created a `multi_intersection_env` example. Here, two different maps and multiple different traffic flows are cycled through upon each episode reset call. The two maps consists of (i) `1lane_left_turn` and (ii) `1lane_to_2lane_left_turn` scenarios. The `1lane_to_2lane_left_turn` scenario is simply the new scenario proposed in PR #1418.
> 
> ```
> git checkout sb3-1
> python3.7 ./examples/multi_intersection.py 
> ```

Could you please add a variation comment to Issue #1424?  If that's satisfactory to Patric and Soheil, I'm fine with not merging this PR.  (Instead, I would then vote for removing the `--seed` command-line argument entirely.)

qianyi-sun:
> > I think this may not be needed.
> > a) In order to cycle through different maps upon each episode reset, we could instantiate the smarts env with a list of scenarios, instead of a single scenario. b) If a given scenario has multiple traffic flows defined in it, the traffic flow is cycled upon each episode reset call.
> > Variations in map and traffic flow, as yielded by (a) and (b) above, is already achievable without changing the seed.
> > The general examples in smarts should be lean and easy for users to get started with, and hence the `seed` parameter was previously removed. Re-adding `seed` parameter bloats these examples and users may not necessarily understand what it actually does. It is better for us to explain `seed` usage in the documents for interested users.
> 
> I agree with this point (keeping the examples simple) but did not realize this was why `seed` was (partially) removed before, as it was still left as an option in `argument_parser.py` (which all of the examples share). Should we just remove it from there too then? (As it is now, specifying a seed on the command line will not have had any effect for most of the examples.)
> 
> > For the various smarts based scenarios being designed for the competition, I believe we should build the appropriate `scenario.py` and its corresponding `env` file together in tandem. For example, see PR #1274 for `intersection-v0` environment. Next, we could have `cut-off-v0` and `merge-v0` task environments.
> > Situations for variable seed number, include: (i) For competition purposes, these task environments could have a configurable seed as input parameter to allow us to test the RL models. (ii) Different seed number is needed when running parallel environments. But this should be seamlessly taken care by the vectorized env wrapper which is used to run the multiple env in parallel.
> > I have created a `multi_intersection_env` example. Here, two different maps and multiple different traffic flows are cycled through upon each episode reset call. The two maps consists of (i) `1lane_left_turn` and (ii) `1lane_to_2lane_left_turn` scenarios. The `1lane_to_2lane_left_turn` scenario is simply the new scenario proposed in PR #1418.
> > ```
> > git checkout sb3-1
> > python3.7 ./examples/multi_intersection.py 
> > ```
> 
> Could you please add a variation of this comment to Issue #1424? If that's satisfactory to Patric and Soheil, I'm fine with not merging this PR. (Instead, I would then vote for removing the `--seed` command-line argument entirely.)

Thank you both! Adai just explained and showed me how his `multi_intersection,py` works. It looks good to me and I'm going to create more scenarios and envs (for merging, straight roads...) according to this and discuss to Soheil see if this is good enough.

Gamenot:
> Could you please add a variation of this comment to Issue #1424? If that's satisfactory to Patric and Soheil, I'm fine with not merging this PR. (Instead, I would then vote for removing the `--seed` command-line argument entirely.)

Truthfully if you need a different random value you could run the example like `PHYTHONHASHSEED=882 python single_agent.py ...`.

Adaickalavan:
@sah-huawei

I concur in removing the `seed` option from `examples/argument_parser.py` .

Gamenot:
We found no need for this change-set.

PR #1427, opened 2022-05-09
Intersection-v0

Adaickalavan:
+ Unprotected left turn task environment with all-way stop traffic, named `intersection-v0`, is added in `smarts/env/intersection_env.py`.
+ Since `gym.Space` does not support dataclass, `StdObs` type is changed from a dataclass to a dictionary.
+ Increased test coverage of `smarts/env/wrappers/format_obs.py`
+ Added an online RL example in `examples/rl/intersection`, for solving the `intersection-v0` environment, using PPO algorithm from Stable Baselines3 library. 
+ An accompanying Colab tutorial is also provided.
+ Old Stable Baselines3 based example is removed in favour of the new online RL example.


smarts/env/intersection_env.py
```diff
@@ -0,0 +1,196 @@
...
+
+    scenario = [
+        str(
+            pathlib.Path(__file__).absolute().parents[2]
+            / "scenarios"
+            / "intersections"
+            / "1lane_left_turn"
+        )
+    ]
+    build_scenario(scenario)
```
Gamenot, line 106: My concern here is that this scenario does not exist in the package version. We could add a `smarts.scenario` module for it.

saulfield (approved):

smarts/env/intersection_env.py
Adaickalavan: In order to enable users to easily instantiate an environment via
```
env = gym.make("smarts.env:intersection-v0")
```
we would need to automatically include and build the scenario. 

To do so, we may
1) include the scenario into smarts package
    + This pull request currently includes the relevant scenario into `smarts` package.
    https://github.com/huawei-noah/SMARTS/blob/41af666cbd26c9c0eb098c7d1f89ccc858b07913/MANIFEST.in#L6
2) include a `smarts.scenario` module
    + I believe we attempted this method previously, but it required folder structure changes and more `setup.py` files.
3) `curl` install the scenario from another github repository during the call to `gym.make()`
    + Alternatively, we may try this method.

Gamenot: 1) I think might not be viable, including a directory called `scenarios` has the possibility to interact badly with other modules.
2) If possible we can include `smarts.scenarios` such like `recursive-include smarts/scenarios/intersections/1lane_left_turn *.xml *.py`. In this case it would not require a new `setup.py`.
3) This seems reasonable, although it perhaps breaks packaging standards so it is only useful if we make it a resource standard within SMARTS.

Gamenot: I am not comfortable with merging this until the scenario issue is resolved.

Adaickalavan: I have repackaged intersection-v0 scenario to be located inside `smarts.scenarios` package.

Gamenot (approved):

PR #1429, opened 2022-05-13
Bugfix ULTRA failure

Gamenot:
- Restore use of the ULTRA benchmark on the current repository.

### DIFF
--- a/ultra/ultra/scenarios/generate_scenarios.py
+++ b/ultra/ultra/scenarios/generate_scenarios.py
@@ -92,8 +92,8 @@ def ego_mission_config_to_route(
         mission_end_lane_index = route_lanes[ego_mission_config["end"]] - 1
         mission_start_offset = (
             random.randint(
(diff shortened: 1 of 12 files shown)


ultra/ultra/scenarios/generate_scenarios.py
```diff
@@ -716,104 +718,118 @@ def build_scenarios(
...
-        ), "Test scenarios must have one ego mission."
-
-        prev_split = 0
-        main_seed_count = 0
-        # sort inverse by percents
-        intersection_types = level_config[mode]["intersection_types"]
-        intersections = sorted(
-            [
+    num_cpus = max(1, psutil.cpu_count(logical=True))
+    with Pool(num_cpus) as process_pool:
```
Gamenot, line 725: This change uses a pool instead to limit number of threads generated.

Gamenot:
These are just fixes to re-enable existing behaviour so I am merging as is.

PR #1443, opened 2022-05-26
Fix early merge case to avoid `off_route`

Gamenot:
- [x] Logic
- [x] Tested


Gamenot:
@Adaickalavan These are the changes I mentioned that are needed to fix the issue I mentioned if a vehicle were to try to cut-in early.

qianyi-sun (approved):

Issue #1155, opened 2021-12-07
Error! Vehicle 'car-flow-random-route' is not known.


Adaickalavan:
## **BUG REPORT**
**High Level Description**
SUMO released a new version 1.11.0 on 21 Nov 2021, which throws the following error when running with SMARTS:
`Error! Vehicle 'car-flow-random-route-bc8cc765-7266489842092764092--0-0.0' is not known.`

**SMARTS version**
0.4.18

**Steps to reproduce the bug**
Reproduce error with new sumo==1.11
```bash
$ git clone https://github.com/huawei-noah/SMARTS.git
$ git checkout sumo-version
$ docker build --no-cache --network=host --file=<path>/SMARTS/examples/Dockerfile.111 --tag=sumo_111 <path>/SMARTS
$ docker run --rm -it --network=host sumo_111
(container) $ sumo # Check sumo==1.11
(container) $ scl scenario build-all ./scenarios/loop/
(container) $ python3.7 ./examples/multi_agent.py ./scenarios/loop --headless
```

Compare with old sumo==1.10.0
```bash
$ git clone https://github.com/huawei-noah/SMARTS.git
$ git checkout sumo-version
$ docker build --no-cache --network=host --file=<path>/SMARTS/examples/Dockerfile.110 --tag=sumo_110 <path>/SMARTS
$ docker run --rm -it --network=host sumo_110
(container) $ sumo # Check sumo==1.10
(container) $ scl scenario build-all ./scenarios/loop/
(container) $ python3.7 ./examples/multi_agent.py ./scenarios/loop --headless
```

**Resulting and expected behaviour**
No error messages should be printed.

**System information**
- Operating system: Ubuntu 18.04
- Python version: 3.7
- SUMO version: 1.10 and 1.11
- Date when bug occurred: 7 Dec 2021

jieye-ericx:
hello,may I ask if you have solved this problem, I can only install sumo 1.15 using apt-get on ubuntu.

Adaickalavan:
Hi @jieye-ericx, 
+ sumo is now installed automatically via the `setup.py` file in our `smarts` package while pip installing the `smarts` package.
+ see `setup.py` file here: https://github.com/huawei-noah/SMARTS/blob/13bfeb91be46bebcbd5c832ff55e1bd8b20abeba/setup.py#L28
+ to manually install sumo, do `pip install eclipse-sumo==1.10.0`.
[...]

PR #1444, opened 2022-05-26
update sumo version in setup.py and changelog.md

qianyi-sun:
With the previous version of sumo `1.10.0`, when running the NGSIM simulation on `peachtree` map, it crashes every time at time step 23.80. Just tested the latest `1.13.0`. there's no crash anymore and it works just fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 # [Unreleased]
 
 ### Changed
+- Updated sumo to 1.13.0 to increase stability.
 - Updated license to 2022 version.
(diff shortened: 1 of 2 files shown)


Gamenot:
The crash might be related to the following: https://github.com/eclipse/sumo/issues/5647

Adaickalavan:
From the Github Actions page, https://github.com/huawei-noah/SMARTS/runs/6614710274?check_suite_focus=true, the failing test shows that upgrading the sumo version still does not fix the error `Error! Vehicle 'car-flow-random-route' is not known.` previously reported in issue #1155 .

Gamenot:
We should resolve this error and get this update merged.

PR #1454, opened 2022-05-31
Update Waymo scenario and utilities

saulfield:
- update the Waymo scenario to use the new support for Waymo maps instead of a generated SUMO map, as well as the new support for traffic histories in the scenario DSL
- fix and cleanup the `gen_sumo_map.py` script and move it to `smarts/waymo/` with the other Waymo utilities
- various documentation updates

### DIFF
--- a/examples/observation_collection_for_imitation_learning.py
+++ b/examples/observation_collection_for_imitation_learning.py
@@ -52,8 +52,10 @@ def main(script: str, scenarios: Sequence[str], headless: bool, seed: int):
 
     # Make sure we can use SUMO traffic for these scenarios
     all_sumo = Scenario.supports_traffic_simulation(scenarios)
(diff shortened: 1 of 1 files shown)


scenarios/waymo/README.md
```diff
@@ -47,13 +30,7 @@ This will build the map and generate the SQLite database of trajectory data (wit
...
 ## Using the traffic histories in SMARTS
 
-There are currently 2 examples that use traffic histories. We explain their use below. First, start the envision server:
-
-```bash
-scl envision start &
-```
-
-Visit http://localhost:8081/ in your browser to visualize the replay examples.
+There are currently 2 examples that use traffic histories. We explain their use below.
```
Gamenot, line 33: nit: I would prefer not being so specific.
```suggestion
There are examples that use traffic histories. We explain their use below.
```

Gamenot (approved):

PR #1461, opened 2022-06-09
Script to extract Waymo data for offline RL

saulfield:


smarts/waymo/waymo_data_extraction.py
```diff
@@ -0,0 +1,266 @@
...
+    for veh_id in valid_vehicles:
+        try:
+            smarts.attach_sensors_to_vehicles(AGENT_SPEC.interface, {veh_id})
+        except ControllerOutOfLaneException:
+            logger.warning(f"{veh_id} out of lane, skipped attaching sensors")
+            off_road_vehicles.add(veh_id)
+
+    # Get observations from each vehicle and record them
+    obs, _, _, _ = smarts.observe_from(list(valid_vehicles))
+    obs = std_obs_wrapper.observation(obs)
```
saulfield, line 104: This is the only part that still feels pretty hacky. I didn't think it was worth refactoring the `FormatObs` class for this PR, but we may want to in the future. Let me know if there is a better way to achieve this for this PR.

Adaickalavan: + Why did we not just use the `smarts.sensors.Observation` directly?
+ Why did we use the `StdObs` from `FormatObs` wrapper here to reformat the observation?
+ By using `FormatObs` wrapper here, we would be making this standard format mandatory for all users.

smarts/waymo/waymo_data_extraction.py
```diff
@@ -0,0 +1,310 @@
...
+    scenarios_dir: Path,
+    vehicle_ids: Optional[List[int]] = None,
+    max_time: float = 20.0,
+):
+    """Extract top-down RGB images and observations from a Waymo Motion Dataset scenario.
+    Args:
+        dataset_file (Path):
+            Path to the TFRecord file.
+        scenario_id (str):
+            ID for the scenario in the TFRecord file.
```
Adaickalavan, line 144: Can we extract Waymo data into Smarts observations for multiple scenarios in one go?

smarts/waymo/waymo_data_extraction.py
Adaickalavan: Do we have mission goals here?

saulfield:
Closing since this has been superseded by #1545.

PR #1462, opened 2022-06-11
[Feature] bubble capture takeover

Gamenot:
Utilities added to allow for taking over vehicles that pass into bubbles.

- [x] Test with traffic sim
- [x] Test with playback
- [x] All tests work


YuechengLiu:
Thanks for the work! @Gamenot 

Does the `Bubble` work in history-replay cases ? (e.g. NGSIM i80) And could you please provide a toy example for this case？(just like https://github.com/huawei-noah/SMARTS/blob/master/examples/history_vehicles_replacement_for_imitation_learning.py) Thanks！

Gamenot:
> Thanks for the work! @Gamenot
> 
> Does the `Bubble` work in history-replay cases ? (e.g. NGSIM i80) And could you please provide a toy example for this case？(just like https://github.com/huawei-noah/SMARTS/blob/master/examples/history_vehicles_replacement_for_imitation_learning.py) Thanks！

Hello @YuechengLiu , yes, the intention is to test for that today and put up an example.

Gamenot:
@YuechengLiu I have run into a setback with some edge cases around the visualisation. I am working out those currently.

YuechengLiu:
@Gamenot Thanks! 

And How can I use a moving bubble? (make it follow a specific ego vehicle)

Gamenot:
@YuechengLiu I have updated the example and provided an option for the bubble to alternatively follow a vehicle id. Following an agent id would be similar syntax but like:

```
    "i80": [
        Bubble(
            ...
            follow_actor_id=AGENT_ID,
            follow_offset=(0, 0),
            margin=10
        ),
    ],
```

examples/bubble_example.py
```diff
@@ -0,0 +1,237 @@
...
+            zone=PositionalZone(pos=(0, 0), size=(20, 100)),
+            actor=SocialAgentActor(
+                name="dummy0", agent_locator="examples:dummy_agent-v0"
+            ),
+            follow_vehicle_id="history-vehicle-314",
+            exclusion_prefixes=("history-vehicle-314",),
+            follow_offset=(0, 0),
+            margin=10
+        ),
+    ],
```
Gamenot, line 87: Referring to this.

YuechengLiu: ok, thanks, I will have a try.

examples/bubble_example.py
```diff
@@ -0,0 +1,237 @@
...
+
+        for episode in range(episodes):
+            logger.info(f"starting episode {episode}...")
+            smarts.reset(scenario, start_time=start_time)
+            traffic_history_provider: TrafficHistoryProvider = smarts.get_provider_by_type(TrafficHistoryProvider)
+            used_history_ids = set()
+
+            bubble_manager: BubbleManager = smarts._bubble_manager
+            bubbles = bubble_manager.bubbles
+            agent = UsedAgent(logger=logger)
```
YuechengLiu, line 159: It seems that the `agent` here is not used in this example. And I have some questions:

1. What is the vehicle being followed controlled by? (whose vehicle id="follow_vehicle_id"). Does it serve as a `social agent` or a `ego vehicle`？
2. If the vehicle being followed is a ego vehicle, should I call `scenario.set_ego_missions` and `smarts.switch_ego_agents`?

Gamenot: I will try and answer the questions.

> It seems that the `agent` here is not used in this example.

Yes and no, in this case we are operating with bubble captured vehicles. These cannot come through the normal smarts interface. This is more a mock-up of the utility but the interface needs to be resolved. There is a parallel effort ongoing for work with traffic https://github.com/huawei-noah/SMARTS/pull/1440 that will help with resolving that interface.

The agent is used here in a different way than before by overriding a social agent action:
https://github.com/huawei-noah/SMARTS/blob/55453bb77f53a35e1829ac44d8912f4519559809/examples/bubble_example.py#L167-L169

If the logging is turned on you get the following:
```bash
INFO:history-vehicles-replacement-example:starting episode 0...
INFO:history-vehicles-replacement-example:Agent act called first time.
INFO:history-vehicles-replacement-example:starting episode 1...
INFO:history-vehicles-replacement-example:Agent act called first time.
```

relating to:
https://github.com/huawei-noah/SMARTS/blob/55453bb77f53a35e1829ac44d8912f4519559809/examples/bubble_example.py#L53-L61

> 1. What is the vehicle being followed controlled by? (whose vehicle id="follow_vehicle_id"). Does it serve as a `social agent` or a `ego vehicle`？

There are two cases for this:
  - If using `follow_vehicle_id` this bubble will follow the vehicle with the given id. The followed vehicle could be any vehicle in the dataset regardless of if it is controlled by an `agent` or simply a data replay vehicle.
  - If using `follow_actor_id` this bubble will follow the agent with that ID when it shows up in the simulation.

The following was to prevent specific vehicles from being taken over by a social vehicle:
```python
exclusion_prefixes=("history-vehicle-314",),
```

> 2. If the vehicle being followed is a ego vehicle, should I call `scenario.set_ego_missions` and `smarts.switch_ego_agents`?

That would be the case at current. Using the `exclusion_prefixes` you can ensure that the vehicle is not already controlled by a social agent agent.

assuming you are taking over vehicles in the dataset using `scenario.set_ego_missions` & `smarts.switch_ego_agents` 

> 3. If the followed vehicle's  agent id is not provided by user (by `AgentInterface`) , we do not know its agent id, so it is unrealistic to define a bubble in advance.

I am not quite what to respond to since this is not a question but there is a design issue here that requires some explanation.
  - The current bubbles are defined in advance but adding them into the scenario dynamically during simulation is not yet supported (but easily implemented.) But since you would be considering making modifications to the sim state before reset (i.e. `scenario.set_ego_missions` & `smarts.switch_ego_agents`) it is possible to set the bubbles in the scenario at the same time.
```python
# TODO: interface to be improved.
scenario.bubbles.clear()
scenario.bubbles.extend([Bubble(..)])
scenario.set_ego_missions(..)
...
smarts.switch_ego_agents(..)
...
smarts.reset(scenario, ..)
```
  - Bubbles acquire vehicles for social agents at runtime. The idea experimented here is that these social agents can be controlled because bubbles were not originally intended to capture ego vehicles.
  - The capture of ego vehicles could result in `smarts.reset(..)` and `smarts.step(..)` having to generate new ids, or needing a supply of ids for the bubble to assign by injecting bubble information into the agent interface.

I do not yet understand what your intention is yet for use of the travelling bubble. Part of what this rough implementation is meant to uncover your intentions. I have some of my own questions:

- Do you need the moving bubble to follow an ego agent previously added by `scenario.set_ego_missions` & `smarts.switch_ego_agents`?
- Is your intention that bubble just provides information about vehicles entering/exiting without automated takeover of those vehicles by social agents?
  - You would want to find vehicles in the bubble?
  - Would you want to manage take-over of vehicles directly?
- Do you need to know the agent ids from the bubble before-hand or are dynamic ids (agent ids provided by smarts) reasonable?

YuechengLiu: @Gamenot Thanks for the replay.

> - Do you need the moving bubble to follow an ego agent previously added by scenario.set_ego_missions & smarts.switch_ego_agents?

Yes, exactly.  What I need is just a interface like this,
```python
start_time = 50
scenario = Scenario(...)
smarts = SMARTS()
ego_vehicle_id = "vehicle-10"
ego_agent_id = "agent-10"

ego_missions = {ego_agent_id: get_vehicle_mission(ego_vehicle_id, start_time)}
agent_interfaces = {ego_agent_id: AgentInterface(...)}
scenario.set_ego_missions(ego_missions)
smarts.switch_ego_agents(agent_interfaces)
scenario.bubbles.clear()
scenario.bubbles.extend([Bubble(follow_actor_id=ego_agent_id)])
# or, scenario.bubbles.extend([Bubble(follow_vehicle_id=ego_vehicle_id)])
agent_manager: AgentManager = smarts.agent_manager
agent_manager.add_social_observation_callback(
    obs_state.observation_callback, "bubble_watcher"
)

class EgoAgent:
    def act(self, observation):
        ...
        return action

class SocialAgent:
    def act(self, observation):
        ...
        return action

ego_agent = EgoAgent()
social_agent = SocialAgent()
ego_observations = smarts.reset(scenario, start_time)
...
for social_agent_id in obs_state.last_observations:
        agent_manager.reserve_social_agent_action(
            social_agent_id, social_agent.act(obs_state.last_observations[agent_id])
        )
ego_action = {ego_agent_id: ego_agent.act(obs) for ego_agent_id, obs in ego_observations.items()}
smarts.step(ego_action)
```

> Is your intention that bubble just provides information about vehicles entering/exiting without automated takeover of those vehicles by social agents?
> - You would want to find vehicles in the bubble?
> - Would you want to manage take-over of vehicles directly?

I do not need to take-over vehicles by myself now.

> Do you need to know the agent ids from the bubble before-hand or are dynamic ids (agent ids provided by smarts) reasonable?

I want to specify the ego_vehicle's (vehicle to be followed) agent id, by I don't need to pre-define social agents' agent ids (because the vehicles in bubble are dynamically changing).

Gamenot:
This needs a merge of #1440 and features from #1300.

Gamenot:
Current status is that example bubble captures and releases to the @sah-huawei's local traffic sim for the current example. I should have full update of the interface by end of tomorrow.

Gamenot:
@YuechengLiu `examples/bubble_example.py` has been updated.

YuechengLiu:
Thanks for the work! @Gamenot  @sah-huawei .

We have tried the `bubble` and observed some problems.

1. Some of the vehicles disappeared after getting into the `Bubble`.
2. The simulation seems to be slow (slower than the case where we control 6~7 ego vehicles). 
3. Some of the vehicles do not disappear after collision.

And I also have some questions,
1.  What are the vehicles controlled by after leaving the bubble? We observe that most of them are still social agents but some of them become history-replay vehicles (in envision visualization).
2.  Sometimes there are two bubble bounding boxes in envision visualization. is this a bug?

I will send **patrick** some videos about these cases, and you can ask him for the videos.

Thanks!

qianyi-sun:
I've posted the videos Yuecheng mentioned above in the Platform chat group. Btw, somehow I couldn't upload them in the comment, they are just 1.* MB mp4 files, I'm not sure why.

Gamenot:
> We have tried the `bubble` and observed some problems.
> 
> 1. Some of the vehicles disappeared after getting into the `Bubble`.

Yes, that is because the agent interface for those social agents specify that they can be done and removed when they collide.

> 2. The simulation seems to be slow (slower than the case where we control 6~7 ego vehicles).

One issue is that bubbles have not been optimized with a spatial tree but it would understandably be slower than dealing with 6~7 ego vehicles. The use case for bubbles is not to be faster at similar numbers of vehicles but to avoid using 100 ego vehicles where 10 might well enough for interaction.

> 3. Some of the vehicles do not disappear after collision.

I assume those are the grey box traffic vehicles after they have exited the bubble. History vehicles will ignore traffic from other sources causing pass-through (as expected.)

> 
> And I also have some questions,
> 
> 1. What are the vehicles controlled by after leaving the bubble? We observe that most of them are still social agents but some of them become history-replay vehicles (in envision visualization).

In this case the vehicles are controlled by the traffic provider the alternative provider is allow such as the `SumoTrafficSimulation` traffic provider or a custom traffic provider so long as it implements the `TrafficProvider` interface.

https://github.com/huawei-noah/SMARTS/blob/9e5a19f2088f0303bf7542d0149c1c97cf62870b/examples/bubble_example.py#L155-L160

Noting that traffic is removed from the history in this way so that it does not get controlled again by the traffic history:

https://github.com/huawei-noah/SMARTS/blob/9e5a19f2088f0303bf7542d0149c1c97cf62870b/examples/bubble_example.py#L226

> 2. Sometimes there are two bubble bounding boxes in envision visualization. is this a bug?

For this example it should not be a bug if there are 2 bubble bounding shapes because there two (and 2 agents) in the example.

> 
> I will send **Patrick** some videos about these cases, and you can ask him for the videos.
> 
> Thanks!

CyberSY:
Thanks for the explanation @Gamenot!
However, there are still some confusing parts about those problems.

> Some of the vehicles disappeared after getting into the Bubble.

We have observed cases where the vehicle didn't collide but still disappeared after getting into the bubble.
Problem1_example2-1
Problem1_example2-2
From the example,the left top vehicle in figures.

> Some of the vehicles do not disappear after collision.

Problem2_example2-1
Problem2_example2-2
Problem2_example2-3
Problem2_example2-4
From the example,the grey box vehicle left the bubble, then it collided with a car behind. However,the vehicle didn't disappear like the car behind it. Instead, it stopped in place and then re-controlled by the moving-forward bubble.
Is it because the lack of control logic? Can we use IDM to control vehicles after they leave the bubble?

> Sometimes there are two bubble bounding boxes in envision visualization. is this a bug?

For this, we observed that there seem to be two bounding boxes for one agent in visualization. We are not sure if this is a bug.
Problem2_example2-4

You can check the videos we sent earlier for details.
Thanks!

Gamenot:
> Thanks for the explanation @Gamenot! However, there are still some confusing parts about those problems.
> 
> > Some of the vehicles disappeared after getting into the Bubble.
> 
> We have observed cases where the vehicle didn't collide but still disappeared after getting into the bubble. Problem1_example2-1 Problem1_example2-2 From the example,the left top vehicle in figures.
> 

OK, yes, that would be a bug. I will look into it. Vehicles should not disappear when entering a bubble.

> > Some of the vehicles do not disappear after collision.
> 
From the example,the grey box vehicle left the bubble, then it collided with a car behind. However,the vehicle didn't disappear like the car behind it. Instead, it stopped in place and then re-controlled by the moving-forward bubble. Is it because the lack of control logic? Can we use IDM to control vehicles after they leave the bubble?
> 

This is due to the traffic control logic as you guessed. Traffic vehicles from the `LocalTrafficProvider` become "stranded" when they have a serious control error such as collision. We do nothing with this yet but ideally (for the entire simulation) we would switch to a collision model when vehicles collide.

> > Sometimes there are two bubble bounding boxes in envision visualization. is this a bug?
> 
> For this, we observed that there seem to be two bounding boxes for one agent in visualization. We are not sure if this is a bug. 

I think I misunderstood you initially, in the normal situation there should be two boxes visualised per bubble; a box for the bubble and a box for the "airlock" region where control is to be handed off. If you see only box per bubble that would be a bug.

> You can check the videos we sent earlier for details. Thanks!

Thanks for the follow-up.

Gamenot:
I am having difficulty capturing the exact situation.

Gamenot:
The cause of the disappearing vehicles is found roughly:

The traffic history provider is removing the vehicles before they come under control of the agent.

The state of the vehicle at that point is:
```
vehicle_id: history-vehicle-101
actor_id: history-vehicle-101
shadowed_by: {'BUBBLE-AGENT-history-vehicle-101'}
```

My assumption is that there is an ordering issue in removing the vehicle from the history. Observations will still come in for agents while they are just shadowing the history vehicle but a shadowing agent does not keep the vehicle alive if it is removed from the traffic history provider:

```python
# i.e.
traffic_history_provider.set_replaced_ids(used_history_ids | agent_vehicles)
```

The need now is that the vehicle must be removed from the traffic history provider conditional on the vehicle actually being controlled by a bubble agent rather than just shadowed.

YuechengLiu:
@Gamenot 
Thanks for the work! 
Has the bug been fixed now?

smarts/core/agent_manager.py
```diff
@@ -348,15 +351,46 @@ def _filter_social_agent_actions_for_controlled_vehicles(
...
+            callback(
+                dict(
+                    filter(
+                        lambda k: k[0] in self._remote_social_agents,
+                        observations.items(),
+                    )
+                )
+            )
         for agent_id, remote_agent in self._remote_social_agents.items():
+            if self._remote_social_agents_action.get(agent_id) != None:
```
sah-huawei, line 383:
```suggestion
            if self._remote_social_agents_action.get(agent_id) is not None:
```

smarts/core/agent_manager.py
```diff
@@ -348,15 +351,46 @@ def _filter_social_agent_actions_for_controlled_vehicles(
...
+            future_action = futures.Future()
+            future_action.set_result(action)
+            self._remote_social_agents_action[agent_id] = future_action
+        self._reserved_social_agent_actions.clear()
+        for callback in self._social_agent_observation_callbacks.values():
+            callback(
+                dict(
+                    filter(
+                        lambda k: k[0] in self._remote_social_agents,
+                        observations.items(),
```
sah-huawei, line 378:
```suggestion
                        lambda k: k in self._remote_social_agents,
                        observations.keys(),
```

smarts/core/agent_manager.py
```diff
@@ -348,15 +351,46 @@ def _filter_social_agent_actions_for_controlled_vehicles(
...
-        #       vehicles.
+    def add_social_observation_callback(self, callback, id):
+        """Suscribe to observe social agents."""
+        self._social_agent_observation_callbacks[id] = callback
+
+    def remove_social_observation_callback(self, id):
+        """Remove a subscription to social agents."""
+        del self._social_agent_observation_callbacks[id]
+
+    def reserve_social_agent_action(self, agent_id, action):
```
sah-huawei, line 362:
```suggestion
    def reserve_social_agent_action(self, agent_id: str, action):
```

smarts/core/agent_manager.py
```diff
@@ -348,15 +351,46 @@ def _filter_social_agent_actions_for_controlled_vehicles(
...
-        """Forwards observations to managed social agents."""
-        # TODO: Don't send observations (or receive actions) from agents that have done
-        #       vehicles.
+    def add_social_observation_callback(self, callback, id):
+        """Suscribe to observe social agents."""
+        self._social_agent_observation_callbacks[id] = callback
+
+    def remove_social_observation_callback(self, id):
+        """Remove a subscription to social agents."""
+        del self._social_agent_observation_callbacks[id]
```
sah-huawei, line 360:
```suggestion
    def remove_social_observation_callback(self, agent_id: str):
        """Remove a subscription to social agents."""
        del self._social_agent_observation_callbacks[agent_id]
```

smarts/core/agent_manager.py
```diff
@@ -348,15 +351,46 @@ def _filter_social_agent_actions_for_controlled_vehicles(
 
         return social_agent_actions
 
-    def send_observations_to_social_agents(self, observations):
-        """Forwards observations to managed social agents."""
-        # TODO: Don't send observations (or receive actions) from agents that have done
-        #       vehicles.
+    def add_social_observation_callback(self, callback, id):
+        """Suscribe to observe social agents."""
+        self._social_agent_observation_callbacks[id] = callback
```
sah-huawei, line 356:
```suggestion
    def add_social_observation_callback(self, callback, agent_id: str):
        """Suscribe to observe social agents."""
        self._social_agent_observation_callbacks[agent_id] = callback
```

smarts/core/agent_manager.py
```diff
@@ -348,15 +351,46 @@ def _filter_social_agent_actions_for_controlled_vehicles(
 
         return social_agent_actions
 
-    def send_observations_to_social_agents(self, observations):
-        """Forwards observations to managed social agents."""
-        # TODO: Don't send observations (or receive actions) from agents that have done
-        #       vehicles.
+    def add_social_observation_callback(self, callback, id):
```
sah-huawei, line 354: I'm not sure I like the terminology "social observation" because we already have confusion throughout SMARTS between "social _vehicles_" and "social _agents_".

examples/bubble_example.py
```diff
@@ -0,0 +1,343 @@
...
+    run_steps = int(run_time / timestep)
+    social_interface = AgentInterface.from_type(
+        AgentType.Direct,
+        done_criteria=DoneCriteria(
+            not_moving=False,
+            off_road=False,
+            off_route=False,
+            on_shoulder=False,
+            wrong_way=False,
+        ),
```
sah-huawei, line 153: Maybe we could add a `ColiisionOnlyDoneCriteria` to `agent_interface.py` as a short-cut for this, as this may be a common use case?

sah-huawei (commented):
Did a first pass over everything _but_ `bubble_manager.py`; I will need another pass now for that file specifically.

smarts/core/bubble_manager.py
```diff
@@ -528,7 +625,10 @@ def _hijack_social_vehicle_with_social_agent(
         else:
             agent_id = BubbleManager._make_social_agent_id(vehicle_id)
 
-        agent_interface = sim.agent_manager.agent_interface_for_agent_id(agent_id)
+        try:
+            agent_interface = sim.agent_manager.agent_interface_for_agent_id(agent_id)
+        except KeyError:
+            return
```
sah-huawei, line 631: When can this happen?

smarts/core/bubble_manager.py
```diff
@@ -271,11 +281,41 @@ class Cursor:
...
+            vehicle_ids_per_bubble (Dict[Bubble, Set[str]]):
+                Bubbles associated with vehicle ids.
+            running_cursors (Set["Cursor"]):
+                A set of existing cursors.
+        """
+        was_in_this_bubble = vehicle_id in vehicle_ids_per_bubble[bubble]
+        is_hijacked, is_shadowed = index.vehicle_is_hijacked_or_shadowed(vehicle_id)
+        transition = None
+        if was_in_this_bubble and (is_shadowed or is_hijacked):
+            transition = BubbleTransition.AirlockExited
```
sah-huawei, line 309: I'm not sure I understand the logic here.  This vehicle has already been removed, right?  Maybe add a comment to explain what's going on...

sah-huawei (commented):
did a pass at `bubble_manager.py`.  I may still need to do another one though!

smarts/core/vehicle_index.py
```diff
@@ -280,10 +280,12 @@ def vehicleitems(self) -> Iterator[Tuple[str, Vehicle]]:
         """A list of all vehicle IDs paired with their vehicle."""
         return map(lambda x: (self._2id_to_id[x[0]], x[1]), self._vehicles.items())
 
-    def vehicle_by_id(self, vehicle_id: str):
+    def vehicle_by_id(self, vehicle_id, default=...):
```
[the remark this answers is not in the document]
Gamenot: Vehicle id is still a string, this is just a conflict with the new develop. The reason why I used `Ellipsis` here is because `None` is a common value being used which could indicate an error. I want an exact value to be specified.

smarts/core/sensors.py
```diff
@@ -195,6 +195,8 @@ class Observation:
     elapsed_sim_time/step_count."""
     events: Events
     ego_vehicle_state: EgoVehicleObservation
+    has_vehicle_control: bool
+    """If the agent currently has control of the vehicle."""
```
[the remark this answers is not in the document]
Gamenot: Sorry I should have been clearer here and indicated `"this agent"`.

smarts/core/sensors.py
```diff
@@ -383,12 +386,14 @@ def observe(sim, agent_id, sensor_state, vehicle) -> Tuple[Observation, bool]:
             sim, agent_id, vehicle, sensor_state
         )
 
-        if (
-            done
-            and sensor_state.steps_completed == 1
-            and agent_id in sim.agent_manager.ego_agent_ids
-        ):
-            logger.warning(f"Agent Id: {agent_id} is done on the first step")
+        if done and sensor_state.steps_completed <= 2:
```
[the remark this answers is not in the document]
Gamenot: This is a remainder over from trying to find the bug with this PR. I will revert it.

smarts/core/agent_manager.py
Gamenot: Sorry, no the original was correct because it was to do a filter on the the items and return them as pairs for the construction of the dictionary. Using `keys()` means there are only keys for the reconstruction.

Gamenot:
The breakdown of a worst case smarts step in this example looks like this:

```
INFO:SMARTS:"Fetching agent actions" took: 0.090837ms
INFO:SMARTS:"Physics cost(pybullet provider)" took: 21.163225ms
INFO:SMARTS:"Stepping all providers and harmonizing state" took: 53.357601ms
INFO:SMARTS:"Checking if all agents are active" took: 0.003338ms
INFO:SMARTS:"Syncing vehicle index" took: 1.561880ms
INFO:SMARTS:"Stepping through bubble manager" took: 13.215065ms
INFO:SMARTS:"Stepping through trap manager" took: 0.008106ms
INFO:SMARTS:"Stepping through sensors" took: 0.097990ms
INFO:SMARTS:"Calculating observations and rewards" took: 195.063353ms
INFO:SMARTS:"Filtering response for ego" took: 0.007868ms
INFO:SMARTS:"Sending observations to social agents" took: 0.390530ms
INFO:SMARTS:"Clearing done agents" took: 0.454187ms
INFO:SMARTS:"Trying to emit the envision state" took: 1.458645ms
INFO:SMARTS:"Trying to emit the visdom observations" took: 0.001431ms
INFO:history-vehicles-replacement-example:agent=`agent-269` is done because `Events(collisions=[Collision(collidee_id='BUBBLE-AGENT-history-vehicle-280')], off_road=False, off_route=False, on_shoulder=False, wrong_way=False, not_moving=False, reached_goal=False, reached_max_episode_steps=False, agents_alive_done=False)`...
INFO:history-vehicles-replacement-example:social_agent=BUBBLE-AGENT-history-vehicle-280 collided @ [Collision(collidee_id='agent-269')]
INFO:history-vehicles-replacement-example:"SMARTS simulation/scenario step" took: 271.368742ms
```

Most of my efforts are on reducing the cost of observations (particularly map related calls.)

CyberSY:
Thanks for the work! @Gamenot @sah-huawei
We are now testing bubble with our model. However, there was a error happened in airlock function because of old version map usage.We wonder whether the route planning could be avoided since we will replace vehicles in bubble with our model.
Screenshot from 2022-07-19 19-56-29
### Error Message
ERROR:SMARTS:Simulation crashed with exception. Attempting to cleanly shutdown.
ERROR:SMARTS:Unable to find a route between start=gneE05b and end=gneE51. If either of these are junctions (not well supported today) please switch to roads and ensure there is a > 0 offset into the road if it's after a junction.
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/smarts.py", line 219, in step
    return self._step(agent_actions, time_delta_since_last_step)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/smarts.py", line 282, in _step
    self._bubble_manager.step(self)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 465, in step
    self._handle_transitions(sim, self._cursors)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 537, in _handle_transitions
    sim, cursor.vehicle_id, cursor.bubble.actor, cursor.bubble
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 604, in _airlock_social_vehicle_with_social_agent
    sim, vehicle_id, agent_id, interface, bubble
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 711, in _prepare_sensors_for_agent_control
    plan.create_route(mission)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/plan.py", line 352, in create_route
    "after a junction.".format(start_road.road_id, end_road.road_id)
smarts.core.plan.PlanningError: Unable to find a route between start=gneE05b and end=gneE51. If either of these are junctions (not well supported today) please switch to roads and ensure there is a > 0 offset into the road if it's after a junction.
ERROR:Client:Connection to Envision terminated with: 'Dispatcher' object has no attribute 'read'

Gamenot:
A detailed look shows the following breakdown.

flamegraph

Gamenot:
I have allowed the bubble airlock to be 0 for the case that agents do not need warm-up observations. This reduces a significant amount observations because observations were being generated for agents that were not under control.

Gamenot:
> Thanks for the work! @Gamenot @sah-huawei We are now testing bubble with our model. However, there was a error happened in airlock function because of old version map usage.We wonder whether the route planning could be avoided since we will replace vehicles in bubble with our model. Screenshot from 2022-07-19 19-56-29
> 
> ### Error Message
> ERROR:SMARTS:Simulation crashed with exception. Attempting to cleanly shutdown. ERROR:SMARTS:Unable to find a route between start=gneE05b and end=gneE51. If either of these are junctions (not well supported today) please switch to roads and ensure there is a > 0 offset into the road if it's after a junction. Traceback (most recent call last): File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/smarts.py", line 219, in step return self._step(agent_actions, time_delta_since_last_step) File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/smarts.py", line 282, in _step self._bubble_manager.step(self) File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 465, in step self._handle_transitions(sim, self._cursors) File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 537, in _handle_transitions sim, cursor.vehicle_id, cursor.bubble.actor, cursor.bubble File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 604, in _airlock_social_vehicle_with_social_agent sim, vehicle_id, agent_id, interface, bubble File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 711, in _prepare_sensors_for_agent_control plan.create_route(mission) File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/plan.py", line 352, in create_route "after a junction.".format(start_road.road_id, end_road.road_id) smarts.core.plan.PlanningError: Unable to find a route between start=gneE05b and end=gneE51. If either of these are junctions (not well supported today) please switch to roads and ensure there is a > 0 offset into the road if it's after a junction. ERROR:Client:Connection to Envision terminated with: 'Dispatcher' object has no attribute 'read'

@CyberSY I have addressed that error with [b6e5e08](https://github.com/huawei-noah/SMARTS/pull/1462/commits/b6e5e0823671c070e2132c52bb5c0d498bf28f49)

smarts/core/bubble_manager.py
```diff
@@ -708,7 +708,11 @@ def _prepare_sensors_for_agent_control(
         else:
             goal = EndlessGoal()
         mission = Mission(start=Start(vehicle.position[:2], vehicle.heading), goal=goal)
-        plan.create_route(mission)
+        try:
+            plan.create_route(mission)
+        except PlanningError:
+            plan.mission = None
```
CyberSY, line 738: I have done test and found that bubble works fine after commenting out this line of code, and the speed is quite acceptable. It probably because we defines that `plan.mission = None` in initialization and we don't need to assign value again? 

**Error Message**
> Traceback (most recent call last):
  File "run_scripts/visualize_smarts.py", line 196, in <module>
    raise e
  File "run_scripts/visualize_smarts.py", line 192, in <module>
    experiment(exp_specs)
  File "run_scripts/visualize_smarts.py", line 62, in experiment
    visualize(variant, env)
  File "run_scripts/visualize_smarts.py", line 129, in visualize
    next_observation_n, reward_n, terminal_n, env_info_n = env.step(action_n)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/ILSwiss/rlkit/envs/wrappers.py", line 155, in step
    observation_n, reward_n, done_n, info_n = self._wrapped_env.step(action_n)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/ILSwiss/rlkit/envs/wrappers.py", line 105, in step
    return self._wrapped_env.step(scaled_action_n)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/ILSwiss/rlkit/env_creators/smarts/smarts_env.py", line 44, in step
    return self._env.step(action_n)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/smarts-imitation/smarts_imitation/envs/bubble_core.py", line 371, in step
    raw_observation_n, reward_n, self.done_n, _ = self.smarts.step(scaled_action_n)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/smarts.py", line 220, in step
    return self._step(agent_actions, time_delta_since_last_step)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/smarts.py", line 285, in _step
    self._bubble_manager.step(self)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 487, in step
    self._handle_transitions(sim, self._cursors)
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 554, in _handle_transitions
    sim, cursor.vehicle_id, cursor.bubble.actor, cursor.bubble
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 628, in _airlock_social_vehicle_with_social_agent
    sim, vehicle_id, agent_id, interface, bubble
  File "/home/z50024513/Desktop/SMARTS/SMARTS_Benchmark/SMARTS/smarts/core/bubble_manager.py", line 739, in _prepare_sensors_for_agent_control
    plan.mission = None
AttributeError: can't set attribute

Gamenot: > And it works well if we just remove the line `plan.mission = None` in our experiment.

Very sorry, I think I added that line just after testing even though it is an immutable property, I have addressed it properly.

Gamenot:
@CyberSY That assert does not need to be that strict. It indicates that there are multiple providers attempting to operate on the same vehicle I have downgraded it from an assert to logging. There is already a mechanism in place to resolve multiple requested states.

smarts/core/agent_manager.py
Gamenot: I realise id is misleading and should be `callback_id`

smarts/core/bubble_manager.py
```diff
@@ -315,59 +355,73 @@ def from_pos(
...
-            )
+            vehicles = []
+            if bubble.follow_actor_id is not None:
+                vehicles += self._last_vehicle_index.vehicles_by_actor_id(bubble.follow_actor_id)
+            if bubble.follow_vehicle_id is not None:
+                vehicle = self._last_vehicle_index.vehicle_by_id(bubble.follow_vehicle_id, None)
+                if vehicle is not None:
+                    vehicles += [vehicle]
+            # print(self._last_vehicle_index)
             return len(vehicles) == 1
```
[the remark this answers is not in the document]
Gamenot: No, there should only ever be 1 or 0 vehicles the bubble follows, I have added a log to announce that error should it occur.

smarts/core/bubble_manager.py
```diff
@@ -378,18 +432,32 @@ def _vehicle_ids_divided_by_bubble_state(
...
+    ) -> Set[str]:
+        """Agents generated by this bubble."""
+        bubble_cursors = set(filter(lambda c: c.bubble == bubble, self._cursors))
+
+        agent_ids = []
+        for bc in bubble_cursors:
+            if bc.state != BubbleState.InBubble:
+                continue
+            agent_id = sim.vehicle_index.actor_id_from_vehicle_id(bc.vehicle_id)
+            if agent_id != None:
```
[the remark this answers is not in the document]
Gamenot: I somewhat prefer the explicit `is not None` because of implicit type.

smarts/core/bubble_manager.py
```diff
@@ -408,33 +476,53 @@ def _sync_cursors(self, last_vehicle_index, vehicle_index):
...
+        active_bubbles, inactive_bubbles = self._bubble_groups()
+        inactive_bubbles_to_run = [b for b in inactive_bubbles if len(vehicle_ids_per_bubble[b])]
+        
+        if inactive_bubbles_to_run:
+            for bubble in inactive_bubbles_to_run:
+                for _, vehicle in persisted_vehicle_index.vehicleitems():
+                    cursor = Cursor.for_removed(
+                        vehicle_id=vehicle.id,
+                        bubble=bubble,
+                        index=vehicle_index,
```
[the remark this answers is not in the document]
Gamenot: Yes, that might avoid errors inside the method.

smarts/core/bubble_manager.py
Gamenot: The bubble manager appears to keep track of all vehicles, this is a check that happens just before the vehicle is removed because there is a need to check if the vehicle has left the bubble.

smarts/core/vehicle_index.py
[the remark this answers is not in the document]
sah-huawei: ok, but are you sure about `vehicle_id` stil being marked as a `str` on your branch?  (It's marked right now on `develop`, but when I pull your branch it's not.)

smarts/core/bubble_manager.py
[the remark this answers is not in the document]
sah-huawei: OK.  (... and `is not None` is better than `!= None` in my opinion.)

sah-huawei:
I see that @Gamenot already replied as well as pushed a fix for the issue caused by the bad map for planning the mission.

However, I just wanted to briefly address this portion of the original question:
> We wonder whether the route planning could be avoided since we will replace vehicles in bubble with our model.

In general, notwithstanding this particular use case, it is useful to have routes pre-planned for all vehicles for cases where they may get relinquished upon bubble exit to another provider (for example, SMARTS' local traffic provider), at which point it can attempt to continue the route as it was planned before the hijacking, or at least attempt to still reach the mission goal if continuing the route is no longer possible.

sah-huawei (approved):
I think this can (finally!) be merged.

Also, let's try to talk a bit about the traffic history vehicle-id encapsulation issue on Thursday if we can.

smarts/core/sensors.py
```diff
@@ -248,32 +275,17 @@ def observe(sim, agent_id, sensor_state, vehicle) -> Tuple[Observation, bool]:
...
-                        lane_position=nv_lane_pos,
+                if (
+                    veh_obs.lane_id is not None
+                    and vehicle.subscribed_to_lane_position_sensor
+                ):
+                    nv_lane_pos = vehicle.lane_position_sensor(
+                        sim.road_map.lane_by_id(veh_obs.lane_id), nv
                     )
+                neighborhood_vehicles.append(
+                    veh_obs._replace(lane_position=nv_lane_pos)
```
sah-huawei, line 288: I just noticed these changes (b/c I had a merge conflict trying to merge these into my own branch).  Even though they don't look like they'll break anything (and of course I've already approved this PR and you've already merged it!), I'm curious why you didn't move the `nv_lane_pos` code into the `_make_vehicle_observations()` function with the rest of the code there.

PR #1466, opened 2022-06-14
Single vehicle in traffic

AisenGinn:
Added trip consisted of single vehicle and route in sstudio.type.py. Added code for writing xml of trip in generator.py.


smarts/sstudio/types.py
```diff
@@ -441,6 +466,7 @@ class Traffic:
 
     flows: Sequence[Flow]
     """Flows are used to define a steady supply of vehicles."""
+    trips: Sequence[Trip]
```
Gamenot, line 469: Please add a descriptive docstring here.

smarts/sstudio/generators.py
```diff
@@ -265,6 +264,8 @@ def _writexml(self, traffic: types.Traffic, route_path: str):
                         begin=flow.begin,
                         end=flow.end,
                     )
+            #wrtie trip into xml format
+            tripwritexml(self.resolve_route,traffic,doc)
```
Gamenot, line 268: Please call `black smarts/sstudio/generators.py`.

Gamenot, line 268: Please call `black smarts/sstudio/generators.py`.

smarts/sstudio/types.py
```diff
@@ -384,7 +384,7 @@ class Flow:
     @property
     def id(self) -> str:
         """The unique id of this flow."""
-        return "flow-{}-{}-".format(
+        return "vehicle-{}-{}-".format(
```
Gamenot, line 387: Please revert this particular change. No changes to flow should happen.

```suggestion
        return "flow-{}-{}-".format(
```

Gamenot, line 387: Please revert this particular change. No changes to flow should happen.

```suggestion
        return "flow-{}-{}-".format(
```
```suggestion
        return "vehicle-{}-{}-".format(
```

smarts/sstudio/types.py
```diff
@@ -398,6 +398,31 @@ def __eq__(self, other):
         return self.__class__ == other.__class__ and hash(self) == hash(other)
 
 
+@dataclass(frozen=True)
+class Trip:
+    """A route with an actor type emitted at a given rate."""
+
+    actor: TrafficActor(str)
```
Gamenot, line 405: Please add a docstring to this field.

Gamenot, line 405: Please add a docstring to this field.

Gamenot (changes requested):

smarts/sstudio/generators.py
```diff
@@ -332,3 +333,63 @@ def _resolve_log_dir(self, log_dir):
             log_dir = make_dir_in_smarts_log_dir("_duarouter_routing")
 
         return os.path.abspath(log_dir)
+    
+
+def tripwritexml(resolve_route, traffic, doc):
```
Gamenot, line 338: This tabbing seems wrong.

```suggestion
    def tripwritexml(resolve_route, traffic, doc):
```

smarts/sstudio/generators.py
```diff
@@ -332,3 +333,63 @@ def _resolve_log_dir(self, log_dir):
...
+                decel=actor.decel,
+                vClass=actor.vehicle_type,
+                speedFactor=actor.speed.mean,
+                speedDev=actor.speed.sigma,
+                sigma=sigma,
+                minGap=min_gap,
+                maxSpeed=actor.max_speed,
+                **actor.lane_changing_model,
+                **actor.junction_model,
+                )
```
Gamenot, line 361: I think this section can be combined with the flow actor `vType` generation:

https://github.com/huawei-noah/SMARTS/blob/305b1e0c7cf829978c7efdb49cd85f3dad35b6d4/smarts/sstudio/generators.py#L215-L217

and change that to:

```
 for actor in { 
     actor for flow in traffic.flows for actor in [*flow.actors.keys(), *trip.actors.keys()]
 }: 
```

Gamenot (changes requested):

Gamenot (approved):

smarts/sstudio/generators.py
```diff
@@ -368,3 +371,36 @@ def _resolve_log_dir(self, log_dir):
             log_dir = make_dir_in_smarts_log_dir("_duarouter_routing")
 
         return os.path.abspath(log_dir)
+
+
+def tripwritexml(resolve_route, traffic, doc):
```
saulfield, line 376: A couple minor things here:
- The name should probably match the style convention -- maybe `write_trip_xml`?
- I don't think this should be a top-level function. It should either be encapsulated in the `TrafficGenerator` class or the code should be inlined where it's called (I don't have a strong opinion either way).

saulfield (changes requested):
Just one minor comment, otherwise looks good to merge.

smarts/sstudio/generators.py
AisenGinn: I will change the name into write_trip_xml and encapsulated it into a class method using self.resolve_route instead of passing resolve_route as an argument.

smarts/sstudio/types.py
```diff
@@ -501,6 +501,37 @@ def __hash__(self):
...
+        """The unique id of this trip."""
+        return self.vehicle_name
+
+    def __hash__(self):
+        # Custom hash since self.actors is not hashable, here we first convert to a
+        # frozenset.
+        return _pickle_hash((self.route, self.actor))
+
+    def __eq__(self, other):
+        return self.__class__ == other.__class__ and hash(self) == hash(other)
```
AisenGinn, line 534: Vehicle name can no longer be assigned in this case in Trip. Considering modifying this dataclass into standard format without __init__() and using __post_init__() to assign the actor.

saulfield (approved):

smarts/sstudio/generators.py
```diff
@@ -236,6 +236,21 @@ def _writexml(
                     **actor.junction_model,
                 )
 
+            if traffic.trips is not None:
+                for actor in {trip.actor for trip in traffic.trips}:
```
Gamenot, line 240: I think what we should do is combine the trip actors and flow actors to ensure they are unique.

```
actors_for_vtype = {
   actor for flow in traffic.flows for actor in flow.actors.keys()
}
if traffic.trips:
    actors_for_vtype |= {trip.actor for trip in traffic.trips}

for actor in actors_for_vtype:
    # Generate vType
    ...
```

smarts/sstudio/generators.py
```diff
@@ -289,6 +296,38 @@ def _writexml(
                 )
             )
 
+    def write_trip_xml(self, traffic, doc, fill_in_gaps):
+        """Wrtes a traffic spec into a route file. Typically this would be the source
+        data to Sumo's DUAROUTER.
```
Gamenot, line 301:
```suggestion
        """Writes a trip spec into a route file. Typically this would be the source
        data to SUMO's DUAROUTER.
```

smarts/sstudio/types.py
```diff
@@ -502,6 +502,42 @@ def __eq__(self, other):
...
+    vehicle_name: str
+    """The name of the vehicle. It must be unique. """
+    route: Union[RandomRoute, Route]
+    """The route for the actor to attempt to follow."""
+    vehicle_type: str = "passenger"
+    """The type of the vehicle"""
+    depart: float = 0
+    """Start time in seconds."""
+    actor: TrafficActor = field(init=False)
+    """The traffic actor for the vehicle"""
```
Gamenot, line 518:
```suggestion
    """The traffic actor model (usually vehicle) that will be used for the trip."""
```

smarts/sstudio/generators.py
```diff
@@ -289,6 +296,38 @@ def _writexml(
...
+
+            # We don't de-dup flows since defining the same flow multiple times should
+            # create multiple traffic flows. Since IDs can't be reused, we also unique
+            # them here.
+        for trip_idx, trip in enumerate(traffic.trips):
+            route = resolved_routes[trip.route]
+            actor = trip.actor
+            doc.stag(
+                "vehicle",
+                id="{}".format(trip.vehicle_name),
```
Gamenot, line 320: The one thing we should do is ensure that the vehicle ids are unique by giving an error if there are multiple trips with the same vehicle name.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - Added ego's mission details into the `FormatObs` wrapper.
 - Added `SmartsLaneChangingModel` and `SmartsJunctionModel` to types available for use with the new smarts traffic engine within Scenario Studio.
 - Added option to `AgentInterface` to include traffic signals (lights) in `EgoVehicleObservation` objects.
+- Added single vehicle `Trip` into type.
```
Gamenot, line 18: Please move this to a newer section.

Gamenot (changes requested):
Some final changes, after this is addressed we can merge it.

smarts/sstudio/generators.py
Gamenot: @AisenGinn Thanks.

Gamenot:
Good work.

Gamenot (approved):

PR #1469, opened 2022-06-17
TargetPose Example/Road waypoints bug

Gamenot:
This is an example using TargetPose.

Previous commit shows waypoint bug: https://github.com/huawei-noah/SMARTS/commit/e6dfc2cfed882631e9423787603eff0b12b4fc81

### DIFF
--- a/examples/target_pose_example.py
+++ b/examples/target_pose_example.py
@@ -62,7 +62,7 @@ class NonInteractiveAgent(Agent):
                     print(wp.lane_id + " but actually: "+lane_waypoints[-1].lane_id)
         dist_to_wp = wp.dist_to(obs.ego_vehicle_state.position)
         print(f"selected {wp.lane_id} {dist_to_wp / self.speed}")
-        return np.array([*wp.pos, wp.heading, dist_to_wp / self.speed])
+        return np.array([*wp.pos, wp.heading, 0.1])
 
 
 def main(scenarios, headless, num_episodes, max_episode_steps=None):
--- a/examples/target_pose_example.py
+++ b/examples/target_pose_example.py
@@ -1,0 +1,112 @@
+import logging
+import pathlib
+
+import gym
+
+from examples.argument_parser import default_argument_parser
+from smarts.core.agent import Agent
+from smarts.core.agent_interface import AgentInterface, AgentType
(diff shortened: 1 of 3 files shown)


Gamenot:
@YuechengLiu I have added an example. The `TargetPose` action space works as intended. (the mentioned bug is related to something else)

YuechengLiu:
Thanks for the help！

SXKai:
Hi， @Gamenot. I have a question about this feature. If I want to control the vehicle to reach a target point with targetPose (x, y, heading，speed) in one step, how can I implement it? Do I need to set the targetPose (x, y, heading，speed)  into obs firstly, and then use .act() function to get the action?

Gamenot:
Hello @SXKai, the description of the TargetPose is as follows:

* `position x` value in meters with +x facing right.
* `position y` value in meters with +y facing front.
* `heading` in radians with 0 facing north and counterclockwise rotation.
* `time horizon` in seconds as to when the vehicle should arrive at the given pose.

Provide the same value as the timestep of the simulation (generally 0.1) if you want the vehicle to arrive at the pose on the next step.

SXKai:
Thanks for your help.

YuechengLiu:
Hello @Gamenot . 

If the agent take a action `TargetPos` at step `t`, will the agent accurately reach the specified location at step `t+1`? Are there any dynamic constraints in this process？

YuechengLiu:
Thanks @Gamenot , I just want to know how it works. 

So under the action space `TargetPos`, is this dynamics constraints realized by `pybullet` or `SMARTS`? Is this constraints same as the cases under action space `Imitation`?

Gamenot:
@YuechengLiu From my understanding and from double-checking the `ActionSpaceType.TargetPose` related vehicle control, the physics related constraints are the exact same as are seen with `Imitation`/`Direct` and fall under `SMARTS` control.

If attempting to reach to a target trajectory using pure dynamics that would require something like Continuous, Lane, ActuatorDynamic, LaneWithContinuousSpeed, Trajectory, or MPC. Such as described https://github.com/huawei-noah/SMARTS/blob/develop/docs/sim/agent.rst#agentinterface.

PR #1475, opened 2022-06-23
[Bugfix] Address lost distance in trip meter sensor.

Gamenot:
This fixes a greater than 1/2 lost distance in the trip meter sensor and adds a test for it.

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -362,7 +362,7 @@ class Sensors:
         )
 
         if waypoint_paths:
-            vehicle.trip_meter_sensor.append_waypoint_if_new(waypoint_paths[0][0])
+            vehicle.trip_meter_sensor.update_distance_wps_record(waypoint_paths)
         distance_travelled = vehicle.trip_meter_sensor(sim)
 
         vehicle.driven_path_sensor.track_latest_driven_path(sim)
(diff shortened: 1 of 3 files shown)


smarts/core/sensors.py
```diff
@@ -938,17 +932,17 @@ def append_waypoint_if_new(self, new_wp):
...
-    def _compute_additional_dist_travelled(recent_wp, waypoint):
-        heading_vec = recent_wp.heading.direction_vector()
-        disp_vec = waypoint.pos - recent_wp.pos
-        direction = np.sign(np.dot(heading_vec, disp_vec))
-        distance = np.linalg.norm(disp_vec)
+    def _compute_additional_dist_travelled(recent_wp: Waypoint, new_waypoint: Waypoint, vehicle_pose: Pose):
+        wp_disp_vec = new_waypoint.pos - recent_wp.pos
+        pose_disp_vec = vehicle_pose.position[:2] - recent_wp.pos[:2]
+        direction = np.sign(np.dot(pose_disp_vec, wp_disp_vec))
+        distance = np.linalg.norm(wp_disp_vec)
```
sah-huawei, line 945: Unfortunately, this still won't fix the "zig-zagging" problem.
Consider:  if `new_waypoint` is on an adjacent lane to `recent_waypoint`, `distance` will be greater than if it is the next waypoint within the same lane.  So if there are multiple lanes within a road along the route, then the way to maximize the distance traveled on that road is to continually change lanes, alternating waypoints between the lanes.

This is not a simple problem to solve, as you also don't want to penalize legitimate lane changes.  I do not think it can be done accurately with our current _waypoints_.  Instead, here's some (inefficient) code that gets at what I believe it should be more like.  Roughly:
```python
    # assuming cur_pose contains our current pose and cur_lane our current lane
    if self._recent_lane_on_route:
        # note: the following computes distance along the _previous_ (not the current) lane!
        # It is deliberately ignoring the t-coordinate in both lanes.
        self.distance_travelled += self._recent_lane_on_route.offset_along_lane(cur_pose.point) - self._recent_offset
    if cur_lane.road in self._plan.route.roads:
        self._recent_lane_on_route = current_lane   
        self._recent_offset = current_lane.offset_along_lane(cur_pose.point)
    else:
        self._recent_lane_on_route = None
```

Note that this would not work "as is" (it needs to be made more robust against off-route driving), and will also be less efficient computationally, but so far it's the only thing I can come up with that attempts to address the zig-zag concern.

I think we should meet about this.  I'm guessing it will require some concerted brainstorming.

smarts/core/tests/test_sensors.py
```diff
@@ -79,6 +79,41 @@ def scenarios():
...
+    )
+    vehicle.length = 3.68
+
+    mission = scenario.missions[AGENT_ID]
+    plan = Plan(scenario.road_map, mission)
+
+    sensor = TripMeterSensor(vehicle, scenario.road_map, plan)
+    waypoints_sensor = WaypointsSensor(vehicle, plan)
+    
+
```
sah-huawei, line 100: minor: extra vertical whitespace

sah-huawei (changes requested):

smarts/core/sensors.py
sah-huawei: ah ha!  thank you for clarifying!!  Yes, that makes a big difference and I agree, makes this solution ok.  I withdraw my objection and suggestion here.

sah-huawei (approved):
I think this can be merged once the CI tests are all made to pass.

Gamenot:
Interesting, determinism seems to be failing often in this PR.

Adaickalavan (approved):

PR #1479, opened 2022-06-25
add Segment Cache

sah-huawei:
This optimization provides more than a 30x speed up to the steps-per-second rate when using SMARTS to provide background traffic on both Waymo and OpenDrive maps.

(The main optimization is to avoid having to re-compute the distances between successive points in the center polylines for Lanes.)

Because lane direction vectors can be based on the correct segment, no matter how long it is, this technique also give a more accurate estimation of curvature and deals more accurately with vectors at the end of lanes too.  (Hence the updates to the unit tests, which were each checked by hand.)


Gamenot (approved):
I have no real issues with this solution.

saulfield (approved):

PR #1483, opened 2022-06-30
Compute intersecting lanes for Waymo maps

saulfield:
- When loading a Waymo map, we now compute a set of intersecting lanes for each lane in the map. This is to be used for identifying junctions and foes for traffic simulation.
- We now cache TFRecord file data to speed up Waymo map loading for cases where multiple scenarios are being loaded consecutively from the same TFRecord file.


smarts/core/waymo_map.py
```diff
@@ -234,6 +245,98 @@ def _raycast_boundaries(
...
+                if cand_id in intersections[feat_id]:
+                    continue
+
+                lanes_to_test.append(cand_id)
+
+            # Main loop -- check each segment of the lane polyline against the
+            # polyline of each candidate lane
+            line1 = np.array(self._polyline_cache[feat_id][0])
+            for cand_id in lanes_to_test:
+                line2 = np.array(self._polyline_cache[cand_id][0])
```
sah-huawei, line 311: I don't think the `np.array()` is necessary here (as `np.roll()` will make an array out of it).

smarts/core/waymo_map.py
```diff
@@ -234,6 +245,98 @@ def _raycast_boundaries(
...
+                C = np.roll(line2, 0, axis=0)[:-1]
+                D = np.roll(line2, -1, axis=0)[:-1]
+                len_c = len(C)
+                for i in range(len(line1) - 1):
+                    A = np.tile(line1[i], (len_c, 1))
+                    B = np.tile(line1[i + 1], (len_c, 1))
```
sah-huawei, line 317: Re our discussion about how to avoid the repeated allocations from `np.tile()` here, I think this page has answers:  https://numpy.org/doc/stable/user/basics.broadcasting.html

Long story short, I don't think you need the `tile()` calls here.

smarts/core/waymo_map.py
```diff
@@ -93,6 +99,11 @@ class WaymoMap(RoadMap):
     DEFAULT_LANE_SPEED = 16.67  # in m/s
     DEFAULT_LANE_WIDTH = 4
 
+    # For caching tfrecord data
+    _tfrecord_path = None
+    _tfrecord_generator = None
+    _scenario_cache = None
```
sah-huawei, line 105:
```suggestion
    _tfrecord_path: Optional[str] = None
    _tfrecord_generator = Optional[Generator[bytes, None, None]] = None
    _scenario_cache: Optional[Dict[str, Any]] = None
```

smarts/core/waymo_map.py
```diff
@@ -705,14 +811,27 @@ def parse_source_to_scenario(source: str):
...
-        for record in dataset_records:
+
+        # Reset cache if this is a new TFRecord file
+        if not WaymoMap._tfrecord_path or WaymoMap._tfrecord_path != dataset_path:
+            WaymoMap._tfrecord_path = dataset_path
+            WaymoMap._tfrecord_generator = read_tfrecord_file(dataset_path)
+            WaymoMap._scenario_cache = dict()
+
+        if scenario_id in WaymoMap._scenario_cache:
+            return WaymoMap._scenario_cache[scenario_id]
```
sah-huawei, line 822: It's minor here, I know, but this pattern avoids double lookups in dicts:
```suggestion
        parsed_scenario = WaymoMap._scenario_cache.get(scenario_id)
        if parsed_scenario:
            return parsed_scenario
```

smarts/core/waymo_map.py
```diff
@@ -234,6 +245,98 @@ def _raycast_boundaries(
...
+                len_c = len(C)
+                for i in range(len(line1) - 1):
+                    A = np.tile(line1[i], (len_c, 1))
+                    B = np.tile(line1[i + 1], (len_c, 1))
+                    if line_intersect_vectorized(A, B, C, D):
+                        intersections[feat_id].add(cand_id)
+                        intersections[cand_id].add(feat_id)
+                        break
+
+        # remove "fake" incoming/outgoing lanes that aren't true intersections
```
sah-huawei, line 323: good!

sah-huawei (commented):
I think I can now use `self._intersections` to implment `foes()` on the segmented lanes.  I'm wondering if I should do that on this branch/PR or wait until it's merged and then do a different one.

smarts/core/waymo_map.py
saulfield: Yes you're right -- good catch! It looks like this does indeed give us an additional 10-20% speedup too.

sah-huawei (approved):

Gamenot (approved):

Gamenot:
@sah-huawei Sorry, I reviewed this and forgot to actually hit the button.

PR #1490, opened 2022-07-06
endless traffic refactor

sah-huawei:
This PR removes the `endless_traffic` parameter from being set at the `TrafficProvider` (simulation) level to instead be set at the scenario level within `scenario.py` using the new `repeat_route` attribute on the `Flow`(s).

Whether or not social vehicles "wrap around" or "teleport back to the beginning of their routes" is something that affects the traffic patterns and density of a given scenario in relatively-complex ways.  This PR gives control of that to the scenario _creator_.


smarts/core/sumo_traffic_simulation.py
```diff
@@ -546,9 +541,8 @@ def _sync(self, provider_state: ProviderState):
             self._traci_conn.vehicle.setSpeedMode(vehicle_id, all_checks)
             self._traci_conn.vehicle.setSpeed(vehicle_id, -1)
 
-        if self._endless_traffic:
-            self._reroute_vehicles(traffic_vehicle_states)
-            self._teleport_exited_vehicles()
+        self._reroute_vehicles(traffic_vehicle_states)
+        self._teleport_exited_vehicles()
```
Gamenot, line 545: This calls the `traci_conn.simulation.getSubscriptionResults().` every `sync(...)` in addition to the other time it is called in `step(...)`. If possible it would be best if we only ever do that call once.

Gamenot (approved):

saulfield (approved):

smarts/core/sumo_traffic_simulation.py
sah-huawei: Good call!  I just pushed something to avoid calling that more than once.

qianyi-sun (approved):
Thanks Steve for implementing this! I've read the changes and looks like to use it, just simply add `repeat_route=False` into the `Flow`, is this correct?

sah-huawei:
> Thanks Steve for implementing this! I've read the changes and looks like to use it, just simply add `repeat_route=False` into the `Flow`, is this correct?

Yes, exactly.
(... although -- unlike before -- `repeat_route=False` is now the default behaviour, so really you only need to add something to the `Flow` if you want `repeat_route=True`.)

PR #1494, opened 2022-07-07
Fix animation bug in waymo browser tool

saulfield:
This fixes an issue where only valid vehicle states were added to the trajectories that get animated, resulting in unaligned states being rendered for vehicles that have "missing" states in the scenario. Now all states will be timestep-aligned, and will simply not be rendered if it is missing for a given timestep.

### DIFF
--- a/smarts/waymo/waymo_browser.py
+++ b/smarts/waymo/waymo_browser.py
@@ -211,10 +211,11 @@ def get_trajectory_data(waymo_scenario: scenario_pb2.Scenario) -> Dict:
             # First pass -- extract data and yield trajectory at every timestep
             for j in range(num_steps):
                 obj_state = scenario.tracks[i].states[j]
-                if not obj_state.valid:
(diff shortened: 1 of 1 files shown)


smarts/waymo/waymo_browser.py
```diff
@@ -591,7 +599,7 @@ def save_plot(
         def update(i):
             drawn_pts = []
             for (xs, ys), point in zip(data, points):
-                if i < len(xs):
+                if i < len(xs) and xs[i] and ys[i]:
```
sah-huawei, line 602: I think we need to allow `xs[i]` and `ys[i]` to be 0.0.
```suggestion
                if i < len(xs) and xs[i] is not None and ys[i] is not None:
```

saulfield: Good catch! (facepalm)

Gamenot (approved):

PR #1495, opened 2022-07-07
cutin options

sah-huawei:
This adds the `slow_down_after` and `hold_period` options to the SMARTS lane changing model.  

This is so we can simulate situations where the agent vehicle being cutin-upon definitely must react.

Edit:  upon further discussion with Patrick, I've also now added the `multi_lane_cutin` option.  And this led me to discover a bug in the handing of boolean parameters (cf. the previous `dogmatic` option) that I've now fixed as well.


saulfield (approved):

saulfield (approved):

qianyi-sun (approved):
Understood, thanks Steve!

PR #1496, opened 2022-07-08
Account for units in opendrive lane speeds

### LINKED ISSUE #1480: road/lane speeds on OpenDrive maps should take into account units
The `map.xodr` file in `scenarios/open_drive/od_newmarket` contains elements like:
```xml
      <type type="town" s="0">
         <speed unit="mph" max="2.68016"/>                                                                                                                      
      </type>  
```

Here the units are specified as being in miles-per-hour (I'm assuming), but in `opendrive_road_network.py`, the `unit` attribute is not considered, so the above would be considered to be `2.68016` meters-per-second.

saulfield:
Fixes #1480


sah-huawei (approved):
thanks!

PR #1498, opened 2022-07-11
Bubble Handoff Fixes

sah-huawei:
Fixes a couple bugs that were originally discovered and diagnosed by Tucker on the moving bubbles PR.  (PR #1462)

### DIFF
--- a/smarts/core/plan.py
+++ b/smarts/core/plan.py
@@ -328,6 +328,15 @@ class Plan:
             self._mission.start.point,
             include_junctions=False,
         )
+        if not start_lane:
+            # it's possible that the Mission's start point wasn't explicitly
+            # specified by a user, but rather determined during the scenario run
+            # from the current position of a vehicle, in which case it may be
+            # in a junction.  But we only allow this if the previous query fails.
+            start_lane = self._road_map.nearest_lane(
+                self._mission.start.point,
+                include_junctions=True,
+            )
         assert start_lane, "route must start in a lane"
         start_road = start_lane.road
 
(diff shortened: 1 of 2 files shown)


smarts/core/plan.py
```diff
@@ -328,6 +328,15 @@ def create_route(self, mission: Mission) -> Mission:
...
         )
+        if not start_lane:
+            # it's possible that the Mission's start point wasn't explicitly
+            # specified by a user, but rather determined during the scenario run
+            # from the current position of a vehicle, in which case it may be
+            # in a junction.  But we only allow this if the previous query fails.
+            start_lane = self._road_map.nearest_lane(
+                self._mission.start.point,
+                include_junctions=True,
+            )
```
Gamenot, line 339: This feels like a lot of additional cost but I can somewhat understand why.

Gamenot (approved):

smarts/core/plan.py
Gamenot: > Actually, it might not be too hard to replace the `include_junctions` parameter to the `nearest_lane` API with something like `bias_away_from_junction_prob` (probably with a shorter name). When this parameter is 0, then it is equivalent to having specified `include_junctions=True`. And when this parameter is 1.0, then it is equivalent to `include_junctions=False`, but when the parameter is something in between -- say 0.3 -- then it scales the score of junction lanes in the sort (which is currently only done by distance) by (one minus) this value. In this case, I would use something like `0.25`. Does this seem useful?

It seems like it might be useful although I am slightly hesitant though to go with a metric like that since it could add complications for debugging.

sah-huawei: Good point.  Then I think I'll leave it as I have it for now.

PR #1499, opened 2022-07-11
Video recording utilities

AisenGinn:
Added gif_recorder.py and recorder_wrapper.py into smarts.env.wrappers folder for video recording function.


smarts/env/wrappers/gif_recorder.py
```diff
@@ -0,0 +1,40 @@
+import os
+
+from moviepy.editor import *
+import gym.envs
+import shutil
+
+
+class GifRecorder:
+    def __init__(self, dir, env):
+        self.dir = dir
```
Gamenot, line 10: An improvement I could suggest is to add a timestamp or `uuid` component to the directory so there is less likelihood of collisions if there are multiple of these running.

smarts/env/wrappers/gif_recorder.py
```diff
@@ -0,0 +1,40 @@
...
+
+        self.dir_name = None
+        if "/" not in dir:
+            self.dir_name = dir
+        else:
+            last_index = dir.rindex("/")
+            self.dir_name = dir[-(len(dir) - last_index - 1) :]
+
+    def capture_frame(self, step_num):
+        image = self.env.render(mode="rgb_array")
```
Gamenot, line 26: There are three directions we could go with this, we can keep as is:

```suggestion
    def capture_frame(self, step_num):
        image = self.env.render(mode="rgb_array")
```

We can remove `self.env` and pass in the environment on render:

```suggestion
    # pass in env
    def capture_frame(self, step_num, env):
        image = env.render(mode="rgb_array")
```

or we could only pass in the image (my preference):

```suggestion
    # elsewhere call 
    # > `gif_recorder.capture_frame(
    # >     get_next_id(),
    # >     env.render(mode="rgb_array"),
    # > )  
    def capture_frame(self, step_num, image):
```

smarts/env/wrappers/recorder_wrapper.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def next_frame_id(self):
+        self.current_frame += 1
+        return self.current_frame
+
+    def close(self):
+        self.gif_recorder.close_recorder()
+
+    def __del__(self):
+        self.gif_recorder.close_recorder()
```
Gamenot, line 56:
```suggestion
    def __del__(self):
        self.close()
```

smarts/env/wrappers/recorder_wrapper.py
```diff
@@ -0,0 +1,60 @@
...
+            self.gif_recorder.capture_frame(self.next_frame_id())
+
+        return observations, rewards, dones, infos
+
+    def next_frame_id(self):
+        self.current_frame += 1
+        return self.current_frame
+
+    def close(self):
+        self.gif_recorder.close_recorder()
```
Gamenot, line 53: I could recommend removing `close_recorder()` and just moving the `gif` generation to here.

```suggestion
    def close(self):
        if self.gif_recorder is not None:
            self.gif_recorder.generate_gif()
            self.gif_recorder.close_recorder()
            self.gif_recorder = None
            self.recording = False
```

Gamenot (changes requested):
Please also remove the `SMARTS` submodule included in this branch.

Gamenot (approved):

Gamenot:
@AisenGinn Please bug people to review this so it does not sit too long.

smarts/env/wrappers/gif_recorder.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        with ImageClip(image) as image_clip:
+            image_clip.save_frame(f"{self.dir}/{self._dir_name}_{step_num}.jpeg")
+
+    def generate_gif(self):
+        """
+        Use the images in the same folder to create a gif file.
+        """
+        with ImageSequenceClip(self.dir, fps=10) as clip:
+            clip.write_gif(f"videos/{self._dir_name}.gif")
```
saulfield, line 66: Should this be relative to `self.dir` as well?

smarts/env/wrappers/recorder_wrapper.py
```diff
@@ -0,0 +1,104 @@
...
+
+class RecorderWrapper(gym.Wrapper):
+    """
+    A Wrapper that interacts the gym environment with the GifRecorder to record video step by step.
+    """
+
+    def __init__(self, dir, env):
+
+        try:
+            os.mkdir("videos")
```
saulfield, line 38: Same feedback about the try/catch here. Also, I think this directory should be explicit about where it is created.

saulfield (changes requested):

smarts/env/wrappers/gif_recorder.py
```diff
@@ -0,0 +1,83 @@
...
+import gym
+import numpy as np
+
+try:
+    from moviepy.editor import ImageClip, ImageSequenceClip
+except (ImportError, ModuleNotFoundError):
+    print(sys.exc_info())
+    print(
+        "You may not have installed the [gym] dependencies required to capture the video. Install them first using the command `pip install -e .[gym]` at the source directory."
+    )
```
Gamenot, line 35: I think my last concern here is that this `try/catch` swallows the import error. I think we still want to raise an error here. Aside from that we should not expect that the user is working from a dev platform.
```suggestion
try:
    from moviepy.editor import ImageClip, ImageSequenceClip
except (ImportError, ModuleNotFoundError):
    logging.warning(sys.exc_info())
    logging.warning(
        "You may not have installed the [gym] dependencies required to capture the video. Install them first with the `smarts[gym]` extras."
    )
    raise
```

Gamenot (changes requested):
One set of issues to figure out. Good Work!

saulfield (approved):

Gamenot (approved):
Good work!

PR #1506, opened 2022-07-18
Detect overpasses in Waymo maps

saulfield:
Detect overpasses in Waymo maps and remove them from the set of lanes tagged as intersections, since this will cause problems for traffic simulation. We now also expose a `has_overpasses` property in the `WaymoMap` class so users have the option of filtering out maps with overpasses.


smarts/core/waymo_map.py
```diff
@@ -112,6 +116,7 @@ def __init__(self, map_spec: MapSpec, waymo_scenario):
         self._default_lane_width = WaymoMap.DEFAULT_LANE_WIDTH
         self._lane_rtree = None
         self._no_composites = False  # for debugging purposes
+        self.has_overpasses = False
```
sah-huawei, line 119: I don't think we want to allow class users to be able to set this member.  This, I would suggest using `self._has_overpasses` and then exposing that as a property.  Also, I think this property is worth adding to the `RoadMap` interface too (with a default implementation there that just returns `False` so the other maps don't have to implement it themselves).

smarts/core/waymo_map.py
```diff
@@ -305,17 +310,28 @@ def _compute_lane_intersections(self) -> Dict[int, Set[int]]:
                         intersections[cand_id].add(feat_id)
                         break
 
-        # Remove "fake" incoming/outgoing lanes that aren't true intersections
+        # Remove lanes that aren't true intersections
         mappings_to_remove = []
         for feat_id, intersect_ids in intersections.items():
-            lane_pts = self._polyline_cache[feat_id][0]
+            lane_pts = np.array(self._polyline_cache[feat_id][0])
```
sah-huawei, line 316: Since we're also doing this for every lane above (in the loops where we set `line1` and `line2`), we may as well just save the result of doing this once somewhere (probably in the `Lane` class or in the `_polyline_cache` itself?) and not repeat the work.

smarts/core/waymo_map.py
```diff
@@ -305,17 +310,28 @@ def _compute_lane_intersections(self) -> Dict[int, Set[int]]:
...
+            z_avg = np.average(lane_pts[:, 2])
             for intersect_id in intersect_ids:
-                intersect_lane_pts = self._polyline_cache[feat_id][0]
-                if tuple(lane_pts[0]) == tuple(intersect_lane_pts[-1]) or tuple(
-                    lane_pts[-1]
-                ) == tuple(intersect_lane_pts[0]):
+                intersect_lane_pts = np.array(self._polyline_cache[intersect_id][0])
```
sah-huawei, line 319: note:  repeated/redundant work

smarts/core/waymo_map.py
```diff
@@ -172,7 +172,7 @@ def _raycast_boundaries(
                     if not (n.self_start_index <= i <= n.self_end_index):
                         continue
                     feature = self._waymo_features[n.feature_id]
```
sah-huawei, line 174: `feature` is no longer needed (neither here nor on line 174).

sah-huawei (approved):

smarts/core/road_map.py
```diff
@@ -61,6 +61,11 @@ def bounding_box(self) -> Optional[BoundingBox]:
         """
         raise NotImplementedError()
 
+    @property
+    def has_overpasses(self) -> bool:
+        """Whether the map has lanes with overlapping z-coordinates."""
+        return False
```
sah-huawei, line 67: I think we should report this value in the waymo browser tool (when a scenario is "explored").

Gamenot: I think this might be a bit misleading terminology-wise in the case of tunnels. (although perhaps any 3d dimension is an overpass technically)

smarts/core/waymo_map.py
```diff
@@ -305,17 +302,28 @@ def _compute_lane_intersections(self) -> Dict[int, Set[int]]:
...
             for intersect_id in intersect_ids:
-                intersect_lane_pts = self._polyline_cache[feat_id][0]
-                if tuple(lane_pts[0]) == tuple(intersect_lane_pts[-1]) or tuple(
-                    lane_pts[-1]
-                ) == tuple(intersect_lane_pts[0]):
+                intersect_lane_pts = self._polyline_cache[intersect_id][0]
+                intlane_z_avg = np.average(intersect_lane_pts[:, 2])
+
+                # Remove "overpasses" that have large z-coordinate differences
+                if abs(z_avg - intlane_z_avg) > WaymoMap.OVERPASS_THRESHOLD:
```
Gamenot, line 315: Sorry, why is this needed?

Gamenot (approved):

PR #1508, opened 2022-07-18
Log step time

Gamenot:


saulfield (approved):

sah-huawei (commented):
This seems like overkill.  In my opinion, a better approach would just be to configure `logging` such that our default log message format includes the timestamp with _every_ message.

Something like:
```python
logging.basicConfig(level=logging.WARNING, format="%(asctime)s [%(levelname)s %(name)s] %(message)s")
```

(Of course, if we wanted to, we could get more fancy with this and add our own global  `config_logging(level, format)` function (with various other "convenience parameters" and default values) that  wraps the one in the `logging` module, but at least for this particular problem, I don't think we need that.)

Gamenot:
> This seems like overkill. In my opinion, a better approach would just be to configure `logging` such that our default log message format includes the timestamp with _every_ message.
> 
> Something like:
> 
> ```python
> logging.basicConfig(level=logging.WARNING, format="%(asctime)s [%(levelname)s %(name)s] %(message)s")
> ```
> 
> (Of course, if we wanted to, we could get more fancy with this and add our own global `config_logging(level, format, path, ...)` function (with various other "convenience parameters" and default values) that wraps the one in the `logging` module, but at least for this particular problem, I don't think we need that.)

Disagree since I think it simplifies finding issues within the simulator at least if isolated to the `step()` level. Time-step is already configured in the way you mention but it is not clear without the additional effort of subtracting between specific where simulation cost is.

However, I think I will revert the physics of it because it resides at a different level of abstraction.

sah-huawei:
> > This seems like overkill. In my opinion, a better approach would just be to configure `logging` such that our default log message format includes the timestamp with _every_ message.
> > Something like:
> > ```python
> > logging.basicConfig(level=logging.WARNING, format="%(asctime)s [%(levelname)s %(name)s] %(message)s")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (Of course, if we wanted to, we could get more fancy with this and add our own global `config_logging(level, format, path, ...)` function (with various other "convenience parameters" and default values) that wraps the one in the `logging` module, but at least for this particular problem, I don't think we need that.)
> 
> Disagree since I think it simplifies finding issues within the simulator at least if isolated to the `step()` level. Time-step is already configured in the way you mention but it is not clear without the additional effort of subtracting between specific where simulation cost is.
> 
> However, I think I will revert the physics of it because it resides at a different level of abstraction.

OK.
Re your point above, then I would suggest we do _both_ things.  That is, include the timestamps as part of our default log message format and _also_ do what you have here for the specific key parts of `step()`.

sah-huawei (approved):

PR #1511, opened 2022-07-19
Junction traffic

sah-huawei:
Updates to the `LocalTrafficProvider` to try to avoid collisions in junctions.  Also adds `SmartsJunctionModel` to `smarts.sstudio.types`.

As a necessary precondition, support for finding a lane's `foes` was added to all map types.  A few bugs were fixed in the `WaymoMap` loading along the way, as well as 

This also brings in part of PR #1083 to add the _start_ of _basic_ traffic light support for Waymo and Sumo maps.  (OpenDRIVE is still TBD.)  These are not yet exposed to agents via Observations, although it shouldn't be too hard to do at this point.  As part of this, I fleshed out the `Feature` part of the `RoadMap` class a bit more, as well as added the `SignalProvider` class.


smarts/core/utils/math.py
```diff
@@ -336,11 +336,17 @@ def line_intersect(a, b, c, d) -> Union[np.ndarray, None]:
...
+    a: np.ndarray,
+    b: np.ndarray,
+    C: np.ndarray,
+    D: np.ndarray,
+    ignore_start_pt: bool = False,
 ) -> bool:
     """Vectorized version of `line_intersect(...)`, where C and D represent
     the segment points for an entire line, and a and b are points of a single
     line segment to be tested against.
+    If ignore_start_pt is True, then two diverging lines that *only* intersect at
```
saulfield, line 348: What was the need for this? It looks like this is only used when determining intersections in Opendrive.

smarts/sstudio/types.py
```diff
@@ -172,6 +172,35 @@ def __init__(
...
+        yield_to_agents:
+            There are 3 possible values for this:
+                - "always" - traffic actors will yield to Ego and Social
+                  agents within juncions;
+                - "never" - traffic actors will never yield to Ego or Social
+                  agents within junctions;
+                - "normal" - traffic actors will attempt to honor normal
+                  right-of-way conventions, only yielding when an agent
+                  has the right-of-way.  Examples of such conventions include:
+                    - Vehicles going stright have the right-of-way over
```
saulfield, line 187:
```suggestion
                    - Vehicles going straight have the right-of-way over
```

smarts/sstudio/types.py
```diff
@@ -172,6 +172,35 @@ def __init__(
...
+                      turning vehicles;
+                    - Vehicles on roads with more lanes have the right-of-way
+                      relative to vehicles on intersecting roads with less lanes;
+                    - All other things being equal, the vehicle to the right
+                      (in the counter-clockwise direction) has the right-of-way.
+        wait_to_restart:
+            The amount of time in seconds after stopping at a signal or stop sign
+            befor this vehicle will start to go again.  Default:  0.0
```
saulfield, line 195:
```suggestion
            before this vehicle will start to go again.  Default:  0.0
```

smarts/core/utils/math.py
sah-huawei: Yeah, it's for OpenDrive lanes that start at the same point but diverge from there; we don't want to count these as intersections (because the traffic in these two lanes does not interfere with each other).  This usually seems to happen\ when a single lane is being split into two.

smarts/sstudio/genhistories.py
```diff
@@ -42,8 +43,10 @@
 
 try:
     from waymo_open_dataset.protos import scenario_pb2  # pytype: disable=import-error
+    from waymo_open_dataset.protos.map_pb2 import (
+        TrafficSignalLaneState,
+    )  # pytype: disable=import-error
```
Gamenot, line 48: It is surprising this is needed inside a catch for import error. I could suggest wrapping the import block.
```suggestion
    # pytype: disable=import-error
    from waymo_open_dataset.protos import scenario_pb2  
    from waymo_open_dataset.protos.map_pb2 import (
        TrafficSignalLaneState,
    )
    # pytype: enable=import-error
```

smarts/core/vehicle_index.py
```diff
@@ -558,7 +558,7 @@ def _switch_control_to_agent_recreate(
 
         # Reserve space inside the traffic sim
         for traffic_sim in sim.traffic_sims:
-            if traffic_sim.manages_vehicle(vehicle.id):
+            if traffic_sim.manages_actor(vehicle.id):
```
Gamenot, line 561: This now seems slightly odd, `vehicle.id` might work but it is not specifically the actor id.

smarts/core/vehicle.py
```diff
@@ -89,13 +87,6 @@ def bbox(self) -> Polygon:
         )
         return shapely_rotate(poly, self.pose.heading, use_radians=True)
 
-    def __lt__(self, other):
-        """Allows ordering VehicleStates for use in sorted data-structures."""
-        assert isinstance(other, VehicleState)
-        return self.vehicle_id < other.vehicle_id or (
-            self.vehicle_id == other.vehicle_id and id(self) < id(other)
-        )
```
Gamenot, line 97: This would appear to be a major change to how the vehicle states are compared, does this have to do with the ordering issues?

sah-huawei: This has just been moved to the parent class (`ActorState` in `actor.py`).  I think it should be inherited from there and have the same functionality as before.

smarts/core/traffic_history.py
```diff
@@ -310,6 +310,26 @@ def vehicles_active_between(
...
+        sim_time: float
+        state: int
+        stop_point_x: float
+        stop_point_y: float
+        lane_id: int
+
+    def traffic_light_states_between(
+        self, start_time: float, end_time: float
+    ) -> Generator[TrafficHistory.TrafficLightRow, None, None]:
+        """Find all traffic light states betwen the given history times."""
```
saulfield, line 325:
```suggestion
        """Find all traffic light states between the given history times."""
```

smarts/sstudio/genhistories.py
```diff
@@ -889,6 +932,41 @@ def constrain_angle(angle: float) -> float:
...
+            return SignalLightState.CAUTION | SignalLightState.FLASHING
+        return SignalLightState.UNKNOWN
+
+    @property
+    def traffic_light_rows(self) -> Generator[Dict, None, None]:
+        scenario = self._get_scenario()
+        num_steps = len(scenario.timestamps_seconds)
+        for i in range(num_steps):
+            dynamic_states = scenario.dynamic_map_states[i]
+            for j, lane_state in enumerate(dynamic_states.lane_states):
```
saulfield, line 960: Looks like `j` is unused. Also we can probably pull the `sim_time` computation out of this loop.

smarts/core/local_traffic_provider.py
```diff
@@ -1039,20 +1112,423 @@ def _slow_for_curves(self):
...
+        # TODO:  check non-fixed-location signals here too
+        if not upcoming_feats:
+            return
+        for feat in upcoming_feats:
+            if feat.type == RoadMap.FeatureType.SPEED_BUMP:
+                self._target_speed *= 0.5
+                continue
+            # if feat.type == RoadMap.FeatureType.CROSSWALK and TODO:
+            #     self._target_speed = 0
+            #     continue
```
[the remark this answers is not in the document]
sah-huawei: I left that in as a place-holder to indicate where cross-walks would be handled.
But I agree that it's not good to leave commented-out code in.  I've removed it.

smarts/sstudio/genhistories.py
sah-huawei: Done.

saulfield (approved):

saulfield:
This looks good to me. One last suggestion would be to add some basic tests so we have at least partial coverage of this code moving forward (if you're able to).

sah-huawei:
> This looks good to me. One last suggestion would be to add some basic tests so we have at least partial coverage of this code moving forward (if you're able to).

Good call.  I just added a basic test of collision avoidance -- at least for a little while -- by the traffic provider in a T junction.  (Hopefully, you guys can add to this over time too.)

Gamenot (approved):

PR #1512, opened 2022-07-19
Numpy Points

sah-huawei:
Optimization and refactor/cleanup to streamline the frequently-used conversion between `Point` and `numpy.array`.

### DIFF
--- a/smarts/core/coordinates.py
+++ b/smarts/core/coordinates.py
@@ -106,7 +106,6 @@ class Point(NamedTuple):
         # So we use the private global _numpy_points as a cache instead.
         # Here we are relying on CPython's implementation of dict
         # to be thread-safe.
-        cached = _shapely_points.get(self)
         cached = _numpy_points.get(self)
         if cached is not None:
             return cached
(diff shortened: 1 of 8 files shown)


smarts/core/coordinates.py
```diff
@@ -85,6 +86,34 @@ class Point(NamedTuple):
...
+        # with a cache of the result.
+        # Note that before python3.8, @cached_property was not thread safe,
+        # nor can it be used in a NamedTuple (which doesn't have a __dict__).
+        # (Points can be used by multi-threaded client code, even when
+        # SMARTS is still single-threaded, so we want to be safe here.)
+        # So we use the private global _numpy_points as a cache instead.
+        # Here we are relying on CPython's implementation of dict
+        # to be thread-safe.
+        cached = _shapely_points.get(self)
+        cached = _numpy_points.get(self)
```
Adaickalavan, line 110: Do we need line 109, as the `cached` in line 109 is overwritten by line 110?

sah-huawei: Good catch.  You're right.
That was an accidental paste error.
Removed.

saulfield (approved):

PR #1521, opened 2022-07-22
Numpy array cache fix

sah-huawei:
Fixes a bug where a numpy array corresponding to a `Point` was getting modified and thus corrupting the `Point` cache.

### DIFF
--- a/smarts/core/coordinates.py
+++ b/smarts/core/coordinates.py
@@ -89,7 +89,7 @@ class Point(NamedTuple):
     @classmethod
     def from_np_array(cls, np_array: np.ndarray):
         """Factory for constructing a Point object from a numpy array."""
-        assert 2 <= len(np.array) <= 3
+        assert 2 <= len(np_array) <= 3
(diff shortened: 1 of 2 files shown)


saulfield (approved):

smarts/core/coordinates.py
```diff
@@ -90,12 +90,12 @@ class Point(NamedTuple):
     def from_np_array(cls, np_array: np.ndarray):
         """Factory for constructing a Point object from a numpy array."""
         assert 2 <= len(np.array) <= 3
-        z = np_array[2] if len(np_array) > 2 else 0.0
+        z = np_array[2] if len(np.array) > 2 else 0.0
```
Adaickalavan, line 93: Should this be `np.array` or `np_array`?

sah-huawei: doh!!! yes.  Thanks.
I did the "fix" in reverse (b/c I thought it hadn't been made yet).

Adaickalavan (approved):

PR #1529, opened 2022-07-27
expose traffic signal state to agents

sah-huawei:
This PR is meant to address issue #1079 (along with parts of PR #1511).  Together they replace PR #1083.

Also note that although I did add the interface to emit traffic signal state to Envision, I didn't actually add anything to render that  at this point.  (That will need to be in another PR.)


Gamenot:
The error I am seeing is this:
```bash
============================================================================= FAILURES ==============================================================================
_____________________________________________________________________ test_signal_observations ______________________________________________________________________

smarts = <smarts.core.smarts.SMARTS object at 0x7f482a4d26a0>
scenario = Scenario(
  _root=scenarios/sumo/intersections/2lane,
  _traffic_specs=['scenarios/sumo/intersections/2lane/traffic/ve...ic(wait_to_hijack_limit_s=0, zone=None, exclusion_prefixes=(), default_entry_speed=1.0), via=(), vehicle_spec=None)},
)

    def test_signal_observations(smarts, scenario):
        observations = smarts.reset(scenario)
    
        # go REAL SLOW so the light can change...
        agent = Agent.from_function(lambda _: (1.0, 0))
    
        # Let the agent step for a while
        for step in range(900):
            agent_obs = observations[AGENT_ID]
            agent_action = agent.act(agent_obs)
            observations, _, dones, _ = smarts.step({AGENT_ID: agent_action})
            if dones[AGENT_ID]:
                break
            my_obs = observations[AGENT_ID]
            signals = my_obs.signals
            pos = my_obs.ego_vehicle_state.position
            if step < 155:
                # it's out of range...
                assert len(signals) == 0, f"step={step}, pos={pos}"
            else:
>               assert len(signals) == 1, f"step={step}, pos={pos}"
E               AssertionError: step=155, pos=[ 1.87697251e+01  6.83999922e+01 -2.65268871e-02]
E               assert 0 == 1
E                +  where 0 = len([])

smarts/core/tests/test_observations.py:298: AssertionError
========================================================================= warnings summary ==========================================================================
.venv/lib/python3.8/site-packages/trimesh/curvature.py:12
  /home/dev/Desktop/repos/SMARTS/.venv/lib/python3.8/site-packages/trimesh/curvature.py:12: DeprecationWarning: Please use `coo_matrix` from the `scipy.sparse` namespace, the `scipy.sparse.coo` namespace is deprecated.
    from scipy.sparse.coo import coo_matrix

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
====================================================================== short test summary info ======================================================================
FAILED smarts/core/tests/test_observations.py::test_signal_observations - AssertionError: step=155, pos=[ 1.87697251e+01  6.83999922e+01 -2.65268871e-02]
```

Gamenot:
Same environment passed when using `python3.7.13`.

sah-huawei:
> The error I am seeing is this under python3.8.10:
> ...
> FAILED smarts/core/tests/test_observations.py::test_signal_observations - AssertionError: step=155, pos=[ 1.87697251e+01  6.83999922e+01 -2.65268871e-02]
> ```

interestingly, those are different values than what it fails with in the CI version.  The CI appears to be using python version `3.7.11`.

So it passes on python versions `3.7.5` (my laptop) and `3.7.13` (@Gamenot's machine), but not versions `3.7.11` (the CI runner) nor `3.8.10` (@Gamenot's machine).  And when it fails, it has different values on each of the failures.  So odd!!

sah-huawei:
> Same environment passed when using `python3.7.13`.

So I "fixed" the test by moving the starting point nearer to the traffic signal, so there is less steps for any floating-point error/imprecision between python versions to accumulate.  After all, this test really has nothing to do with position imprecision while cruising, but rather just properly detecting signal state transitions.  

However, despite tests all passing here now, I would recommend that at some point in the future, the SMARTS team try to understand where the python version differences manifest themselves in the code.  This test could be used to explore that simply by recording agent vehicle positions on each step and comparing these positions across versions.  Once a divergence is noticeable, you can "zoom in" on that step within the code and try to figure out which mathematical _operations_ diverge between the python versions.  Once that's understood, it might be possible to improve the code there to avoid the imprecise math and/or file a bug with the python development community if it looks like the newer versions are more incorrect relative to the older versions.

Gamenot:
> > Same environment passed when using `python3.7.13`.
> 
> So I "fixed" the test by moving the starting point nearer to the traffic signal, so there is less steps for any floating-point error/imprecision between python versions to accumulate. After all, this test really has nothing to do with position imprecision while cruising, but rather just properly detecting signal state transitions.
> 
> However, despite tests all passing here now, I would recommend that at some point in the future, the SMARTS team try to understand where the python version differences manifest themselves in the code. This test could be used to explore that simply by recording agent vehicle positions on each step and comparing these positions across versions. Once a divergence is noticeable, you can "zoom in" on that step within the code and try to figure out which mathematical _operations_ diverge between the python versions. Once that's understood, it might be possible to improve the code there to avoid the imprecise math and/or file a bug with the python development community if it looks like the newer versions are more incorrect relative to the older versions.

Hmm, looking into it, I think you are right to consider that this may be a case of different handling of floating point values. I think the other source of problem could potentially come through `numpy`.

smarts/env/wrappers/format_obs.py
```diff
@@ -357,7 +378,12 @@ def get_spaces() -> Dict[str, Callable[[Any], gym.Space]]:
             "lane_width": gym.spaces.Box(low=0, high=1e10, shape=_WAYPOINT_SHP, dtype=np.float32),
             "pos": gym.spaces.Box(low=-1e10, high=1e10, shape=_WAYPOINT_SHP + (3,), dtype=np.float64),
             "speed_limit": gym.spaces.Box(low=0, high=1e10, shape=_WAYPOINT_SHP, dtype=np.float32),
-        })
+        }),
+        "signals": lambda _: gym.spaces.Dict({
+            "state": gym.spaces.Box(low=0, high=32, shape=(_SIGNALS_SHP,), dtype=np.int8),
+            "stop_point": gym.spaces.Box(low=-1e10, high=1e10, shape=(_SIGNALS_SHP, 2), dtype=np.float64),
```
Gamenot, line 384: I could suggest making `_SIGNALS_SHP` a tuple since shape deals with dimensions rather than scalar.

```suggestion
            "state": gym.spaces.Box(low=0, high=32, shape=_SIGNALS_SHP, dtype=np.int8),
            "stop_point": gym.spaces.Box(low=-1e10, high=1e10, shape=_SIGNALS_SHP  + (2,), dtype=np.float64),
```

or

```suggestion
            "state": gym.spaces.Box(low=0, high=32, shape=_SIGNALS_SHP, dtype=np.int8),
            "stop_point": gym.spaces.Box(low=-1e10, high=1e10, shape=(_SIGNALS_SHP[0], 2), dtype=np.float64),
```

smarts/core/sensors.py
```diff
@@ -191,6 +208,7 @@ class Observation:
     """Amount of simulation time the last step took."""
     step_count: int
     """Number of steps taken by SMARTS thus far."""
+    steps_completed: int
```
Gamenot, line 211: I am not sure how this observation got in without describing it but:
```suggestion
    steps_completed: int
    """The number of steps this agent has taken within SMARTS."""
```

smarts/core/sensors.py
```diff
@@ -1209,3 +1239,91 @@ def closest_point_on_lane(position, lane_id):
...
+        lane_pos: RefLinePoint,
+        state,  # VehicleState
+        plan: Plan,
+        provider_state,  # ProviderState
+    ) -> List[SignalObservation]:
+        result = []
+        if not lane:
+            return result
+        upcoming_signals = []
+        half_len = 0.5 * state.dimensions.length
```
Gamenot, line 1272: Is this necessary? By the time the car is half-way past the signal, there is already consequence for violating or obeying the signal. Or the signal still affects the rules even though the vehicle has passed it.

smarts/core/sensors.py
```diff
@@ -1209,3 +1239,91 @@ def closest_point_on_lane(position, lane_id):
...
+        if lookahead <= 0:
+            return
+        for ogl in lane.outgoing_lanes:
+            if route and ogl.road not in route.roads:
+                continue
+            upcoming_signals += [
+                feat for feat in ogl.features if self._is_signal_type(feat)
+            ]
+            lookahead -= lane.length
+            self._find_signals_ahead(ogl, lookahead, route, upcoming_signals)
```
Gamenot, line 1326: I kind of wish we had a convention for marking recursion since this is important for road map scaling.

smarts/core/sensors.py
sah-huawei: Yeah, it's probably not.  I was trying to be a bit "safe" with respect to the "sampling rate" (i.e., stepping rate) of the vehicle whereby it might move adjacent to the signal within one step if it's still moving too fast to stop on time.  I.e., I was trying to allow for an agent to not be so good at stopping at signals precisely.

But this might not be what someone who's training an agent to stop at signals would want.  So I'm fine with taking this out.   Will do, but will add a comment about this....

smarts/core/sensors.py
sah-huawei: Yeah, that would have been good.  What would you suggest?  I can at least start using that here.

envision/data_formatter.py
```diff
@@ -330,6 +340,28 @@ def _format_traffic_actor_type(
...
+        SignalLightState.Go: 3,
+    }
+    data_formatter.add_primitive(mapping[obj])
+
+
+def _format_signal_state(obj: SignalState, data_formatter: EnvisionDataFormatter):
+    assert type(obj) is SignalState
+    data_formatter.add(obj.signal_id, op=Operation.REDUCE)
+    assert type(obj.signal_light_state) is SignalLightState
+    data_formatter.add(obj.signal_light_state)
```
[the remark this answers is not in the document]
sah-huawei: Yeah, that might be better overall.   Something to consider changing in the future I guess.

saulfield (approved):

Adaickalavan (approved):

smarts/env/wrappers/format_obs.py
```diff
@@ -177,6 +179,22 @@
...
+            See smarts.core.signal_provider.SignalLightState for interpretation.
+            Defaults to np.array([0]) per signal.  shape=(3,), dtype=np.int8.
+        "stop_point":
+            The stopping point for traffic controlled by the signal, i.e., the
+            point where actors should stop when the signal is in a stop state.
+            Defaults to np.array([0, 0]) per signal.  shape=(3,2), dtype=np.float64.
+        "last_changed":
+            If known, the simulation time this signal last changed its state.
+            Defaults to np.array([0]) per signal.  shape=(3,), dtype=np.float32.
+    })
```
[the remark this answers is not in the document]
sah-huawei: Thanks! :)

Gamenot:
Great work!

PR #1537, opened 2022-07-30
Add `bubble_env` contribution

Gamenot:


competition/evaluation/copy_data.py
```diff
@@ -10,7 +10,10 @@ def __init__(self):
         self._agent_names = None
 
     def __call__(self, **kwargs):
-        self._data = copy.deepcopy(dict(**kwargs))
+        try:
+            self._data = copy.deepcopy(dict(**kwargs))
+        except RecursionError:
+            self._data = copy.copy(dict(**kwargs))
```
Adaickalavan, line 16: `deepcopy` was used to prevent external users from modifying some of the `infos` objects later using additional env wrappers.   

How does `RecursionError` occur and would a shallow copy be a sufficient tamper protection here?

Gamenot: I am unsure how the `RecursionError` occurred but some part of the SMARTS `Observation` type causes it. I intend to fix this during the competition.

Gamenot: https://github.com/huawei-noah/SMARTS/issues/1540

Gamenot: @Adaickalavan To be clear, I think this is fine to update because it is only useful to us.

competition/track1/train/train/info.py
```diff
@@ -28,6 +28,6 @@ def step(
         obs, reward, done, info = self.env.step(action)
 
         for agent_id in info.keys():
-            info[agent_id]["is_success"] = bool(info[agent_id]["score"])
+            info[agent_id]["is_success"] = bool(info[agent_id].get("score", True))
```
Adaickalavan, line 31: The `is_success` is set here to enable the SB3 library to count the number of successfully completed episodes and to print it during evaluation. An episode is considered to be successfully completed if the agent reaches the goal, i.e., `obs.events.reached_goal=True`.

Why is the `score` being set to `True` by default here, as this is misleading?

competition/competition_bundle/track1_evaluation/config.yaml
```diff
@@ -2,3 +2,4 @@ phase: track1
 eval_episodes: 50
 seed: 42
 scenarios: []
+bubble_env_evaluation_seeds: []
```
Adaickalavan, line 5: If `bubble_env_evaluation_seeds` is an empty list, then the additional bubble_env will not be run during Codalab evaluation (i.e., not included in competition scoring and ranking). It only runs during local evaluation by participants as the default config sets `bubble_env_evaluation_seeds=[6]`.

Was this the intended outcome?

competition/evaluation/evaluate.py
```diff
@@ -15,6 +15,7 @@
     "eval_episodes",
     "seed",
     "scenarios",
+    "bubble_env_evaluation_seeds",
```
Adaickalavan, line 18: Consider using short variable names. Here, we could use `bubble_env_seeds` instead of `bubble_env_evaluation_seeds`.

PR #1541, opened 2022-08-01
traffic histories to observations

sah-huawei:
Revamp of `traffic_histories_to_observations.py` script (not yet fully tested).

Many of the features of the script from PR #1461 are made here, however this is intended to not be specific to only Waymo datasets like #1461 was.

Still TBD: 
- document how to use the script
- test more thoroughly


examples/traffic_histories_to_observations.py
```diff
@@ -1,130 +1,302 @@
...
+            if ego_state.lane_index is None:
+                del obs[id_]
+            else:
+                mission = self._missions[ego_state.id]
+                if mission:
+                    # doh! ego_state is immutable!
+                    new_ego_state = ego_state._replace(mission=mission)
+                    obs[id_] = replace(obs[id_], ego_vehicle_state=new_ego_state)
+        # TODO: handle case where neighboring vehicle has lane_index of None too
+        obs = self._std_obs_wrapper.observation(obs)
```
Adaickalavan, line 232: - Why did we not just return the `smarts.sensors.Observation` directly?
- Why did we use the `FormatObs` wrapper here to reformat the observation to `StdObs`?
- By using `FormatObs` wrapper here, we would be making this standard format mandatory for all users. Some users may not realise that this wrapper has been applied internally.
- Can we simply return `smarts.sensors.Observation`?

sah-huawei: I did it this way b/c this is what @saulfield did originally in PR #1461.  So I'll let him answer these questions.

saulfield:
Closing since this has been superseded by #1545.

PR #1545, opened 2022-08-03
Waymo tools and offline dataset extraction

saulfield:
This PR supersedes #1541 and #1461.

### Changes
- Add `waymo` commands to `scl` to:
  - Display information about all the scenarios within a TFRecord file
  - Plot the map, and optionally show vehicles with their IDs or animate their trajectories
  - Export a scenario to a SMARTS scenario that can be replayed
- Update the `traffic_histories_to_observations.py` example to work with any traffic history format (e.g. NGSIM, Waymo), and to output the full `Observation` structure and set of birds-eye-view images (e.g. for use in offline RL)
- Updates to `traffic_histories_vehicle_replacement.py` to make it compatible with the new data format


examples/traffic_histories_vehicle_replacement.py
```diff
@@ -183,6 +198,7 @@ def custom_filter(vehs: Iterable[VehicleWindow]) -> Iterable[VehicleWindow]:
             ego_missions = {}
             sample = set()
 
+            print(scenario.traffic_history.dataset_source)
```
Gamenot, line 201: We should be using logging rather than prints generally, especially since we already have logger being used.

examples/traffic_histories_vehicle_replacement.py
```diff
@@ -61,27 +61,42 @@ def act(self, obs: Observation) -> Tuple[float, float]:
...
+            cur_state.position[1], exp_state.position[1], abs_tol=2
+        ), f"vid={self._vehicle_id}: {cur_state.position[1]} != {exp_state.position[1]} @ {obs_time}"
 
         # Then get and return the next set of control inputs
         atime = self._rounder(obs_time + self._fixed_timestep_sec)
-        data = self._data.get(atime, {"acceleration": 0, "angular_velocity": 0})
-        return (data["acceleration"], data["angular_velocity"])
+        next_obs = self._data.get(atime)
+        if not next_obs:
+            return (0.0, 0.0)
```
Gamenot, line 86: Should this ever happen?

examples/traffic_histories_vehicle_replacement.py
saulfield: This was actually a debug statement I forgot to remove. Thanks for catching!

examples/traffic_histories_vehicle_replacement.py
saulfield: Yes I think so, perhaps on the last timestep? This guard was in the original script, so I'm assuming it was there for a reason.

examples/traffic_histories_to_observations.py
```diff
@@ -1,130 +1,330 @@
...
 
-    smarts.destroy()
+        if not self._output_dir:
+            return
+
+        # Write top-down RGB image to a file for each vehicle if we have one
+        for agent_id, agent_obs in obs.items():
+            if agent_obs.top_down_rgb is not None:
+                rgb_data = agent_obs.top_down_rgb.data
+                h, w = rgb_data.shape[0], rgb_data.shape[1]
```
Gamenot, line 279: Numpy arrays unpack.

```suggestion
                h, w = rgb_data.shape
```

examples/traffic_histories_to_observations.py
```diff
@@ -1,130 +1,330 @@
...
+    def collect(
+        self, vehicles_with_sensors: Optional[Sequence[int]], headless: bool = True
+    ):
+        """Generate Observations from the perspective of one or more
+        social/history vehicles within a SMARTS scenario.
+
+        Args:
+            vehicles_with_sensors (Sequence[int], optional):
+                A list of vehicle_ids within the scenario to which to attach
+                sensors and record Observations.  If not specified, this will default
```
Gamenot, line 132: I think there is a question of if there should be a difference in how `None` is treated in comparison to `[]`.

Consistency should say that `None` should have this behaviour but `[]` should just be empty.

Gamenot (approved):
Some comments but I think this can go in. Good work.

Adaickalavan:
1. `TRAJECTORY_HANDLES` and `MAP_HANDLES` appear to be the same in `waymo_browser.py` and `waymo_utils.py`. Can they be combined?

2. Several functions appear similar between `waymo_browser.py` and `waymo_utils.py`, e.g., `_plot_trajectories`, `_get_map_features`, `_plot_map_features`. Can they be combined? 

3. Can the `gen_smarts_scenario_code` function in `waymo_utils.py` and `smarts/waymo/templates/scenario_template.py` be combined?

examples/traffic_histories_to_observations.py
```diff
@@ -1,130 +1,330 @@
...
+        orig_missions = self._scenario.discover_missions_of_traffic_histories()
+        for v_id, mission in orig_missions.items():
+            veh_goal = self._scenario._get_vehicle_goal(v_id)
+            # TODO: get prefixed vehicle_id from TrafficHistoryProvider
+            veh_id = f"history-vehicle-{v_id}"
+            self._missions[veh_id] = replace(
+                mission, goal=PositionalGoal(veh_goal, radius=3)
+            )
+
+    def _create_agent_interface(
```
Adaickalavan, line 81: Consider extracting the `AgentInterface` and making it as an input to the `ObservationRecorder` class.

examples/traffic_histories_to_observations.py
saulfield: In general I would agree, but I think in this case there isn't a meaningful distinction since the result will be the same in both cases (script terminating due to having no valid vehicles to record observations for).

cli/studio.py
```diff
@@ -187,58 +187,7 @@ def replay(directory: Sequence[str], timestep: float, endpoint: str):
             )
 
 
-@scenario_cli.command(
-    name="browse-waymo",
```
Adaickalavan, line 191: The command `scl scenario browse-waymo` was referred to in https://github.com/huawei-noah/SMARTS/tree/saul/waymo-utils/smarts/waymo#running-the-utility . 

What should the equivalent new explanation be in the readme or should the scl command example be removed from the readme?

saulfield: I removed it from `scl` to keep it distinct from the newly added `scl waymo` commands. I've now removed that part from the README.

examples/traffic_histories_to_observations.py
saulfield: The `AgentInterface` is now an optional parameter to the `ObservationRecorder` class, and will default to this one if none is provided.

saulfield:
> 1. `TRAJECTORY_HANDLES` and `MAP_HANDLES` appear to be the same in `waymo_browser.py` and `waymo_utils.py`. Can they be combined?
> 
>     2. Several functions appear similar between `waymo_browser.py` and `waymo_utils.py`, e.g., `_plot_trajectories`, `_get_map_features`, `_plot_map_features`. Can they be combined?
> 
>     3. Can the `gen_smarts_scenario_code` function in `waymo_utils.py` and `smarts/waymo/templates/scenario_template.py` be combined?

Thanks @Adaickalavan -- I've implemented all of these suggestions except for the 2nd one, since they are different enough in each tool and I don't think it's worth unifying them at this time. Perhaps in the future though.

Adaickalavan (approved):

Issue #1548, opened 2022-08-04
Fixed Bubble Feature


CyberSY:
**Is your feature request related to a problem? Please describe.**
Since moving bubble(following specific ego/histroy vehicle) has been developed, it would be helpful if bubble can be fixed in specified map area, which can observe interaction behavior in typical road structure. 

**Describe the solution you'd like**
Just like the `create_moving_bubble` function, it would be great to have a similar `create_fixed_bubble` function with position argument. When the histroy vehicle enter the fixed bubble, the bubble should take control of it; when the ego agent entering bubble,it should not be affected.

Gamenot:
Hello @CyberSY, we have this feature already, it just leaves out the `follow_offset` and `follow_vehicle_id`/`follow_agent_id` options:


```python
from smarts.sstudio.types import Bubble, MapZone, PositionalZone, Zone

def create_fixed_bubble(
        zone: Zone,
        exclusion_prefixes: Optional[Tuple[str]] = None,
        margin=0,
        social_agent_name="dummy_agent-v0"
    ):
    bubble = Bubble(
        zone=zone,
        actor=SocialAgentActor(
            name="dummy0", agent_locator=f"examples:{social_agent_name}"
        ),
        exclusion_prefixes=exclusion_prefixes or tuple(),
        margin=margin,
    )
    return bubble

# If wanting to place the bubble by position
fixed_bubble = create_fixed_bubble(PositionalZone(pos=(4, 2), size=(20, 7))
scenario.bubbles.extend(fixed_bubble)
# If wanting to place the bubble by 
if "i80" in scenario.root_filepath:
    lane_index=0
    offset = 20
    map_bubble = create_fixed_bubble(MapZone(start=("gneE01.132", lane_index, offset), length=100, nlanes=3)
    scenario.bubbles.extend(fixed_bubble)
```

CyberSY:
Hi @Gamenot ! We have tested this feature and it works as expected.
For placing the bubble by position, could we add `rotation angle` argument to make bubble have certain angle?

Gamenot:
Hello @CyberSY, we do not have that capability at the moment that will take a bit of work but I can add that capability tomorrow.

Gamenot:
Hello @CyberSY, I have put up a solution. If you try [tucker/rotatable_positional_zones](https://github.com/huawei-noah/SMARTS/tree/tucker/rotatable_positional_zones) or `git cherry-pick 164697c` the following should be possible:


```python
[...]

PR #1549, opened 2022-08-05
Allow rotation of positional zones for bubbles.

Gamenot:
After a lot of different options I settled on setting the base rotation to be frozen into the geometry.


Gamenot:
This is confirmed to work: https://github.com/huawei-noah/SMARTS/issues/1548

Issue #1446, opened 2022-05-27
Get different waypoint heading in NGSIM US101


YuechengLiu:
## **BUG REPORT**

I observed that in `NGSIM us101` environment,  `waypoint.heading` could be `0` or `pi (opposite to the direction of the vehicle, which is not expected)`. It's not an occasional problem and can be easily reproduced.

But I have not observed this problem in  `i80` environment.

PR #1557, opened 2022-08-15
fix us101 map

qianyi-sun:
This is a map improvement for fixing https://github.com/huawei-noah/SMARTS/issues/1446, Judy's team had already validated this and it's just a simple map file change, maybe we can consider their validation as one approval and one more review is needed to merge this


Gamenot (approved):
I looked at the map. Good work!.

PR #1558, opened 2022-08-16
Add alternative for grpc if timeout occurs

Gamenot:
This should provide an alternative to keep the simulation going if `grpc` starts to mess up.


Gamenot:
This is no longer needed.

PR #1566, opened 2022-08-17
Fix import issue in examples

### LINKED ISSUE #1565: ModuleNotFoundError: No module named 'examples.argument_parser'
**High Level Description**
I tested the expample scl command. However, the chrome browser is opened with a black background and no active simulations is started. 

**Desired SMARTS version**
[0.6.1]

**Operating System**
ubuntu 20.04

**Problems**
The command line and outputs are attached as follows:

(smarts) ros2@ros2:~/SMARTS$ scl scenario build --clean scenarios/sumo/loop
build-scenario scenarios/sumo/loop
/home/ros2/anaconda3/envs/smarts/lib/python3.8/site-packages/trimesh/curvature.py:12: DeprecationWarning: Please use `coo_matrix` from the `scipy.sparse` namespace, the `scipy.sparse.coo` namespace is deprecated.
  from scipy.sparse.coo import coo_matrix

(smarts) ros2@ros2:~/SMARTS$ scl run --envision examples/single_agent.py scenarios/sumo/loop 
  File "examples/single_agent.py", line 6, in <module>
    from examples.argument_parser import default_argument_parser
ModuleNotFoundError: No module named 'examples.argument_parser'
Killed

saulfield:
Fixes #1565 

This resolves the inconsistency around relative/absolute imports between `scl run` and running from the tests.


Adaickalavan:
The argument parser is only for running the examples and does not seem related to smarts core. Hence, I am hesitant about moving the argument parser into smarts core.

One approach could be the following:
1. Change the import statement from `from examples.argument_parser import default_argument_parser` to `from argument_parser import default_argument_parser` in the examples. 
2. Include `import sys; sys.path.insert(0,str(Path(__file__).parents[1]))` inside the `pytest examples/tests/test_examples.py` so that it can find the `default_argument_parser` while testing the examples.

examples/argument_parser.py
```diff
@@ -1,3 +1,22 @@
+# Copyright (C) 2022. Huawei Technologies Co., Ltd. All rights reserved.
```
Adaickalavan, line 1: Do we need headers for examples folder? Currently headers in examples are not checked by our CI.

Adaickalavan (approved):

examples/argument_parser.py
saulfield: I had to add it when I moved this into the core. I'll remove it.

Gamenot:
> The argument parser is only for running the examples and does not seem related to smarts core. Hence, I am hesitant about moving the argument parser into smarts core.
> 
> One approach could be the following:
> 
> 1. Change the import statement from `from examples.argument_parser import default_argument_parser` to `from argument_parser import default_argument_parser` in the examples.
> 2. Include `import sys; sys.path.insert(0,str(Path(__file__).parents[1]))` inside the `examples/tests/test_examples.py` so that pytest can find the `argument_parser` while testing the examples.

I believe I am getting swayed too much by arguments. I agree about `argument_parser` not being in the `core.utils`. It could still go in `smarts` as the standard SMARTS example interface as long as it is not in `smarts.core` (i.e. it could be in something like `smarts.examples.utils`.) 

The current approach also works.

@saulfield @Adaickalavan

Gamenot (approved):

PR #1570, opened 2022-08-18
game_of_tag

AisenGinn:
Converting the game of tag into jupyter notebook. Merging the game_of_tag.py with run_checkpoint.py into game_of_tag.ipynb.


Gamenot:
@AisenGinn Please ensure that the `test-types succeeds.

Gamenot:
I am unsure if the checkpoints should be included in this PR. I feel like the trained model should not be included in the repository since it takes up so much space. This means, though, that we would need to remove the model from history rather than deleting it in a commit.

examples/game_of_tag/game_of_tag.py
```diff
@@ -0,0 +1,282 @@
+"""Let's play tag!
```
Gamenot, line 1: I do not think it is necessary to have the python example anymore if we already have the working notebook version.

examples/game_of_tag/scenarios/game_of_tag_demo_map/scenario.py
```diff
@@ -0,0 +1,42 @@
+import random
+from pathlib import Path
+
+from smarts.sstudio import gen_scenario
+from smarts.sstudio import types as t
+from smarts.core import seed
+
+seed(42)
+
+# traffic = t.Traffic(
```
Gamenot, line 10: There is a question as to if we should re-enable the traffic in this scenario.

examples/game_of_tag/README.md
```diff
@@ -0,0 +1,49 @@
...
+    Collision with predator: -10
+    Distance to predator(d): 0.5/(d-COLLIDE_DISTANCE)^2 
+
+### Action:
+Speed selection in m/s: [0, 3, 6, 9]
+
+Lane change selection relative to current lane: [-1, 0, 1]
+
+## Output a model:
+Currently Rllib does not have implementation for exporting a pytorch model.
```
Gamenot, line 33: Please check to see if this is still true.

examples/game_of_tag/tag_adapters.py
```diff
@@ -0,0 +1,230 @@
...
+    rew += dominant_reward(distance_to_target)
+
+    events = observations.events
+    for c in observations.events.collisions:
+        if _is_vehicle_wanted(c.collidee_id, PREY_IDS):
+            rew += global_rewards.collesion_with_target
+            print(
+                f"predator {ego.id} collided with prey {c.collidee_id} distance {distance_to_target}"
+            )
+        # # keeping this commented code for expanding to mutiple preys and predators in the future
```
Gamenot, line 173: I think we can pass by with 1 prey and 1 predator for now but we would want to do a follow-up that implements multiple prey and predators. Please make an issue for that improvement.

Gamenot (commented):
There is a bit of work to do but what is already here is awesome work. :)

PR #1574, opened 2022-08-19
Ensure ego observation road, lane, lane index fields are filled.

Gamenot:
This addresses an issue where there are values in the ego observation that are not optional but return `None`.


saulfield (approved):

Adaickalavan (approved):

Adaickalavan:
While running a Waymo scenario, the following error popped up. Error message is truncated for brevity. 
```text
  File "/home/adai/workspace/SMARTS/competition/evaluation/.venv/lib/python3.8/site-packages/smarts/env/wrappers/format_obs.py", line 291, in observation
    val = func(getattr(agent_obs, self._stdob_to_ob[stdob]))
  File "/home/adai/workspace/SMARTS/competition/evaluation/.venv/lib/python3.8/site-packages/smarts/env/wrappers/format_obs.py", line 568, in _std_neighbors
    lane_index = np.array(lane_index, dtype=np.int8)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
Exception ignored in: <function SMARTS.__del__ at 0x7f960ffb7c10>
```

The `VehicleObservation.lane_index` in `obs.neighborhood_vehicle_states` appears to be `None`. We have to make the same fix to the neighbours, as was made to the ego, to ensure neighbour vehicles don't have a `lane_index=None`.

PR #1576, opened 2022-08-20
Bugfix `on_shoulder` and `offroad` events in junctions

Gamenot:
The check could not find the junction lanes so we check for a road using the standard lanes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - Renamed `examples/observation_collection_for_imitation_learning.py` to `examples/traffic_histories_to_observations.py`.
 - Renamed `examples/history_vehicles_replacement_for_imitation_learning.py` to `examples/traffic_histories_vehicle_replacement.py`.
 - `SumoTrafficSimulation` will now try to hand-off the vehicles it controls to the new SMARTS background traffic provider by default if the Sumo provider crashes.
+- SMARTS now gives a error about a suspected lack of junction edges in sumo maps on loading of them.
 
 ### Removed
 - Removed support for deprecated json-based and YAML formats for traffic histories.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - Renamed `examples/observation_collection_for_imitation_learning.py` to `examples/traffic_histories_to_observations.py`.
 - Renamed `examples/history_vehicles_replacement_for_imitation_learning.py` to `examples/traffic_histories_vehicle_replacement.py`.
 - `SumoTrafficSimulation` will now try to hand-off the vehicles it controls to the new SMARTS background traffic provider by default if the Sumo provider crashes.
+- SMARTS now gives a error about a suspected lack of junction edges in sumo maps on loading of them.
```
saulfield, line 30:
```suggestion
- SMARTS now gives an error about a suspected lack of junction edges in sumo maps on loading of them.
```

saulfield (approved):

smarts/core/sumo_road_network.py
```diff
@@ -796,6 +824,8 @@ def road_with_point(self, point: Point) -> RoadMap.Road:
         for nl, dist in self.nearest_lanes(point, radius):
             if dist < 0.5 * nl._width + 1e-1:
                 return nl.road
+        else:
+            print(f"Point is off at {point}")
```
Adaickalavan, line 828: Is it possible for the `self.nearest_lanes()` to ever return an empty list?

Adaickalavan (approved):

smarts/core/sumo_road_network.py
Gamenot: This keeps sneaking back in, I thought I removed this debugging.

Gamenot: Yes, it is possible for nearest lanes to be empty if a vehicle strays far enough from the road.

In this case the `else` was entirely unnecessary because there is a `return` rather than a `break`...

PR #1608, opened 2022-08-25
Address division issues in local traffic provider.

Gamenot:
Addresses issues with division by 0 in the local traffic provider.

### DIFF
--- a/smarts/core/local_traffic_provider.py
+++ b/smarts/core/local_traffic_provider.py
@@ -28,7 +28,7 @@ from bisect import bisect_left, bisect_right, insort
 from collections import defaultdict, deque
 from dataclasses import dataclass
 from functools import lru_cache
-from typing import Any, Deque, Dict, List, Optional, Sequence, Set, Tuple
+from typing import Any, Deque, Dict, List, Optional, Set, Tuple
 
 import numpy as np
 from cached_property import cached_property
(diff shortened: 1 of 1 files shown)


smarts/core/local_traffic_provider.py
```diff
@@ -1238,13 +1270,15 @@ def predict_crash_in(self, veh_id: str, bumper: int) -> float:
                 prev_range = rvi.dist
                 prev_bearing = rvi.bearing
                 prev_heading = rvi.heading
-            range_del /= len(window) - 1
-            bearing_del /= len(window) - 1
+            range_del /= len(window) - 1 or 1
+            bearing_del /= len(window) - 1 or 1
```
Adaickalavan, line 1274: Just checking that the desired outcome when `len(window)=1`, is indeed
```python
range_del = range_del / ( len(window) -1 or 1)
          = range_del                
```

Adaickalavan (approved):

smarts/core/local_traffic_provider.py
Gamenot: Thanks for bringing this up, there is no defined behaviour previously except for a crash. Perhaps it is better to handle the problem explicitly as you mention.

Gamenot: Looking at it again I should remove this condition. L1257 covers this case, I believe I should revert this condition.

Gamenot, line 1274:
```suggestion
            range_del /= len(window) - 1
            bearing_del /= len(window) - 1
```

PR #1613, opened 2022-08-29
Loosen dependencies

Gamenot:
- Loosen `cloudpickle` dependency

### DIFF
--- a/examples/env/create_run_visualize.ipynb
+++ b/examples/env/create_run_visualize.ipynb
@@ -22,7 +22,7 @@
    "outputs": [],
    "source": [
     "!git clone https://github.com/huawei-noah/SMARTS 2> /dev/null\n",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -41,7 +41,7 @@
         # The following are for Scenario Studio
         "yattag>=1.14.0",
         # The following is for both SS and Envision
-        "cloudpickle>=1.3.0,<1.4.0",
+        "cloudpickle>=1.3.0,<=2.1.0",
```
Adaickalavan, line 44: What was the reason for constraining cloudpickle to between 1.3.0 and 2.1.0?

Adaickalavan (approved):

setup.py
Gamenot: I have no idea, I tried it out and had no issues (there may have been a bug in a version.) But considering there were no issues it looks like it was just constrained that way.

PR #1620, opened 2022-09-06
Set initial camera position when CameraSensor is created

### LINKED ISSUE #1619: waymo dataset and dummy dataset has no background information for the first several time steps
The waymo dataset and dummy dataset has no background information for the first several time steps. We can only see ego-vehicle in those images. 
C4CD04BF-53A4-48CA-90E8-EE6BE92AC108

saulfield:
Fixes #1619.

In the `traffic_histories_to_observations.py` example, the first RGB image was not rendered correctly because the observations are taken before the camera sensor calls `step()` and moves the camera to the correct position.

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -743,6 +743,7 @@ class CameraSensor(Sensor):
             height,
             resolution,
         )
+        self._follow_vehicle()  # ensure we have a correct initial camera position
 
     def teardown(self):
         self._camera.teardown()


Gamenot (approved):

Gamenot:
This seems wrong in general if the camera position is not being updated with the current target location. It sounds like it would be a step behind its target.

saulfield:
> This seems wrong in general if the camera position is not being updated with the current target location. It sounds like it would be a step behind its target.

I believe it's only an issue in this example script (and only on the first step) because of how observations are collected independent of `smarts.step()`.

Adaickalavan (approved):

Issue #1597, opened 2022-08-23
Envision: Scenario Studio Preview


Gamenot:
Issue:
It is difficult to debug scenarios in development.

Request:
Add an `scl scenario preview` command. It should build and run the scenario for viewing in envision.

PR #1621, opened 2022-09-06
Add `scl scenario preview` command to preview the scenario for better debugging experience

qianyi-sun:
resolve issue https://github.com/huawei-noah/SMARTS/issues/1597

### DIFF
--- a/cli/studio.py
+++ b/cli/studio.py
@@ -205,15 +205,38 @@ def replay(directory: Sequence[str], timestep: float, endpoint: str):
     help="Port on which Envision will run.",
     default=None,
 )
+@click.option(
+    "--clean",
+    is_flag=False,
+    default=True,
+    help="Clean previously generated artifacts first",
+)
+@click.option(
+    "--allow-offset-map",
+    is_flag=False,
+    default=False,
+    help="Allows road network to be offset from the origin. If not specified, creates a new network file if necessary.",
+)
+@click.option(
+    "--seed",
+    type=int,
+    default=42,
+    help="Set the base seed of the scenario.",
+)
 @click.argument("scenario", type=click.Path(exists=True), metavar="<scenario>")
-def preview_scenarios(envision, envision_port, scenario):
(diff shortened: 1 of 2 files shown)


cli/studio.py
```diff
@@ -185,7 +190,60 @@ def replay(directory: Sequence[str], timestep: float, endpoint: str):
...
+            )
+            # Just in case: give Envision a bit of time to warm up
+            time.sleep(2)
+            url = "http://localhost:" + str(envision_port)
+            webbrowser.open_new_tab(url)
+
+        if (not envision) and envision_port:
+            click.echo(
+                "Port passed without starting up the envision server. Use the --envision option to start the server along with the --envision port option."
+            )
```
Gamenot, line 238: I have a thought that this might be overthinking that having these two together is an issue. Assuming envision refreshes properly (in the future) it should be entirely reasonable to have this particular case.

smarts/env/run_env.py
```diff
@@ -0,0 +1,59 @@
...
+        scenarios=scenarios,
+        agent_specs={},
+        headless=headless,
+        sumo_headless=True,
+    )
+
+    if max_episode_steps is None:
+        max_episode_steps = 1000
+
+    for episode in episodes(n=num_episodes):
```
Gamenot, line 22: I have some thought here that we should keep this particular scenario looping until the kill key is pressed (whatever key that is.)

smarts/env/run_env.py
```diff
@@ -0,0 +1,59 @@
...
+    env = gym.make(
+        "smarts.env:hiway-v0",
+        scenarios=scenarios,
+        agent_specs={},
+        headless=headless,
+        sumo_headless=True,
+    )
+
+    if max_episode_steps is None:
+        max_episode_steps = 1000
```
Gamenot, line 20: I think this might need some configuration (although ideally I think this max step information should also be inherent in the scenario.)

PR #1622, opened 2022-09-07
Limit TargetPose action range

Adaickalavan:
+ This PR prevents the unconstrained motion range of the agent while using TargetPose action space.
+ An env wrapper is introduced over `multi-scenario-v0` env to cap the delta-x and delta-y, such that the next xy coordinate in TargetPose action is range constrained and the agent speed is constrained to a maximum speed of 100.8km/h.
+ The env wrapper also fixes the `time-delta` in TargetPose action space to 0.1.
+ In the long run, these physical constraints on vehicle motion should preferably be performed inside SMARTS TargetPose action and not at the environment wrapper level.


smarts/env/multi_scenario_env.py
```diff
@@ -282,6 +290,101 @@ def resolve_agent_interface(
...
+        )
+        speed_max = 28  # 28m/s = 100.8 km/h. Maximum speed should be >0.
+        dist_max = speed_max * time_delta
+
+        # Set time-delta
+        if not math.isclose(action[3], time_delta, abs_tol=1e-3):
+            logger.warning(
+                f"{name}: Expected time-delta={time_delta}, but got time-delta={action[3]}. "
+                f"Action time-delta automatically changed to {time_delta}."
+            )
```
Gamenot, line 370: We can skip string operations when logging level is higher than warning by using old c format substitution.

```suggestion
            logger.warning(
                "%s: Expected time-delta=%s, but got time-delta=%s. "
                "Action time-delta automatically changed to %s.",
                name,
                time_delta,
                action[3],
                time_delta 
            )
```

smarts/env/multi_scenario_env.py
```diff
@@ -282,6 +290,101 @@ def resolve_agent_interface(
...
+        vector = next_coord - prev_coord
+        dist = np.linalg.norm(vector)
+        if dist > dist_max:
+            unit_vector = vector / dist
+            limited_action[0], limited_action[1] = prev_coord + dist_max * unit_vector
+            logger.warning(
+                f"{name}: Allowed max speed={speed_max}, but got speed={dist/time_delta}. "
+                f"Next x-coordinate and y-coordinate automatically changed from {next_coord} "
+                f"to {limited_action[:2]}."
+            )
```
Gamenot, line 383: Same here as above.
```suggestion
            logger.warning(
                "%s: Allowed max speed=%s, but got speed=%s. "
                "Next x-coordinate and y-coordinate automatically changed from %s "
                "to %s.",
                name,
                speed_max,
                dist/time_delta,
                next_coord,
                limited_action[:2]
            )
```

Gamenot (approved):
Is there any compelling reason if we should apply acceleration caps for linear acceleration and angular velocity? 

I think there is a case we should limit it for some amount over what is desired in the metrics.

Regardless this does not block the application of velocity limits.

saulfield (approved):

PR #1623, opened 2022-09-07
updated CQL training for dummy dataset

alfredgu001324:


competition/track2/train/train.py
```diff
@@ -88,7 +88,7 @@ def train(input_path, output_path):
             print(f"Adding data for vehicle id {id} in scenario {scenario}.")
 
             with open(
-                scenario_path / (f"Agent-history-vehicle-{id}.pkl"),
+                scenario_path / (f"sumo_sumo_1_Agent-history-vehicle-{id}.pkl"),
```
Adaickalavan, line 91: If you hard code "sumo_sumo_*" string, what happens when a differently named hidden dataset is used during docker-based re-training?

competition/track2/train/train.py
```diff
@@ -87,36 +87,30 @@ def train(input_path, output_path):
...
-
-            goal_pos_x = vehicle_data[float(image_names[-1].split("_Agent")[0])]["ego"][
-                "pos"
-            ][0]
-            goal_pos_y = vehicle_data[float(image_names[-1].split("_Agent")[0])]["ego"][
-                "pos"
-            ][1]
+            
+            goal_pos_x = vehicle_data[float(image_names[-1].split("_Agent")[0])].ego_vehicle_state.mission.goal.position.x
+            goal_pos_y = vehicle_data[float(image_names[-1].split("_Agent")[0])].ego_vehicle_state.mission.goal.position.y
```
Adaickalavan, line 102: @cz0204 @alfredgu001324 
1) Try to avoid using "Agent" to get the timestamp.
2) Try to only use the file naming format, shown below, as the only information available in order to extract any info.
    image
3) With respect to
    ```python
    vehicle_data[float(image_names[-1].split("_Agent")[0])].ego_vehicle_state.mission.goal.position.x
    ```
    + Why did we use the data from the last time point to get the `mission.goal.position` value?
    + I believe that the `mission.goal.position` value is the same at all time points for a given vehicle.
    + Could you check on this matter and simplify the code if necessary?

competition/track2/train/train.py
```diff
@@ -87,36 +87,30 @@ def train(input_path, output_path):
...
             threshold = 3
 
             for i in range(len(image_names) - 1):
                 with Image.open(scenario_path / image_names[i], "r") as image:
                     image.seek(0)
                     sim_time = image_names[i].split("_Agent")[0]
                     sim_time_next = image_names[i + 1].split("_Agent")[0]
```
Adaickalavan, line 109: 1) Try to avoid using "Agent" to get the timestamp.
2) Try to only use the file naming format, shown below, as the only information available in order to extract any info.
    image

competition/track2/train/train.py
alfredgu001324: 1. fixed
2. fixed
3. The mission.goal.position is indeed the same at all times, so picking any arbitrary time would be fine. I changed it to the initial starting time but it doesn't really matter that much.

competition/track2/train/train.py
alfredgu001324: 1. fixed
2. fixed

competition/track2/train/train.py
```diff
@@ -129,8 +122,12 @@ def train(input_path, output_path):
...
-                    if all(value == 0 for value in events.values()):
+                    events = vehicle_data[float(sim_time)].events
+                    attributes = [a for a in dir(events) if not a.startswith('__')]
+                    events_values = []
+                    for a in attributes:
+                        events_values.append(getattr(events, a))
+                    if all(value == 0 for value in events_values):
                         terminal = 0
                     else:
                         terminal = 1
```
Adaickalavan, line 133: 1) Is this code correctly doing what it is expected to do?
2) Consider simplifying this code.

Adaickalavan, line 133: @alfredgu001324 @cz0204 
1) Is this code correctly doing what it is expected to do?
2) Consider simplifying this code.

competition/track2/train/train.py
```diff
@@ -87,36 +86,30 @@ def train(input_path, output_path):
...
-
-            goal_pos_x = vehicle_data[float(image_names[-1].split("_Agent")[0])]["ego"][
-                "pos"
-            ][0]
-            goal_pos_y = vehicle_data[float(image_names[-1].split("_Agent")[0])]["ego"][
-                "pos"
-            ][1]
+            
+            goal_pos_x = vehicle_data[float(image_names[0].split("_")[0])].ego_vehicle_state.mission.goal.position.x
+            goal_pos_y = vehicle_data[float(image_names[0].split("_")[0])].ego_vehicle_state.mission.goal.position.y
```
Adaickalavan, line 101:
```suggestion
            goal_pos_x, goal_pos_y = vehicle_data[float(image_names[0].split("_")[0])].ego_vehicle_state.mission.goal.position.as_np_array[0:2]
```

competition/track2/train/train.py
```diff
@@ -87,36 +86,30 @@ def train(input_path, output_path):
...
 
-            with open(
-                scenario_path / (f"Agent-history-vehicle-{id}.pkl"),
-                "rb",
-            ) as f:
-                vehicle_data = pickle.load(f)
+            for pkl_file in os.listdir(scenario_path):
+                if pkl_file == (id + '.pkl'):
+                    with open(scenario_path / pkl_file, "rb") as f:
+                        vehicle_data = pickle.load(f)
```
Adaickalavan, line 92:
```suggestion
            with open(scenario_path / f"{id}.pkl", "rb") as f:
                vehicle_data = pickle.load(f)
```

competition/track2/train/train.py
alfredgu001324: 1. It is iterating through all the object's attributes and checking their values, since the current data format is no longer a dictionary so cannot check for dict.values()
2. Not too sure how to simplify this

Adaickalavan: @alfredgu001324 @cz0204 
1. Can you check and confirm whether the code is correctly performing its intended functionality?
2. A potential simplification is as follows:
```suggestion
                    events = vehicle_data[float(sim_time)].events
                    terminal = 0
                    for name, value in events._asdict().items():
                        if bool(value):
                            terminal = 1
                            break
```

Adaickalavan: @alfredgu001324 @cz0204 
1. Can you check and confirm whether the code is correctly performing its intended functionality?
2. When should the `terminal=1` and when should `terminal=0`?
3. Must there be at least one instance of `terminal=1` in `terminals` for the CQL training to work?
4. A potential simplification is as follows:
```suggestion
                    events = vehicle_data[float(sim_time)].events
                    terminal = 0
                    for name, value in events._asdict().items():
                        if bool(value):
                            terminal = 1
                            break
```

alfredgu001324: 1. Yes
2 & 3. I am not too sure. Maybe @cz0204  will know this better
3. Does event object has _asdict()?

cz0204: 2. if anyone of the events is True, then terminal is 1, otherwise it is 0.
3. yes, terminal has to contain at least 1 in offline training to work. Otherwise we have to setup another feature in CQL, which is not used here in our dataset. So this feature is not included in current code.

Adaickalavan: @cz0204 @alfredgu001324 

1. The `events` contains `collision` which is a list. The equality check `all(value == 0 ... )` will always evaluate to `False` because comparing a list to an int gives false.
2. After making necessary corrections, could you rectify this CQL code such that it can be trained using the [example dataset](https://github.com/smarts-project/smarts-project.offline-datasets/blob/master/offline_dataset.rar)?

alfredgu001324: @cz0204 @Adaickalavan 
3. It seems that the current dummy data set only contains terminal 0 (total of 143 pieces of data), and thus reporting an 'index out of range' error

Adaickalavan:
The `comp-1` branch has been updated with the CQL fix.

PR #1626, opened 2022-09-13
Fix `macos` test (naive)

Gamenot:
- According to https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#language-and-runtime the version of python on the `macos-11` runner is `python3.10`. 
- Actions allows for setup of python in a different way as indicated here: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python

### DIFF
--- a/.github/workflows/ci-auto-commit-mac.yml
+++ b/.github/workflows/ci-auto-commit-mac.yml
@@ -17,11 +17,15 @@ jobs:
         uses: actions/checkout@v2   
         with:
           ref: develop 
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
       - name: Setup Python
         run: |
           brew update
-          brew install python@3.7
-          brew unlink python@3.9
+          brew install python@3.8
+          brew unlink python@3.10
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-auto-commit-mac.yml
```diff
@@ -17,16 +17,14 @@ jobs:
...
-      - name: Setup Python
-        run: |
-          brew update
-          brew install python@3.7
-          brew unlink python@3.9
-          brew link --force --overwrite python@3.7
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
```
Adaickalavan, line 23: This is a very useful GitHub action.

Adaickalavan (approved):

.github/workflows/ci-auto-commit-mac.yml
Gamenot: Yes. What resources are available can be checked by looking at the tools available for each image such as here:

https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#python

PR #1628, opened 2022-09-13
Fix local traffic

Gamenot:
- This fixes two issues with the `local_traffic_provider`
  - Issue 1: Vehicles were stopping randomly. This was due to an optimization that was storing old caches individually in the internal traffic actors.
  - Issue 2: Newest actor was removed when an actor left the traffic provider. Similar to previous issue this had to do with using a local reference from a loop rather than recovering the agent from the id(which did no cause an error from a quirk of python scope rules.)
  - Minor issue: Vehicles that recently enter now hold lane for a short amount of time.

This should be merged into both `develop` and `comp-1` asap.

### DIFF
--- a/smarts/core/local_traffic_provider.py
+++ b/smarts/core/local_traffic_provider.py
@@ -68,7 +68,6 @@ class LocalTrafficProvider(TrafficProvider):
 
     def __init__(self):
         self._logger = logging.getLogger(self.__class__.__name__)
-        self._logger.setLevel(logging.DEBUG)
         self._sim = None
         self._scenario = None
         self.road_map: RoadMap = None
@@ -457,7 +456,6 @@ class _TrafficActor:
 
     def __init__(self, flow: Dict[str, Any], owner: LocalTrafficProvider):
         self._logger = logging.getLogger(self.__class__.__name__)
-        self._logger.setLevel(logging.DEBUG)
         self._owner = weakref.ref(owner)
 
         self._state = None
--- a/smarts/core/local_traffic_provider.py
+++ b/smarts/core/local_traffic_provider.py
@@ -68,6 +68,7 @@ class LocalTrafficProvider(TrafficProvider):
 
     def __init__(self):
         self._logger = logging.getLogger(self.__class__.__name__)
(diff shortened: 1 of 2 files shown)


smarts/scenarios/straight/3lane_cut_in/scenario.py
```diff
@@ -76,7 +77,7 @@
                     end=("gneE3", end_lane, "max"),
                 ),
                 # Random flow rate, between x and y vehicles per minute.
-                rate=60 * random.uniform(15, 25),
+                rate=40* random.uniform(10, 20),
```
Adaickalavan, line 80:
```suggestion
                rate=60* random.uniform(6, 14),
```
`rate=60 * random.uniform(15, 25),` means the number of vehicles emitted per minuted is between 15 to 25. Changing the first number to `40` makes the command harder to comprehend. If needed, consider reducing the number range inside the `uniform(x,y)` while keeping the first number at `60`.

smarts/core/local_traffic_provider.py
```diff
@@ -275,23 +278,26 @@ def step(self, actions, dt: float, elapsed_sim_time: float) -> ProviderState:
...
                 # so we need to give teleporting vehicles a new id and thus a new chassis.
                 actor.bump_id()
                 remap_ids[actor_id] = actor.actor_id
-        for actor in losts:
+        for actor in losts - removed:
+            removed.add(actor.actor_id)
             self._relinquish_actor(actor.state)
-        for actor_id in dones:
+        for actor_id in dones - removed:
+            actor = self._my_actors.get(actor_id)
```
Adaickalavan, line 300: Fix for pytype.
```suggestion
            actor = self._my_actors[actor_id]
```

Adaickalavan (approved):
Need to fix the pytype errors before merging.

saulfield (approved):

qianyi-sun:
> Need to fix the pytype errors before merging.

@Adaickalavan After this gets merged into `develop`, please update `comp-1` accordingly, thanks!

PR #1630, opened 2022-09-14
Defer only to right of current lane

Gamenot:
- This increases the default lock-in time for holding lane after cut-in
- Also actors will only now yield to right to behind vehicles.

### DIFF
--- a/smarts/core/local_traffic_provider.py
+++ b/smarts/core/local_traffic_provider.py
@@ -1108,20 +1108,13 @@ class _TrafficActor:
             checks = []
 
         ## TODO: Determine how blocked lane changes should be addressed
-        # cut_in_is_real_lane = self._cutting_into and self._cutting_into.index < len(
-        #     self._lane_windows
-        # )
         ## Idea is to keep lane if blocked on right, slow down if blocked on left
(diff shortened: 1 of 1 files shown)


smarts/core/local_traffic_provider.py
```diff
@@ -1095,23 +1095,22 @@ def _pick_lane(self, dt: float):
...
         # hold lane for some time if added recently
         if self._forward_after_added < self._after_added_hold_secs:
             self._forward_after_added += dt
             # skip checks
             checks = []
 
         ## TODO: Determine how blocked lane changes should be addressed
+        # cut_in_is_real_lane = self._cutting_into and self._cutting_into.index < len(
+        #     self._lane_windows
+        # )
```
saulfield, line 1113: Just want to confirm this wasn't left commented unintentionally.

saulfield (approved):

smarts/core/local_traffic_provider.py
Gamenot: It is commented out intentionally, this was a consideration of how to address blocking when lanes to right and left are unavailable for periods of time.

Gamenot: I should convert this to pseudo-code rather than keeping it in as code.

Issue #1631, opened 2022-09-15
2022 NeurIPS Driving SMARTS Competition - track 2 - Problem while generating SMARTS dataset from the Waymo Motion Open dataset


tunglm2203:
**High Level Description**
2022 NeurIPS Driving SMARTS Competition - track 2 - Problem while generating SMARTS dataset from the Waymo Motion Open dataset

**Desired SMARTS version**
0.6.1

**Operating System**
ubuntu 18.04 (docker)

**Problems**
As mentioned in _https://github.com/huawei-noah/SMARTS/issues/1577#issuecomment-1247482020_, while generating the data from Waymo Motion Open, I faced some errors. I was trying to narrow the problem in the file `validation.tfrecord-00000-of-00150` (it's also happened in other files), and found it happened in the scenario IDs:

- 43f06b0743dc9355
- 45c6cd9309b3ce87

The error from scenario `43f06b0743dc9355`: 
```
Waiting on offline_dataset/43f06b0743dc9355/. ...
WARNING:WaymoMap:ignoring 'entry_lane' feature=147 for feature=297 as it only has a single point in its polyline.
WARNING:WaymoMap:ignoring 'entry_lane' feature=165 for feature=297 as it only has a single point in its polyline.
Process Process-1:
Traceback (most recent call last):
  File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/root/workspace/rlbench/neurips22_comp/SMARTS/cli/studio.py", line 79, in _build_single_scenario_proc
    build_single_scenario(clean, allow_offset_map, scenario, seed, click.echo)
  File "/root/workspace/rlbench/neurips22_comp/SMARTS/smarts/sstudio/build_scenario.py", line 66, in build_single_scenario
    road_map, _ = map_spec.builder_fn(map_spec)
  File "/root/workspace/rlbench/neurips22_comp/SMARTS/smarts/core/default_map_builder.py", line 148, in get_road_map
    road_map = map_class.from_spec(map_spec)
  File "/root/workspace/rlbench/neurips22_comp/SMARTS/smarts/core/waymo_map.py", line 933, in from_spec
    return cls(map_spec, waymo_scenario)
  File "/root/workspace/rlbench/neurips22_comp/SMARTS/smarts/core/waymo_map.py", line 124, in __init__
[...]

PR #1632, opened 2022-09-15
Waymo - bugfixes and improved error messages

saulfield:
Addresses #1631

- Fix bug related to glb generation in Waymo maps (the `assert lane_to_left` error)
- Introduce custom `WaymoDatasetError` exception which will be raised when trying to load a Waymo scenario that SMARTS is not able to handle. Currently there are 2 possible reasons this could be raised: either the recorded timesteps deviate by more that 1 timestep (which is not supported by the traffic history provider), or the map contains lanes with single-point polylines (which is not supported by the Waymo map class). Any other exception raised related to Waymo should be considered a bug.

This is intended to make it more explicit when a Waymo scenario should be ignored by detecting these issues when building the scenario (rather than waiting for them to pop up during runtime). However, it now appears these cases are more common than we previously thought -- around 30% of Waymo scenarios have one of these issues.

### DIFF
--- a/smarts/waymo/waymo_utils.py
+++ b/smarts/waymo/waymo_utils.py
@@ -24,9 +24,12 @@ import numpy as np
 from matplotlib import pyplot as plt
 from matplotlib.animation import FuncAnimation
 from matplotlib.lines import Line2D
+from smarts.core.utils.file import read_tfrecord_file
+
+# pytype: disable=import-error
 from waymo_open_dataset.protos import scenario_pb2
 
-from smarts.core.utils.file import read_tfrecord_file
+# pytype: enable=import-error
 
 
 class WaymoDatasetError(Exception):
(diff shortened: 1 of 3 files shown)


Adaickalavan (approved):

smarts/waymo/waymo_utils.py
```diff
@@ -24,10 +24,14 @@
 from matplotlib import pyplot as plt
 from matplotlib.animation import FuncAnimation
 from matplotlib.lines import Line2D
-from waymo_open_dataset.protos import scenario_pb2
-
 from smarts.core.utils.file import read_tfrecord_file
-from smarts.core.waymo_map import WaymoMap
```
Gamenot, line 30: Right, this removal should help with the bleeding import errors related to waymo.

Gamenot (approved):

PR #1633, opened 2022-09-16
Patch orphaned social agents

Gamenot:
- Bubble agents occasionally get orphaned.
- This change-set converts the assertion to an error log and automatically cleans up orphaned social agents.

### DIFF
--- a/smarts/core/agents_provider.py
+++ b/smarts/core/agents_provider.py
@@ -90,7 +90,7 @@ class AgentsProvider(Provider):
     def reset(self):
         self.teardown()
 
-    def _remove_missing_actors(self, agent_actions: Dict[str, Any]):
+    def _remove_actors_without_actions(self, agent_actions: Dict[str, Any]):
         missing_agents = self._my_agent_actors.keys() - agent_actions.keys()
         for agent_id in missing_agents:
             self._log.info(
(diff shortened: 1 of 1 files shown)


smarts/core/agents_provider.py
```diff
@@ -122,19 +122,29 @@ def perform_agent_actions(self, agent_actions: Dict[str, Any]):
...
         """
-        self._remove_missing_actors(agent_actions)
-        actions_without_agents = agent_actions.keys() - self._my_agent_actors.keys()
-        assert (
-            not actions_without_agents
-        ), f"actions specified for non-tracked agents:  {actions_without_agents}"
+        self._remove_actors_without_actions(agent_actions)
+        agents_without_actors = agent_actions.keys() - self._my_agent_actors.keys()
+        if agents_without_actors:
+            if self._log.level >= logging.ERROR:
```
saulfield, line 128: Why do we need to check the log level? Won't it be filtered out already based on whatever level is set?

qianyi-sun (approved):

smarts/core/agents_provider.py
Gamenot: Yes, thanks for pointing it out. I meant to only lock the error warning behind the logging level.

smarts/core/agents_provider.py
```diff
@@ -122,19 +122,29 @@ def perform_agent_actions(self, agent_actions: Dict[str, Any]):
...
+                self._log.error(
+                    "actions specified for an agent without an actor: %s. Cleaning up social agents.",
+                    agents_without_actors,
+                )
```
Gamenot, line 132:
```suggestion
            self._log.error(
                "actions specified for an agent without an actor: %s. Cleaning up social agents.",
                agents_without_actors,
            )
```

Issue #1599, opened 2022-08-23
Envision: Fix Playback Rate (realtime/uncapped)


Gamenot:
Issue:
Envision plays in an uncontrolled manner. Realtime used to allow for envision to be played in a realtime mode. This would play as if it is 1 second in simulation time corresponds to 1 second in real time.

Request:
Fix this feature!

PR #1651, opened 2022-09-30
Envision fixes

### LINKED ISSUE #1598: Envision: Fix Pause and Rewind
Issue:
There is a scrubbing bar at the bottom of the simulation that should allow for envision to go back in time. This used to work.

Request:
Fix this feature. Clicking on a location should send the simulation back in time.

saulfield:
Fixes #1598 and #1599.

- Fix issue where frame time was not being passed properly from the envision server to the web client
- Fix issue in web client where `prevElapsedTime` was always null, preventing realtime mode from ever being enabled
- Switch playback bar callback to use `onChange` instead of `onAfterChange`, which is not functional (see open issue here: https://github.com/ant-design/ant-design/issues/14119)
- Change server loop to be non-blocking, to enable seeking and playback after episode end

### DIFF
--- a/envision/server.py
+++ b/envision/server.py
@@ -317,10 +317,9 @@ class BroadcastWebSocket(tornado.websocket.WebSocketHandler):
         frame_time = None
         # Find the first number value, which will be the frame time.
         for prefix, event, value in it:
-            if not prefix or event != "number":
-                continue
-            frame_time = float(value)
-            break
+            if prefix and event == "number":
+                frame_time = float(value)
+                break
         assert isinstance(frame_time, float)
         self._frames.append(Frame(timestamp=frame_time, data=message))
 
--- a/envision/server.py
+++ b/envision/server.py
@@ -223,28 +223,24 @@ class WebClientRunLoop:
             self._log.debug("First frame ready.")
 
             while True:
-                # Attempt to get new frames
+                # Handle seek
                 if self._seek is not None and self._frames.start_time is not None:
-                    # Handle seek
                     frame_ptr = self._frames(self._frames.start_time + self._seek)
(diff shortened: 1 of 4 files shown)


envision/server.py
```diff
@@ -318,6 +320,7 @@ async def on_message(self, message):
             if not prefix or event != "number":
                 continue
             frame_time = float(value)
+            break
```
Gamenot, line 323: I remember doing this. This might be a simplified logic.
```suggestion
            if prefix and event == "number":
                frame_time = float(value)
                break
```

envision/server.py
```diff
@@ -231,14 +234,24 @@ async def run_loop():
...
+                    # If not seeking, try to get next frames normally
+                    delay = self._calculate_frame_delay(frame_ptr)
+                    time.sleep(delay)
+                    frames_to_send = []
+                    while (
+                        frame_ptr.next_
+                        and len(frames_to_send) <= self._message_frame_volume
+                    ):
+                        frame_ptr = frame_ptr.next_
+                        frames_to_send.append(frame_ptr)
```
Gamenot, line 247: Is there a reason why this should not always happen? (assuming `frames_to_send` is instead reset at the beginning of every loop)

Gamenot (commented):
Some comments.

envision/server.py
saulfield: I refactored the loop to take out this if/else branch. I also moved the sleep to happen if there are no new frames. This seems to work well from testing (I also experimented with removing the sleep completely, which appeared to break some things).

Gamenot (approved):

Adaickalavan (approved):

PR #1661, opened 2022-10-05
Fix signal observations

### LINKED ISSUE #1650: Outputs for the Signal state obserbation is always empty list
## **BUG REPORT**
**High Level Description**
The  ```obs.signals``` is always empty list ```[]```, even in scenarios that being tested for having signals
**SMARTS version**
[e.g. 0.4.12]

When I tried to add signals=True in the agent_interface in ```traffic_history_to_observation.py``` , there is always []
The scenario_id: ```a7ea3da73ebb0ac7```, which is a scenario in ```waymo_scneario_20s/00000.tfrecords``` And we have tested that there is obviously signal information in this scneario.

All step returns ```[]``` for obs.signals and reported Warnings:
'''WARNING:SignalsSensor:could not find signal state corresponding with feature_id={signal.feature_id}'''

saulfield:
Fixes #1650

- Fix an issue where signals would not be checked for empty routes (which affected traffic history vehicles, which do not have proper routes)
- Change the signal lookahead distance to decrease on the recursive calls to outgoing lanes rather than iteratively when looping over the outgoing lanes

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -1323,13 +1323,14 @@ class SignalsSensor(Sensor):
         if lookahead <= 0:
             return
         for ogl in lane.outgoing_lanes:
-            if route and ogl.road not in route.roads:
+            if route and route.road_length > 0 and ogl.road not in route.roads:
                 continue
             upcoming_signals += [
                 feat for feat in ogl.features if self._is_signal_type(feat)
             ]
-            lookahead -= lane.length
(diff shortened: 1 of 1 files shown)


Adaickalavan (approved):

Gamenot (approved):

Gamenot:
That is a funny kind of bug. Please add a changlog entry.

PR #1669, opened 2022-10-13
Pat/simulation benchmark

qianyi-sun:


cli/benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+        [
+            "scl",
+            "scenario",
+            "build-all",
+            "--clean",
+            "smarts/benchmark/"
+        ]
+    )
+
+    all_scenarios = glob('./scenarios/benchmark/*/*/', recursive=True)
```
Gamenot, line 61: This needs to target the module location of `smarts/benchmark`. You get that further up using `f"{smarts.__path__[0]}/benchmark"`. 

I could instead recommend adding a `smarts/benchmark/__init__.py` and then `import smarts.benchmark` and then you can do `smarts.benchmark.__path__[0]` instead.

cli/benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+    benchmark.main(scenarios=[f"{smarts.__path__[0]}/benchmark/{scenario}"], headless=True, num_episodes=10, max_episode_steps=None)
+
+def build_all_scenarios_and_run(scenario):
+    subprocess.run(
+        [
+            "scl",
+            "scenario",
+            "build-all",
+            "--clean",
+            "smarts/benchmark/"
```
Gamenot, line 57: You will want to pass in `smarts.benchmark.__path__[0]` as default here. If a configuration is given it can defer to that.

cli/benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+@click.group(
+    name="benchmark",
+    help="Utilities for benchmarking the simulation performance. See `scl benchmark COMMAND --help` for further options.",
+)
+
+
+def benchmark_cli():
+    pass
+
+@click.command("start", help="Start benchmarking.")
```
Gamenot, line 35: This command could be `scl benchmark run`.

cli/benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+
+    from smarts.sstudio.scenario_construction import build_single_scenario
+
+    build_single_scenario(False, True, f"{smarts.__path__[0]}/benchmark/{scenario}", 42, click.echo)
+
+    # Run the scenario
+    from smarts.benchmark import benchmark
+    benchmark.main(scenarios=[f"{smarts.__path__[0]}/benchmark/{scenario}"], headless=True, num_episodes=10, max_episode_steps=None)
+
+def build_all_scenarios_and_run(scenario):
```
Gamenot, line 50: This command could be `scl benchmark run-all`.

cli/benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+    all_scenarios = glob('./scenarios/benchmark/*/*/', recursive=True)
+    for scenario in all_scenarios:
+        subprocess.run(
+            [
+            "scl",
+            "run",
+            "examples/egoless.py",
+            f"{scenario}",
```
Gamenot, line 68: I do not remember if it needs to be stringified but this is a better way.
```suggestion
            str(scenario),
```

Gamenot (changes requested):
You will need to add do the following so that the benchmark gets packaged with smarts:

1) Update `MANIFEST.in` to include the benchmark files.
```
# <SMARTS>/MANIFEST.ini
recursive-include smarts/benchmark *.xml *.py
```
2) Ensure that `smarts/benchmark` is up to date
3) Remove `scenarios/benchmark`
4) Because the files are big, remember to squash and merge after everything is done (or edit large files out of the git history.)

Issue #1587, opened 2022-08-23
Maintenance: Untether Actors From Lanes


Gamenot:
@AisenGinn
Issue:
Agents are locked into being close to a lane. Essentially they are on a tether.

Request:
Agents should be able to move far off-road without causing the simulation to crash.

AisenGinn:
Currently the `lookahead` in `smarts.core.road_map.curevature_radius_at_offset` only accept `lookahead>=1`, but `lookahead` is calculated by `lookahead = math.ceil(1 + math.log(self._target_speed))` in `smarts.core.local_traffic_provider._slow_for_curves`. The error will occur when the `self.target_speed` is very small which make the ` math.log(self._target_speed)<-1`. This will cause `lookahead` to be negative and thus raise the assertion error. Try to fixed it in PR1673:https://github.com/huawei-noah/SMARTS/pull/1673

PR #1673, opened 2022-10-18
Fixed iss1587 smarts simulation crash

AisenGinn:
This PR attempts to fix https://github.com/huawei-noah/SMARTS/issues/1587. Currently the `lookahead` in `smarts.core.road_map.curevature_radius_at_offset` only accept `lookahead>=1`, but `lookahead` is calculated by `lookahead = math.ceil(1 + math.log(self._target_speed))` in `smarts.core.local_traffic_provider._slow_for_curves`. The error will occur when the `self.target_speed` is very small which make the ` math.log(self._target_speed)<-1`. This will cause `lookahead` to be negative and thus raise the assertion error. I added code that set `lookahead=1` when it is negative due to small `self.target_speed` which ensure the `smarts.core.road_map.curevature_radius_at_offset`  doesn't raise assertion error.

### DIFF
--- a/smarts/core/local_traffic_provider.py
+++ b/smarts/core/local_traffic_provider.py
@@ -1237,6 +1237,7 @@ class _TrafficActor:
     def _slow_for_curves(self):
         # XXX:  this may be too expensive.  if so, we'll need to precompute curvy spots for routes
         lookahead = math.ceil(1 + math.log(self._target_speed))
+        lookahead = max(1, lookahead)
         # we round the offset in an attempt to reduce the unique hits on the LRU caches...
         rounded_offset = round(self._offset)
         radius = self._lane.curvature_radius_at_offset(rounded_offset, lookahead)


smarts/core/local_traffic_provider.py
```diff
@@ -1237,6 +1237,8 @@ def _get_radius(lane: RoadMap.Lane) -> float:
     def _slow_for_curves(self):
         # XXX:  this may be too expensive.  if so, we'll need to precompute curvy spots for routes
         lookahead = math.ceil(1 + math.log(self._target_speed))
+        if lookahead < 1:
+            lookahead = 1
```
Gamenot, line 1241:
```suggestion
        lookahead = math.ceil(1 + math.log(self._target_speed))
        lookahead = max(1, lookahead)
```

scenarios/sumo/loop/scenario.py
```diff
@@ -30,7 +31,8 @@
             },
         )
         for route in [("445633931", "445633932"), ("445633932", "445633931")] * 12
-    ]
+    ],
+    engine="SMARTS"
```
Gamenot, line 35: We should not change this. Please remove the `SMARTS` traffic backing here.

examples/test_agent.py
```diff
@@ -0,0 +1,77 @@
+import logging
```
Gamenot, line 1: Please remove this file.

Gamenot (changes requested):
Some changes are needed.

Gamenot (approved):

saulfield (approved):

PR #1679, opened 2022-10-21
Fix interpolation of headings in Waymo

saulfield:


smarts/sstudio/genhistories.py
```diff
@@ -851,7 +851,7 @@ def constrain_angle(angle: float) -> float:
                 row["sim_time"] = scenario.timestamps_seconds[j]
                 row["position_x"] = obj_state.center_x
                 row["position_y"] = obj_state.center_y
-                row["heading_rad"] = obj_state.heading - math.pi / 2
+                row["heading_rad"] = (obj_state.heading - math.pi / 2) % (2 * math.pi)
```
Gamenot, line 854: A solution like this is going to act oddly in the case of heading being less than `2pi`.

```python
>>> (-2 * math.pi - 10) % (2 * math.pi)
10
```

We have a utility for this in our math utilities.

```python
import smarts.core.utils.math
math.constrain_angle(obj_state.heading)
```

Gamenot: This depends on if we want the heading to be between `[0:2pi]` or `[-pi:pi]`

saulfield: We want it to be in [0, 2pi] for the interpolation (the main change in this PR). We already convert to [-pi, pi] after the fact, and that remains unchanged.

Gamenot: OK.

Adaickalavan (approved):

Gamenot (approved):

Issue #1601, opened 2022-08-23
Envision: Vehicle Debugging Information


Gamenot:
Issue:
It is hard to diagnose the exact vehicles in the simulation that we are working with.

Request:
Mousing over a vehicle should show its information. Alternatively implement a dropdown that moves camera to focus on a specific vehicle or agent.

PR #1690, opened 2022-10-31
Envision - debugging info for vehicles and lanes

### LINKED ISSUE #1600: Envision: Road Debugging Information
Issue:
It is hard to diagnose the exact roads in the simulation that we are working with.

Request:
Mousing over a road should show its information. Alternatively, implement a dropdown that causes the camera to focus on a road or lane.

AC:
- [ ] Road id is shown
- [ ] Lane id is shown

saulfield:
Closes #1600 and #1601. Also includes changes to the vehicle model glb files to apply their scale transforms to get the correct mesh.


Gamenot:
From what I can tell, it looks like there is some rare race condition with the `test_data_replay`. We are losing record of a frame when recording or losing a frame when loading.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - Added ego's mission details into the `FormatObs` wrapper.
 - Added `SmartsLaneChangingModel` and `SmartsJunctionModel` to types available for use with the new smarts traffic engine within Scenario Studio.
 - Added option to `AgentInterface` to include traffic signals (lights) in `EgoVehicleObservation` objects.
+- Added the ability to hover over vehicles and roadmap elements in Envision to see debug info.
```
Gamenot, line 27: Please add in record of fixing the vehicle sizes.

envision/web/src/components/vehicles.js
```diff
@@ -187,31 +202,48 @@ export default function Vehicles({
...
+      let boxSize = boundingInfo.boundingBox.extendSize.scale(2);
+      let box = MeshBuilder.CreateBox(
+        `boundingbox-${meshId}`,
+        {
+          height: boxSize.y + 0.1,
+          width: boxSize.x + 0.1,
+          depth: boxSize.z + 0.1,
+        },
+        scene
+      );
```
Gamenot, line 215: I think we may have to update this in the future to show the actual bounding box of the vehicle. Right now the bounding box is calculated for the vehicle mesh.

envision/web/src/components/vehicles.js
```diff
@@ -187,31 +202,48 @@ export default function Vehicles({
...
+      boxMaterial.diffuseColor = new Color4(...SceneColors.Selection);
+      boxMaterial.specularColor = new Color3(0, 0, 0);
+      boxMaterial.alpha = 0.0;
+
+      box.material = boxMaterial;
+      box.position = boundingInfo.boundingBox.center;
+      box.actionManager = new ActionManager(scene);
+      box.actionManager.registerAction(
```
Gamenot, line 228: I can note there is a different display now for social vehicles than before. They used to show up as boxes all the time.

envision/web/src/components/vehicles.js
Gamenot: To be clear, the bounding box in sumo and SMARTS does not match the bounding box shown in envision at the moment.

This can cause the scenario to end in envision without an apparent collision.

envision/web/src/components/simulation.js
```diff
@@ -354,6 +387,18 @@ export default function Simulation({
...
+            data={debugInfo}
+            attrName="Selected Vehicle"
+          />
+        ) : null}
+        {mapElementSelected ? (
+          <DebugInfoDisplay
+            data={debugInfo}
+            attrName="Selected Map Element"
+          />
+        ) : null}
```
Gamenot, line 401: Good.

Gamenot (changes requested):
Good work, we may have to squash this PR because of the `main.js` changes.

I also slightly wonder about how the `glb`s managed to increase in size when the animation is removed. Although, it looks like the blend files have shrunk due to deleting some information.

Gamenot (approved):
`main.js` and `main.map.js` may need to be updated before you merge. This will require a rebase and then you will either want to remove the `main.js` files from history or squash.

Adaickalavan (approved):

PR #1693, opened 2022-11-01
Fixed bubble rendering issue, added ConfigurableZone

AisenGinn:
Fixed Envision bubble rendering issue, added ConfigurableZone in to bubble.


smarts/sstudio/types.py
```diff
@@ -813,6 +813,36 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
             poly = shapely_rotate(poly, self.rotation, use_radians=True)
         return shapely_translate(poly, xoff=x, yoff=y)
 
+@dataclass(frozen=True)
+class ConfigurableZone(Zone):
+    """A descripter that defines a specific configurableZone defined by user"""
+
+    ext_coordinates: List[Tuple[float, float]]
+    """external coordinates of the polygon"""
```
Gamenot, line 821: Please describe how many points are valid for what kind of shape is desired. 

- <2 points causes an error
- 2 points generates a box
- \>2 points generates a shape

smarts/sstudio/types.py
```diff
@@ -813,6 +813,36 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
             poly = shapely_rotate(poly, self.rotation, use_radians=True)
         return shapely_translate(poly, xoff=x, yoff=y)
 
+@dataclass(frozen=True)
+class ConfigurableZone(Zone):
+    """A descripter that defines a specific configurableZone defined by user"""
+
+    ext_coordinates: List[Tuple[float, float]]
```
Gamenot, line 820: Please add a `__post_init__` that can check that assert that there are a minimum of 2 points provided.

smarts/sstudio/types.py
```diff
@@ -813,6 +813,36 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
...
+        if (
+            len(self.ext_coordinates) == 2
+        ):  # if user only specified two points, create a box
+            x_min = min(self.ext_coordinates[0][0], self.ext_coordinates[1][0])
+            x_max = max(self.ext_coordinates[0][0], self.ext_coordinates[1][0])
+            y_min = min(self.ext_coordinates[1][0], self.ext_coordinates[1][1])
+            y_max = max(self.ext_coordinates[1][0], self.ext_coordinates[1][1])
+            poly = Polygon(
+                [(x_min, y_min), (x_max, y_min), (x_max, y_max), (x_min, y_max)],
+            )
```
Gamenot, line 837: Will this cause an error if you provide a situation where it describes an axis-parallel line: `p1.x == p2.x` or `p1.y == p2.y`?

The box would have a width or height of 0 then.

Gamenot (changes requested):
Please add a CHANGLOG record.

smarts/sstudio/types.py
AisenGinn: Maybe I could add the check for this in __post__init__ together with the point numbers check

smarts/sstudio/types.py
```diff
@@ -814,6 +815,51 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
         return shapely_translate(poly, xoff=x, yoff=y)
 
 
+@dataclass(frozen=True)
+class ConfigurableZone(Zone):
+    """A descripter that defines a specific configurableZone defined by user"""
```
saulfield, line 820:
```suggestion
    """A descriptor for a zone with user-defined geometry."""
```

smarts/sstudio/types.py
```diff
@@ -814,6 +815,51 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
...
+
+    ext_coordinates: List[Tuple[float, float]]
+    """external coordinates of the polygon
+    < 2 points provided: error
+    = 2 points provided: generates a box using these two points as diagonal
+    > 2 points provided: generates a polygons according to the coordinates"""
```
saulfield, line 826:
```suggestion
    > 2 points provided: generates a polygon according to the coordinates"""
```

smarts/sstudio/types.py
```diff
@@ -814,6 +815,51 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
...
+            raise ValueError(
+                "Two points or more are needed to create a polygon. (less than two points are provided)"
+            )
+
+        x_set = set(point[0] for point in self.ext_coordinates)
+        y_set = set(point[1] for point in self.ext_coordinates)
+        if len(x_set) == 1 or len(y_set) == 1:
+            raise ValueError(
+                "Parallel line cannot form a polygon. (points provided form a parallel line)"
+            )
```
saulfield, line 841: This message is a bit confusing. I think this could be incorporated into the first check above, just add a check for `len(self.ext_coordinates) < 2`.

smarts/sstudio/types.py
```diff
@@ -814,6 +815,51 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
...
+            y_min = min(self.ext_coordinates[0][1], self.ext_coordinates[1][1])
+            y_max = max(self.ext_coordinates[0][1], self.ext_coordinates[1][1])
+            poly = box(x_min, y_min, x_max, y_max)
+
+        else:  # else create a polygon according to the coordinates
+            poly = Polygon(self.ext_coordinates)
+
+        if self.rotation is not None:
+            poly = shapely_rotate(poly, self.rotation, use_radians=True)
+        return
```
saulfield, line 860:
```suggestion
        return poly
```

smarts/sstudio/types.py
```diff
@@ -890,6 +936,12 @@ def __post_init__(self):
                 "Only boids can have keep_alive enabled (for persistent boids)"
             )
 
+        poly = None
+        if not isinstance(self.zone, MapZone):
+            poly = self.zone.to_geometry(road_map=None)
+        if poly is not None and (poly.is_valid == False):
+            raise ValueError("The Zone Polygon is not a valid closed loop")
```
saulfield, line 943: What does it mean if it is `None`? Is that check needed? In any case, this could be simplified:

```python
if not isinstance(self.zone, MapZone):
    poly = self.zone.to_geometry(road_map=None)
    if poly is not None and not poly.is_valid:
        raise ValueError("The Zone Polygon is not a valid closed loop")
```

envision/web/src/helpers/state_unpacker.js
```diff
@@ -33,9 +33,9 @@ const WorldState = Object.freeze({
   SCENARIO_ID: 1,
   SCENARIO_NAME: 2,
   TRAFFIC: 3,
-  BUBBLES: 4,
-  SCORES: 5,
-  EGO_AGENT_IDS: 6,
+  BUBBLES: 5,
+  SCORES: 6,
+  EGO_AGENT_IDS: 7,
```
saulfield, line 38: Why was this needed?

saulfield (changes requested):
Looks good, just a few comments.

envision/web/src/helpers/state_unpacker.js
AisenGinn: This fixed the bubble rendering issue. There is an empty array in front of the bubbles in 4th position. This causes the bubbles, scores and ego_agent_ids arrays capture the wrong information. (The bubbles capture the empty array, scores are actually bubbles information, and ego_agent_ids are actually scores.)

AisenGinn: I am not sure what the empty array is used for. For now I did not see any usage of that array, but I am not sure whether it contains information in other scenarios, so I just move one position forwards for the bubbles, scores and ego_agent_ids.

smarts/sstudio/types.py
AisenGinn: This is to check whether user provides multiple coordinates with identical x_axis or y_axis. This will form a parallel line which is not a polygon.

smarts/sstudio/types.py
AisenGinn: I think you are right, the check for poly is None is not needed, will use the simplified version, thanks!

smarts/sstudio/types.py
saulfield: Ah yes, I missed the `set`, sorry. I still think we need an explicit check for len < 2 above though.

envision/web/src/helpers/state_unpacker.js
saulfield: I just looked into this, and it's because traffic signals were added at index 4 without updating this enum (see `_format_state()` in `envision/data_formatter.py`). I think we should add a line like `SIGNALS: 4,` here so this is clear in the future.

smarts/sstudio/types.py
```diff
@@ -890,6 +936,11 @@ def __post_init__(self):
                 "Only boids can have keep_alive enabled (for persistent boids)"
             )
 
+        if not isinstance(self.zone, MapZone):
+            poly = self.zone.to_geometry(road_map=None)
+            if poly.is_valid:
```
saulfield, line 941:
```suggestion
            if not poly.is_valid:
```

envision/web/src/helpers/state_unpacker.js
AisenGinn: I will add `SIGNALS: 4` in this branch, thanks!

smarts/sstudio/types.py
AisenGinn: There is a problem of using `len < 2` to check numbers of coordinates. If the user only provides one coordinate, the ext_coordinate will become something like this: ((90,90)). The `len(ext_coordinate)` should be 2 but python somehow unpacks it to be `ext_coordinate = (90,90)` and its `len = 2` again, this may not be a valid check.

saulfield: Are you sure? As long as it's always a list it should work. From the REPL:

```
>>> len([(90,90)])
1
```

AisenGinn: oh I tried it with this:
```
test = ((90,90))
print(len(test))
```
and it outputs 2

saulfield: That would be expected, since you are assigning a tuple of length 2. The outside parentheses are ignored.

```
>>> test = ((90,90))
>>> test
(90, 90)
```

I think we can assume that `ext_coordinates` is always a list.

AisenGinn: Normally the user should input a list, but () also works for multiple coordinates, and `len(((90,90),(110,110))) = 2`. Maybe I can convert it to list if index 0 is not a tuple and do the len check?

saulfield: This should actually work for tuples as well with no change. But the code would be:

```
>>> test = ((90,90),)
>>> len(test)
1
```

AisenGinn: I see, I will add the len check as well, thanks for your patience

saulfield (approved):

smarts/sstudio/types.py
```diff
@@ -852,6 +853,55 @@ def to_geometry(self, road_map: Optional[RoadMap] = None) -> Polygon:
...
+    = 2 points provided: generates a box using these two points as diagonal
+    > 2 points provided: generates a polygon according to the coordinates"""
+    rotation: Optional[float] = None
+    """The heading direction of the bubble(radians, clock-wise rotation)"""
+
+    def __post_init__(self):
+        if (
+            not self.ext_coordinates
+            or not isinstance(self.ext_coordinates[0], tuple)
+            or len(self.ext_coordinates) < 2
```
Gamenot, line 872: We should swap these conditions to short circuit and prevent an index error.
```suggestion
            or len(self.ext_coordinates) < 2
            or not isinstance(self.ext_coordinates[0], tuple)
```

smarts/sstudio/types.py
```diff
@@ -928,6 +978,11 @@ def __post_init__(self):
                 "Only boids can have keep_alive enabled (for persistent boids)"
             )
 
+        if not isinstance(self.zone, MapZone):
+            poly = self.zone.to_geometry(road_map=None)
+            if not poly.is_valid:
+                raise ValueError("The Zone Polygon is not a valid closed loop")
```
Gamenot, line 984: Is there any context we have that we can used to determine which bubble is having an issue?

Gamenot (changes requested):

Gamenot (approved):

Gamenot:
Good work!

PR #1702, opened 2022-11-07
Remove PyMARL

Adaickalavan:
+ [PyMARL](https://github.com/oxwhirl/pymarl) appears stale.
+ Our example to interface SMARTS with PyMARL appears non-functioning.
+ It is better to maintain a commonly accepted robust interface for SMARTS to ensure its compatibility with other frameworks, compared to maintaining individual adapter code for each external framework.
+ Hence, PyMARL content is removed in its entirety.


Gamenot (approved):
OK.

saulfield (approved):

PR #1708, opened 2022-11-09
Bubble distance check

AisenGinn:
Added distance check between vehicle and bubbles to avoid unnacessary creation of bubble.


smarts/core/bubble_manager.py
```diff
@@ -538,17 +550,31 @@ def _sync_cursors(self, last_vehicle_index, vehicle_index):
...
-                    index=persisted_vehicle_index,
-                    vehicle_ids_per_bubble=vehicle_ids_per_bubble,
-                    running_cursors=cursors,
+            point = vehicle.pose.point
+            v_radius = (
+                math.sqrt(
+                    vehicle.width * vehicle.width + vehicle.length * vehicle.length
                 )
-                cursors.add(cursor)
+            )
```
saulfield, line 558: These outer parentheses seem to be unnecessary.

examples/single_agent.py
```diff
@@ -70,6 +71,7 @@ def main(scenarios, headless, num_episodes, max_episode_steps=None):
 if __name__ == "__main__":
     parser = default_argument_parser("single-agent-example")
     args = parser.parse_args()
+    sys.path.append(".")
```
saulfield, line 74: Why was this needed?

AisenGinn: I cannot run the simulation sometimes without this line, I will remove this once other changes are approved

saulfield (approved):

smarts/core/bubble_manager.py
```diff
@@ -76,6 +77,13 @@ class Bubble:
     def __init__(self, bubble: SSBubble, road_map: RoadMap):
         geometry = bubble.zone.to_geometry(road_map)
 
+        bounding = geometry.bounds
+        self.radius = math.sqrt(
+            (bounding[2] - bounding[0]) * (bounding[2] - bounding[0])
+            + (bounding[3] - bounding[1]) * (bounding[3] - bounding[1])
+        )
```
Gamenot, line 84: This does not seem quite right... 

`pow(maxx - minxx, 2) + pow(maxy - miny, 2)`

I think this ends up significantly larger than it should be. I think it should be something like `pow(maxx - centroidx, 2) + pow(centroidy - miny, 2) + 0.1` perhaps?

Gamenot (changes requested):
This ended up more compact than I expected.

Gamenot (approved):

PR #1714, opened 2022-11-16
Add foe traffic signals to observation signals.

Gamenot:
This change is intended to expose foe traffic signals through `AgentInterface` configuration.


saulfield (approved):

Adaickalavan (approved):

smarts/core/sensors.py
```diff
@@ -1300,8 +1304,18 @@ def __call__(
...
         lookahead = self._lookahead - lane.length + lane_pos.s
-        self._find_signals_ahead(lane, lookahead, plan.route, upcoming_signals)
+        self._find_signals_ahead(
+            lane,
+            lookahead,
+            plan.route,
+            self._include_foes,
+            used_features,
+            upcoming_signals,
+        )
```
Adaickalavan, line 1318: It would be better to have functions with return values, compared to relying on functions with side effects.

Issue #1174, opened 2021-12-11
Remove adapters in favour of wrappers for obs/action/reward/info


Adaickalavan:
**Is your feature request related to a problem? Please describe.**
Adapters fall short, such as not allowing to propagate `spaces`. Currently we have adapter interface to shape observation, action, reward, and info.

https://github.com/huawei-noah/SMARTS/blob/7564644104e2eb2869641ec528a3067fa690c2df/smarts/core/agent.py#L109-L116

**Describe the solution you'd like**
Remove adapter interface in favour of wrappers. Wrapper interface to shape observation, action, reward, and info, appears to be the widely adopted norm. It is also present in Gym specifications.

https://github.com/openai/gym/blob/180d8ddd5cce0d23cdc4cada1faa52f3ae45313c/gym/core.py#L328-L368

**Describe alternatives you've considered**
Currently we have implemented both approaches, i.e., adapters and some selected wrappers, in our code base. I propose to completely remove or deprecate adapters.

**Additional context**
Use of wrappers will make SMARTS friendlier to users, who are familiar with usage of wrappers from other RL algorithms and environments.

Adaickalavan:
With the removal of adapters, we may consider removing the entire `AgentSpec` class as all of its contents are either deprecated or unused. The only exception being the `interface` attribute which could be elevated to top level such that users may access it directly instead of being nested under `AgentSpec` class.

https://github.com/huawei-noah/SMARTS/blob/7564644104e2eb2869641ec528a3067fa690c2df/smarts/core/agent.py#L72-L118

Adaickalavan:
Closed by #1779

PR #1724, opened 2022-11-24
Competition agent integration with zoo

AisenGinn:
competition_agent integration with zoo.

run:`scl run --envision competition_agent_demo.py scenarios/sumo/loop` to see competition integration with loop scenarios. To see other scenarios, replace `scenarios/sumo/loop` with other scenarios path.


zoo/policies/__init__.py
```diff
@@ -100,3 +104,86 @@ def human_keyboard_entrypoint(*arg, **kwargs):
 
 
 register(locator="human-in-the-loop-v0", entry_point=human_keyboard_entrypoint)
+
+
+from smarts.env.multi_scenario_env import resolve_agent_interface
```
saulfield, line 109: Is there a reason this isn't at the top of the file? If not, it should be moved.

zoo/policies/__init__.py
```diff
@@ -100,3 +104,86 @@ def human_keyboard_entrypoint(*arg, **kwargs):
...
+root_path = str(Path(__file__).absolute().parents[2])
+
+register(
+    "competition_agent-v0",
+    entry_point=competition_entry,
+    policy_path=os.path.join(root_path, "competition/track1/submission_1"),
+)
+
+register(
+    "competition_agent-v1",
```
saulfield, line 186: I wonder if these should have distinct names, and then have `v0` for both.

AisenGinn: The competition_agent-v1 will be removed in the end after testing some competition policies.

zoo/policies/competition_agent.py
```diff
@@ -0,0 +1,73 @@
...
+                    "install",
+                    "-t",
+                    self._sub_env_path,
+                    "-r",
+                    req_file,
+                ]
+            )
+            sys.path.append(self._sub_env_path)
+        except:
+            print("Failed to install requirement for Competition Agent")
```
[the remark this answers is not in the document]
AisenGinn: Sure, that's reasonable, should I use sys.exit to print the message and exit the program?

smarts/zoo/agent_spec.py
```diff
@@ -74,6 +74,8 @@ class AgentSpec:
     """An adaptor that allows shaping of the reward (default lambda obs, reward: reward)"""
     info_adapter: Callable = lambda obs, reward, info: info
     """An adaptor that allows shaping of info (default lambda obs, reward, info: info)"""
+    adapt_env: Optional[Callable]= lambda env: env
+    """And adaptor that allows configurating the env (default lambda env:env)"""
```
Adaickalavan, line 78: Adding more adapters counters efforts to remove the adapters #1174 .

Adaickalavan:
+ Why is this pull request being merged into `comp-1` branch?
+ We may grandfather `comp-1` branch and add refactored competition code into `develop` branch such that it can run any external users' agents in general.

AisenGinn:
> * Why is this pull request being merged into `comp-1` branch?
> * We may grandfather `comp-1` branch and add refactored competition code into `develop` branch such that it can run any external users' agents in general.

Currently only comp-1 branch has the competition folder and the require policies/requirements. My work is based on that and merging directly into develop may cause some issues I think.

Gamenot:
> > * We may grandfather `comp-1` branch and add refactored competition code into `develop` branch such that it can run any external users' agents in general.
> 
> Currently only comp-1 branch has the competition folder and the require policies/requirements. My work is based on that and merging directly into develop may cause some issues I think. Maybe we should try to merge comp-1 into develop first?

It is possible that comp-1 does not get merged in. This agent, however, lets us run competition agents directly as were created in the competition so that we will not have to modify them.

zoo/policies/competition_agent.py
```diff
@@ -0,0 +1,91 @@
...
+                    "install",
+                    "-t",
+                    self._sub_env_path,
+                    "-r",
+                    req_file,
+                ]
+            )
+            sys.path.append(self._sub_env_path)
+        except:
+            sys.exit("Failed to install requirement for Competition Agent")
```
Gamenot, line 43: Is this necessary, wouldn't the original exception be enough?

```suggestion
        except:
            logging.error("Failed to install requirement for Competition Agent")
            raise
```

zoo/policies/competition_agent.py
```diff
@@ -0,0 +1,91 @@
...
+
+        self._policy_dir = policy_path
+        self._comp_env_path = str(os.path.join(root_path, "competition_env"))
+        self._sub_env_path = str(os.path.join(self._comp_env_path, env_name))
+
+        Path.mkdir(Path(self._comp_env_path), exist_ok=True)
+
+        if Path(self._sub_env_path).exists():
+            shutil.rmtree(self._sub_env_path)
+        Path.mkdir(Path(self._sub_env_path))
```
Gamenot, line 26: This should not be necessary since you can use `parents=True` on line 26.
```suggestion
        if Path(self._sub_env_path).exists():
            shutil.rmtree(self._sub_env_path)
        Path.mkdir(Path(self._sub_env_path), parents=True)
```

Gamenot (changes requested):

zoo/policies/__init__.py
```diff
@@ -100,3 +108,126 @@ def human_keyboard_entrypoint(*arg, **kwargs):
...
+        logging.error(
+            f"Failed to install requirement for Competition Agent in folder {Path(policy_path).name}"
+        )
+        raise
+
+    # insert submission path
+    while policy_path in sys.path:
+        sys.path.remove(policy_path)
+
+    sys.path.insert(0, policy_path)
```
Gamenot, line 166: Why is this necessary? Is there some sort of leak?
```suggestion
    if policy_path not in sys.path:
        sys.path.insert(0, policy_path)
```

zoo/policies/__init__.py
```diff
@@ -100,3 +108,126 @@ def human_keyboard_entrypoint(*arg, **kwargs):
...
+        adapt_env=env_wrapper,
+        agent_builder=CompetitionAgent,
+    )
+
+    # delete competition policy module and remove related path
+    while policy_path in sys.path:
+        sys.path.remove(policy_path)  # prevent duplicate path remain in sys.path
+
+    while sub_env_path in sys.path:
+        sys.path.remove(sub_env_path)  # prevent duplicate path remain in sys.path
```
Gamenot, line 213: How does this end up in the path?

zoo/policies/competition_agent.py
```diff
@@ -0,0 +1,40 @@
...
+        self._policy_dir = policy_path
+        self._comp_env_path = str(os.path.join(root_path, "competition_env"))
+        self._sub_env_path = str(os.path.join(self._comp_env_path, env_name))
+
+        self._policy = policy
+
+    def act(self, obs):
+        return self._policy.act(obs)
+
+    def close(self, remove_all_env=False):
```
Gamenot, line 26: I do not see where this is called.

zoo/policies/__init__.py
```diff
@@ -100,3 +108,126 @@ def human_keyboard_entrypoint(*arg, **kwargs):
...
+            "policy_path": policy_path,
+            "policy": policy,
+        },
+        adapt_env=env_wrapper,
+        agent_builder=CompetitionAgent,
+    )
+
+    # delete competition policy module and remove related path
+    while policy_path in sys.path:
+        sys.path.remove(policy_path)  # prevent duplicate path remain in sys.path
```
Gamenot, line 210: Why is it necessary to remove duplicates like this?

zoo/policies/competition_agent.py
```diff
@@ -0,0 +1,40 @@
...
+            sys.path.remove(self._policy_dir)
+        for key, module in list(sys.modules.items()):
+            if "__file__" in dir(module):
+                module_path = module.__file__
+                if module_path and (
+                    self._policy_dir in module_path or self._sub_env_path in module_path
+                ):
+                    sys.modules.pop(key)
+        if remove_all_env:
+            shutil.rmtree(self._comp_env_path, ignore_errors=True)
```
Gamenot, line 40: This cleanup seems to be related to the requirements for all agents of this type rather than the agent instance-specific resources. Maybe this can be registered as an `atexit` callback in the entrypoint?

zoo/policies/__init__.py
```diff
@@ -100,3 +108,126 @@ def human_keyboard_entrypoint(*arg, **kwargs):
...
+    # remove all modules related to policy_path
+    for key, module in list(sys.modules.items()):
+        if "__file__" in dir(module):
+            module_path = module.__file__
+            if module_path and (
+                policy_path in module_path or sub_env_path in module_path
+            ):
+                sys.modules.pop(key)
+
+    del policy_module
```
Gamenot, line 224: Why are the modules removed at this point?

zoo/policies/__init__.py
AisenGinn: There are some users that insert the submission path inside the policy.py they wrote, some do not. If they insert there will be two submission paths inside sys.path (which is a list that allow duplication). This is to remove all possible policy paths inside sys.path.

test_agent.py
```diff
@@ -0,0 +1,27 @@
+import gym
+from smarts.zoo import registry
+import sys
+
+test_agent = registry.make_agent("zoo.policies:competition_agent-v0")
+test_agent_spec = registry.make("zoo.policies:competition_agent-v0")
```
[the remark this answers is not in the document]
AisenGinn: Is it possible to add an optional argument default to false to let the user decide whether they want to keep the spec?

zoo/policies/__init__.py
```diff
@@ -100,3 +108,126 @@ def human_keyboard_entrypoint(*arg, **kwargs):
...
+
+    return spec
+
+
+register(
+    "competition_agent-v0",
+    entry_point=competition_entry,
+    policy_path=os.path.join(root_path, "competition/track1/submission"),
```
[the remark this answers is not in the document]
AisenGinn: Will do.

zoo/policies/__init__.py
AisenGinn: Just want to ensure that policy_path is been searched first in the sys.path to avoid potential issues. If the policy is already in sys.path and there are some other paths contain `policy.py` in front of it, it may cause some import errors.

zoo/policies/competition_agent.py
AisenGinn: It should be called when all works in done inside ` test_agent.py`. I will modify `test_agent.py` to include this.

zoo/policies/__init__.py
AisenGinn: The sub_env_path is appended when the dependency is installed.

zoo/policies/__init__.py
AisenGinn: If the user doesn't called competition_agent.close(), this will ensure they don't have import issues when they import the next competition_agent.

zoo/policies/competition_agent.py
AisenGinn: That sound reasonable, will do.

test_agent.py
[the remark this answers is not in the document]
Gamenot: No, I think we just cache the spec if it is not already cached.

zoo/policies/__init__.py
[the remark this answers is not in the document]
Gamenot: I think we will want to either do it by referencing the winning agents like `"competition_aid_v0"`, `"competition_vcr_v0"`, and `"competition_tju-fanta_v0"`. or by removing a default to force `policy_path` to be specified when initializing an agent.

zoo/policies/__init__.py
Gamenot: Oh, interesting, the imports resolve in the order they occur in `sys.path`? I think I understand now.

zoo/policies/__init__.py
Gamenot: OK, that makes more sense.

zoo/policies/__init__.py
Gamenot: Then one final question: Is the policy module not needed anymore such that the object can be deleted?

test_agent.py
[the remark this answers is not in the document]
AisenGinn: ok

zoo/policies/__init__.py
AisenGinn: Will do.

zoo/policies/__init__.py
AisenGinn: Will do.

zoo/policies/__init__.py
AisenGinn: I think after the policy & submitted wrapper is once been stored by a variable then the module is not needed anymore. Although keeping the policy_module may not affect the import of next agent.

PR #1734, opened 2022-11-25
Use build for map spec to avoid error prone syntax.

Gamenot:
This avoids the potential issue of passing in an improper value to the map spec builder function.

### DIFF
--- a/smarts/sstudio/types.py
+++ b/smarts/sstudio/types.py
@@ -382,6 +382,12 @@ class MapSpec:
     If not specified, this currently defaults to a function that creates
     SUMO road networks (get_road_map()) in smarts.core.default_map_builder."""
 
+    def build(self) -> Tuple[Optional[RoadMap], Optional[str]]:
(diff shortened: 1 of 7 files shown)


smarts/sstudio/types.py
```diff
@@ -382,6 +382,12 @@ class MapSpec:
     If not specified, this currently defaults to a function that creates
     SUMO road networks (get_road_map()) in smarts.core.default_map_builder."""
 
+    def build(self) -> Tuple[Optional[RoadMap], Optional[str]]:
+        """This should return an object derived from the RoadMap base class
+        and a hash that uniquely identifies it (changes to the hash should signify
+        that the map is different enough that map-related caches should be reloaded)."""
+        return self.builder_fn(self)
+
```
Adaickalavan, line 390: We could try keeping data and methods separately.

Gamenot:
This is not necessary at this time.

Issue #1722, opened 2022-11-23
Smarts map coordinate system


yuant95:
**Problems**
I noticed that smarts map is not centered but rather started at left bottom corner? I am wondering whether I can center the map, or get the bounding box of the map so I can convert the agent location into a centered map. 

I need to figure this out to get a third part library to generate actions for some custom agents I implemented. They centered the map, so I need to do the conversion. 

Let me know if my question is not clear or you need more information. Thanks!

Gamenot:
Hello @yuant95, if you have access to the underlying `SMARTS` instance you can use it to get the map:

```python
smarts_instance = hiway_env._smarts
hiway_env.reset()
## the current road map is available after SMARTS.reset(scenario) or HiwayEnv.reset() 
center = smarts_instance.road_map.bounding_box.center
```

another option is to get the road map centers beforehand

```python
scenario_roots = smarts.core.Scenario.get_scenario_list(scenarios_or_scenarios_dirs)
for root in scenario_roots:
    map_spec = smarts.core.Scenario.discover_map(scenario_root, 1.0, None)
    road_map, road_map_hash = map_spec.builder_fn(map_spec)
    center = road_map.bounding_box.center
```

Gamenot:
Another option is to use the metadata from the glb https://github.com/huawei-noah/SMARTS/pull/1716/files.

```python
meta = road_map.map_glb_metadata
bounding_box = meta["bounding_box"]
```

PR #1750, opened 2022-11-25
Expose underlying map source

Gamenot:
@saulfield This might address the common issue where people want access to the underling map data.

A similar request was made [here](https://github.com/huawei-noah/SMARTS/issues/1722) as well as internally.

I am thinking we may wish to also expose the scenario directory as well. That way we can allow users to generate the RoadMap directly through `Scenario.discover_scenario_roots`.

### DIFF
--- a/smarts/core/sensors.py
+++ b/smarts/core/sensors.py
@@ -234,6 +234,7 @@ class Observation:
     top_down_rgb: Optional[TopDownRGB]
     road_waypoints: Optional[RoadWaypoints]
     via_data: Vias
+    map_source: str
     signals: Optional[List[SignalObservation]] = None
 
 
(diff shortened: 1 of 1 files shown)


Gamenot:
@Adaickalavan @saulfield Any counter-proposals to this solution?

Adaickalavan:
1. Does the `Observation(map_source=sim.road_map.source)` return the same value at every time step? If yes, it might be redundant and go against expectation of receiving observation produced at each time step.
3. On another note, the latest gym spaces appears to accommodate texts.

Gamenot:
It looks like `gymnasium` or a minimum of `gym==0.21` would let us get a dictionary called `infos` from `env.reset`. That would solve the issue here since it would no longer have to come out through the observations.

Adaickalavan:
The `info` from `obs, info = env.reset()` should be analogous to the `info` returned by `obs, reward, terminated, truncated, info = step()`.

Adaickalavan:
Curently in `develop` branch the scenario is exposed via `def scenario(self) -> Scenario:` method from which the `RoadMap` can be obtained.

PR #1752, opened 2022-11-25
Update of rl example: racing

Adaickalavan:
1. Updated to use `smarts[camera_obs]==0.7.0rc0`
2. Added action space
3. Updated loop scenario to use continuous flowing traffic
4. Avoided subprocess in scenario build
5. Simplified example folder structure


setup.cfg
```diff
@@ -79,7 +79,7 @@ doc = sphinx>=4.4.0
     sphinxcontrib-apidoc>=0.3.0
 extras = pynput>=1.7.4  # Used by HumanKeyboardAgent
 gym = moviepy == 1.0.3
-remote_agent = grpcio==1.32.0
+remote-agent = grpcio==1.32.0
```
Gamenot, line 82: Is this change necessary? The direction we are going is to have all of the extras remove dash.

Gamenot (approved):
This works, we will need to update the format of the main repository extras as a different PR.

setup.cfg
Adaickalavan: Reverted accidental change.

Gamenot: On second review, this is the cause of all of the extras having the `-` -> `_` conversion.

Adaickalavan: Shall we simply convert all extras to use `_`, to avoid confusion and to avoid auto name changes?

Gamenot: I think we can do that because the next release is a major version change.

Gamenot: For now we should use `remote-agent` and then have an issue to convert the rest.

Adaickalavan: Conversion of `-` to `_` in `options.extras_require` is done in #1756 .

examples/rl/racing/setup.py
```diff
@@ -15,10 +15,10 @@
     packages=find_packages(),
     include_package_data=True,
     zip_safe=True,
-    python_requires=">=3.7",
+    python_requires="~=3.7",
```
Gamenot, line 18: `3.8` does not work?

saulfield (approved):

examples/rl/racing/setup.py
Gamenot: We can merge this

PR #1753, opened 2022-11-28
Performance metrics

Adaickalavan:
1. An easy to use `Metrics` env wrapper is introduced to compute agents' performance metrics.
2. Fine grained modular metric computation and reporting for each agent in each scenario.
3. An overall performance score is also computed.
4. Access to private variables is restricted to ensure security of metrics computed.
5. Immutable data structures are favoured, such as frozen `Counts` and `Costs`, to facilitate scalability, understandability, and to prevent accidental erroneous modifications.


smarts/env/wrappers/metrics.py
```diff
@@ -0,0 +1,273 @@
...
+
+    Raises:
+        AttributeError: Upon accessing an attribute beginning with an underscore.
+
+    Returns:
+        gym.Env: A wrapped gym.Env which computes agents' performance metrics.
+    """
+
+    def __init__(self, env: gym.Env):
+        env = _Metrics(env)
```
Gamenot, line 69: If the _Metrics utility is wrapped like this, how do we get at score and records?

The `gym` way to get this information might be `env.render("score")` perhaps?

Adaickalavan: We still can get the score and records by calling `env.score()` and `env.records()`.

Gamenot: Ah, right, it forwards through `__getattr__`.

smarts/env/wrappers/metrics.py
```diff
@@ -0,0 +1,273 @@
...
+                "compute its metrics. Current interface is "
+                "{1}.".format(agent_name, intrfc)
+            )
+
+
+T = TypeVar("T", Costs, Counts)
+
+
+def _add_dataclass(first: T, second: T) -> T:
+    new = {}
```
Gamenot, line 246: A reasonable check would be to test that the instances are of the same class.
```suggestion
def _add_dataclass(first: T, second: T) -> T:
    assert type(first) is type(second)
    new = {}
```

smarts/env/wrappers/metrics.py
```diff
@@ -0,0 +1,273 @@
...
+                if reason == termination.Reason.Goal:
+                    steps_adjusted = self._steps[agent_name]
+                    goals = 1
+                elif reason == termination.Reason.Crash:
+                    steps_adjusted = _MAX_STEPS
+                    crashes = 1
+                else:
+                    raise Exception(
+                        f"Unsupported agent done reason. Events: {agent_obs.events}."
+                    )
```
Gamenot, line 123: I think we should raise a more specific error than `Exception` since `Exception` makes the error hard to isolate for purposes of debugging and `try-catch`.

Gamenot (changes requested):

smarts/env/wrappers/metrics.py
```diff
@@ -0,0 +1,273 @@
...
+from dataclasses import dataclass, fields
+from typing import Any, Dict, Set, TypeVar
+
+import gym
+
+from smarts.env.wrappers.metric import termination
+from smarts.env.wrappers.metric.costs import CostFuncs, Costs
+from smarts.env.wrappers.metric.counts import Counts
+
+_MAX_STEPS = 800
```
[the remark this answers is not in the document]
Adaickalavan: Now, the max episode steps is obtained directly from each agent spec's `max_episode_steps` attribute. This will be used to normalise the number of steps taken by each agent in each episode.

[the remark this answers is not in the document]
Adaickalavan: Now, the max episode steps is obtained directly from each agent spec's `max_episode_steps` attribute. The `_MAX_STEPS` constant has been removed.

smarts/env/wrappers/metrics.py
Adaickalavan: A custom `MetricsError` class has been added.

smarts/env/wrappers/metric/termination.py
```diff
@@ -0,0 +1,56 @@
...
+        Reason: Reason for agent termination.
+    """
+    if obs.events.reached_goal:
+        return Reason.Goal
+    elif (
+        len(obs.events.collisions) > 0
+        or obs.events.off_road
+        or obs.events.reached_max_episode_steps
+    ):
+        return Reason.Crash
```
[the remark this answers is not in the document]
Adaickalavan: Now, the events are checked directly without any prior grouping.

smarts/env/wrappers/metrics.py
Adaickalavan: Added type assertion.

smarts/env/wrappers/metrics.py
```diff
@@ -0,0 +1,284 @@
...
+                counts = Counts(
+                    episodes=1,
+                    steps=self._steps[agent_name],
+                    steps_adjusted=steps_adjusted,
+                    goals=goals,
+                )
+                self._records[self._cur_scen][agent_name].record.counts = _add_dataclass(
+                    counts, 
+                    self._records[self._cur_scen][agent_name].record.counts
+                )
```
saulfield, line 138: I have mixed feelings about this approach. Though the immutability is nice, it seems like a bit of a burdensome way to maintain state. Creating a new dict and dataclass each step also seems inefficient. I don't think it's too critical as long as the performance is ok though.

saulfield (approved):

smarts/env/wrappers/metric/counts.py
```diff
@@ -0,0 +1,41 @@
...
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class Counts:
+    """Performance count values."""
+
+    goals: int = 0
+    """ Number of episodes completed succesfully by achieving the goal.
```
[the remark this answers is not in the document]
Adaickalavan: Fixed the spelling mistake.

Gamenot (approved):

PR #1761, opened 2022-11-29
Envision: "All Simulation" Button

qianyi-sun:
1) Add "All Simulations" button next to the dropdown in the header to navigate to the "All Simulation" overview page.
2) Make each small-windowed simulations in the "All Simulation" page clickable and can be maximized by clicking any of them.


Gamenot (approved):
I am confident in it after seeing it. There is some consideration to add back the list of simulations but as a feature it works well.

saulfield (approved):

PR #1767, opened 2022-12-06
Speed up rllib test

Gamenot:
Shorten the time of the `test_rllib_hiway_env` test.

### DIFF
--- a/smarts/env/tests/test_rllib_hiway_env.py
+++ b/smarts/env/tests/test_rllib_hiway_env.py
@@ -36,6 +36,7 @@ from smarts.core.agent_interface import AgentInterface, AgentType
 from smarts.core.utils.file import make_dir_in_smarts_log_dir
 from smarts.env.rllib_hiway_env import RLlibHiWayEnv
 from smarts.zoo.agent_spec import AgentSpec
+from smarts import sstudio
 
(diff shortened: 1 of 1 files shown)


saulfield:
What is the speed improvement for this? I think this test took around 30 mins on my machine originally.

qianyi-sun:
Not sure if this is reasonable, how about we create a even simpler scenario or duplicate an existing scenario just for the test and maybe gitignore it to prevent possible accidentally removal of it? I just have a feeling that in the future somehow we might forget that we have set this scenario for the test and delete it for some reason..

Gamenot:
> Not sure if this is reasonable, how about we create a even simpler scenario or duplicate an existing scenario just for the test and maybe gitignore it to prevent possible accidentally removal of it? I just have a feeling that in the future somehow we might forget that we have set this scenario for the test and delete it for some reason..

I do not think that is possible since it is linked to the CI.

Gamenot:
> What is the speed improvement for this? I think this test took around 30 mins on my machine originally.

I had the test take 3 minutes with this version.

saulfield (approved):

Gamenot:
@saulfield I think the next step after may be to limit the number of CPUs.

PR #1768, opened 2022-12-06
add build process in scl run command

qianyi-sun:
With the "Scenario Caching" PR https://github.com/huawei-noah/SMARTS/pull/1764, scenarios are only built when needed so maybe we could still leave the `scl scenario build` command as is for manual scenario building usage, but call `build_scenarios()` in the `scl run` command so that we don't need two commands (`scl scenario build` and `scl run`) to run a simulation.

The new command would work like `scl run --clean --envision <example_path> [scenario_path]`

This PR should be in after https://github.com/huawei-noah/SMARTS/pull/1764 get merged.


Gamenot (approved):

cli/run.py
```diff
@@ -43,24 +44,42 @@ def kill_process_group_afterwards():
...
-def run_experiment(envision, envision_port, script_path, script_args):
+@click.argument("scenario_path", nargs=-1, type=click.UNPROCESSED)
+def run_experiment(clean, envision, envision_port, example_path, scenario_path,seed):
     with kill_process_group_afterwards():
+        build_scenarios(
+        allow_offset_maps=False,
+        clean=clean,
+        scenarios=scenario_path,
+        seed=seed,
+        )
```
Gamenot, line 82:
```suggestion
        build_scenarios(
            allow_offset_maps=False,
            clean=clean,
            scenarios=scenario_path,
            seed=seed,
        )
```

cli/run.py
```diff
@@ -43,24 +44,42 @@ def kill_process_group_afterwards():
     help="Run an experiment on a scenario",
     context_settings=dict(ignore_unknown_options=True),
 )
+@click.option(
+    "--clean",
+    is_flag=True,
+    default=False,
+    help="Clean previously generated artifacts first",
+)
```
Gamenot, line 52: @saulfield Is this necessary now? I thought it is no longer necessary to call clean.

saulfield: It's not necessary, but it is still an option.

cli/run.py
```diff
@@ -43,24 +44,42 @@ def kill_process_group_afterwards():
...
-    "script_path", type=click.Path(exists=True), metavar="<script>", required=True
+    "example_path", type=click.Path(exists=True), metavar="<example>", required=True
 )
-@click.argument("script_args", nargs=-1, type=click.UNPROCESSED)
-def run_experiment(envision, envision_port, script_path, script_args):
+@click.argument("scenario_path", nargs=-1, type=click.UNPROCESSED)
+def run_experiment(clean, envision, envision_port, example_path, scenario_path,seed):
     with kill_process_group_afterwards():
+        build_scenarios(
+        allow_offset_maps=False,
```
saulfield, line 78: `allow_offset_maps` is no longer a parameter

cli/run.py
```diff
@@ -43,24 +44,42 @@ def kill_process_group_afterwards():
...
+        clean=clean,
+        scenarios=scenario_path,
```
saulfield, line 80: I think this needs to be a list of scenarios

cli/run.py
```diff
@@ -43,24 +44,42 @@ def kill_process_group_afterwards():
...
+)
 @click.option(
     "-p",
     "--envision_port",
     help="Port on which Envision will run.",
     default=None,
 )
 @click.argument(
-    "script_path", type=click.Path(exists=True), metavar="<script>", required=True
+    "example_path", type=click.Path(exists=True), metavar="<example>", required=True
```
Adaickalavan, line 72: Perhaps it could still be called as `script`.
```suggestion
    "script_path", type=click.Path(exists=True), metavar="<script>", required=True
```

cli/run.py
```diff
@@ -58,9 +71,14 @@ def kill_process_group_afterwards():
 @click.argument(
     "script_path", type=click.Path(exists=True), metavar="<script>", required=True
 )
-@click.argument("script_args", nargs=-1, type=click.UNPROCESSED)
-def run_experiment(envision, envision_port, script_path, script_args):
+@click.argument("scenario_path", nargs=-1, type=click.UNPROCESSED)
```
Adaickalavan, line 74: It would be helpful to add "scenario_path" as a separate argument in addition to maintaining multiple optional arguments to be fed to the script.

Gamenot, line 74: I agree with @Adaickalavan's take on this. Please make sure that the change does not prevent forwarding options to the underlying script.

Gamenot (changes requested):

PR #1772, opened 2022-12-13
Performance metrics - normalized, recomputed, overall score

Adaickalavan:
+ Example usage of `Metrics` env wrapper to compute agents' performance metrics.
    ```python
    from smarts.env.wrappers.metrics import Metrics
    env = gym.make( . . .)
    env = Metrics(env)
    . . .
    print(env.records()) # Fine grained performance metric for each agent in each scenario.
    print(env.score()) # Overall performance score achieved on the wrapped environment.
    ```
+ Normalized performance costs.
+ Introduced a `Completion` metric to compute the percentage of scenario tasks completed. `Completion` currently consists of distance remaining to reach goal location. More tasks could be added in the future.
+ Modified `smarts.core.plan::Plan.create_route` such that multiple nearest starting lanes are considered in sequence to find a route. This helps route calculation for agents with starting position inside intersections, where the `nearest_lanes` method might not return the correct road as the first choice.
+ A single overall performance score is now made available.


saulfield (approved):

smarts/core/scenario.py
```diff
@@ -558,7 +559,7 @@ def get_vehicle_start_at_time(
             Start(
                 np.array([pos_x + hhx, pos_y + hhy]),
                 Heading(heading),
-                from_front_bumper=False,
+                from_front_bumper=True,
```
Gamenot, line 562: This will have a minor affect on traffic history.

smarts/core/plan.py
```diff
@@ -335,45 +335,41 @@ def create_route(self, mission: Mission):
...
+        # When an agent is in an intersection, the `nearest_lanes` method might
+        # not return the correct road as the first choice. Hence, nearest
+        # starting lanes are tried in sequence until a route is found or until
+        # all nearby starting lane options are exhausted.
+        for start_lane, _ in start_lanes:
+            self._route = self._road_map.generate_routes(
+                start_lane.road, end_lane.road, via_roads, 1
+            )[0]
+            if self._route.road_length > 0:
+                break
```
Gamenot, line 363: This breaks on the first route. Is seems to go counter to the comment above.

Gamenot (changes requested):

smarts/core/plan.py
Adaickalavan: The `nearest_lane()` method returns nearby `start_lanes` sorted in increasing distance from the given point. Previously, we tried to find a route using the first `start_lane` returned, which might not yield a route if the `start_lane` was incorrect. Hence, now we try each `start_lane` in a loop until we find a legal route and immediately break upon finding one with `_route.road_length > 0` . We do not wait to compute routes for all the `start_lanes` and pick the shortest overall route, because the further `start_lanes` are away from the given point the higher the chances of it also being an incorrect starting road.

smarts/core/scenario.py
Adaickalavan: The `from_front_bumper` was set to `True` as the start position `np.array([pos_x + hhx, pos_y + hhy])` was at the front bumper.

smarts/core/plan.py
Gamenot: OK, I just note that there is the following potential problem.
selection_problem

smarts/core/plan.py
```diff
@@ -335,45 +335,41 @@ def create_route(self, mission: Mission):
 
         assert isinstance(self._mission.goal, PositionalGoal)
 
-        start_lane = self._road_map.nearest_lane(
+        start_lanes = self._road_map.nearest_lanes(
             self._mission.start.point,
-            include_junctions=False,
+            include_junctions=True,
         )
```
Adaickalavan, line 341: Here, `radius` is not specified in call to `self._road_map.nearest_lanes()`. By default `radius=None` and thus the radius is set to `radius = self._default_lane_width` inside `self._road_map.nearest_lanes()`.

https://github.com/huawei-noah/SMARTS/blob/4482821e0fbf44d83b0f627738d7753666902a49/smarts/core/sumo_road_network.py#L792-L797

Gamenot (approved):

PR #1773, opened 2022-12-13
Updated use of shapely.geometry.Polygon and shapely.geometry.collection.GeometryCollection 

Adaickalavan:
1. Newly released `shapely==2.0.0` includes type check which fails when `Polygon` is instantiated using sequence of `Point` objects.
    ```python
    # File: shapely/geometry/polygon.py
    if not np.issubdtype(shell.dtype, np.number):
        # conversion of coords to 2D array failed, this might be due
        # to inconsistent coordinate dimensionality
        raise ValueError("Inconsistent coordinate dimensionality")
    ```
    + Hence, `Polygon` is instantiated using sequence of coordinate pairs instead of sequence of `Point` objects.

2. Newly released `shapely==2.0.0` deprecates iteration over multi-part geometries. 
    + Hence, `geoms` property is used to access the constituent parts of a multi-part `GeometryCollection` object.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     rich>=11.2.0
     Rtree>=0.9.7
     sh>=1.14.2
-    shapely>=1.8.1
+    shapely>=2.0.0
     tableprint>=0.9.1
     trimesh==3.9.29  # Used for writing .glb files
     visdom>=0.1.8.9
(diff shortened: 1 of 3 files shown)


Gamenot:
Does this generalise to previous versions?

saulfield (approved):

qianyi-sun (approved):

PR #1774, opened 2022-12-14
Suppress the `DeprecationWarning` caused by the `trimesh` package

qianyi-sun:
Suppress the ```/root/SMARTS/.venv/lib/python3.8/site-packages/trimesh/curvature.py:12: DeprecationWarning: Please use `coo_matrix` from the `scipy.sparse` namespace, the `scipy.sparse.coo` namespace is deprecated.```  warning.

### DIFF
--- a/smarts/core/scenario.py
+++ b/smarts/core/scenario.py
@@ -66,6 +66,11 @@ from smarts.sstudio.types import Via as SSVia
 
 VehicleWindow = TrafficHistory.TrafficHistoryVehicleWindow
 
+# Suppress trimesh deprecation warning
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=DeprecationWarning)
+    import trimesh # only suppress the warnings caused by trimesh
 
 class Scenario:
     """The purpose of the Scenario is to provide an aggregate of all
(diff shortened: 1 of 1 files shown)


saulfield (approved):

smarts/core/scenario.py
```diff
@@ -66,6 +66,11 @@
 
 VehicleWindow = TrafficHistory.TrafficHistoryVehicleWindow
 
+# Suppress trimesh deprecation warning
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=DeprecationWarning)
+    import trimesh # only suppress the warnings caused by trimesh
```
Adaickalavan, line 73: + Does the catch warning also cater to `import trimesh` in other files, such as `opendrive_road_network.py`, `sumo_road_network.py`, `waymo_map.py`, and `geometry.py`? 
+ Can catch warning be implemented inside `map_glb_meta_for_file()` method below in this file where `trimesh` is being imported?

Gamenot: @qianyi-sun Could you make the warning more specific? It is possible to pattern match warnings to ignore.

Adaickalavan (approved):

smarts/core/scenario.py
qianyi-sun: @Adaickalavan I checked running waymo/opendrive/sumo scenarios, seems like the warning is gone in all three type of scenarios. 
@Gamenot If I understood "specfic" correctly, I just added a `message=` option in the `warnings.filterwarning` to only suppress this specific message.

Gamenot: Yes, that is what I meant.

Gamenot (approved):

PR #1775, opened 2022-12-14
Benchmark Improvement

qianyi-sun:
There are two things can be improved on the benchmark:

- [x] Sort the result in ascending order to ensure that the xlabel of output plot is always in order rather than random order like "1_actors, 200_actors, 50_actors..."
- [ ] Support of Spaghetti Plot (a line plot with many lines displayed together, in case of running the benchmark like `scl benchmark run n_sumo_actors/1_actors n_agents/1_agents`)


setup.cfg
```diff
@@ -66,6 +66,7 @@ console_scripts =
 [options.extras_require]
 benchmark = py-cpuinfo==9.0.0
     mdutils==1.4.0
+    natsort==8.2.0
```
saulfield, line 69: We should avoid adding any new dependencies if we can. I think we can use Python's native sorting functionality for this.

Gamenot: I agree.

Gamenot:
I don't think we can merge this with `natsort`.

PR #1782, opened 2022-12-22
Move dataset converter tool

Adaickalavan:
Moved `traffic_histories_vehicle_replacement.py` from `examples` folder to `smarts/waymo/` folder as it is more of a tool to generate SMARTS compatible observations from offline datasets rather than an example.


qianyi-sun (approved):

qianyi-sun:
I agree to move this script from examples to a tool but maybe we should re-consider to put it into `smarts/waymo` but to a more general, non-waymo folder since we also use it for converting NGSIM datasets too.

Gamenot:
> I agree to move this script from examples to a tool but maybe we should re-consider to put it into `smarts/waymo` but to a more general, non-waymo folder since we also use it for converting NGSIM datasets too.

I think I agree on this point as well, I am not sure if this goes into something like `smarts/dataset_core`.

Adaickalavan:
@qianyi-sun @Gamenot 
What about moving `traffic_histories_to_observations.py` into a new folder named `smarts/dataset`?

Gamenot:
> @qianyi-sun @Gamenot What about moving `traffic_histories_to_observations.py` into a new folder named `smarts/dataset`?

That seems reasonable.

Gamenot (approved):

PR #1793, opened 2023-01-10
Mainline competition

Gamenot:
- This is intended to bridge the competition into SMARTS through upgrading to `gymnasium`.


smarts/env/gymnasium/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from gymnasium.envs.registration import register
```
Adaickalavan, line 21: Gymnasium provides API for single agent RL env. SMARTS supports multi-agent RL env currently by building upon Gymnasium to follow the observation and action space interface from multi-agent RLlib. All our env could be simply placed under `smarts/env` instead of `smarts/env/gymnasium`.

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Solved requirement:
+
+    .. note::
+
+        If agent successfully completes the mission then ``info["score"]`` will
+        equal 1, else it is 0. Considered solved when ``info["score"] == 1`` is
+        achieved over 500 consecutive episodes.
+
+    :param scenario: Scenario name or path to scenario folder.
```
Adaickalavan, line 112: Consider using Google styled dcostrings for standardized style and easy readability. Documentation standard suggestion is available [here]( https://smarts.readthedocs.io/en/rtd/resources/contributing.html#documentation).

docs/sim/agent.rst
```diff
@@ -72,6 +72,7 @@ different max_episode_len for each agent.
 - `ActionSpaceType.TrajectoryWithTime`: `(Sequence[float], Sequence[float], Sequence[float], Sequence[float], Sequence[float])` continuous action space using trajectory as times, x coordinates, y coordinates, headings, and speeds to interpolate the vehicle along the trajectory.
 - `ActionSpaceType.MPC`: `(Sequence[float], Sequence[float], Sequence[float], Sequence[float])` continuous action space using trajectory as x coordinates, y coordinates, headings, and speeds to adaptively perform controls on the vehicle model in an attempt to match the given trajectory. 
 - `ActionSpaceType.TargetPose`: `Sequence[float, float, float, float]` continuous action space with a single vehicle x coordinate, y coordinate, heading, and time delta to reach the given pose.
+- `ActionSpaceType.TargetPoseFixedTime`: `Sequence[float, float, float]` same as `ActionSpaceType.TargetPose` but the time delta is fixed to `0.1s`.
```
Adaickalavan, line 75: Perhaps include `ActionSpaceType.RelativeTargetPose`.

examples/control/single_agent.ipynb
```diff
@@ -56,7 +56,7 @@
    "outputs": [],
    "source": [
     "from smarts.core.agent import Agent\n",
-    "from smarts.core.sensors import Observation\n",
+    "from smarts.core.observationsmport Observation\n",
```
Adaickalavan, line 59:
```suggestion
    "from smarts.core.observations import Observation\n",
```

examples/rl/racing/reward.py
```diff
@@ -1,7 +1,7 @@
 import gym
 import numpy as np
 
-from smarts.core.sensors import Observation
+from smarts.core.observations import Observation
```
Adaickalavan, line 4: The `examples/rl/racing` setup.py uses SMARTS version 0.7.0rc0 which might need to be upgraded and the example should be re-checked with any changes to the example.

smarts/core/agents_provider.py
```diff
@@ -222,7 +222,7 @@ def __init__(self, sim):
         super().__init__(sim)
 
     @property
-    def action_spaces(self) -> Set[ActionSpaceType]:
+    def actions(self) -> Set[ActionSpaceType]:
```
Adaickalavan, line 225: Why did we rename `action_spaces` to `actions`?

smarts/core/road_map.py
```diff
@@ -625,6 +625,9 @@ def __eq__(self, other) -> bool:
             """Required for set usage; derived classes may override this."""
             return self.__class__ == other.__class__ and hash(self) == hash(other)
 
+        def _add_road(self, road: RoadMap.Road):
```
Adaickalavan, line 628: Should we have an underscored method to be implemented by the child?

smarts/env/wrappers/format_obs.py
```diff
@@ -214,72 +214,69 @@ class FormatObs(gym.ObservationWrapper):
     def __init__(self, env: gym.Env):
         """
         Args:
-            env (gym.Env): SMARTS environment to be wrapped.
+            env (E): SMARTS environment to be wrapped.
```
Adaickalavan, line 217: What is `E`?

smarts/env/wrappers/format_obs.py
```diff
@@ -318,7 +315,7 @@ def get_spaces() -> Dict[str, Callable[[Any], gym.Space]]:
     """Returns the available gym spaces of a `StdObs`.
 
     Returns:
-        Dict[str, Callable[[Any], gym.Space]]:
+        Dict[str, Callable[[Any], S]]:
```
Adaickalavan, line 318: What is `S`?

smarts/env/tests/test_rgb_image.py
```diff
@@ -36,13 +36,13 @@ def _make_agent_specs(topdown_rgb):
     if topdown_rgb == "rgb":
         rgb = RGB()
     elif topdown_rgb == "false":
-        rgb = False
+        gb = False
```
Adaickalavan, line 39:
```suggestion
        rgb = False
```

baselines/driving_smarts/example.py
```diff
@@ -0,0 +1,25 @@
...
+
+observation, info = env.reset(seed=42)
+while not info.get("reached_episode_limit"):
+    # Generally [x-coordinate, y-coordinate, heading]
+    action = env.action_space.sample()
+    observation, reward, terminated, truncated, info = env.step(action)
+    env.observation_space.contains(observation)
+
+    if terminated or truncated:
+        observation, info = env.reset()
```
Adaickalavan, line 22: Can we avoid introducing `EpisodeLimit` env wrapper by directly counting the number of `env.reset()` called?

smarts/env/gymnasium/wrappers/episode_limit.py
```diff
@@ -0,0 +1,42 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+from typing import Any, Dict, Optional
+
+import gymnasium as gym
+
+
+class EpisodeLimit(gym.Wrapper):
+    """Wraps an environment to provide an ending after `episode_limit` resets signalled by `info["reached_episode_limit"]`."""
```
Adaickalavan, line 28: Can we avoid introducing `EpisodeLimit` env wrapper as most RL libraries use step counts to break during training instead of episode counts? For breaks depending on step count, our smarts env should be compatible with gymanisum such that the [`TimeLimit`](https://gymnasium.farama.org/api/wrappers/misc_wrappers/#gymnasium.wrappers.TimeLimit) could be used. For breaks depending on episode count, the user could implement themselves by counting the number of `env.reset()` calls made.

baselines/driving_smarts/benchmark.py
```diff
@@ -0,0 +1,140 @@
...
+                logging.error("Policy robustness failed.")
+                # env.mark_policy_failure()
+                terminated, truncated = True, True
+            else:
+                observation, reward, terminated, truncated, info = env.step(action)
+            if terminated or truncated:
+                observation, info = env.reset()
+    finally:
+        score = env.score()
+        env.close()
```
Adaickalavan, line 54: Our current `metrics` is designed by assuming that an episode ends when the agent reaches goal, collides, goes off road, or reaches max episode steps. The fairness or consistency of scores computed is not carefully brainstormed in other episode termination cases, such as policy failure.

Here, we might need better handling method. Perhaps policy failure is simply a fatal user error.

baselines/driving_smarts/benchmark.py
```diff
@@ -0,0 +1,140 @@
...
+            try:
+                # action: [global x-coordinate, global y-coordinate]
+                action = {
+                    agent_id: agent.act(obs) for agent_id, obs in observation.items()
+                }
+                # assert env.action_space.contains(action)
+            except Exception:
+                logging.error("Policy robustness failed.")
+                # env.mark_policy_failure()
+                terminated, truncated = True, True
```
Adaickalavan, line 47: If an episode ends due to error, then I guess:
`terminated = False`
`truncated = True`

smarts/core/road_map.py
Gamenot: The issue here was that the children were expected to have it so I had to require this to be implemented or any further implementation will break. Perhaps this should only be necessary in the `RouteWithCache` class rather than the base `RoadMap.Route`.

Gamenot: On second look, it seems that all implementations of `RoadMap` assume this method exists on `Route`.

baselines/driving_smarts/benchmark.py
```diff
@@ -0,0 +1,140 @@
...
+    for name, score in task_iterator(
+        env_configs=env_configs, benchmark_config=benchmark_config
+    ):
+        total_score = sum_scores(total_score, score)
+        print(f"Scoring {name}...")
+
+    print("Evaluation complete...")
+    print()
+    print("`Driving SMARTS V0` result:")
+    print("\n".join(f"- {k}: {v}" for k, v in total_score.items()))
```
Adaickalavan, line 115: The individual value for each key in the `total_score` should be averaged over the number of environments tested to ensure scores remain within [0, 1] range.

baselines/driving_smarts/benchmark.py
```diff
@@ -0,0 +1,140 @@
...
+    #     env_configs[f"bubble_env_{seed}"] = partial(
+    #         env=config["bubble_env"]["env"],
+    #         scenario=seed,
+    #         shared_params=config["shared_env_params"],
+    #     )
+    # TODO MTA: naturalistic
+    total_score = {}
+
+    def sum_scores(s, o):
+        return {k: s.get(k, 0) + o.get(k, 0) for k in set(o)}
```
Adaickalavan, line 104: Why do we use a default value of `0` and what is the impact on the score's fairness?

baselines/driving_smarts/benchmark.py
```diff
@@ -0,0 +1,140 @@
+# smarts/benchmark/driving_smarts_competition_benchmark.py
+import argparse
+import logging
+from functools import partial
+
+import gymnasium as gym
+import ray
+
+from smarts.benchmark import auto_install
+from smarts.benchmark.driving_smarts import DEFAULT_CONFIG, load_config
```
Adaickalavan, line 10: Could we just directly import `from smarts.core.utils.resources import load_yaml_config` instead of introducing `load_config`?

smarts/benchmark/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+from typing import Dict
+
+
+def auto_install(config: Dict[str, str]):
+    """Install dependencies as specified by the configuration given."""
+    pass
```
Adaickalavan, line 27: Should we populate `auto_install` with some code?

baselines/driving_smarts/benchmark.py
```diff
@@ -0,0 +1,140 @@
+# smarts/benchmark/driving_smarts_competition_benchmark.py
```
Adaickalavan, line 1: Perhaps change comment as file does not exist.

Makefile
```diff
@@ -140,7 +140,7 @@ clean:
 .PHONY: format
 format:
 	echo "isort, version `isort --version-number`"
-	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/marl_benchmark/ ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./zoo
+	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines/marl_benchmark/ ./baselines/driving_smarts ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./zoo
```
Adaickalavan, line 143: Will this work?
```suggestion
	isort -m VERTICAL_HANGING_INDENT --skip-gitignore --ac --tc --profile black ./baselines ./cli ./envision ./examples/ ./utils/ ./scenarios/ ./smarts ./zoo
```

baselines/driving_smarts/benchmark.py
Adaickalavan, line 1: Many files named `benchmark`, thus perhaps simply name this file `SMARTS/baselines/driving_smarts/run.py` .

baselines/driving_smarts/demo_observation_space.py
```diff
@@ -0,0 +1,17 @@
...
+    sumo_headless=False,
+)
+
+print("Action space:")
+print(env.action_space)
+print("")
+print("Observation space:")
+print(env.observation_space)
+
+env.close()
```
Adaickalavan, line 17: Is this file to be removed?

baselines/driving_smarts/example.py
```diff
@@ -0,0 +1,25 @@
+import gymnasium as gym
```
Adaickalavan, line 1: Suggested to add a CI test case for the `SMARTS/baselines/driving_smarts/benchmark.py`. Contents of this file could be added to the test case.

Adaickalavan, line 1: Suggested to add a CI test case for the `SMARTS/baselines/driving_smarts/benchmark.py`. 

Contents of this file could be moved into the new test case.

baselines/driving_smarts/benchmark.py
Adaickalavan, line 1: Suggested to add a CI test case for the `SMARTS/baselines/driving_smarts/benchmark.py`.

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+        dones["__all__"] = self._dones_registered >= len(self._agent_interfaces)
+
+        assert all("score" in v for v in infos.values())
+        return (
+            self.observations_formatter.format(observations),
+            sum(r for r in rewards.values()),
+            dones["__all__"],
+            dones["__all__"],
+            infos,
+        )
```
Adaickalavan, line 226: Following RLlib, extending `terminated` and `truncated` to multi-agent case results in a dictionary.

https://github.com/ray-project/ray/blob/d909ac2b5fe5afe65feb192b49aca53abe962cee/rllib/env/multi_agent_env.py#L90-L132

smarts/env/multi_scenario_env.py
```diff
@@ -344,6 +354,8 @@ def step(
 
         limited_actions: Dict[str, np.ndarray] = {}
         for agent_name, agent_action in action.items():
+            if not agent_name in self._prev_obs:
+                continue
```
Adaickalavan, line 358: What is the purpose of these two lines?

smarts/env/hiway_env.py
```diff
@@ -237,7 +282,11 @@ def step(
...
 
         infos = {
-            agent_id: {"score": value, "env_obs": observations[agent_id]}
+            agent_id: {
+                "score": value,
+                "env_obs": observations[agent_id],
+                "done": dones[agent_id],
+            }
             for agent_id, value in extras["scores"].items()
         }
```
Adaickalavan, line 291: Why are observation and done duplicated, and returned again in `infos`? Can they be removed from `infos`?

smarts/env/hiway_env.py
```diff
@@ -98,6 +100,8 @@ def __init__(
             zoo_addrs (Optional[str], optional): List of (ip, port) tuples of
                 zoo server, used to instantiate remote social agents. Defaults
                 to None.
+            agent_specs (Dict[str, AgentInterface]): Interfaces of the agents
```
Adaickalavan, line 103:
```suggestion
            agent_interfaces (Dict[str, AgentInterface]): Interfaces of the agents
```

smarts/env/gymnasium/utils/observation_conversion.py
```diff
@@ -0,0 +1,1017 @@
...
+
+    def active(self, agent_interface: AgentInterface) -> bool:
+        return True
+
+    @property
+    def name(self):
+        return "active"
+
+
+class ObservationSpaceFormat(DictSpaceFormat):
```
Adaickalavan, line 958: Preferably, we should strive to avoid multi-tiered inheritance. Numerous inheritance makes maintaining and refactoring the code harder in the long run. 

Here, we have: 
`BaseSpaceFormat` -> `ConfigurableSpaceFormat` -> `DictSpaceFormat` -> `ObservationSpaceFormat`

smarts/env/gymnasium/utils/observation_conversion.py
```diff
@@ -0,0 +1,1017 @@
...
+_WAYPOINT_SHP = (4, 20)
+_SIGNALS_SHP = (3,)
+_POSITION_SHP = (3,)
+_WAYPOINT_NAME_LIMIT = 50
+_TEXT_PAD_CHAR = " "
+_WAYPOINT_CHAR_SET = frozenset(
+    "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_=+.,;\"' "
+)
+
+## TODO MTA: use constants instead these classes
```
Adaickalavan, line 44: Using constants here sounds better.

smarts/env/hiway_env.py
Gamenot: Right, done for a `gym` environment should relate to the `gym` environment itself rather than the agents in it. If you notice a bit further down I now only return a `bool` rather than the original dictionary.

I remember that the observations here are duplicated as the raw observations in the case that there are wrappers applied over the environment. In my memory it seems to have been for logging purposes.

smarts/env/multi_scenario_env.py
Gamenot: There were cases where the agent was not in the previous observation. There was no way to then constrain the agent's action because there needs to be at least one history frame.

smarts/env/wrappers/format_obs.py
Gamenot: It was a template type that I had reverted.

smarts/env/gymnasium/hiway_env_v1.py
Gamenot: Right, but this is a gym environment rather than a `MultiAgentEnv` or `pettingzoo` environment. If we are going to do that we should first create a multi-agent interface or inherit from one of the two aforementioned environment types.

Since this is `gym` it should return gym environment level values rather than agent level values.

baselines/driving_smarts/example.py
Gamenot: Yes. This was only for demonstration.

baselines/driving_smarts/demo_observation_space.py
Gamenot: Yes. This was only for demonstration.

baselines/driving_smarts/benchmark.py
Gamenot: I moved it to `smarts/benchmark/entrypoints/benchmark_runner_v0.py` so that `scl benchmark` can use it.

Gamenot: It is removed.

baselines/driving_smarts/benchmark.py
Gamenot: I made this abstraction because I saw the potential need to modify the configuration during loading that would be beyond just loading a `.yaml`.

baselines/driving_smarts/benchmark.py
Gamenot: I have done this now.

baselines/driving_smarts/benchmark.py
Gamenot: Should all keys already exist by the time this is run?

smarts/zoo/__init__.py
```diff
@@ -17,3 +17,32 @@
...
+def entry_point(speed=10, **kwargs):
+    return AgentSpec(
+        AgentInterface(
+            action=ActionSpaceType.RelativeTargetPose,
+        ),
+        agent_builder=TestAgent,
+    )
+
+
+register("competition-baseline-v0", entry_point)
```
Adaickalavan, line 48: + Too many use of `baseline` and `benchmark` phrases in numerous names. Perhaps, rename this simply as a `random-agent`.
+ Do we add agents to `SMARTS/smarts/zoo` or `SMARTS/zoo`?

smarts/env/hiway_env.py
```diff
@@ -110,22 +114,46 @@ def __init__(
...
         if not fixed_timestep_sec:
             fixed_timestep_sec = timestep_sec or 0.1
 
-        self._agent_specs = agent_specs
+        def assert_type_and_return(v, types, target):
+            if not isinstance(v, types):
+                raise TypeError(
+                    f"{target}must be supplied with only {target} in values."
+                )
+            return v
```
Adaickalavan, line 130: Is this type check needed or can it be removed?

smarts/env/gymnasium/wrappers/metrics.py
```diff
@@ -296,6 +288,28 @@ def check_intrfc(agent_intrfc: AgentInterface):
             )
 
 
+class CompetitionMetrics(gym.Wrapper):
```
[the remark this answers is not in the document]
Gamenot: I am slightly concerned about versioning here. I know that the metrics should be possible to be used in other circumstances but they are very specific to the competition. As you mentioned later on they do not handle well for cases where an agent terminates for reasons other than those that are intended in the competition.

smarts/env/gymnasium/driving_smarts_competition_env.py
Gamenot: I will be cleaning up the docstrings tomorrow.

baselines/driving_smarts/benchmark.py
Adaickalavan: + `env.score()` should give values for all keys.
+ so all keys should be available in `o`.
+ keys will be missing in `s` in the very first call since `total_score={}`.
+ to avoid this, we can define a namedtuple for score in `metrics.py`.
+ then initiate total_score = empty_score_namedtuple

smarts/env/gymnasium/hiway_env_v1.py
Adaickalavan: The problem:
+ Gym and Gymnasium are only for single-agent.
+ Previously, we extended upon Gym to create a multi-agent SMARTS env `HiWayEnv`.
+ Previously, our multi-agent extension was in line with the multi-agent interface of other libraries such as RLlib.
+ RLlib also extends upon Gymnasium to create their multi-agent env `MultiAgentEnv`.
+ With the current changes, we have essentially extended upon Gymnasium to create a new multi-agent interface which is not in line with any other multi-agent libraries.
+ This is very likely to create more confusion.

smarts/env/multi_scenario_env.py
Adaickalavan: I think actions were only expected for agents with observation. These two lines here might allow errors to slip by unnoticed. Anyway, I think we should simply remove this older `multi_scenario_env.py` since we now have `RelativeTargetPose` action space.

docs/sim/agent.rst
Gamenot: I have removed `TargetPoseFixedTime` since it is satisfied by just requiring `0.1` through asserting on the action space.

Gamenot: I added `RelativeTargetPose`'s description.

baselines/driving_smarts/benchmark.py
Gamenot: Right. Or I could record all of them and sum them all at the end. I will just add a breakdown.

Gamenot: I am seeing negative completion sometimes. I assume that is because the agent is going backwards?

smarts/benchmark/__init__.py
Gamenot: I intend to fill this out shortly, the only dependencies appear to be `SMARTS` and `ray` for now.

smarts/core/agents_provider.py
Gamenot: This was to bring it in line with the agent interface which had deprecated `action_space` for `action`.

smarts/env/gymnasium/wrappers/metrics.py
```diff
@@ -120,16 +110,16 @@ def step(self, action: Dict[str, Any]):
             # Update stored costs.
             self._records[self._scen_name][agent_name].record.costs = costs
 
-            if dones[agent_name]:
+            if info[agent_name]["done"]:
```
Adaickalavan, line 123: If `terminated` or `truncated` were dictionaries, they could have been used here instead of `info[agent_name]["done"]`.

smarts/env/gymnasium/wrappers/metrics.py
```diff
@@ -120,16 +110,16 @@ def step(self, action: Dict[str, Any]):
...
                 self._done_agents.add(agent_name)
                 if not (
-                    agent_obs.events.reached_goal
-                    or len(agent_obs.events.collisions) > 0
-                    or agent_obs.events.off_road
-                    or agent_obs.events.reached_max_episode_steps
+                    agent_obs["events"]["reached_goal"]
+                    or agent_obs["events"]["collisions"]
```
Adaickalavan, line 127: What will be the output of `agent_obs["events"]["collisions"]` ?

smarts/env/gymnasium/wrappers/metrics.py
```diff
@@ -100,12 +82,20 @@ def __init__(self, env: gym.Env):
...
+        obs, _, terminated, truncated, info = result
 
         # Only count steps in which an ego agent is present.
         if len(obs) == 0:
-            return obs, rewards, dones, info
+            return result
 
+        done = False
+        if terminated or truncated:
+            done = True
```
Adaickalavan, line 106:
```suggestion
        done = terminated or truncated
```

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+        self._dones_registered = 0
+        observations = self._smarts.reset(scenario)
+        info = {}
+
+        if self._env_renderer is not None:
+            self._env_renderer.reset(observations)
+
+        if seed is not None:
+            smarts_seed(seed)
+        return self.observations_formatter.format(observations), info
```
Adaickalavan, line 269: The (i) contents (with all options enabled), (ii) types, and (iii) structure, of the observation returned by the env (i.e. `self.observations_formatter.format(observations)`) should desirably be defined and documented in a manner easy to refer for users.

smarts/env/gymnasium/wrappers/metric/completion.py
```diff
@@ -122,16 +131,16 @@ def _dist_remainder():
...
-        if obs.events.reached_goal:
+        if obs["events"]["reached_goal"]:
             dist = 0
         else:
-            cur_pos = Point(*obs.ego_vehicle_state.position)
+            cur_pos = Point(*obs["ego_vehicle_state"]["position"])
             # pytype: disable=attribute-error
-            goal_pos = obs.ego_vehicle_state.mission.goal.position
+            goal_pos = Point(*obs["mission"]["goal_position"])
             # pytype: enable=attribute-error
```
Adaickalavan, line 144: Probably no need to disable attribute error anymore (i.e., `# pytype: disable=attribute-error`).

smarts/benchmark/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+        ) from err
+
+    benchmark = _benchmark_at_version(target, benchmark_version)
+
+    module, _, name = benchmark["entrypoint"].rpartition(".")
+    entrypoint = _get_entrypoint(module, name)
+    entrypoint(**benchmark.get("params", {}), agent_config=agent_config)
+
+
+def list_bechmarks(benchmark_listing):
```
saulfield, line 80:
```suggestion
def list_benchmarks(benchmark_listing):
```

smarts/__init__.py
```diff
@@ -19,7 +19,8 @@
 # THE SOFTWARE.
 
 
-import pkg_resources
+# import pkg_resources
 
 # The full version, including alpha/beta/rc tags
-VERSION = pkg_resources.get_distribution("smarts").version
+# VERSION = pkg_resources.get_distribution("smarts").version
+VERSION = "1.0.0"
```
saulfield, line 26: Is this meant to be left in?

smarts/core/plan.py
```diff
@@ -64,7 +64,7 @@ def from_pose(cls, pose: Pose):
         )
 
 
-@dataclass(unsafe_hash=True)
+@dataclass(frozen=True, unsafe_hash=True)
```
saulfield, line 67: Why was this needed? I remember we removed `frozen=True` because it broke when running with Python 3.8

smarts/env/gymnasium/wrappers/metric/completion.py
```diff
@@ -113,7 +113,16 @@ def _get_dist(start: Point, end: Point) -> float:
...
+        else:
+            raise
+    except CompletionError:
+        dist_tot = 1e10
+        logger.info(
+            "completion.get dist(): Did not find a route from "
+            "%s to %s, because too far off road. Score set to minimum.",
+            point_a,
+            point_b,
+        )
```
Adaickalavan, line 125: Previously, after considering the following,
+ Vehicle might not be more than 5 meters off-road from nearest road as 150km/h is 4.2m in 0.1s.
+ All dead-end roads should be two-way roads, since the cars entering them have to exit somehow.

I thought the first distance computation (i.e., dist_tot = _get_dist(point_a, point_b)) is sufficient by itself (perhaps with a slightly increased radius=7), even without the need for the secondary reversed route computation. In short, realistically we should always be able to compute the remainder-distance cost without errors.

I think the error due to cars appearing spuriosly far off-road should either be avoided or handled elsewhere, as handling the repercussions here affects the realism of our cost computation code. Moreover, allowing this far off-road vehicle to propagate inside our cost computation code makes the values of other costs computed questionable.

smarts/__init__.py
Gamenot: No, I was testing and I think this found its way into the changes. The main cause was that when you import `pkg_resources` it raises an error if you do not have the exactly compatible packages installed. This was causing issues for me when I was testing out packages.

Gamenot: If possible I would get rid of it. There is a bit of a problem in that if we put the version in `smarts` then it fails to load but if we put it in here then `smarts` gets locked into requiring all packages to always be within the versions in the requirements.

smarts/core/plan.py
Gamenot: It seems to work without issues somehow. I know the original issue was that you could not have a non-`frozen` dataclass inherit a `frozen` data class

smarts/env/gymnasium/hiway_env_v1.py
Gamenot: That was part of the reason for applying it to the action space. In that case the users should just be able to call `action_space` to see the format of the data.

I will add documentation on Thursday/Friday to describe the observations in full.

smarts/env/gymnasium/__init__.py
Gamenot: Right, I will meet you halfway. The `observation_space` and `action_space` will still remain the the maximum action. However, `HiWayV1` will get configuration to toggle between multi-agent compliant and strict gym compliant.

smarts/env/gymnasium/wrappers/metric/completion.py
```diff
@@ -122,17 +131,15 @@ def _dist_remainder():
...
-        if obs.events.reached_goal:
+        if obs["events"]["reached_goal"]:
             dist = 0
         else:
-            cur_pos = Point(*obs.ego_vehicle_state.position)
-            # pytype: disable=attribute-error
-            goal_pos = obs.ego_vehicle_state.mission.goal.position
-            # pytype: enable=attribute-error
+            cur_pos = Point(*obs["ego_vehicle_state"]["position"])
+            goal_pos = Point(*obs["mission"]["goal_position"])
```
Adaickalavan, line 142: We have moved `mission` to `obs["mission"]`. So should we also similarly move `obs.ego_vehicle_state.mission` to `obs.mission` in `Observation` class in `smarts/core/observations.py`?

smarts/env/gymnasium/utils/observation_conversion.py
```diff
@@ -0,0 +1,1050 @@
...
+    def active(self, agent_interface: AgentInterface) -> bool:
+        raise NotImplementedError()
+
+    @property
+    def name(self):
+        raise NotImplementedError()
+
+    @property
+    def space(self):
+        return gym.spaces.Dict({s.name: s.space for s in self._spaces})
```
Adaickalavan, line 181: Is the space composition executed multiple times when called repeatedly and, if yes, can it be prestored?

smarts/env/gymnasium/utils/observation_conversion.py
```diff
@@ -0,0 +1,1050 @@
...
+        ego_vehicle_obs = obs.ego_vehicle_state
+        if hasattr(ego_vehicle_obs.mission.goal, "position"):
+            goal_pos = np.array(ego_vehicle_obs.mission.goal.position, dtype=np.float64)
+        else:
+            goal_pos = np.zeros((3,), dtype=np.float64)
+
+        return {"goal_position": goal_pos}
+
+    def active(self, agent_interface: AgentInterface) -> bool:
+        return True
```
Adaickalavan, line 672: We have `EndlessGoal`, `PositionalGoal`, and `TraverseGoal`, but only the `PositionalGoal` is visible (through its desired final position) in the returned observation. Should boolean return value of `active()` be dependent on whether the goal is a `PositionalGoal`?

smarts/env/gymnasium/utils/observation_conversion.py
```diff
@@ -0,0 +1,1050 @@
...
+                    low=-1e10, high=1e10, shape=_SIGNALS_SHP + (2,), dtype=np.float64
+                ),
+                "last_changed": gym.spaces.Box(
+                    low=0, high=1e10, shape=_SIGNALS_SHP, dtype=np.float32
+                ),
+            }
+        )
+
+
+class EnabledSpaceFormat(BaseSpaceFormat):
```
Adaickalavan, line 975: What is the purpose of `EnabledSpaceFormat` ?

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+            visualization_client = visualization_client_builder(
+                headless=headless,
+                sim_name=sim_name,
+            )
+
+        self._env_renderer = None
+
+        visdom_client = None
+        if visdom:
+            visdom_client = VisdomClient()
```
Adaickalavan, line 130:
```suggestion
        visdom_client = VisdomClient() if visdom else None
```

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+        reset this environment's state for the next episode.
+        .. versionchanged:: 0.26
+            The Step API was changed removing ``done`` in favor of ``terminated`` and ``truncated`` to make it clearer
+            to users when the environment had terminated or truncated which is critical for reinforcement learning
+            bootstrapping algorithms.
+        Args:
+            action (ActType): an action provided by the agent to update the environment state.
+        Returns:
+            observation (ObsType): An element of the environment's :attr:`observation_space` as the next observation due to the agent actions.
+                An example is a numpy array containing the positions and velocities of the pole in CartPole.
```
Adaickalavan, line 176:
```suggestion
        Returns:
            observation (ObsType): An element of the environment's :attr:`observation_space` as the next observation due to the agent actions.
```

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+            reward (SupportsFloat): The reward as a result of taking the action.
+            terminated (bool): Whether the agent reaches the terminal state (as defined under the MDP of the task)
+                which can be positive or negative. An example is reaching the goal state or moving into the lava from
+                the Sutton and Barton, Gridworld. If true, the user needs to call :meth:`reset`.
```
Adaickalavan, line 180:
```suggestion
            terminated (bool): Whether the agent reaches the terminal state (as defined under the MDP of the task)
                which can be positive or negative. An example is reaching the goal state. If true, the user needs to call :meth:`reset`.
```

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+            truncated (bool): Whether the truncation condition outside the scope of the MDP is satisfied.
+                Typically, this is a timelimit, but could also be used to indicate an agent physically going out of bounds.
+                Can be used to end the episode prematurely before a terminal state is reached.
+                If true, the user needs to call :meth:`reset`.
+            info (dict): Contains auxiliary diagnostic information (helpful for debugging, learning, and logging).
+                This might, for instance, contain: metrics that describe the agent's performance state, variables that are
+                hidden from observations, or individual reward terms that are combined to produce the total reward.
+                In OpenAI Gym <v26, it contains "TimeLimit.truncated" to distinguish truncation and termination,
+                however this is deprecated in favour of returning terminated and truncated variables.
```
Adaickalavan, line 189:
```suggestion
            info (dict): Contains auxiliary diagnostic information (helpful for debugging, learning, and logging).
                This might, for instance, contain: metrics that describe the agent's performance state, variables that are
                hidden from observations, or individual reward terms that are combined to produce the total reward.
```

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+            done (bool): (Deprecated) A boolean value for if the episode has ended, in which case further :meth:`step` calls will
+                return undefined results. This was removed in OpenAI Gym v26 in favor of terminated and truncated attributes.
+                A done signal may be emitted for different reasons: Maybe the task underlying the environment was solved successfully,
+                a certain timelimit was exceeded, or the physics simulation has entered an invalid state.
```
Adaickalavan, line 193:
```suggestion
```

examples/control/single_agent_lw.py
```diff
@@ -0,0 +1,63 @@
+import sys
```
Adaickalavan, line 1: Instead of naming examples `single_agent_lw.py` and `multi_agent.py`, we could consider giving them more descriptive names, such as `hiwayenvv1_lane_follower.py`.

smarts/env/gymnasium/wrappers/metrics.py
Gamenot: This has been set to configure the dictionaries version of done.

smarts/env/gymnasium/wrappers/metrics.py
Gamenot: It would be a boolean. Truthfully this should be called "collided", or be a list of agent ids.

smarts/env/gymnasium/wrappers/metrics.py
Gamenot: It has been changed to include both.

smarts/env/gymnasium/wrappers/metrics.py
```diff
@@ -100,12 +82,20 @@ def __init__(self, env: gym.Env):
 
     def step(self, action: Dict[str, Any]):
         """Steps the environment by one step."""
-        obs, rewards, dones, info = super().step(action)
+        result = super().step(action)
+
+        terminated, truncated = False, False
```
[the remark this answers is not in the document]
Gamenot: This was a leftover from when I was attempting to support both `gym` and `gymnasium` in the same class.

smarts/env/hiway_env.py
Gamenot: It is not really needed but was used to ensure that if someone is using `agent_specs` that they are not using `None` in the optional `AgentSpec.agent_interface`. I suppose we are deprecating `agent_spec` anyway so I will remove this check.

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -137,25 +139,37 @@ def multi_scenario_v0_env(
     env_specs = _get_env_specs(scenario)
     sstudio.build_scenario(scenario=[env_specs["scenario"]])
```
Adaickalavan, line 140: `sstudio` appears undefined.

smarts/__init__.py
Gamenot: I have reverted this.

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -0,0 +1,390 @@
...
+        .. versionchanged:: v0.25
+            The ``return_info`` parameter was removed and now info is expected to be returned.
+        Args:
+            seed (optional int): The seed that is used to initialize the environment's PRNG (`np_random`).
+                If the environment does not already have a PRNG and ``seed=None`` (the default option) is passed,
+                a seed will be chosen from some source of entropy (e.g. timestamp or /dev/urandom).
+                However, if the environment already has a PRNG and ``seed=None`` is passed, the PRNG will *not* be reset.
+                If you pass an integer, the PRNG will be reset even if it already exists.
+                Usually, you want to pass an integer *right after the environment has been initialized and then never again*.
+                Please refer to the minimal example above to see this paradigm in action.
```
Adaickalavan, line 279:
```suggestion
```

smarts/env/gymnasium/wrappers/metric/completion.py
Gamenot: I would say primarily that it is unacceptable for the metrics to cause the program to crash.

It is also not that vehicles appear far off-road. Vehicles can drive there naturally if off-road is not set to cause done, especially when I revert the distance check for finding the route back to 5 meters. 

There may not be a realistic way of calculating the remainder-distance cost when off-road without introducing a logical error because the metric has already assumed that the agent is following the road graph. We can always increase the radius in which the vehicle can find a route from off-road but that adds in performance inefficiency and still does not solve the issue that the agent has escaped the graph (and potentially attempted to cheat the graph.)

To handle a failure of the metrics from a higher level, the caller would need to know about the radius restrictions that the metric has (e.g. 5 meters).

For now I have added a warning specifically mentioning this.

smarts/env/gymnasium/utils/observation_conversion.py
Gamenot: Goal is determined from the scenario, so it is not possible to determine it from the interface.

smarts/env/gymnasium/utils/observation_conversion.py
Gamenot: The intention here is to determine if the agent is available in the case that the observations are returned as a boid system (i.e. `ObservationOptions.full`.)

smarts/env/gymnasium/utils/observation_conversion.py
```diff
@@ -0,0 +1,1050 @@
...
+        return out_obs
+
+    @property
+    def space(self):
+        return gym.spaces.Dict(
+            {
+                agent_id: space_format.space
+                for agent_id, space_format in self.space_formats.items()
+            }
+        )
```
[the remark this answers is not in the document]
Gamenot: I have set it to be a cached property.

examples/control/single_agent_lw.py
Gamenot: This has been renamed.

On another note, there is a major problem now with determining which of `terminated` and `truncated` to place `"__all__"`.

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -269,6 +283,34 @@ def _get_env_specs(scenario: str):
         raise Exception(f"Unknown scenario {scenario}.")
 
 
+def resolve_agent_action_space(agent_interface: AgentInterface):
+    if agent_interface.action == ActionSpaceType.RelativeTargetPose:
+        return gym.spaces.Box(
+            low=np.array([-28, -28, -np.pi]),
+            high=np.array([28, 28, np.pi]),
```
saulfield, line 290: Where did the number 28 come from?

smarts/env/gymnasium/wrappers/metric/completion.py
Gamenot: That was not intended, I will revert it to be under "ego_vehicle_state".

smarts/env/gymnasium/utils/observation_conversion.py
Gamenot: Duplicate. https://github.com/huawei-noah/SMARTS/pull/1793#discussion_r1081643799

examples/control/single_agent_lw.py
Adaickalavan: There are instances in RLlib code where `"__all__"` is present in both `terminated` and `truncated`.

But I guess for the time being, we could put the `"__all__"` in `terminated` as required here:
https://github.com/ray-project/ray/blob/d909ac2b5fe5afe65feb192b49aca53abe962cee/rllib/env/multi_agent_env.py#L104-L105

smarts/env/gymnasium/wrappers/metric/completion.py
Adaickalavan: + `Metrics` verifies the environment suitability during instantiation. 
+ Here, `Metrics` cannot be used on an environment which has an agent with `DoneCriteria(off_road=False)`. 
+ On another comment, since environment suitability is checked by `Metrics`, I thought we don't have to call it `CompetitionMetrics`.

https://github.com/huawei-noah/SMARTS/blob/5d2dff7e0a91454cfd4722481ebfe5d99a298b29/smarts/env/gymnasium/wrappers/metrics.py#L276-L307

Adaickalavan:
https://github.com/huawei-noah/SMARTS/blob/684e6a1f255b0ea75adcf2dae9c9a2c2abb73eb6/smarts/env/gymnasium/driving_smarts_competition_env.py#L115-L119

This could be deleted as the `_InfoScore` wrapper was removed. It was previously only used to assist in the use of Stable Baselines3 library.

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -23,12 +23,15 @@
 import math
 import os
 import pathlib
-from typing import Any, Dict, Optional, Tuple
+import typing
+from functools import partial
+from typing import Any, Dict, Tuple
```
Adaickalavan, line 28:
```suggestion
from functools import partial
from typing import Any, Dict, Optional, Tuple
```

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -65,17 +70,16 @@ def multi_scenario_v0_env(
     Action space for each agent:
     .. note::
 
-        A ``smarts.core.controllers.ActionSpaceType.TargetPose``, which is a
-        sequence of [x-coordinate, y-coordinate, heading, and time-delta].
-        Use time-delta = 0.1 .
+        A ``smarts.core.controllers.ActionSpaceType.RelativeTargetPose``, which is a
+        sequence of [x-coordinate, y-coordinate, heading].
```
Adaickalavan, line 74:
```suggestion
        A ``smarts.core.controllers.ActionSpaceType.RelativeTargetPose``, which is a
        sequence of [Δx, Δy, heading].
```

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -39,22 +42,24 @@
...
 )
 from smarts.core.controllers import ActionSpaceType
-from smarts.env.hiway_env import HiWayEnv
-from smarts.zoo.agent_spec import AgentSpec
+from smarts.env.gymnasium.hiway_env_v1 import HiWayEnvV1, SumoOptions
 
 logger = logging.getLogger(__file__)
 logger.setLevel(logging.WARNING)
 
+SUPPORTED_ACTION_TYPES = (ActionSpaceType.RelativeTargetPose,)
```
Adaickalavan, line 50:
```suggestion
SUPPORTED_ACTION_TYPES = (ActionSpaceType.RelativeTargetPose, ActionSpaceType.TargetPose)
```

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -269,6 +283,34 @@ def _get_env_specs(scenario: str):
...
+        {
+            a_id: resolve_agent_action_space(a_inter)
+            for a_id, a_inter in agent_interfaces.items()
+        }
+    )
+
+
 def resolve_agent_interface(
     img_meters: int = 64, img_pixels: int = 256, action_space="TargetPose", **kwargs
 ):
```
Adaickalavan, line 316:
```suggestion
def resolve_agent_interface(
    img_meters: int = 64, img_pixels: int = 256, action_space="RelativeTargetPose", **kwargs
):
```

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -269,6 +283,34 @@ def _get_env_specs(scenario: str):
...
+            high=np.array([1e10, 1e10, np.pi, 0.1]),
+            dtype=np.float32,
+        )
+
+    assert (
+        agent_interface.action in SUPPORTED_ACTION_TYPES
+    ), f"Unsupported action type `{agent_interface.action}` not in supported actions `{SUPPORTED_ACTION_TYPES}`"
+
+
+def resolve_env_action_space(agent_interfaces: typing.Dict[str, AgentInterface]):
```
Adaickalavan, line 305:
```suggestion
def resolve_env_action_space(agent_interfaces: Dict[str, AgentInterface]):
```

baselines/driving_smarts/benchmark.py
Adaickalavan: We should check why it is happening. Ideally, `0 <= Score.completion <= 1`.

smarts/env/gymnasium/driving_smarts_competition_env.py
Gamenot: This issue was hidden because the check for supported action spaces came after actually applying the action space based on the given action type.

smarts/env/gymnasium/driving_smarts_competition_env.py
Gamenot: This was a constant based on the maximum safe deceleration... Oddly enough I am not certain why we are using it here since 280m/s is nonsensically fast.

Gamenot: Oh, it was intended to be 2.8m/0.1s.

saulfield: Ok, maybe worth leaving a comment for clarity.

smarts/benchmark/entrypoints/benchmark_runner_v0.py
```diff
@@ -0,0 +1,188 @@
...
+    agent = agent_registry.make_agent(
+        locator=agent_config["locator"],
+        **agent_config["params"],
+    )
+
+    observation, info = env.reset()
+    try:
+        while not info.get("reached_episode_limit"):
+            try:
+                # action: [global x-coordinate, global y-coordinate]
```
Adaickalavan, line 61: Comment needs update?

smarts/env/gymnasium/driving_smarts_competition_env.py
Gamenot: I have declared the value as a commented constant.

baselines/driving_smarts/example.py
Gamenot: I have removed the wrapper because it does not make for an improvement and causes confusion compared to `TimeLimit` as you mentioned elsewhere.

smarts/env/gymnasium/driving_smarts_competition_env.py
Gamenot: Sorry, I had this commented out locally to test faster.

smarts/env/gymnasium/driving_smarts_competition_env.py
```diff
@@ -269,8 +287,40 @@ def _get_env_specs(scenario: str):
...
+def resolve_agent_action_space(agent_interface: AgentInterface):
+    assert (
+        agent_interface.action in SUPPORTED_ACTION_TYPES
+    ), f"Unsupported action type `{agent_interface.action}` not in supported actions `{SUPPORTED_ACTION_TYPES}`"
+
+    if agent_interface.action == ActionSpaceType.RelativeTargetPose:
+        max_speed = MAXIMUM_SPEED_MPS / 0.1  # assumes 0.1 timestep
+        return gym.spaces.Box(
+            low=np.array([-max_speed, -max_speed, -np.pi]),
+            high=np.array([max_speed, max_speed, np.pi]),
```
Adaickalavan, line 299:
```suggestion
        max_dist = MAXIMUM_SPEED_MPS * 0.1  # assumes 0.1 timestep
        return gym.spaces.Box(
            low=np.array([-max_dist, -max_dist, -np.pi]),
            high=np.array([max_dist, max_dist, np.pi]),
```

Gamenot: Right, thanks for spotting that.

smarts/benchmark/__init__.py
Gamenot: SMARTS has to already be installed to run this so I added configuration for other dependencies.

smarts/core/agents_provider.py
Gamenot: Truthfully, I want to rename `ActionSpaceType` to `ActionType` because we are dealing with shaped actions directly without regard to limitations.

smarts/env/gymnasium/utils/observation_conversion.py
Gamenot: I understand the issues with inheritance although it is helped by the fact that `BaseSpaceFormat` is an interface and `ConfigurableSpaceFormat` is interface except for the initializer, and `DictSpaceFormat` is an abstract class. It means that the former two can be removed or replaced very easily without a lot of the troubles involved from inheriting from implemented classes.

Gamenot:
> https://github.com/huawei-noah/SMARTS/blob/71672c84d6cb04ab4d90bf38e549e921d31934b2/smarts/core/agent_manager.py#L246-L248
> 
> 1. Where and how is the above `dones["__sim__"]` used?
> 2. Does the addition of `dones["__sim__"]` affect the computation of `dones["__all__"]` here below?
>    https://github.com/huawei-noah/SMARTS/blob/71672c84d6cb04ab4d90bf38e549e921d31934b2/smarts/env/gymnasium/hiway_env_v1.py#L214-L217

I forgot to reply to this earlier. This occurs uniquely for the SMARTS `gym` environments if the simulation has an issue. If you look just above ["__all__"] every agent is set to `done` so ["__all__"] also becomes `True` at the same time.

smarts/env/gymnasium/wrappers/metric/completion.py
Gamenot: Odd, somehow an agent was able to drive off-road without terminating. I think this is a bug.

baselines/driving_smarts/benchmark.py
Gamenot: I have made it default now that it causes a fatal error.

Gamenot:
I still need to to a bit more documentation but it is late. I will merge this in and release tomorrow.

PR #1795, opened 2023-01-11
test_environment with gymnasium

AisenGinn:
test_env.py is the new environment compatible with SingleAgent wrapper and gymnasium. register_test.py is the test file.


Gamenot:
This is already addressed.

PR #1806, opened 2023-01-25
Automatically supply scenario path to Envision server

### LINKED ISSUE #1625: Avoid manually specifying scenario path to Envision server
+ Currently, we need to manually specify the scenario path when starting the Envision server. 
+ This is not very user friendly. 
+ Consider the case where we would like to run a simulation inside a docker container and visualize it in Envision. Inside the docker container, the command to start the Envision server would be something as follows.
    ```bash
    $ scl envision start -s /opt/.venv/lib/python3.8/site-packages/smarts/scenarios/ & 
    ``` 
    Here, the scenario path supplied to the Envision server should refer to location of the scenario inside the docker container. This path is not intuitive and could be complex.

It would be better if we don't have to manually specify the scenario location to the Envision server. The scenario path should be supplied directly by SMARTS to the Envision server.

**Additional context**
For more context, see issue #1624 and the reply.

Adaickalavan:
+ No need to manually specify the scenario location to the Envision server during setup. 
+ The scenario path is sent automatically by SMARTS to the Envision server during simulation startup phase.
+ Closes #1625 .


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 ## [Unreleased]
 ### Added
 ### Changed
+- Scenario paths need not be supplied manually to Envision server while setup. Scenario paths are automatically sent to Envision server from SMARTS during simulation startup phase.
```
Gamenot, line 14: This is not quite the case.

```suggestion
- Scenario paths is no longer manually supplied to Envision server while setup. Scenario paths are automatically sent to Envision server from SMARTS during simulation startup phase.
```

Gamenot (approved):

CHANGELOG.md
Adaickalavan: I have changed it.

Issue #1735, opened 2022-11-25
Fix the Waymo waypoints issue


Gamenot:
- See https://github.com/huawei-noah/SMARTS/issues/1706

Gamenot:
See #1777

PR #1807, opened 2023-01-25
Fix waypoint observations for offline dataset generation

### LINKED ISSUE #1706: Competition -- Waypoint issue in Waymo scenarios
This is an issue reported by participants, one team is using RL to do high-level decision making of lane-changing/car-following and the agent itself basically moves along the waypoints. The concern is, their solution can perform well in the interactive scenarios, but on the /naturalistic (Waymo) scenarios, they realized that in some scenarios, all waypoints in the observation lead to wrong direction (i.e., far from the end point), which means that their current model couldn't handle such situation if there's no waypoint is leading to the end point.

We might want to look into the Waymo scenarios to verify this issue.

saulfield:
Closes #1706, #1735.

## Context

When generating offline dataset observations with the `traffic_histories_to_observations.py` script, history vehicles do not have routes associated with their missions. This results in the waypoints sensor returning waypoints for the nearest lanes, somewhat erratically (see below -- vehicle positions are in red, waypoints for each position are plotted as lines):

The desired behaviour is for the waypoints to consistently lead to the final position in the recorded trajectory:

## Solution
- When adding sensors to history vehicles, create a mission with a route leading to the final position in the trajectory
- Various changes to get the `traffic_histories_to_observations` script working again

### DIFF
--- a/smarts/core/traffic_history.py
+++ b/smarts/core/traffic_history.py
@@ -131,6 +131,12 @@ class TrafficHistory:
         return ego_id
 
     @lru_cache(maxsize=32)
+    def vehicle_initial_time(self, vehicle_id: str) -> float:
+        """Returns the final time the specified vehicle is seen in the history data."""
+        query = "SELECT MIN(sim_time) FROM Trajectory WHERE vehicle_id = ?"
+        return self._query_val(float, query, params=(vehicle_id,))
+
+    @lru_cache(maxsize=32)
     def vehicle_final_exit_time(self, vehicle_id: str) -> float:
         """Returns the final time the specified vehicle is seen in the history data."""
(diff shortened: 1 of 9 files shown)


Gamenot (approved):

smarts/core/traffic_history.py
```diff
@@ -130,6 +130,12 @@ def ego_vehicle_id(self) -> int:
         ego_id = self._query_val(int, query)
         return ego_id
 
+    @lru_cache(maxsize=32)
+    def vehicle_initial_time(self, vehicle_id: str) -> float:
+        """Returns the final time the specified vehicle is seen in the history data."""
```
Adaickalavan, line 135:
```suggestion
        """Returns the initial time the specified vehicle is seen in the history data."""
```

Adaickalavan (approved):

PR #1808, opened 2023-01-25
Action space formatting

Gamenot:
See `CHANGELOG.md`.

### DIFF
--- a/smarts/env/gymnasium/utils/action_conversion.py
+++ b/smarts/env/gymnasium/utils/action_conversion.py
@@ -254,7 +254,7 @@ class ActionSpacesFormatter:
             out_actions[agent_id] = formatted_action
 
         if self._action_options == ActionOptions.full:
-            assert set(actions) == set(self.space.spaces)
+            assert actions.keys() == self.space.spaces.keys()
 
         return out_actions
 
--- a/smarts/env/gymnasium/utils/action_conversion.py
+++ b/smarts/env/gymnasium/utils/action_conversion.py
@@ -195,7 +195,7 @@ class ActionOptions(IntEnum):
     """Action must map to full action space. Inactive and active agents are included."""
     unformatted = 2
     """Actions are not reformatted or constrained to action space. Actions must directly map to
-    underling SMARTS actions."""
+    underlying SMARTS actions."""
     default = 0
     """Defaults to :attr:`multi_agent`."""
 
--- a/smarts/env/gymnasium/utils/action_conversion.py
+++ b/smarts/env/gymnasium/utils/action_conversion.py
@@ -1,0 +1,183 @@
+import math
+from dataclasses import dataclass, field
+from functools import lru_cache
+from typing import Any, Callable, Dict
+
+import gymnasium as gym
+import numpy as np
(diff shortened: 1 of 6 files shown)


smarts/env/gymnasium/utils/action_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+class ActionOptions(IntEnum):
+    """Defines the options for how the formatting matches the action space."""
+
+    multi_agent = 0
+    """Action must map to partial action space. Only active agents are included."""
+    full = 1
+    """Action must map to full action space. Inactive and active agents are included."""
+    unformatted = 2
+    """Actions are not reformatted or constrained to action space. Actions must directly map to
+    underling SMARTS actions."""
```
saulfield, line 198:
```suggestion
    underlying SMARTS actions."""
```

saulfield (approved):

smarts/env/gymnasium/utils/action_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class ActionSpacesFormatter:
+    """Formats actions to adapt SMARTS to `gym` environment requirements."""
+
+    def __init__(
+        self, agent_interfaces: Dict[str, AgentInterface], action_options: ActionOptions
+    ) -> None:
+        self._agent_interfaces = agent_interfaces
+        self._action_options = action_options
```
Adaickalavan, line 210: + Consider explaining inputs (agent_interfaces, action_options) in the docstring.
+ Helpful VSCode extension for generating docstrings:
    https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

Adaickalavan (approved):

smarts/env/gymnasium/utils/action_conversion.py
```diff
@@ -0,0 +1,254 @@
...
+            agent_interface = self._agent_interfaces[agent_id]
+            format_ = formatting_groups[agent_interface.action]
+            assert format_.space.contains(
+                action
+            ), f"Action {action} does not match space {format_.space}!"
+            formatted_action = format_.formatting_func(action)
+            out_actions[agent_id] = formatted_action
+
+        if self._action_options == ActionOptions.full:
+            assert self.space.contains(out_actions)
```
Adaickalavan, line 237: + If `self._action_options == ActionOptions.full`, all the active and inactive agent action spaces were checked in the preceeding `for` loop.
+ So instead of checking `assert self.space.contains(out_actions)` here, is a simple check whether `actions.keys()` matches `self._agent_interfaces.keys()` be sufficient here?

Gamenot: From what I can see I did not actually check for inactive agents in the former loop. This makes sense because I can only check for actions that exist.

However, as you mention it makes more sense to just do a check on the `agent_ids` to test if the action is fully utilised instead of using `space.contains` because in every case where the `agent_ids` are fully utilised, they would have already been checked in the previous loop.

Gamenot: I am comparing the keys instead.

PR #1809, opened 2023-01-26
Metrics use env obs

Gamenot:
This allows the metrics wrapper to be used over top of other wrappers instead of directly on `hiway-v1` by using `info["env_obs"]`.

This can also be placed over top of a `hiway-v0` using the `gymnasium.wrappers.compatibility.EnvCompatibility` wrapper and then converted back to a `gym` environment using our `smarts.env.gymnasium.wrappers.api_reversion.Api021Reversion` wrapper.

### DIFF
--- a/smarts/env/gymnasium/wrappers/metric/completion.py
+++ b/smarts/env/gymnasium/wrappers/metric/completion.py
@@ -133,16 +133,19 @@ def _dist_remainder():
     step: int = 0
 
     def func(
-        road_map: RoadMap, obs: Dict[str, Any], initial_compl: Completion
+        road_map: RoadMap, obs: Observation, initial_compl: Completion
     ) -> Completion:
         nonlocal mean, step
 
-        if obs["events"]["reached_goal"]:
+        if obs.events.reached_goal:
             dist = 0
         else:
-            cur_pos = Point(*obs["ego_vehicle_state"]["position"])
-            goal_pos = Point(*obs["ego_vehicle_state"]["mission"]["goal_position"])
-            dist = get_dist(road_map=road_map, point_a=cur_pos, point_b=goal_pos)
+            cur_pos = Point(*obs.ego_vehicle_state.position)
+            goal_pos = Point(*getattr(obs.ego_vehicle_state.mission.goal, "position", None))
+            if goal_pos is not None:
+                dist = get_dist(road_map=road_map, point_a=cur_pos, point_b=goal_pos)
+            else:
+                dist = 0
 
         # Cap remainder distance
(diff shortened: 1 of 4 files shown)


smarts/env/gymnasium/wrappers/metric/completion.py
```diff
@@ -133,16 +133,21 @@ def _dist_remainder():
...
-            goal_pos = Point(*obs["ego_vehicle_state"]["mission"]["goal_position"])
-            dist = get_dist(road_map=road_map, point_a=cur_pos, point_b=goal_pos)
+            cur_pos = Point(*obs.ego_vehicle_state.position)
+            goal_pos = Point(
+                *getattr(obs.ego_vehicle_state.mission.goal, "position", None)
+            )
+            if goal_pos is not None:
+                dist = get_dist(road_map=road_map, point_a=cur_pos, point_b=goal_pos)
+            else:
+                dist = 0
```
Adaickalavan, line 150: `goal_pos` cannot be `None` as we check each scenario's suitability, i.e., availability of `PositionalGoal`.

https://github.com/huawei-noah/SMARTS/blob/84637f4c185f2c105d8ea5a7c56a0d8ec13ff5d7/smarts/env/gymnasium/wrappers/metrics.py#L358-L375

Gamenot: I have changed it to assertions. The issue here is that the type requirement is not indicated through the interface.

If the method required a position or a positional goal as part of the arguments that would be possible but this method has no relation to the previous goal constraint.

Adaickalavan (approved):

smarts/env/gymnasium/wrappers/metric/completion.py
```diff
@@ -133,16 +133,20 @@ def _dist_remainder():
...
             dist = 0
         else:
-            cur_pos = Point(*obs["ego_vehicle_state"]["position"])
-            goal_pos = Point(*obs["ego_vehicle_state"]["mission"]["goal_position"])
-            dist = get_dist(road_map=road_map, point_a=cur_pos, point_b=goal_pos)
+            cur_pos = Point(*obs.ego_vehicle_state.position)
+            goal_position = getattr(obs.ego_vehicle_state.mission.goal, "position")
+            assert (
+                goal_position is not None
+            ), f"Ego using `{obs.ego_vehicle_state.id}` cannot use cost func without a positional goal."
```
Adaickalavan, line 147:
```suggestion
            ), f"Ego `{obs.ego_vehicle_state.id}` cannot use cost func without a positional goal."
```

PR #1821, opened 2023-01-30
Fix issue with unformatted metrics

Gamenot:
This fixes an issue where "unformatted" observations could not be used with metrics.

### DIFF
--- a/smarts/env/gymnasium/wrappers/metrics.py
+++ b/smarts/env/gymnasium/wrappers/metrics.py
@@ -109,7 +109,8 @@ class MetricsBase(gym.Wrapper):
                 dones["__all__"] = True
             dones.update({a: d["done"] for a, d in info.items()})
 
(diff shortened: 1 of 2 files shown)


smarts/env/gymnasium/wrappers/metrics.py
```diff
@@ -109,7 +109,8 @@ def step(self, action: Dict[str, Any]):
                 dones["__all__"] = True
             dones.update({a: d["done"] for a, d in info.items()})
 
-        obs = {agent_id: o for agent_id, o in obs.items() if o["active"]}
+        if len(obs) > 0 and isinstance(next(iter(obs.values())), dict):
```
Adaickalavan, line 112: `len(obs)` is checked to be greater than zero on line 101 above.

Gamenot: Right thanks for pointing that out.

PR #1828, opened 2023-02-02
Engine configuration utility

Gamenot:
This adds an engine configuration utility to SMARTS.

- [X] Add configuration utility.
- [x] Add tests.
- [x] Confirm that configuration can be shared between processes.
- [x] Add issue to convert engine constants to use config. https://github.com/huawei-noah/SMARTS/issues/1854
- [x] Update changelog.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -46,6 +46,7 @@ If you use SMARTS in your research, please cite the `paper <https://arxiv.org/ab
    sim/simulator.rst
    sim/scenario_studio.rst
    sim/bubbles.rst
+   sim/configuration.rst
 
 .. toctree::
    :hidden:
@@ -69,7 +70,6 @@ If you use SMARTS in your research, please cite the `paper <https://arxiv.org/ab
    :caption: API
 
    api/modules.rst
-   sim/configuration.rst
 
 .. toctree::
    :hidden:
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -69,6 +69,7 @@ If you use SMARTS in your research, please cite the `paper <https://arxiv.org/ab
    :caption: API
 
    api/modules.rst
+   sim/configuration.rst
 
 .. toctree::
    :hidden:
@@ -88,3 +89,4 @@ If you use SMARTS in your research, please cite the `paper <https://arxiv.org/ab
    resources/diagnostic.rst
    resources/faq.rst
    resources/contributing.rst
+   resources/todo.rst
--- a/docs/sim/configuration.rst
+++ b/docs/sim/configuration.rst
@@ -5,7 +5,7 @@ Configuration
 
 You can change the behavior of the underlying SMARTS engine.
 
-Configuration of the engine can come from several sources. These locations take precidence as noted:
+Configuration of the engine can come from several sources. These locations take precedence as noted:
 
 1. Individual ``SMARTS_`` prefixed environment variables (e.g. ``SMARTS_SENSOR_WORKER_COUNT``)
 2. Local directory engine configuration (./smarts_engine.ini )
--- a/docs/sim/configuration.rst
+++ b/docs/sim/configuration.rst
@@ -1,0 +1,27 @@
+.. _configuration:
+
+Configuration
+=============
+
+You can change the behavior of the underlying SMARTS engine.
+
+Configuration of the engine can come from several sources. These locations take precidence as noted:
+
+1. Individual ``SMARTS_`` prefixed environment variables (e.g. ``SMARTS_SENSOR_WORKER_COUNT``)
+2. Local directory engine configuration (./smarts_engine.ini )
+3. Local user engine configuration, ``~/.smarts/engine.ini``, if local directory configuration is not found.
+4. Global engine configuration, ``/etc/smarts/engine.ini``, if local configuration is not found.
+5. Package default configuration, ``$PYTHON_PATH/smarts/engine.ini``, if global configuration is not found.
+
+Note that configuration files resolve all settings at the first found configuration file (they do not layer.)
+
+Options
+-------
+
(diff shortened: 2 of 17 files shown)


Gamenot:
I am somewhat struggling with figuring out a simple way to propagate the configuration. I understand that environment variables would allow that by propagation if the initial configuration just looped through and set them.

Gamenot:
## Discussion

We had some discussion regarding this issue.

> 1. Loop through the configuration and use them to set environment variables. (this should pass to subprocesses and workers but it would also mean only a single engine configuration of SMARTS could ever run on the same thread)

This seems viable, as it may allow us to make the engine configuration propagate across machines if a program propagates across workers (using something like `ray`) as long as the environment configuration resolves before the workers are generated (before something like `ray.init()`.) It looks like this might be viable to implement.

With that said, any user using a cluster management system could just inline the configuration using environment variables.

> 2. Explicitly send over the configuration to workers. This feels like it would be clunky but it would be consistent.

This option is likely too clunky and there is some question about performance. We can enforce sending over the configuration by accessing workers only through utilities we provide. If a user decides not to use a utility they may get confused as to why the engine configuration does not carry over.

> 3. Share an instance id with all workers/subprocesses, this allows reference to a shared resource location, all workers/subprocesses intialise using information from that location.

This is probably the most intensive solution, because we would need to provide a central or distributed location for the processes to find the engine settings. We could use an event processing platform like `apache kafka` or `ros`, or a cluster management system `ray`, or a web server. All of these seem like too much work for the immediate need.

> 4. Create and pass an environment capture object. When it deserialises it restores the global configuration.

This is very similar to option 2, it still requires us to route the user through an engine utility but it ends up being perhaps even more mysterious to end users.

> 5. Set a series of configuration paths (global user, local user, local directory) that are resolved in order of closeness to the project. Sub-processes will know about it inherently and each process lazy loads the configuration independently.

This seems like the most reasonable option for now. The exact implementation will resolve each of the configuration in this priority:

1. Individual environment variables (e.g. SMARTS_SENSOR_WORKER_COUNT)
2. Local directory engine configuration (`./smarts_engine.ini` )
3. Local user engine configuration (`~/.smarts/smarts_engine.ini` ) if local directory configuration is not found.
4. Global engine configuration (`/etc/smarts/smarts_engine.ini` ) if local configuration is not found.
5. Package default configuration (`${PYTHON_ENV}/lib/${PYTHON_VERSION}/site-packages/smarts/core/smarts_engine.ini`).


> 6. Configure the engine through `scl`.

This is viable regardless of other features. It allows us in the case of `scl run` to provide a `--configuration` flag which we could pass in a configuration from a custom location and immediately resolve it before anything else is evaluated.

## Direction
We decided to go

smarts/core/__init__.py
```diff
@@ -25,9 +25,14 @@
 
 import random
 import uuid
+from functools import lru_cache, partial
+from pathlib import Path
 
 import numpy as np
 
+import smarts
```
Gamenot: This caused an odd bug. It could not resolve the difference between `smarts` and `smarts.core.smarts`. I believe should change `smarts.core.smarts` to `smarts.core.engine`.

docs/index.rst
```diff
@@ -69,6 +69,7 @@ If you use SMARTS in your research, please cite the `paper <https://arxiv.org/ab
    :caption: API
 
    api/modules.rst
+   sim/configuration.rst
```
Adaickalavan, line 72: Maybe keep files under a given toctree in the same folder?

docs/sim/configuration.rst
```diff
@@ -0,0 +1,27 @@
+.. _configuration:
+
+Configuration
+=============
+
+You can change the behavior of the underlying SMARTS engine.
+
+Configuration of the engine can come from several sources. These locations take precidence as noted:
```
Adaickalavan, line 8:
```suggestion
Configuration of the engine can come from several sources. These locations take precedence as noted:
```

Adaickalavan:
Do check whether the docs build properly without any warnings before merging.

docs/index.rst
Gamenot: It is hard to find where to put this since `api/*` is generated dynamically but environment variables and configuration files are related to `sim` but are relevant to API. Maybe not relevant enough. I am going to move it to resources.

Gamenot: I am thinking `resources` is not quite the right spot for it either so I am moving it to next steps.

docs/sim/configuration.rst
```diff
@@ -0,0 +1,27 @@
...
+3. Local user engine configuration, ``~/.smarts/engine.ini``, if local directory configuration is not found.
+4. Global engine configuration, ``/etc/smarts/engine.ini``, if local configuration is not found.
+5. Package default configuration, ``$PYTHON_PATH/smarts/engine.ini``, if global configuration is not found.
+
+Note that configuration files resolve all settings at the first found configuration file (they do not layer.)
+
+Options
+-------
+
+All settings demostrated as environment variables are formatted to ``UPPERCASE`` and prefixed with ``SMARTS_`` (e.g. ``[core] logging`` can be configured with ``SMARTS_CORE_LOGGING``)
```
saulfield, line 21:
```suggestion
All settings demonstrated as environment variables are formatted to ``UPPERCASE`` and prefixed with ``SMARTS_`` (e.g. ``[core] logging`` can be configured with ``SMARTS_CORE_LOGGING``)
```

smarts/core/smarts.py
```diff
@@ -1281,7 +1283,18 @@ def fixed_timestep_sec(self) -> float:
...
-            self._rounder = rounder_for_dt(round(1 / MAX_PYBULLET_FREQ, 6))
+            self._rounder = rounder_for_dt(
+                round(
+                    1
+                    / config()(
+                        "physics",
+                        "max_pybullet_freq",
+                        default=MAX_PYBULLET_FREQ,
+                        cast=int,
+                    ),
```
saulfield, line 1294: I think pulling this out as a separate variable would make the expression more readable.

saulfield (approved):

PR #1862, opened 2023-02-14
Add reset options to "hiway-v1".

Gamenot:
See CHANGELOG.md

### DIFF
--- a/smarts/env/gymnasium/hiway_env_v1.py
+++ b/smarts/env/gymnasium/hiway_env_v1.py
@@ -338,7 +338,9 @@ class HiWayEnvV1(gym.Env):
                 If you pass an integer, the PRNG will be reset even if it already exists.
                 Usually, you want to pass an integer *right after the environment has been initialized and then never again*.
             options (optional dict): Additional information to specify how the environment is reset (optional,
-                depending on the specific environment)
+                depending on the specific environment). Forwards to :meth:`~smarts.core.smarts.SMARTS.reset`.
+                - "scenario" (smarts.sstudio.Scenario): An explicit scenario to reset to. The default is a scenario from the scenario iter.
+                - "start_time" (float): Forwards the start time of the current scenario. The default is 0.
 
         Returns:
             observation (dict): Observation of the initial state. This will be an element of :attr:`observation_space`
@@ -347,10 +349,15 @@ class HiWayEnvV1(gym.Env):
                 the ``info`` returned by :meth:`step`.
         """
         super().reset(seed=seed, options=options)
-        scenario = next(self._scenarios_iterator)
+        options = options or {}
+        scenario = options.get("scenario")
+        if scenario is None:
+            scenario = next(self._scenarios_iterator)
 
         self._dones_registered = 0
-        observations = self._smarts.reset(scenario)
+        observations = self._smarts.reset(
+            scenario, start_time=options.get("start_time", 0)
+        )
         info = {"map_source": self._smarts.scenario.road_map.source}
 
         if self._env_renderer is not None:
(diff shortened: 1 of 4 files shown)


smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -497,3 +504,12 @@ def scenario(self) -> Scenario:
...
         return self._smarts.scenario
+
+    @property
+    def smarts(self):
+        """Gives access to the underlying simulator. Use this carefully.
+
+        Returns:
+            smarts.core.smarts.SMARTS: The smarts simulator instance.
+        """
+        return self._smarts
```
Adaickalavan, line 515: Considering we have previously provided `def scenario()`, what is the use case for additionally providing `def smarts()`?

smarts/env/tests/test_hiway_env_v1.py
```diff
@@ -0,0 +1,100 @@
...
+    agent_ids = set(agent_interfaces)
+    env: HiWayEnvV1 = gym.make(
+        "smarts.env:hiway-v1",
+        scenarios=["scenarios/sumo/loop"],
+        agent_interfaces=agent_interfaces,
+        action_options="unformatted",
+        observation_options="unformatted",
+        headless=True,
+        visdom=False,
+        fixed_timestep_sec=0.01,
```
Adaickalavan, line 48: Maybe we should simply make the `fixed_timestep_sec=0.01` as the default value in `HiWayEnvV1`.

smarts/env/tests/test_hiway_env_v1.py
```diff
@@ -0,0 +1,100 @@
...
+        MAX_EPISODES - 1
+    ), "Simulation must cycle through to the final episode."
+
+
+def test_hiway_env_v1_reset_with_scenario(env: HiWayEnvV1):
+    scenarios = ["scenarios/sumo/loop"]
+    scenario: Scenario = next(Scenario.scenario_variations(scenarios, [AGENT_ID]))
+
+    env.reset(options={"scenario": scenario, "start_time": 100})
+    assert "loop" in env.scenario.root_filepath
```
Adaickalavan, line 98: Given the env originally only had one scenario, namely `scenarios/sumo/loop`, testing reset functionality by resetting to the same scenario might be an insufficient test. We could try resetting the env to a different scenario to check whether the functionality works.

smarts/env/tests/test_hiway_env_v1.py
```diff
@@ -0,0 +1,100 @@
...
+    assert isinstance(env.unwrapped, HiWayEnvV1)
+    assert not (agent_ids - set(env.agent_interfaces))
+    matching_items = [
+        env.agent_interfaces[k] == agent_interfaces[k]
+        for k in env.agent_interfaces
+        if k in agent_interfaces
+    ]
+    assert all(matching_items)
+    assert len(env.agent_interfaces) == len(agent_interfaces)
+    assert not (agent_ids - env.agent_ids)
```
Adaickalavan, line 60: Wondering whether these checks could be simplified.

smarts/env/gymnasium/hiway_env_v1.py
Gamenot: Generally just for direct access to SMARTS, I am somewhat cautious about allowing this but if we do not allow it there may be little exposure to actually using SMARTS directly.

smarts/env/tests/test_hiway_env_v1.py
Gamenot: Why `0.01`? The fixed time-step for the environment is `0.1` if not already given.

smarts/env/tests/test_hiway_env_v1.py
Gamenot: Right I was grabbing something quick, I will try "cloverleaf".

smarts/env/tests/test_hiway_env_v1.py
Gamenot: I also note that these checks should not be in the fixture. I am moving and simplifying them.

smarts/env/tests/test_hiway_env_v1.py
Gamenot: Used `figure_eight`.

saulfield (approved):

PR #619, opened 2021-02-25
Bugtest sumo crash reproduction

Gamenot:
This is a branch that contains scripts for generating a crash and reproducing it.

PR #1884, opened 2023-02-27
Attempt simplification of the traffic sims.

Gamenot:


smarts/core/sumo_traffic_simulation.py
```diff
@@ -390,14 +397,16 @@ def _handle_traci_exception(
...
-        if self._remove_agents_only_mode:
+        if self._remove_agents_only_mode == AgentRemovalMode.REMOVE_AGENTS:
             vehicles_to_remove = self._non_sumo_vehicle_ids
-        else:
+        elif self._remove_agents_only_mode == AgentRemovalMode.REMOVE_ALL:
             vehicles_to_remove = self._non_sumo_vehicle_ids.union(
                 self._sumo_vehicle_ids
             )
+        else:
+            vehicles_to_remove = []
```
saulfield, line 409: Minor, but this should be a set if we want consistency across each path.

saulfield (approved):

Adaickalavan (approved):

Gamenot:
This changeset makes me nervous, I stated that we should not have separate cases for vehicle insertion but I do remember how many issues SUMO caused and even swapping over to the local traffic simulation can cause significant disruption.

Gamenot:
This is abandoned for now because we should just find the underlying sumo issue and fix it.

Gamenot:
The underlying need is to resolve #619

PR #1891, opened 2023-03-02
Return done value and rewards for all agents in HiWayEnvV1 when observation options are set to 'full"

ajlangley:
Hi, I wanted to propose a simple change to the `step` method of `HiWayEnvV1`. 

Currently, when `observation_options` are set to `full`, the `step` method returns only a single value for both `done` and `reward`, respectively. I think, however, that this behavior is undesirable for a few reasons:

1. (I think that) this deviates from the PettingZoo parallel API, which makes it less convenient to pass a `HiWayEnvV1` instance to a wrapper intended for PettingZoo environments. 
2. The docs don't specify that the reward and done values will be affected.
3. There could be separate formatting options for `reward` and `done`

### DIFF
--- a/smarts/env/gymnasium/hiway_env_v1.py
+++ b/smarts/env/gymnasium/hiway_env_v1.py
@@ -293,9 +293,9 @@ class HiWayEnvV1(gym.Env):
         if self._observations_formatter.observation_options == ObservationOptions.full:
             return (
                 self._observations_formatter.format(observations),
-                sum(r for r in rewards.values()),
-                dones["__all__"],
-                dones["__all__"],
+                rewards,
+                dones,
+                dones,
                 info,
             )
         elif self._observations_formatter.observation_options in (


Gamenot:
Good morning @ajlangley, thank you for bringing up the issue. We realize this is the case.

As context. The HiWay environments are incomplete by design. We want them to be a base template to configure as a specific environment.

I believe here that `"multi_agent"` already has the  behaviour you expect. There is a case that multi-agent should be the default and that `"full"` does not properly describe a particular intention.

The particular issue that generated this is the conflict in standard between `gym` and multi-agent environments found in `RLlib` and `PettingZoo`.

• `gym` expects the reward and terminal returns to express the state of the environment.
• `RLlib` and `PettingZoo` which expresses the reward and terminal state of agents through the environment interface.

We did consider that we should implement an environment for `PettingZoo` explicitly and not provide multi-agent through the `gym/gymnasium` environment.

I think what should be done here for now is to split out the interface configuration from the observation configuration.

smarts/env/gymnasium/hiway_env_v1.py
```diff
@@ -293,9 +293,9 @@ def step(
         if self._observations_formatter.observation_options == ObservationOptions.full:
             return (
                 self._observations_formatter.format(observations),
-                sum(r for r in rewards.values()),
-                dones["__all__"],
-                dones["__all__"],
+                rewards,
+                dones,
+                dones,
```
Gamenot, line 298: With this change there is no difference between the current branch and the branch immediately following.

It may be sufficient to get rid of the branching for now.

ajlangley: Yeah, I see that now. The main motivation for using `"full"` rather than `"multi_agent"` was that, as I understood from the docs, `"multi_agent"` will not return the observation for every agent. So what was need in  my case was the observation for inactive agents and done values + rewards for inactive agents. 

I think that splitting the interface is the better option as well. It occurs to me now that we would still want an option like what `"full"` currently provides, for cases where we may want to easily model driving as a cooperative task (whether or not this is a well-motivated use case). 

Should I just close this PR perhaps and make a new one with the interface options mentioned?

Gamenot: > Yeah, I see that now. The main motivation for using `"full"` rather than `"multi_agent"` was that, as I understood from the docs, `"multi_agent"` will not return the observation for every agent. So what was need in my case was the observation for inactive agents and done values + rewards for inactive agents.

I see, your intention would be multi-agent interface with fixed sized observations|rewards|dones. I can understand why the current implementation of "full" does not address that need.

> I think that splitting the interface is the better option as well. It occurs to me now that we would still want an option like what `"full"` currently provides, for cases where we may want to easily model driving as a cooperative task (whether or not this is a well-motivated use case).

Yes, there is possibility that this will never be a real use case for anything other than evaluation. Mainly, we have a second concern (traffic simulators/providers) which we may want to evaluate against agents in the future. And the environment score is likely more interesting than any individual agent for that case.

> Should I just close this PR perhaps and make a new one with the interface options mentioned?

If you have the time to, that would be greatly appreciated.

Issue #1902, opened 2023-03-10
Problem running smarts in Jupyter notebooks with Python 3.8


saulfield:
### High Level Description

When upgrading the minimum Python version to 3.8, I ran into an error with the `smarts/core/tests/test_notebook.py::test_notebook1` test. To reproduce, run the following on the master branch:

```sh
python3.8 -m venv .venv
. .venv/bin/activate
pip install --upgrade pip
pip install --upgrade wheel
pip install -e .[test,test_notebook]
pytest -v smarts/core/tests/test_notebook.py::test_notebook1
```

Output:
```
.venv/lib/python3.8/site-packages/nbclient/client.py:953: DeadKernelError
------------------------------------------------------------------------------------ Captured stderr call ------------------------------------------------------------------------------------
Assertion failed: ok (src/mailbox.cpp:99)
------------------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------------------
ERROR    pytest_notebook.execution:client.py:808 Kernel died while waiting for execute reply.
...
FAILED smarts/core/tests/test_notebook.py::test_notebook1 - nbclient.exceptions.DeadKernelError: Kernel died
```

I matched the installed package versions to those used in a successful run with Python 3.7, but still got the same error. The only difference appears to be the Python version.

This can also be reproduced by running the notebook standalone (not as a test). The kernel crashes when executing the second cell, and Jupyter does not provide any helpful error messages.

### Version

1.0.7

PR #1904, opened 2023-03-10
Upgrade minimum Python version to 3.8

saulfield:
This PR bumps the minimum supported Python version to 3.8. This has been planned for some time, but was made necessary during the Argoverse 2 integration which requires >=3.8.

Some package version changes were needed:
- Upgraded `ray[rllib]` from `1.0.1.post1` to `1.4.0` to get rid of `gym[atari]` dependency which had problems installing in 3.8
- Put a `numpy<1.24.0` upper bound to fix an issue with `ray[rllib]==1.4.0` (see https://github.com/ray-project/ray/issues/31258)

Note also that an issue with notebooks popped up: #1902. We have decided to ignore the test for now to unblock the Argoverse integration, but this should be investigated.

### DIFF
--- a/utils/setup/install_deps.sh
+++ b/utils/setup/install_deps.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-function check_python_version_gte_3_7 {
+function check_python_version_gte_3_8 {
     echo "Checking for >=python3.8"
     # running through current minor versions
     hash python3.8 2>/dev/null \
@@ -8,7 +8,7 @@ function check_python_version_gte_3_7 {
     || hash python3.9 2>/dev/null;
 }
 
-function install_python_3_7 {
+function install_python_3_8 {
     echo "Installing python3.8"
     sudo apt-get install $1 software-properties-common
     sudo add-apt-repository $1 ppa:deadsnakes/ppa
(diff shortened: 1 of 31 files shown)


Gamenot:
Please add a changelog entry.

utils/setup/install_deps.sh
```diff
@@ -1,18 +1,18 @@
...
     || hash python3.9 2>/dev/null;
 }
 
 function install_python_3_7 {
-    echo "Installing python3.7"
+    echo "Installing python3.8"
     sudo apt-get install $1 software-properties-common
     sudo add-apt-repository $1 ppa:deadsnakes/ppa
-    sudo apt-get install $1 python3.7 python3.7-tk python3.7-venv
+    sudo apt-get install $1 python3.8 python3.8-tk python3.8-venv
```
Gamenot, line 15: Do you know why we need python Tkinter? It is not in most other requirements.

saulfield: I'm not sure. It seems it's been there for a while.

utils/setup/install_deps.sh
```diff
@@ -1,18 +1,18 @@
...
     # running through current minor versions
-    hash python3.7 2>/dev/null \
+    hash python3.8 2>/dev/null \
     || hash python3.8 2>/dev/null \
     || hash python3.9 2>/dev/null;
 }
 
 function install_python_3_7 {
-    echo "Installing python3.7"
+    echo "Installing python3.8"
```
Adaickalavan, line 12: Maybe the function names can also be update, such as `function install_python_3_8` .

Adaickalavan (approved):

Gamenot:
Is there anything blocking this?

PR #1905, opened 2023-03-13
Eliminate waymo_open_dataset dependency

### LINKED ISSUE #1880: Waymo Dataset Incompatibility
### High Level Description

The `waymo-open-dataset` package is now incompatible with smarts since it is pinned to old versions of tensorflow and numpy. There don't appear to be plans on their end to release a version with updated package versions anytime soon, so we need to find a workaround.

It should be noted that we only use the package for the `waymo_open_dataset.protos` module -- the generated code for parsing their protobuf format.

We discussed some potential solutions:
1. Host the protos ourselves, and compile that to our own module for parsing the dataset (and eliminate the dependency on their package)
2. Write a script that would still use the package, but would convert the scenarios to some other format. The user would do this in a separate venv, and we would have to change the waymo code in smarts to use this other format.

### Version

1.0.0

### Steps to reproduce the bug

_No response_

### System info

_No response_

### Error logs and screenshots

_No response_

### Impact (If known)

_No response_

saulfield:
Fixes #1880


Gamenot:
I am cautious about naming the package the same as the original. It might cause a package conflict.

qianyi-sun (approved):

saulfield:
> I am cautious about naming the package the same as the original. It might cause a package conflict.

I agree in general, but in this case I think it's ok. It makes it easier since we don't have to modify any of the protos or generated files to make the imports work, and since the waymo package is incompatible with smarts anyway there isn't a risk of them being installed in the same venv.

Adaickalavan:
Consider putting up an example, in the example folder, on using Waymo map and traffic. Add the Waymo example to the CI tests. Update how to use SMARTS for Waymo under Ecosystem section in ReadTheDocs.

Gamenot:
OK.

Gamenot (approved):

Adaickalavan:
Some thoughts on presenting SMARTS support for external datasets. We could address this in a separate PR too.

Consider updating how to use SMARTS for Waymo in ReadTheDocs::Ecosystem::Waymo.

In ReadTheDocs::Example::Naturalistic section we could provide intructions 
(e.g., `scl run --envision examples/<script_name>.py scenarios/<scenario_name>`) 
to run a simple ego agent (e.g., `AgentType.Laner`, etc) on a Waymo map with traffic (e.g., dataset="training_20s.tfrecord-00022-of-01000" and scenario_id="62662fab4d63c247"). We could then post a gif of the replayed Waymo scenario showing the expected outcome. To reproduce the result, we could instruct the users to download and edit the dataset path in the provided script prior to executing the instructions.

We could move readmes such as (i)`SMARTS/scenarios/waymo/README.md`, and (ii) `SMARTS/waymo_open_dataset/README.md`, to ReadTheDocs::Ecosystem::Waymo .

Adaickalavan (approved):

saulfield:
I moved the `waymo_open_dataset` module to `smarts/waymo` because there were some issues with finding the module path from within smarts.

I also updated the Waymo section of the docs, and deleted the README in the waymo example scenario.

PR #1906, opened 2023-03-14
Update sumo builder utility

Gamenot:
This is a step toward resolving #619. This updates the sumo source builder to generalize for other Ubuntu versions.

### TODO
- [ ] Target local SUMO repository
- [ ] Add debug build option
- [ ] Option to persist builder container


Gamenot:
This exposed a funny bug in the CI where `pytype` test will run without any python files because the CI looks for `.py` which is in `"README.pypi.md"`.

Not sure it is worth fixing.

PR #1910, opened 2023-03-17
Lanepoint requirement fixes

Gamenot:
Make lanepoint_spacing non-optional and update map classes to lazily load lanepoints.


smarts/core/argoverse_map.py
```diff
@@ -909,6 +904,11 @@ def query(
                     return [path[: (lookahead + 1)] for path in hit]
                 return None
 
+    @cached_property
+    def _lanepoints(self):
+        assert self._map_spec.lanepoint_spacing > 0
+        return LanePoints.from_argoverse(self, spacing=self._map_spec.lanepoint_spacing)
```
Gamenot, line 910: This slightly helps but generally not. The issue I was thinking of is that all waypoints are generated on this line whenever that might be. It will not scale.

smarts/sstudio/types.py
```diff
@@ -352,8 +352,8 @@ class MapSpec:
 
     source: str
     """A path or URL or name uniquely designating the map source."""
-    lanepoint_spacing: Optional[float] = None
-    """If specified, the default distance between pre-generated Lane Points (Waypoints)."""
+    lanepoint_spacing: float = 1.0
+    """The default distance between pre-generated Lane Points (Waypoints)."""
```
Gamenot, line 356: This by itself is useful enough to merge.

smarts/core/argoverse_map.py
saulfield: It helps for the case where we load the roadmap without "using" it, which I believe happens in scenario build.

Adaickalavan (approved):

PR #1916, opened 2023-03-22
remove `init=false` to enable `actor` in `Trip`

qianyi-sun:
see topic for description


smarts/sstudio/types.py
```diff
@@ -511,7 +511,7 @@ class Trip:
     """The type of the vehicle"""
     depart: float = 0
     """Start time in seconds."""
-    actor: TrafficActor = field(init=False)
+    actor: TrafficActor = field
```
Gamenot, line 514:
```suggestion
    actor: TrafficActor = field()
```

smarts/sstudio/types.py
```diff
@@ -511,7 +511,7 @@ class Trip:
     """The type of the vehicle"""
     depart: float = 0
     """Start time in seconds."""
-    actor: TrafficActor = field(init=False)
+    actor: TrafficActor = field(default=None)
```
Gamenot, line 514: Ah, right this messes with typing it should be
```suggestion
    actor: Optional[TrafficActor] = field(default=None)
```

Gamenot (approved):

Adaickalavan (approved):

PR #1919, opened 2023-03-24
Add actor alive interface

Gamenot:
This is intended to allow an agent to become done on actors of interest leaving the simulation.

### DIFF
--- a/envision/tests/test_data_formatter.py
+++ b/envision/tests/test_data_formatter.py
@@ -119,18 +119,7 @@ def covered_data():
                 2,  # z
                 -3.14,  # heading
                 20,  # speed
-                (
-                    [],
-                    0,
-                    0,
-                    0,
-                    0,
-                    0,
-                    0,
-                    0,
-                    1,
-                    0
-                ),  # events
+                ([], 0, 0, 0, 0, 0, 0, 0, 1, 0),  # events
                 [
                     [[4, 5, 3, 2.24, 1, 4, 10, 0, 4.96]],
                     [[9, 5, 3, 1.11, 2, 2, 1.2, 1, 9.96]],
@@ -168,7 +157,16 @@ def complex_data():
                         actor_id="agent_007",
                         lane_id="NE-NW",
                         events=Events(
-                            [], False, False, False, False, False, False, False, True, False
+                            [],
+                            False,
+                            False,
+                            False,
+                            False,
+                            False,
+                            False,
+                            False,
+                            True,
+                            False,
                         ),
                     )
                     for i in range(2)
--- a/envision/tests/test_data_formatter.py
+++ b/envision/tests/test_data_formatter.py
@@ -81,7 +81,7 @@ def covered_data():
                 actor_id="agent_007",
                 lane_id="NE-NW",
                 events=Events(
-                    [], False, False, False, False, False, False, False, True
+                    [], False, False, False, False, False, False, False, True, False
                 ),
                 driven_path=[(4, 4), (2, 2)],
                 point_cloud=[[1, 3], [4, 2]],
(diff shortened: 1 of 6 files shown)


smarts/core/agent_interface.py
```diff
@@ -225,6 +238,8 @@ class DoneCriteria:
     """
     agents_alive: Optional[AgentsAliveDoneCriteria] = None
     """If set, triggers the ego agent to be done based on the number of active agents for multi-agent purposes."""
+    actors_alive: Optional[ActorsAliveDoneCriteria] = None
+    """If set, triggers the ego agent to be done based on actors existing in the simulation."""
```
Adaickalavan, line 242: If possible, it might be better to combine `agents_alive` and `actors_alive` options together into one event configuration, because separate `agents_alive` and `actors_alive` options might conflict.

smarts/core/agent_interface.py
Adaickalavan, line 1: We could modify an exisiting test or add a new test to check for the actors alive done criteria.

smarts/core/agent_interface.py
```diff
@@ -193,6 +193,19 @@ class AgentsAliveDoneCriteria:
     """
 
 
+@dataclass(frozen=True)
+class ActorsAliveDoneCriteria:
+    """Require actors to persist."""
+
+    actors_of_interest: str = r""
```
Adaickalavan, line 200: If `actors_of_interest` was a list of strings, instead of a single string, we could account for multiple vehicles of interest and also probably unify with `agents_alive` done criteria.

smarts/core/sensors.py
```diff
@@ -333,6 +334,37 @@ def _agents_alive_done_check(
...
+
+        sensor_state: SensorState = sensor_state
+        from smarts.core.vehicle_index import VehicleIndex
+
+        vehicle_index: VehicleIndex = vehicle_index
+
+        pattern = re.compile(actors_alive.actors_of_interest)
+        ## TODO optimization to get vehicles that were added and removed last step
+        ## TODO second optimization to check for already known vehicles
+        for vehicle_id in vehicle_index.vehicle_ids:
```
qianyi-sun, line 355: When start a simulation for testing, an `TypeError` occurs here, saying that `'_CacheCallable' object is not iterable`. Maybe that's due to `vehicle_index.vehicle_ids: _CacheCallable`?

envision/tests/test_data_formatter.py
Adaickalavan, line 1: `actors_alive_done` needs to be added in the data formatter in `SMARTS/smarts/env/utils/observation_conversion.py`.

smarts/core/agent_interface.py
Gamenot: This is a pattern string, so it would account based on pattern matching. I could break it up into multiple for ease of use I suppose.

smarts/core/agent_interface.py
Gamenot: I think what you are considering is the difference between AND and OR causing termination. It is slightly difficult to satisfy these complex cases without introducing expressions.

I think we do not have the case for AND currently.

smarts/core/agent_interface.py
Gamenot: Broken up.

smarts/core/agent_interface.py
Adaickalavan: When we use a single env to cater to two types of scenarios, namely (i) one which uses and terminates with the social agent as the leader, and (ii) one which uses and terminates with the sumo vehicle as the leader, we end up with a code which looks like

[Assume lines 151 to 158 are uncommented]
https://github.com/huawei-noah/SMARTS/blob/7cdc5c3dd5b87b51026cfd4256bb2c79c539e2be/smarts/env/gymnasium/platoon_env.py#L151-L163

Here, the env will terminate immediately every time as one of the done criteria, namely `actors_alive` or `agents_alive` will become true immediately at the start of the every scenario.

Another issue here is the name of vehicle of interest differs in both scenarios, because `social-agent-leader` string gets prefixed to the agent name when using social agent as the leader.

PR #1924, opened 2023-03-24
Fix waypoints in SUMO junctions

### LINKED ISSUE #1911: Disappearing waypoints path at intersection
Do we know why the waypoints path appear and then disappear here?

Steps to reproduce:
```
$ git checkout zoo-7
$ cd <path>/SMARTS/examples/rl/baseline
$ python3.8 -m venv ./.venv
$ source ./.venv/bin/activate
$ pip install --upgrade pip wheel
$ pip install -e ./../../../.[camera_obs,dev,doc]
$ pip install -e ./inference/
$ python3.8 train/run.py
```

Here we use the scenario: SMARTS/scenarios/sumo/platoon/merge_exit_agents_1

This example enables sumo-gui by default for debugging. Simulation runs until the lead vehicle (in blue colour) reaches near the exit ramp. Then, a top-down rgb image with the waypoints superimposed in green colour is plotted at each time step. Simply close each of the rgb image to proceed to the next step in the simulation.

The intended outcome is for the social vehicle to turn right into the exit ramp. The social vehicle uses action space `ActionSpaceType.LaneWithContinuousSpeed`. It always executes "change-lane-to-right" action when it is near the exit ramp (i.e., when the rgb images start to be plotted).

Stepping through the simulation, we see that the waypoint path leading towards the exit ramp appear (see first attached image), then disappear for one time step (see second attached image), before re-appearing again. I guess it reappears because the social vehicle continuously attempts to change lane to right. The question is why did the waypoint path leading towards the exit ramp disappear in one of the time step?

Try running a couple of episodes or rebuilding the scenario, if the issue is not seen in the first episode.

saulfield:
Fixes #1911.

See [this comment](https://github.com/huawei-noah/SMARTS/issues/1911#issuecomment-1480054822) for context. I'm opening a PR for this because it appears to be separate from the other waypoint issues that have been reported.


smarts/core/sumo_road_network.py
```diff
@@ -969,14 +974,17 @@ def _resolve_in_junction(self, pose: Pose) -> List[str]:
...
-        assert (
-            len(next_roads) <= 1
-        ), "A junction is expected to have <= 1 outgoing roads"
-        if next_roads:
-            road_ids.append(next_roads[0].road_id)
-        return road_ids
+
+        wp_lanes = []
+        for incoming_lane in lane.incoming_lanes:
+            for junction_lane in incoming_lane.outgoing_lanes:
```
Gamenot, line 980: This may or may not be enough, I remember that sometimes the junction lanes are subdivided so that the incoming lane is still a junction lane. You may want to do a `in_junction` check to get back to an incoming lane that is not in the junction.

saulfield: Good to know. Should we hold off on merging this until we test it a bit more? I think this change will only help.

Gamenot (approved):

PR #1936, opened 2023-03-31
Simplify scenario paths

Adaickalavan:
+ Moved benchmark scenarios into `SMARTS/scenarios` folder.
+ Simplified the scenario loading code in `driving_smarts` benchmark.


scenarios/sumo/straight/3lane_bubble/scenario.py
```diff
@@ -5,12 +5,12 @@
 
 trajectory_boid_agent = t.BoidAgentActor(
     name="trajectory-boid",
-    agent_locator="scenarios.sumo.straight.agent_prefabs:trajectory-boid-agent-v0",
+    agent_locator="scenarios.sumo.straight.3lane_bubble.agent_prefabs:trajectory-boid-agent-v0",
```
Gamenot, line 8: Modules starting with numerals are not Python compliant. This is fine since we support it by loading in dynamically but it is not easy to test this as a regular import.

Gamenot (approved):

Issue #1816, opened 2023-01-27
CI timeout


Gamenot:
## **BUG REPORT**
**High Level Description**
CI noted the timeout of a particular run. This means that we have a rare infinite loop somewhere in the engine. This seems to be recent.

**SMARTS version**
1.0.1

**Previous associated issues**
N/A

**Steps to reproduce the bug**
Unsure.

**Resulting and expected behaviour**
The tests timed out but should roughly pass in under 12 minutes.

**Error logs and screenshots**
https://github.com/huawei-noah/SMARTS/actions/runs/4017689624/jobs/6902359722
 
**System information**
CI
- Date when bug occurred
2023/01/27
 
**Impact [If known]**
Potential issues with long running SMARTS.

Gamenot:
There seems to be an earlier instance of this happening with `test_studio:test_scenario_generation_unchanged`:

https://github.com/huawei-noah/SMARTS/actions/runs/3978346086/jobs/6820060923

Gamenot:
@saulfield It appears that scenario building might be causing issues.

PR #1937, opened 2023-04-03
Implement shape method for Argoverse Road class

saulfield:
This was required to visualize mission roads in envision.


qianyi-sun (approved):
thanks!

Gamenot (approved):

Gamenot:
I notice the CI stalled, is it related to the current change-set? https://github.com/huawei-noah/SMARTS/actions/runs/4598197407/jobs/8145488130

saulfield:
I don't believe so. I got it to pass just now after rerunning a few times. It seems there is some non-determinism. It looks similar to #1816.

Issue #1927, opened 2023-03-26
Strict waypoint paths in Argoverse map


Adaickalavan:
Steps to reproduce:
```
$ git checkout zoo-9
# Place scenario id "c624608b-fd20-43c9-bc2c-c4181ce9dafa" in `SMARTS/argoverse/data/` folder
$ cd <path>/SMARTS/examples/rl/baseline
$ python3.8 -m venv ./.venv
$ source ./.venv/bin/activate
$ pip install --upgrade pip wheel
$ pip install -e ./../../../.[camera_obs,argoverse]
$ pip install -e ./inference/
# Enable envision for debugging
$ scl envision start &
$ python3.8 train/run.py --head
# Scenario used: SMARTS/scenarios/argoverse/platoon/map_3_agents_1/scenario.py
```

Simulation runs until the ego vehicle (in red colour) reaches near the intersection. Then, a top-down rgb image with the waypoints superimposed in green colour is plotted at each time step. Simply close each of the rgb image to proceed to the next step in the simulation.

Stepping through the simulation, we see the ego's turn-left and turn-right waypoints appear as it reaches the intersection. However, both turn-left and turn-right waypoints disappear simultaneously very early on, making the potential waypoints too strict. This is especially true for the turn-left waypoints which disappear too quickly although the turn-left still appears some distance ahead. 

Is it reasonable to make the potential waypoint paths slightly less strict?

Two consecutive images are shown below depicting this phenomenon. The embedded rgb images, in each image below, show three consecutive time points.

![Screenshot from 2023-03-25 21-00-42](https://user-images.githubusercontent.com/24203326/227803745-db5f03d7-ac08-45be-a65f-a8d005c89404.png)

![Screenshot from 2023-03-25 21-01-06](https://user-images.githubusercontent.com/24203326/227803749-08db6791-44b9-475e-b312-5e1f72b05205.png)

Gamenot:
This is potentially related to #1847

Issue #1926, opened 2023-03-26
Malformed waypoints for agents using Mission and EndlessMission in Argoverse maps


Adaickalavan:
After PR #1923, some waypoints appear malformed in Argoverse maps. 

Waypoints appear malformed differently in agents using "Mission" compared to agents using "EndlessMission". Preferably, agents using "Mission" and "EndlessMission" should generate waypoints in the same manner, with the exception of agents using "EndlessMission" should get all legal waypoints at each time point whereas agents using "Mission" should only get the waypoints leading to the goal at each time point.  

A sequence of images showing malformed (i) social agent waypoints using "Mission" and (ii) ego agent waypoints using "EndlessMission", is shown below in Argoverse map `c624608b-fd20-43c9-bc2c-c4181ce9dafa`.

1) Social agent waypoints in the current lane stops abruptly and does not branch to the south, but ego agent waypoints does branch to the south.
    <img src="https://user-images.githubusercontent.com/24203326/227794042-e124f91a-4555-4e4d-89af-8c1bd15024c9.png" width=500px height=500px>
    <img src="https://user-images.githubusercontent.com/24203326/227795581-e83a757b-b1e9-4d33-a110-4a7ee9ade215.png" width=500px height=500px>

2) Social agent waypoints depict a wierd shape like "7" in the intersection.
    <img src="https://user-images.githubusercontent.com/24203326/227795468-318f6418-082e-4601-8c8b-96423d14e90a.png" width=500px height=500px>

3) Ego agent waypoints pointing straight ahead in the current lane disappears and instead waypoints pointing to a left-turn appears. This happens for exactly one time step, after which the left-turn waypoints disappears and the waypoints pointing straight ahead reappears. Two consecutive images are shown below depicting this phenomenon. The embedded rgb images, in each image below, show three consecutive time points.  
    <img src="https://user-images.githubusercontent.com/24203326/227795715-64aeacc8-63dc-4b21-baad-99113f697cd8.png" width=800px height=400px>
     <img src="https://user-images.githubusercontent.com/24203326/227795716-97ac289e-65db-4792-aeae-593472f006c4.png" width=800px height=400px>

saulfield:
For the first issue of the social agent with a `Mission`, the waypoints/route generation are actually "correct", but they look a bit strange because of the road structure. See the below image:

![image](https://user-images.githubusercontent.com/17256344/228032526-947ace04-c5ae-4058-a39b-4b108e29534a.png)

[...]

PR #1944, opened 2023-04-06
Argoverse waypoint improvements in junctions

saulfield:
This addresses #1926, specifically the 3rd point raised in the issue. We now keep track of the last junction lane used for waypoints, to prevent erratic behaviour when the vehicle is on top of multiple lanes.

This also addresses #1927, by generating waypoints for all paths leading into the current junction lane:

https://user-images.githubusercontent.com/17256344/230475359-f3e9283d-068c-4c5e-890a-00e25d1f6166.mp4


Adaickalavan:
Wondering whether it would be possible to slightly tighten the potential waypoint paths which now appear a little too loose?

Screenshot from 2023-04-06 15-52-35

saulfield:
The problem is that there is no road connectivity between that road and where the ego is, so there is no valid path leading back to those waypoints, though they are still part of the junction. Would it be suitable to filter out those waypoints earlier on so that they aren't so far away?

smarts/core/argoverse_map.py
```diff
@@ -948,11 +964,42 @@ def waypoint_paths(
...
             [pose], within_radius=within_radius
         )
-        closest_lane = closest_lps[0].lane
+        closest_lane: RoadMap.Lane = closest_lps[0].lane
+
         waypoint_paths = []
+        if closest_lane.in_junction:
+            junction_lane = closest_lane
+            closest_lanes = [lp.lane for lp in closest_lps]
+            if self._last_junction_lane and self._last_junction_lane in closest_lanes:
```
Gamenot, line 977: We should be able to assume that `None` isn't in the lanes.
```suggestion
            if self._last_junction_lane in closest_lanes:
```

smarts/core/lanepoints.py
```diff
@@ -881,7 +881,7 @@ def closest_lanepoints(
             k=maximum_count,
             filter_composites=True,
         )
-        return [l_lps[0].lp for l_lps in linked_lanepoints]
+        return [l_lps.lp for l_lps in linked_lanepoints[0]]
```
Gamenot, line 884: What is the purpose of switching from row to column selection?

```
Before
+---+---+---+
|p1 |p2 |p3 |
+---+---+---+
| x | x | x |
+---+---+---+
|   |   |   |
+---+---+---+
After
+---+---+---+
|p1 |p2 |p3 |
+---+---+---+
| x |   |   |
+---+---+---+
| x |   |   |
+---+---+---+
```

Am I mistaken in thinking that now all the `lps` will be related to the first pose?

smarts/core/argoverse_map.py
```diff
@@ -948,11 +964,42 @@ def waypoint_paths(
...
+                            np.dot(heading_vec, wp_vec)
+                            / (np.linalg.norm(heading_vec) * np.linalg.norm(wp_vec))
+                        )
+                        return angle
+
+                    # Filter out any waypoints that are behind the vehicle
+                    wps = [wp for wp in path if _angle_between(pose, wp) < np.pi / 1.5]
+                    if len(wps) > 0:
+                        waypoint_paths.append(wps)
+            self._last_junction_lane = junction_lane
```
Gamenot, line 999: Is this intended to be primitive caching? Can this handle agents simultaneously in multiple intersections?

smarts/core/argoverse_map.py
```diff
@@ -948,11 +964,42 @@ def waypoint_paths(
...
+                        heading_vec = pose.heading.direction_vector()
+                        wp_vec = wp.pos - pose.position[:2]
+                        angle = np.arccos(
+                            np.dot(heading_vec, wp_vec)
+                            / (np.linalg.norm(heading_vec) * np.linalg.norm(wp_vec))
+                        )
+                        return angle
+
+                    # Filter out any waypoints that are behind the vehicle
+                    wps = [wp for wp in path if _angle_between(pose, wp) < np.pi / 1.5]
```
Gamenot, line 996: This might not work because the vehicle could get turned around and lose out on seeing waypoints that are ahead of where it should go because it is turned around.

image

Gamenot: Considering @Adaickalavan's comment on the junction waypoints I am not entirely certain what should be done. 

- Filter branches by using the closest similarly near-oriented point's orientation to cull other branches
- Cut off branches which have the nearest point far away and ends oriented away.
- Leave roughly as is

Adaickalavan:
> The problem is that there is no road connectivity between that road and where the ego is, so there is no valid path leading back to those waypoints, though they are still part of the junction. Would it be suitable to filter out those waypoints earlier on so that they aren't so far away?

+ Yes, perhaps we could keep the right-turn path until this point in time (shown below). This would ensure the path options are not too strict. 
+ After this point in time, when the right-turn path becomes too far away, we could filter it out if possible. This would ensure the path options are not too loose.

Screenshot from 2023-04-06 17-05-02

I just realised that the same issue of far-away waypoints occur in the left-turn path too.

Screenshot from 2023-04-06 17-13-29

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 ### Changed
 ### Deprecated
 ### Fixed
+- Fixed issues related to waypoints in junctions on Argoverse maps.
```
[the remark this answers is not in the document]
saulfield: I've added more detail to the changelog entry.

smarts/core/argoverse_map.py
saulfield: I've removed this.

saulfield:
I've made some changes based on the feedback.

- Filtered out waypoint paths that are out of range
- Removed the caching, and instead generated (unique) paths for all junctions lanes that the vehicle is on

This should now give all possible paths, and relevant paths should not suddenly disappear. However, sometimes new paths can suddenly appear if the vehicle drives onto a new lane. I think this is fine, since an agent can determine for itself which waypoints to follow.

See the video below for a demo:

https://user-images.githubusercontent.com/17256344/231545971-e91991c4-2fbf-42cf-ab51-3e7b7d3a463e.mp4

smarts/core/argoverse_map.py
saulfield: We are now filtering based on distance to first lanepoint.

smarts/core/lanepoints.py
saulfield: Good catch. I didn't even realize that this takes in multiple poses. I'm a bit confused by the intention of this function, considering it seems to only ever be called with 1 pose. The original definition only took the first lanepoint of each set of 10 lanepoints for each pose. I think I may just change it to take a single pose and leave it as-is.

smarts/core/argoverse_map.py
```diff
@@ -936,6 +936,21 @@ def _lanepoints(self):
         assert self._map_spec.lanepoint_spacing > 0
         return LanePoints.from_argoverse(self, spacing=self._map_spec.lanepoint_spacing)
 
+    def _resolve_in_junction(self, junction_lane: RoadMap.Lane) -> List[List[str]]:
+        # There are no paths we can trace back through the junction, so return
+        if len(junction_lane.road.incoming_roads) == 0:
+            return []
+
+        # Trace back to the road that leads into the junction
+        inc_road: RoadMap.Road = junction_lane.road.incoming_roads[0]
```
Gamenot, line 945: Is this enough? I remember that some junction roads are split inside the intersection in SUMO. As in junction roads can lead into other junction roads so we would need to regress until the incoming road is not in the junction.

Gamenot (approved):
Only one concern and this can be merged.

smarts/core/argoverse_map.py
saulfield: I have not come across a case like like that so far. Argoverse lane segments tend to be quite long, and to go through the entire intersection. I think we should keep it simple for now.

Adaickalavan (approved):

PR #1954, opened 2023-04-11
Fix ReadTheDocs

saulfield:
Added missing `__init__.py` file.


bin/header_test.sh
```diff
@@ -2,7 +2,7 @@
 
 # Collate all files excluding auto generated files of *_pb2.py and *_pb2_grpc.py.
 # We also exlude files in the waymo_open_dataset module.
-python_files="$(find ./baselines/marl_benchmark ./cli ./envision ./smarts -name '*.py' ! -name '*_pb2.py' ! -name '*_pb2_grpc.py' ! -wholename './smarts/waymo/waymo_open_dataset/__init__.py')"
+python_files="$(find ./baselines/marl_benchmark ./cli ./envision ./smarts -name '*.py' ! -name '*_pb2.py' ! -name '*_pb2_grpc.py' ! -wholename './smarts/waymo/waymo_open_dataset/__init__.py' ! -wholename './smarts/waymo/waymo_open_dataset/protos/__init__.py')"
```
Adaickalavan, line 5: Curious. This path `! -wholename './smarts/waymo/waymo_open_dataset/__init__.py'` is being ignored but it has a Huawei copyright header.

Adaickalavan (approved):

bin/header_test.sh
saulfield: Strange, looks like it got added at some point. I've changed it to be the Waymo copyright header.

PR #1957, opened 2023-04-13
Actor vehicle name same as configuration.

Gamenot:
This ensures that the vehicle name generated from a social agent actor is the exact same as the name of the social agent actor (the agent id will still be the same as before.)

As a result, actors in social agent missions must be unique.


qianyi-sun (approved):

smarts/sstudio/types.py
```diff
@@ -1133,3 +1134,21 @@ class Scenario:
...
+                for o_group in groups:
+                    if group == o_group:
+                        continue
+                    if intersection := set.intersection(
+                        set(map(_get_name, actors)),
+                        map(_get_name, self.social_agent_missions[o_group][0]),
+                    ):
+                        raise ValueError(
+                            f"Social agent mission groups `{group}`|`{o_group}` have overlapping actors {intersection}"
+                        )
```
Gamenot, line 1154: This is to ensure that there are not duplicate actors in the social missions.

smarts/core/vehicle_index.py
```diff
@@ -653,11 +653,13 @@ def build_agent_vehicle(
         surface_patches,
         initial_speed=None,
         boid=False,
+        *,
```
saulfield, line 656: Is this a typo?

Adaickalavan: The single star * syntax is from [PEP3102](https://peps.python.org/pep-3102/).

saulfield: I'm not sure I understand why `vehicle_id` needs to be a "keyword-only argument".

Gamenot: It does not need to be 'keyword-only' but it ensures that the value is not specified accidentally.

I have a preference for keeping it 'keyword-only' but no heavy attachments.

smarts/core/tests/test_smarts_frame_rate.py
```diff
@@ -92,7 +106,7 @@ def smarts():
     agents = {AGENT_1: laner}
     smarts = SMARTS(
         agents,
-        traffic_sims=[SumoTrafficSimulation(headless=True)],
+        traffic_sims=[SumoTrafficSimulation(headless=False)],
```
Adaickalavan, line 109: Why run with sumo-gui ?

Adaickalavan (approved):

saulfield (approved):

smarts/core/tests/test_smarts_frame_rate.py
Gamenot: This is what is causing the failing test.

PR #1959, opened 2023-04-14
Warn character truncation once only

Adaickalavan:
+ Character truncation warning message from observation space formatter tends to fill up the terminal. Example:
  ```bash
  WARNING:smarts.env.utils.observation_conversion:`vehicle id` named `car-flow-route-E1_2_0-E3_2_max--5782264158901839037--6390006830799584855--3-0.2` is more than `50` characters long. It will be truncated and may cause unintended issues with navigation and lane identification.
  WARNING:smarts.env.utils.observation_conversion:`vehicle id` named `car-flow-route-E1_0_0-E3_0_max-6963017027078397547--6390006830799584855--0-0.2` is more than `50` characters long. It will be truncated and may cause unintended issues with navigation and lane identification.
  WARNING:smarts.env.utils.observation_conversion:`vehicle id` named `car-flow-route-E1_0_0-E3_0_max-6963017027078397547--6390006830799584855--0-0.3` is more than `50` characters long. It will be truncated and may cause unintended issues with navigation and lane identification.
  WARNING:smarts.env.utils.observation_conversion:`vehicle id` named `car-flow-route-E1_2_0-E3_2_max--5782264158901839037--6390006830799584855--3-0.2` is more than `50` characters long. It will be truncated and may cause unintended issues with navigation and lane identification.
  ```
+ Hence, warn only on the first occurence.
+ Is there better a way to warn once only?

### DIFF
--- a/smarts/env/utils/observation_conversion.py
+++ b/smarts/env/utils/observation_conversion.py
@@ -67,10 +67,13 @@ _UNSIGNED_INT64_SPACE = gym.spaces.Box(low=0, high=1e10, shape=(), dtype=np.int6
 _DISCRETE2_SPACE = gym.spaces.Discrete(n=2)
 _LANE_ID_SPACE = gym.spaces.Text(_WAYPOINT_NAME_LIMIT, charset=_WAYPOINT_CHAR_SET)
 
+_WARN_CHARACTER_TRUNCATION = True
 
 def _format_id(lane_id: str, max_length, type_):
+    global _WARN_CHARACTER_TRUNCATION
     lane_name = lane_id
-    if len(lane_name) > max_length:
+    if _WARN_CHARACTER_TRUNCATION and len(lane_name) > max_length:
+        _WARN_CHARACTER_TRUNCATION = False
         warnings.warn(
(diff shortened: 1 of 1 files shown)


Gamenot:
It would be better just to add an engine configuration to toggle it on and off.

Gamenot:
My only thought for resolving this is to shorten the id or add configuration to increase the character limit.

Adaickalavan:
> This seems like an important warning message. I am biased towards saying that this should be an error.

+ If this warning is converted into an error, then the simulation might not proceed far as this warning currently pops up very frequently.
+ The character truncation only affects the formatted observations returned to the users model, whereas smarts internally still has access to and may use the full untruncated vehicle ids.

Gamenot:
> > ...
> 
> * If this warning is converted into an error, then the simulation might not proceed far as this warning currently pops up very frequently.

That is fair, if they want the full observations they would then be forced to use SMARTS observation object directly.

> * The character truncation only affects the formatted observations returned to the users model, whereas smarts internally still has access to and may use the full untruncated vehicle ids.

Then I think we should shorten the generated id length. I am going to put up a PR for it.

PR #1964, opened 2023-04-17
Modify "feature request" template for more natural user experience

qianyi-sun:
The structure of current "feature request" issue template is "Proposal" then "Motivation", which is easier for us to read as devs because when we check it we can always see the solution that user would like first. 

But from user's perspective, it might be more logically natural for general writing if motivation/context goes first, and followed by the possible solution. 

Still not sure if we should do this since either way has its pros/cons, sall comments are welcomed!

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -4,6 +4,13 @@ title: "[Feature request] Feature request title"
 labels: ["enhancement"]
 body:
   - type: textarea
+    id: motivation
+    attributes:
+      label: Motivation
+      description: |
+        Is your feature request related to a problem? Please describe. 
+        Give clear and concise description of what the problem is. Ex. I am always frustrated when (...).
+  - type: textarea
     id: proposal
     attributes:
       label: Proposal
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -3,6 +3,13 @@ description: Suggest an idea for this project not related to bugs
 title: "[Feature request] Feature request title"
 labels: ["enhancement"]
 body:
+  - type: textarea
+    id: motivation
+    attributes:
+      label: Motivation
+      description: |
+        Is your feature request related to a problem? Please describe. 
+        Give clear and concise description of what the problem is. Ex. I am always frustrated when (...).
```
Gamenot, line 12: If we are moving it up here we must force the user to justify the change.
```suggestion
  - type: textarea
    id: motivation
    attributes:
      label: Motivation
      description: |
        Is your feature request related to a problem? Please describe. 
        Give clear and concise description of what the problem is. Ex. I am always frustrated when (...).
    validations:
      required: true
```

saulfield (approved):

.github/ISSUE_TEMPLATE/feature_request.yml
Adaickalavan: Requesting users to describe the issue will allow us to evaluate the proposed changes and also help us formulate our own proposals.

Gamenot: It is difficult to prioritize or offer alternatives if we do not know why the user is making the request. My opinion is that the motivation statement is more important than the proposal.

Adaickalavan (approved):

Gamenot:
@qianyi-sun Please merge this when you have the chance.

PR #1965, opened 2023-04-17
Add conditions to `sstudio`

### LINKED ISSUE #1963: [Feature request] New entry tactic whereby the ego appears after leader
### Proposal

Add a new `EntryTactic` to allow that one actor (usually the ego) spawns conditional on a vehicle existing, like:
```
entry_tactic = ExistEntryTactic(vehicle_id -> str)
```

### Motivation

For following scenarios, in some cases that we need to spawn the leader/follower in a moving traffic, with current implemenation it's a bit challenging to ensure that the traffic actors dont interfere the leader from spawning at the exact desired time step and location only by tuning their depart_time. 

With this proposed new entry tactic, an actor can be allowed to be spawned after an actor of interest is spawned, which could guarantee that one vehicle always respawns after another to prevent any possible interference.

### Alternatives

_No response_

### Additional context

_No response_

Gamenot:
This introduces conditions to `sstudio` to give a way of expressing conditions on which something occurs. As part of this change, entry tactics will use conditions instead of the way they currently work.

The condition operators are inspired by the OpenSCENARIO logical operators: https://www.asam.net/static_downloads/public/asam-openscenario/2.0.0/language-reference/expressions_main.html#sec-lc-operators-logical

Closes #1963


smarts/sstudio/types.py
```diff
@@ -595,11 +600,317 @@ class Traffic:
...
+            Else FALSE
+        DISJUNCTION (A OR B)
+            If either condition evaluates TRUE the result is exclusively TRUE.
+            Else if either condition evaluates BEFORE then the result will be BEFORE.
+            Else if either condition evaluates EXPIRED then the result will be EXPIRED.
+            Else FALSE
+        IMPLICATION (A AND B or not A)
+            If the first condition evaluates *not* TRUE the result is exclusively TRUE.
+            Else if the first condition evaluates TRUE and the second condition evaluates TRUE the result is exclusively TRUE.
+            Else FALSE
```
Gamenot, line 849: In terms of combining conditions, this is how they resolve.

smarts/sstudio/types.py
```diff
@@ -595,11 +600,320 @@ class Traffic:
...
+        DISJUNCTION (A OR B)
+            If either condition evaluates TRUE the result is exclusively TRUE.
+            Else if either condition evaluates BEFORE then the result will be BEFORE.
+            Else if both conditions evaluate EXPIRED then the result will be EXPIRED.
+            Else FALSE
+        IMPLICATION (A AND B or not A)
+            If the first condition evaluates *not* TRUE the result is exclusively TRUE.
+            Else if the first condition evaluates TRUE and the second condition evaluates TRUE the result is exclusively TRUE.
+            Else FALSE
```
Gamenot, line 849: This is how the condition operators evaluate.

Notable oddities are:
- On conjunction, if either condition is expired the conjunction will always be expired.
- On disjunction, if both conditions are expired the conjunction will always be expired.
- On implication, the resulting condition can only be TRUE or FALSE. The only theoretical case where BEFORE or EXPIRED can occur is when the first condition evaluates as a constant TRUE (via LiteralCondition and negations and combinations of) but I have not handled that case.

scenarios/sumo/intersections/1_to_2lane_left_turn_t_agents_1/scenario.py
```diff
@@ -23,65 +23,78 @@
...
 traffic = {}
 for name, routes in enumerate(route_comb):
     traffic[str(name)] = Traffic(
         flows=[
             Flow(
                 route=Route(
-                    begin=(start_edge, start_lane, 0),
-                    end=(end_edge, end_lane, "max"),
+                    begin=("gneE3", start_lane, 0),
+                    end=("gneE4", end_lane, "max"),
```
Adaickalavan, line 64: This map `scenarios/sumo/intersections/1_to_2lane_left_turn_t_agents_1/scenario.py` seems to only contain `E0`, `E1`, and `E4` edges. What does `gneE3`, `gneE4`, and `gneE6` refer to?

Gamenot: I think this file was changed and somehow did not get updated when I used rebase. I am going to fix it.

Adaickalavan (approved):

smarts/core/trap_manager.py
```diff
@@ -210,61 +245,52 @@ def largest_vehicle_plane_dimension(vehicle: Vehicle):
...
-                vehicle = self._take_existing_vehicle(
-                    sim,
-                    vehicle_id,
-                    agent_id,
-                    mission,
-                    social=agent_id in sim.agent_manager.pending_social_agent_ids,
-                )
-            elif trap.patience_expired(sim.elapsed_sim_time):
+            if ConditionState.EXPIRED in capture.ready_state:
                 # Make sure there is not a vehicle in the same location
```
Gamenot, line 290: @saulfield The currently failing test has to do with the change in order of the conditions here. I have expired default evaluating first in the exclusive conditions instead of second.

Should I revert the order?

smarts/sstudio/types.py
```diff
@@ -595,11 +603,573 @@ class Traffic:
...
+
+
 @dataclass(frozen=True)
 class EntryTactic:
     """The tactic that the simulation should use to acquire a vehicle for an agent."""
 
-    pass
+    start_time: float
+    # condition: Condition
+    # """The condition to determine if this entry tactic should be used."""
```
saulfield, line 1172: Remove?

saulfield (approved):

smarts/core/id_actor_capture_manager.py
```diff
@@ -58,17 +59,29 @@ def step(self, sim):
...
-                    + f"`{mission.start_time}` and `{patience_expiry}` has expired with no vehicle."
-                    f"`simulation time: {sim.elapsed_sim_time}`"
+                    + f"`Condition `{entry_tactic.condition}` has expired with no vehicle."
                 )
                 used_actors.append(actor_id)
                 sim.agent_manager.teardown_ego_agents({agent_id})
                 continue
+            if not condition_result:
+                continue
+            print(condition_result)
```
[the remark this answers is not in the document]
Gamenot: Yes, it is for debugging. I thought I had removed it.

smarts/core/id_actor_capture_manager.py
```diff
@@ -58,17 +59,29 @@ def step(self, sim):
...
+                entry_tactic.condition.requires,
+            )
+            condition_result = entry_tactic.condition.evaluate(**condition_kwargs)
+            if condition_result == ConditionState.EXPIRED:
+                print(condition_result)
+                self._log.warning(
                     f"Actor aquisition skipped for `{agent_id}` scheduled to start between "
-                    + f"`{mission.start_time}` and `{patience_expiry}` has expired with no vehicle."
-                    f"`simulation time: {sim.elapsed_sim_time}`"
+                    + f"`Condition `{entry_tactic.condition}` has expired with no vehicle."
```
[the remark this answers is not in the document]
Gamenot: Conditions have an `EXPIRED` state which is lowest priority on OR operations and highest on AND operations. So it is still possible to expire. If the condition never generates `EXPIRED` it will never expire.

There are cases like such:

```python console
>>> condition = LiteralCondition(ConditionState.FALSE).expire(20)
>>> condition.evaluate(time=10)
<ConditionState.FALSE: 0>
>>> condition.evaluate(time=21)
<ConditionState.EXPIRED: 2>
```

I have some thoughts that we should add `.then(other_condition)` and/or `.on_hold(count_seconds=5, hold_state=ConditionState.FALSE, emit=ConditionState.EXPIRED)` to handle additional temporal logic.

smarts/core/trap_manager.py
```diff
@@ -286,14 +312,21 @@ def largest_vehicle_plane_dimension(vehicle: Vehicle):
                     trap.default_entry_speed,
                     social=agent_id in sim.agent_manager.pending_social_agent_ids,
                 )
+            elif trap_condition and capture.vehicle_id is not None:
```
Gamenot, line 315: Oh, I just noticed that pytype did not catch that this was a bug since `trap_condition` could be unbound.

smarts/core/trap_manager.py
Gamenot: I don't like that I changed the order around and am still getting the same result. I can only assume that the generation has changed.

Gamenot: Right, it turns out I had changed `zoo_intersection` accidentally.

PR #1975, opened 2023-04-21
Fix memory leak that leads to cumulative flows

qianyi-sun:
This fixes a memory leak issue that leads to cumulative flows across episodes when using smarts as the traffic engine.

### DIFF
--- a/smarts/core/local_traffic_provider.py
+++ b/smarts/core/local_traffic_provider.py
@@ -688,7 +688,7 @@ class _TrafficActor:
 
     def update_route(self, route: RouteWithCache):
         """Update the route (sequence of road_ids) this actor will attempt to take.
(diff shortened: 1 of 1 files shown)


Gamenot (approved):

smarts/core/local_traffic_provider.py
```diff
@@ -687,7 +688,8 @@ def route(self) -> RoadMap.Route:
 
     def update_route(self, route: RouteWithCache):
         """Update the route (sequence of road_ids) this actor will attempt to take.
-        A unique route_key is provided for referencing the route cache in he owner provider."""
+        A unique route_key is provided for referencing the route cache in he owner provider.
```
saulfield, line 691:
```suggestion
        A unique route_key is provided for referencing the route cache in the owner provider.
```

saulfield (approved):

PR #1986, opened 2023-04-26
Get actor of interest from scenario metadata and fix metrics

Adaickalavan:
- Added a specialised metric formula module for Driving SMARTS 2023.1 and 2023.2 benchmark.
- Made all metrics as functions to be minimised, except the overall score which is to be maximised.
- Driving SMARTS 2023.3 benchmark and the metrics module now uses `actor_of_interest_re_filter` from scenario metadata to identify the lead vehicle.
- Minor fix in regular expression compilation of `actor_of_interest_re_filter` from scenario metadata.
- Fixed acceleration and jerk computation in comfort metric, by ignoring vehicle position jitters smaller than a threshold.


smarts/env/gymnasium/wrappers/metric/metrics.py
```diff
@@ -424,17 +432,19 @@ def _check_scen(scenario: Scenario, agent_interfaces: Dict[str, AgentInterface])
         for agent_name, agent_mission in scenario.missions.items()
     }
 
+    aoi = scenario.metadata.get("actor_of_interest_re_filter", None)
     for agent_name, agent_interface in agent_interfaces.items():
         interest_criteria = agent_interface.done_criteria.interest
         if not (
-            (goal_types[agent_name] == PositionalGoal and interest_criteria is None)
+            (goal_types[agent_name] == PositionalGoal and interest_criteria == None)
```
Gamenot, line 439: The standard is to use `is None` for checking against `None` because the reference and values are both equal, but either works.

Gamenot (approved):

PR #1995, opened 2023-04-27
Fix interest done not triggering.

Gamenot:
A mistake that should have been obvious.

### DIFF
--- a/scenarios/sumo/loop/scenario.py
+++ b/scenarios/sumo/loop/scenario.py
@@ -2,6 +2,7 @@ import random
 from pathlib import Path
 
 from smarts.core import seed
+from smarts.core.colors import Colors
 from smarts.sstudio import gen_scenario
 from smarts.sstudio import types as t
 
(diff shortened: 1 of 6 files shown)


qianyi-sun (approved):
have merged this branch on my branch and tested with several scenarios, the episodes now end with the vehicle of interest done, thanks Tucker!

scenarios/sumo/loop/scenario.py
```diff
@@ -62,6 +68,7 @@
                 follow_offset=(-7, 10),
             ),
         ],
+        scenario_metadata=t.ScenarioMetadata(r"(leader)|(target)", Colors.Blue),
```
Adaickalavan, line 71: target or other_interest?

smarts/core/smarts.py
```diff
@@ -1423,7 +1423,7 @@ def _process_collisions(self):
                 actor_id = self._vehicle_index.owner_id_from_vehicle_id(collidee.id)
                 # TODO: Should we specify the collidee as the vehicle ID instead of
                 #       the agent/social ID?
-                collision = Collision(collidee_id=actor_id)
+                collision = Collision(collidee_id=actor_id or collidee.id)
```
Adaickalavan, line 1426: Perhaps choose one instead of using `or`.

Adaickalavan (approved):

smarts/core/smarts.py
Gamenot: @Adaickalavan do you prefer knowing the agent id (or other type of owner) or the vehicle id?

Gamenot: I have made it always the vehicle name.

PR #2008, opened 2023-05-02
Social agent name should map to mission.

Gamenot:
This should make `Scenario.missions[]` contain the social agent name instead of the mangled id.

### DIFF
--- a/smarts/core/scenario.py
+++ b/smarts/core/scenario.py
@@ -388,8 +388,8 @@ class Scenario:
 
         file_match = os.path.join(social_agents_path, "*.pkl")
         for missions_file_path in glob.glob(file_match):
+            count = 0
             with open(missions_file_path, "rb") as missions_file:
-                count = 0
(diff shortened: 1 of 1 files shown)


Gamenot:
It appears like this change causes a problem with combining groups of social agents.

The agent mission ids are marked like `"social-actor-"+<group id>+<actor id>`.

I can strip away the prefix "social-actor-" but then group id is still integrated. This is necessary because missions are combined between groups which may share the underlying actor name.

Gamenot:
@Adaickalavan

Adaickalavan:
Given this case, I have now resorted to searching for the vehicles of interest in `Scenario.missions.keys()` using a substring match criteria. Hence, this PR may be closed. Thanks.

Gamenot:
There is some consideration still here that the actor names should still be exactly named and unique between groups considering the actors are being combined across groups.

Issue #2004, opened 2023-04-29
Scenario issues


Adaickalavan:
### High Level Description

This is on the `merge_exit` map. Expected waypoints to behave like that shown [here](https://smarts.readthedocs.io/en/latest/_images/merge-single.gif). But got the below issues in chronological order:
+ In on-ramp, agent sees three waypoint paths. [ok]
+ In on ramp, three waypoint paths extends across the junction. [ok] 
+ In junction, agent sees single waypoint path which jumps around. [probably scenario issue]
+ After merge in highway, waypoint paths end abruptly. [probably scenario issue]
+ In highway, three short and thick lines appear. [probably scenario issue]
+ In highway, at the very end of the gif, `obs.waypoint_paths` is empty, causing the episode to crash. [probabaly scenario issue]

![Peek 2023-04-28 19-40](https://user-images.githubusercontent.com/24203326/235272307-0aee1160-3187-4206-b867-e5d9790176a1.gif)

### Version

branch: master
version: v1.0.11

### Steps to reproduce the bug

```shell
Replay the `merge_exit` scenario and visualize the waypoints.
```

### Impact (If known)

The `merge_exit` scenario is unusable.

qianyi-sun:
@Adaickalavan The issues are mostly due to overlapped edges around the junctions, if the edges are overlapped, the waypoints are overlapped as well, which leads to:
- confusion on picking the right waypoints path (jump around issue)
- dense, crowded waypoints in a small edge (three thick lines) 

To fix this, 
- modified the geometry endpoints for most of the edges to minimize the overlapping around the junction so that the waypoints are generally clearer and smoother.
- extend edges/junctions a bit for same reason above, and also can give more room/time to access the upcoming waypoints in the observations

Please take a look at the following gif and let me know if this is acceptable, I have already pushed this map onto the repo, thanks!

![merge_fix](https://user-images.githubusercontent.com/59240680/235287603-355b5eac-ec52-4497-8915-873c496554d4.gif)

Adaickalavan:
More scenario issues on straight map.

![straight](https://user-images.githubusercontent.com/24203326/235307213-07656be2-885a-40d3-842c-280b1fd7732f.gif)
![Peek 2023-04-28 22-15](https://user-images.githubusercontent.com/24203326/235307222-1c4a324c-110c-4c99-a54b-3d6573d33963.gif)

+ Use entry tactic method to ensure lead vehicle appears before ego, rather than map design methods.
[...]

PR #2009, opened 2023-05-02
Ensure waypoints are never empty

saulfield:
As mentioned in #2004, there is an edge case where waypoints can be empty when in a junction and all the potential paths have been filtered out. This PR ensures that we fall back to just returning waypoints for the nearest lane in this case.


Gamenot (approved):
Please addd a changlog entry.

Adaickalavan (approved):

Issue #1967, opened 2023-04-18
[Feature request] implement scenario_metadata color config on Envision


qianyi-sun:
### Proposal

The color set in the following code can be also visualized on Envision
```gen_scenario(
    scenario=Scenario(
        traffic=traffic,
        ego_missions=ego_missions,
        scenario_metadata=ScenarioMetadata("Leader-007", Colors.Blue),
    ),
    output_dir=Path(__file__).parent,
)
```

### Motivation

For some use case like following scenarios, users might want to clearly watch some actor(s) in the simulation. Allowing color config works on Envision might be good to improve user experience.

PR #2022, opened 2023-05-16
Add interest color to envision.

Gamenot:
#1967


Adaickalavan (approved):

Gamenot:
This ended up being abnormally tricky.
- The first issue is that all the instanced meshes share the same material (which is good for performance reasons) but it meant that changing the `diffuseColor` would cause all instances to change colour. The solution was recommended to use a colour buffer.
- However, it seems like `gltf` files are loaded with some odd scaling that causes colour buffer values to not reflect on the mesh if it is still parented to the original root node.

envision/web/src/components/vehicles.js
```diff
@@ -192,12 +210,19 @@ export default function Vehicles({
...
+            1
+          );
+        } else if (
           state.actor_type == ActorTypes.SOCIAL_VEHICLE ||
           instancedSubMesh.material.id == "body" || // Change the car body color based on actor type
           childMeshes.length == 1
         ) {
-          instancedSubMesh.material.diffuseColor = new Color3(...color);
+          instancedSubMesh.instancedBuffers.color = new Color4(...color, 1);
+          // instancedSubMesh.material.diffuseColor = new Color3(...color);
```
saulfield, line 225:
```suggestion
```

saulfield (approved):

Issue #2023, opened 2023-05-16
Missing waypoints


Adaickalavan:
### High Level Description

Several expected waypoints are missing.

### Version

1.1.0

### Steps to reproduce the bug

Replay `t3/training/ffd10ec2-715b-48af-a89d-b11f79927f63_agents_1` scenario and observe the waypoints.

### Error logs and screenshots

First an ideal image with all expected waypoints is shown below for reference.
![t_10_1](https://github.com/huawei-noah/SMARTS/assets/24203326/993e2470-c994-4ad1-8d1f-c6eb95d46902)

Screen capture of several different time instances where the expected waypoints are missing are shown below. The expected waypoints are drawn in red.

![t_1](https://github.com/huawei-noah/SMARTS/assets/24203326/d6ee77a1-114f-4a20-899f-4fb8935a4850)
![t_2](https://github.com/huawei-noah/SMARTS/assets/24203326/f58c1a0a-9ce2-4e0d-804e-1385e282e68a)
![t_3](https://github.com/huawei-noah/SMARTS/assets/24203326/7ce84278-43b7-400b-927c-c847d75decd3)
![t_4](https://github.com/huawei-noah/SMARTS/assets/24203326/7efe650c-3a20-4cab-ad9f-70f7cb201461)
![t_5](https://github.com/huawei-noah/SMARTS/assets/24203326/557a5077-6a0e-4ade-a989-5089629f0e39)
![t_6](https://github.com/huawei-noah/SMARTS/assets/24203326/4be09318-e216-4e16-95a4-e6f06a8c2513)
![t_7](https://github.com/huawei-noah/SMARTS/assets/24203326/1d4c59be-40ed-477b-b80a-92f4dba3a2b0)
![t_8](https://github.com/huawei-noah/SMARTS/assets/24203326/07df235f-d9bc-4221-b8d7-77e17908f5cc)
![t_9](https://github.com/huawei-noah/SMARTS/assets/24203326/d5e37425-0a0d-4ad9-b42f-3a60a9f3fe85)

### Impact (If known)

Expected waypoints which are missing affects the RL training and the ego's learning ability.

PR #2024, opened 2023-05-16
Argoverse and SUMO waypoint improvements

### LINKED ISSUE #2018: Abruptly breaking waypoints affect vehicle_gap metric
### High Level Description

Waypoints tend to break abruptly on straight road with junctions. 

### Version

1.1.0

### Steps to reproduce the bug

Replay latest `t3/training/merge_exit_sumo_t_agents_1` scenario with an endless mission ego agent and observe the waypoints.

### System info

_No response_

### Error logs and screenshots

The following is a sequence of ordered images in time. Each image comprise of a map and its corresponding rgb observation. Each rgb observation depicts 3 consecutive time points, with the latest being on the left.

t1

t3

t4

t6

t7

### Impact (If known)

The `vehicle_gap` metrics uses waypoints to traverse across different road segments in order to identify whether the vehicle of interest is in the same lane as the ego. When waypoints abruptly get truncated (except at the end of the map), this curtails the ability to reliably check whether the ego is properly following the vehicle of interest.

Either we fix this abruptly ending waypoints or change the method used to identify lanes across different road segments in `vehicle_gap` metric.

saulfield:
Fixes #2018 and #2023.

There were a number of issues here. One notable one was that cached lanepoints from incorrect lanes were sometimes returned. We now always generate waypoints for each junction lane (rather than for their roads), and without filtering by `road_id`. This ensures the `lookahead` value will be respected.

Videos below:

https://github.com/huawei-noah/SMARTS/assets/17256344/b7f559ff-c8c8-4c99-963c-8934fb59d1c6

https://github.com/huawei-noah/SMARTS/assets/17256344/210cd356-50f9-49e5-a115-b4a42622e43f

### DIFF
--- a/smarts/core/argoverse_map.py
+++ b/smarts/core/argoverse_map.py
@@ -922,7 +922,7 @@ class ArgoverseMap(RoadMapWithCaches):
                 self.point = point
                 self.filter_road_ids = filter_road_ids
                 self._starts = {}
-            self._starts[llp.lp.lane.index] = paths
+            self._starts[llp.lp.lane.lane_id] = paths
 
         def query(
             self,
@@ -933,7 +933,7 @@ class ArgoverseMap(RoadMapWithCaches):
         ) -> Optional[List[List[Waypoint]]]:
             """Attempt to find previously cached waypoints"""
             if self._match(lookahead, point, filter_road_ids):
-                hit = self._starts.get(llp.lp.lane.index, None)
+                hit = self._starts.get(llp.lp.lane.lane_id, None)
                 if hit:
                     # consider just returning all of them (not slicing)?
                     return [path[: (lookahead + 1)] for path in hit]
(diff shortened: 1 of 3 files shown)


smarts/core/argoverse_map.py
```diff
@@ -975,36 +972,36 @@ def waypoint_paths(
...
+            # Get set of all lanes leading through the junction
+            wp_lanes = set()
             for junction_lane in junction_lanes:
-                paths_for_lane = self._resolve_in_junction(junction_lane)
-                for path in paths_for_lane:
-                    paths.add(tuple(path))
-
-            for road_ids in paths:
-                new_paths = self._waypoint_paths_along_route(
-                    pose.point, lookahead, road_ids
```
Gamenot, line 1002: If I remember correctly, this was done to constrain the waypoints to route for cases where there is a mission. Does that still work after this change?

saulfield: That happens above in a different place. It probably never should have been used here in the first place.

Gamenot (approved):

Adaickalavan (approved):

PR #2026, opened 2023-05-17
Remove intersection-v0

Adaickalavan:
Removed `intersection-v0` env and `examples/rl/intersection` as they are superseded by `driving-smarts-v2023` env and `examples/rl/drive`, respectively.


Gamenot (approved):
This should cut down on the security warnings https://github.com/huawei-noah/SMARTS/security/dependabot.

PR #2029, opened 2023-05-18
Remove frame stack environment wrapper

Adaickalavan:
+ Removed `FrameStack` environment wrapper.
+ If needed, users are (i) more likely to build their own frame stacking functions which fit their needs, or (ii) more likely to use frame stacking functions from RL libraries of their choice.


Gamenot:
It looks like `test_rgb_image` is broken by this.

Adaickalavan:
> It looks like `test_rgb_image` is broken by this.

The `smarts/env/wrappers/rgb_image.py` and `smarts/env/tests/test_rgb_image.py` was removed as part of PR #2027 .

Gamenot (approved):

PR #2030, opened 2023-05-18
Add lane error penalty

Adaickalavan:
Here, the goal is on lane 0 but ego terminates on lane 2. 

end_episode

- Previously, the `DistToDestination` cost == `dist_along_road` which is 0 in this situation.
- Now, the `DistToDestination` cost == `dist_along_road + lane_error*lane_width` in this situation.

Changes made:
1. `RoadMap.Route` now optionally stores the start and end lanes of the route.
1. `DistToDestination` metric now adds lane error penalty when agent terminates in different lane but same road as the goal position.


smarts/core/route_cache.py
```diff
@@ -51,6 +51,8 @@ class RouteWithCache(RoadMap.Route):
     def __init__(self, road_map: RoadMap):
         self._map = road_map
         self._logger = logging.getLogger(self.__class__.__name__)
+        self._start_lane: Optional[RoadMap.Lane] = None
```
Gamenot, line 54: Should the new attributes affect the `hash`?

smarts/core/road_map.py
```diff
@@ -650,6 +650,26 @@ def __eq__(self, other) -> bool:
...
 
+        @property
+        def start_lane(self) -> Optional[RoadMap.Lane]:
+            "Route's start lane."
+            return None
+
+        @start_lane.setter
+        def start_lane(self,value:RoadMap.Lane):
+            "Route's start lane."
+            raise NotImplementedError()
```
Gamenot, line 661: I am somewhat averse to mutability here since none of the other attributes of the Route are mutable. It means that the hash has to exclude the lanes because the hash would not remain consistent.

I think the state should only be initialized.

Gamenot (changes requested):

smarts/env/gymnasium/wrappers/metric/costs.py
```diff
@@ -544,33 +544,44 @@ def get_dist(road_map: RoadMap, point_a: Point, point_b: Point) -> float:
...
+        # remaining outside the goal capture radius at all times. Default 
+        # positional goal radius is 2m.
+        dist_tot = abs(dist_tot)
+
+    # Account for agent ending in a different lane but in the same road as
+    # the goal position. 
+    start_lane = plan.route.start_lane
+    end_lane = plan.route.end_lane
+    lane_error = abs(start_lane.index-end_lane.index)
+    if len(plan.route.roads) == 1 and lane_error > 0:
```
saulfield, line 580: Does this need to support routes with more than 1 road?

Adaickalavan: Here we intend to add lane error penalty only when the ego is in the same road as the goal position, but in a different lane. Hence, we check for `len(plan.route.roads) == 1` where the computed `route` is between the final ego position and the goal position. If there are more than one road in the `route`, the lane error penalty becomes ambiguous and thus it is not implemented.

smarts/core/road_map.py
Adaickalavan: As discussed, I have made changes to make the `Route` object immutable.

saulfield (approved):

smarts/core/route_cache.py
Adaickalavan: `start_lane` and `end_lane` is now included in the hash.

PR #2040, opened 2023-05-29
Render traffic lights in camera observations

### LINKED ISSUE #1994: [Feature request] Traffic lights in camera observations
### Motivation

Traffic lights would be useful to render in the RGB or a separate camera observation so that it is possible for an agent to be able to use only cameras to operate.

### Proposal

The traffic lights can be represented by using a coloured bar similar to the proposal in #1993.

### Alternatives

It might be possible to use lane colouring to show how the agent can progress instead:

```
yellow: can cautiously progress to lane
red: lane is closed
green: lane can be entered
```

### Additional context

_No response_

saulfield:
Closes #1994.

traffic_light_obs

Scenario used for testing:
[4lane_t_signals.zip](https://github.com/huawei-noah/SMARTS/files/11602297/4lane_t_signals.zip)

### DIFF
--- a/.github/workflows/ci-auto-commit-linux.yml
+++ b/.github/workflows/ci-auto-commit-linux.yml
@@ -12,7 +12,7 @@ jobs:
   auto-commit-linux:
     runs-on: ubuntu-20.04
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-    container: huaweinoah/smarts:v0.6.1-minimal
+    container: ghcr.io/smarts-project/smarts:v0.6.1-minimal
     steps:
       - name: Install packages
         run: |
--- a/.github/workflows/ci-base-tests-linux.yml
+++ b/.github/workflows/ci-base-tests-linux.yml
@@ -9,7 +9,7 @@ jobs:
   base-tests-linux:
     runs-on: ubuntu-20.04
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-    container: huaweinoah/smarts:v0.6.1-minimal
+    container: ghcr.io/smarts-project/smarts:v0.6.1-minimal
     strategy:
       matrix:
         tests:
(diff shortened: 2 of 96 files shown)


Gamenot (approved):
Good work! My only question now is if we should change the RGB colour of the ego vehicle.

@Adaickalavan

qianyi-sun:
> Good work! My only question now is if we should change the RGB colour of the ego vehicle.
> 
> @Adaickalavan

Then everything regards ego agent color in SMARTS has to be changed, like pictures and etc, and maybe the models have to be retrained as well.

Maybe we could consider to change the color of red light instead, like adding a color in RGB like `trafficlight_red=(254,255,255)`, visually there shouldnt be much difference compared to (255,255,255), but code-wise there's difference and we dont have to touch all the current ego-agent-related stuff at all.

Adaickalavan:
Vision based policies will have difficulty distinguishing between similarly coloured objects, such as the red-coloured ego and red traffic light.

All existing documents and example gifs uses red to represent the ego. Therefore, to minimise any change needed, we may consider changing the red traffic light colour to a different shade of red such as maroon (128, 0, 0) which has non-negligible difference in rgb value compared to that of red (210, 30, 30).

Maroon-traffic-light

If the red traffic light colour is changed in the rgb image, similar change will be needed in Envision.

saulfield:
I have changed the color of the stop signal to burgundy.

{ Type Agent_0-Top-Down RGB }

Adaickalavan:
Internet search claims the following rgb values:
- Maroon = (128, 0, 0)
- Burgundy = (128, 0, 32)

I had preferred Maroon since it is further away from our current Red = (210, 30, 30). 

Our current "Burgundy" = (127, 0, 1) seems to be actually closer to Maroon and thus probably mislabelled. Our current "Burgundy" does not seem to be used anywhere in the code. Hence, we could consider rectifying the label and value.

Adaickalavan (approved):

PR #2042, opened 2023-05-29
Universal DistToDestination metric

Adaickalavan:
`DistToDestination` metric is now computed by summing the 
(i) off-route distance driven by the vehicle from its last on-route position, and 
(ii) the distance to goal from the vehicle's last on-route position.

### DIFF
--- a/smarts/env/gymnasium/wrappers/metric/costs.py
+++ b/smarts/env/gymnasium/wrappers/metric/costs.py
@@ -136,7 +136,7 @@ def _dist_to_destination(
         if not done:
             cur_pos = Point(*obs.ego_vehicle_state.position)
             cur_on_route, cur_route_lane, cur_route_lane_point, cur_route_displacement = on_route(
-                road_map=road_map, route=route, pos=cur_pos
+                road_map=road_map, route=route, point=cur_pos
             )
 
             if cur_on_route:
@@ -151,7 +151,7 @@ def _dist_to_destination(
         else:
             cur_pos = Point(*obs.ego_vehicle_state.position)
             cur_on_route, cur_route_lane, cur_route_lane_point, cur_route_displacement = on_route(
-                road_map=road_map, route=route, pos=cur_pos
+                road_map=road_map, route=route, point=cur_pos
             )
 
             # Step 1: Compute the last off-route distance driven by the vehicle, if any.
@@ -179,7 +179,7 @@ def _dist_to_destination(
 
             # Step 4: Compute lane error penalty if vehicle is in the same road as goal, but in a different lane.
             # TODO: Lane error penalty should be computed. It is not computed
-            # currently because the end lane of a SUMO traffic vehicle of 
+            # currently because the end lane of a SUMO traffic vehicle of
             # interest is currently not accessible.
             lane_error_dist = 0
             # end_lane = route.end_lane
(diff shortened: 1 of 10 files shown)


smarts/env/gymnasium/wrappers/metric/costs.py
```diff
@@ -559,30 +628,55 @@ def get_dist(road_map: RoadMap, point_a: Point, point_b: Point) -> float:
...
-        end_offset = end_lane.offset_along_lane(world_point=point_b)
-        lane_width, _ = end_lane.width_at_offset(end_offset)
-        lane_error_dist = lane_error * lane_width
-        dist_tot += lane_error_dist
-
-    return dist_tot
+    elif dist_tot < 0 and not tolerate:
+        raise CostError(
+            "Route computed in reverse direction from point_b to "
+            f"point_a resulting in negative distance: {dist_tot}."
```
Gamenot, line 642: Perhaps we should be using the term `displacement` rather than distance here since `distance` can also mean magnitude from a point.

smarts/env/gymnasium/wrappers/metric/costs.py
```diff
@@ -559,30 +628,55 @@ def get_dist(road_map: RoadMap, point_a: Point, point_b: Point) -> float:
...
+        )
+
+    return dist_tot, plan.route
+
+
+def on_route(
+    road_map: RoadMap, route: RoadMap.Route, pos: Point, radius: float = 7
```
Gamenot, line 649: Would just recommend using `point`.

```suggestion
    road_map: RoadMap, route: RoadMap.Route, point: Point, radius: float = 7
```

smarts/env/gymnasium/wrappers/metric/metrics.py
```diff
@@ -237,8 +234,22 @@ def reset(self, **kwargs):
                     raise MetricsError(
                         "Unsupported configuration for distance-to-destination cost function."
                     )
+                start_pos = Point(*self._scen.missions[agent_name].start.position)
```
Gamenot, line 237:
```suggestion
                start_pos = self._scen.missions[agent_name].start.point
```

Gamenot (approved):
Good work!

smarts/env/gymnasium/wrappers/metric/costs.py
Adaickalavan: Changed `pos` to `point`.

smarts/env/gymnasium/wrappers/metric/metrics.py
Adaickalavan: I had missed the in-built `point` method. Now changed to use the `point` method.

smarts/env/gymnasium/wrappers/metric/costs.py
Adaickalavan: If the route from `point_b` to `point_a` is not straight, then the route `distance` will be longer than the `displacement` between `point_b` and `point_a`. So rephrasing the error message as "Route computed in reverse direction from point_b to point_a resulting in negative displacement: {dist_tot}." also sounds a little awkward as the displacement might be shorter than the printed `dist_tot`.

saulfield (approved):

PR #2043, opened 2023-05-30
Handle incorrect off_route event from triggering for overlapping lanes

saulfield:


qianyi-sun (approved):

smarts/core/sensors/__init__.py
```diff
@@ -708,14 +708,20 @@ def _vehicle_is_off_route_and_wrong_way(
...
         )
 
         # No road nearby, so we're not on route!
-        if not nearest_lane:
+        if not nearest_lanes:
             return (True, False)
 
+        # Handle case where there are multiple nearest lanes the same dist away
+        min_dist = nearest_lanes[0][1]
+        tied_nearest = [lane for (lane, d) in nearest_lanes if d == min_dist]
```
Gamenot, line 721: I could suggest an allowance for slight deviation.
```suggestion
        tied_nearest = [lane for (lane, d) in nearest_lanes if math.isclose(d, min_dist)]
```

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Copy and pasting the git commit messages is __NOT__ enough.
 - `SumoTrafficSimulation` no longer reports that it manages vehicles when it is disconnected.
 - Fixed waypoints so that they will always be the length of the `lookahead` parameter, even in junctions.
 - Fixed an issue where a single waypoint would appear in off-route lanes for missions with a goal.
+- Fixed an issue where the `off_route` event would be incorrectly triggered when driving on overlapping lanes.
```
Gamenot, line 44: What I will say is that this does not solve the problem fully.

The other issue at the core is that we have overpass lanes overlapping on the main road without being separated by height layer, and without being listed as a junction. This means that just driving through the intersection between the overpass and the other highway can result in ambiguous road association.

Gamenot (approved):

PR #2056, opened 2023-06-06
Attempt to fix the package test

Gamenot:
This fixes an error with the `test-package-conflicts` by using a newer version of the minimal image.

### DIFF
--- a/.github/workflows/ci-format.yml
+++ b/.github/workflows/ci-format.yml
@@ -80,7 +80,7 @@ jobs:
   test-docs:
     runs-on: ubuntu-20.04
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-    container: ghcr.io/smarts-project/smarts:v0.6.1-minimal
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-format.yml
```diff
@@ -80,18 +80,14 @@ jobs:
...
       - name: Add safe directory
         run: git config --global --add safe.directory $GITHUB_WORKSPACE
       - name: Checkout
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Install Enchant
-        run: |
-          apt-get update --fix-missing
-          apt-get install -y libenchant-dev
```
Gamenot, line 94: This would be no longer needed with the newer minimal.

saulfield (approved):
Perhaps we should add `python3.8-dev` to the dockerfile as well, to keep them all in one place? Then we can remove it from the `test-package-conflicts` workflow.

Adaickalavan (approved):

PR #2067, opened 2023-06-15
Update `ray.rllib` to 2.5

Gamenot:


examples/rl/rllib/rllib.py
```diff
@@ -126,69 +196,95 @@ def main(
...
-        local_dir=str(result_dir),
-        resume=resume_training,
-        restore=checkpoint,
-        max_failures=3,
-        num_samples=num_samples,
-        export_formats=["model", "checkpoint"],
-        config=tune_config,
-        scheduler=pbt,
-    )
+    ## Approach 2
```
Gamenot, line 223: This approach could be useful because it makes very evident what is happening with the training since very little is automated in terms of schenduling because it does not use `Tune`. The downside is that it does not get the trialing and hyperparameter manipulation of schedulers like the `PopulationBasedTraining` scheduler that was originally used.

I am not sure if we should completely abandon this approach since it is very clear how it works. I am considering splitting this approach out.

examples/rl/rllib/rllib.py
```diff
@@ -126,69 +196,95 @@ def main(
...
+                checkpoint_dir = get_checkpoint_dir(checkpoint_iteration)
+                print(f"======= Saving checkpoint {checkpoint_iteration} =======")
+                algo.save_checkpoint(checkpoint_dir)
+                checkpoint_iteration += 1
+            current_iteration += 1
+        algo.save_checkpoint(get_checkpoint_dir(checkpoint_iteration))
+    finally:
+        algo.save(get_checkpoint_dir("latest"))
+
+    algo.stop()
```
Gamenot, line 249: This approach could be useful because it makes very evident what is happening with the training since very little is automated in terms of schenduling because it does not use `Tune`. The downside is that it does not get the trialing and hyperparameter manipulation of schedulers like the `PopulationBasedTraining` scheduler that was originally used.

I am not sure if we should completely abandon this approach since it is very clear how it works. I am considering splitting this approach out.

examples/rl/rllib/rllib.py
```diff
@@ -126,69 +196,95 @@ def main(
...
-    best_logdir = Path(analysis.get_best_logdir("episode_reward_max", mode="max"))
-    model_path = best_logdir / "model"
+    # results = tuner.fit()
+    # # Get the best result based on a particular metric.
+    # best_result = results.get_best_result(metric="episode_reward_mean", mode="max")
 
-    copy_tree(str(model_path), save_model_path, overwrite=True)
-    print(f"Wrote model to: {save_model_path}")
+    # # Get the best checkpoint corresponding to the best result.
+    # best_checkpoint = best_result.checkpoint
```
Gamenot, line 287: I have not finished testing this approach but it appears to be the modern way of working with `Tune` and more flexible than the previous two. It uses `Tune` through the `Tuner` interface. I had a bit of difficulty figuring out where the algorithm config goes.

The examples appear to show it goes in `param_space` but that is an unclear argument.

examples/rl/rllib/rllib.py
```diff
@@ -126,69 +196,95 @@ def main(
...
-    experiment_name = "rllib_example_multi"
-    result_dir = Path(result_dir).expanduser().resolve().absolute()
-    if checkpoint_num:
-        checkpoint = str(
-            result_dir / f"checkpoint_{checkpoint_num}" / f"checkpoint-{checkpoint_num}"
-        )
-    else:
-        checkpoint = None
+    # copy_tree(str(model_path), save_model_path, overwrite=True)
+    # print(f"Wrote model to: {save_model_path}")
```
Gamenot, line 221: The original approach still works after updating the configuration. It appears like the approach is not advocated in any way by the `ray` documentation. The flow is a bit different though which makes it harder to reproduce with the other approaches.

examples/rl/rllib/rllib.py
```diff
@@ -103,13 +108,78 @@ def main(
...
+            rllib_agent["observation_space"],
+            rllib_agent["action_space"],
+            {"model": {"custom_model": TrainingModel.NAME}},
+        )
+        for i in range(num_agents)
+    }
+
+    smarts.core.seed(seed)
+    algo_config = (
+        PGConfig()
```
Gamenot, line 135: The typed config appears to help with generating the example from the code side of things. I somewhat wonder about working with it from a different approach.

It also appears possible that as long as we register the algorithm it should be possible to use the [`rllib train`](https://docs.ray.io/en/latest/rllib/rllib-cli.html) to run our examples using a custom configuration.

examples/rl/rllib/rllib_agent.py
```diff
@@ -41,19 +42,38 @@
...
+        self,
+        obs_space,
+        action_space,
+        num_outputs: int,
+        model_config: ModelConfigDict,
+        name: str,
+    ):
+        super().__init__(obs_space, action_space, num_outputs, model_config, name)
+
+    def forward(self, input_dict, state, seq_lens):
```
Gamenot, line 72: One of the issues here is that the observation and action spaces are expected to be formed at this point (these are given at the start).

I could inject the action and observation space adaptors in here but this is already after the tensors have been generated... I am unsure if there is a way to intercept them. I found an experimental `observation_fn` but it appears it does not inject between the observation and the policy.

We may have to keep the action space and observation space adaptors for `RLlibHiWayEnv`.

saulfield (approved):
Looks good to me. I don't have strong opinions about which approach to use. I would probably vote for option 3 if I had to choose.

Gamenot:
> Looks good to me. I don't have strong opinions about which approach to use. I would probably vote for option 3 if I had to choose.

@saulfield OK, I will pursue this option.

I think I will still keep around option 2 but as a different primitive example.

```
rllib
\ tune_example.py
\ base_example.py
```

examples/rl/rllib/model/saved_model.pb
Adaickalavan, line 1: Since the model files were removed, the `SMARTS/examples/rl/rllib/model/README.md` file should be removed too.

Adaickalavan (approved):
We could go with the latest approach 3, and keep approach 2 as a separate example with its own test case.

Gamenot:
I have most things working, the main issue left is that it appears like the parallel environments are not getting unique `worker_index` and `vector_index` which we originally used for environment diversity within a trial.

smarts/env/rllib_hiway_env.py
```diff
@@ -75,9 +116,9 @@ def __init__(self, config):
         a = config.worker_index
         b = config.vector_index
         c = (a + b) * (a + b + 1) // 2 + b
-        smarts.core.seed(seed + c)
+        self._seed = seed + c
+        smarts.core.seed(self._seed + c)
```
Gamenot, line 120: One thing to note from this PR is that the original behaviour which used to result in environment seed diversity no longer works.

Seed used to vary across all environments during a trial but is now the same within a trial. I am unsure if this has to do with how the environment configuration is distributed.

The implication is that this slows down experience gain.

examples/rl/rllib/pg_pbt_example.py
```diff
@@ -0,0 +1,298 @@
...
+        help="Number of times to sample from hyperparameter space",
+    )
+    parser.add_argument(
+        "--save_model_path",
+        type=str,
+        default=str(Path(__file__).expanduser().resolve().parent / "model"),
+        help="Destination path of where to copy the model when training is over",
+    )
+    args = parser.parse_args()
+    build_scenario(scenario=args.scenario, clean=False, seed=42)
```
Adaickalavan, line 281: Here, using a seed, we first build the scenarios and generate a fixed traffic set. Then, we use the same scenarios in each of the the parallel environments to collect experience. Although the parallel environments might have different seeds and the scenario order might be shuffled, the underlying traffic set appears to be the same since they were generated using the same seed. Should we be building the scenarios with different seeds within each parallel environment instance?

Gamenot: That approach will work only partially because unless we use a true-random seed, the seed is managed by `tune` in the `PBT` scheduling. From my investigation, all of the environments are identical within a trial. In that case, if we are building scenarios during the experiment, I am worried about race conditions between the environments because it is not clear how to differentiate between the environments.

The second issue is that the config is specified at the start of the configuration. To do what you ask would require copying and building the scenario at the start of each trial. I think that might be possible through the callbacks, but from what I can tell, the config is intended to be constant and I would need to do some investigation to see if it is possible.

What I am slightly surprised about ([see above comment](https://github.com/huawei-noah/SMARTS/pull/2067#issuecomment-1595354965)) is that the `config.vector_index` and `config.worker_index` values do not appear to be different across workers. And `reset(seed=None)` seems to be the constant case.

The easier way of working with this is to pre-generate a set of scenario variations at the start, I will expand the number of scenarios.

PR #2073, opened 2023-06-26
Fix traffic history state not being reset

Gamenot:
This fixes an issue where the traffic history state was not reset between episodes.


Gamenot:
This was not necessary.

Issue #1981, opened 2023-04-25
Minimize core package dependencies


saulfield:
### Proposal

Most of the packages in `install_requires` should be either removed or moved to be optional dependencies.

```
setuptools>=41.0.0,!=50.0 # TODO: remove
cached-property>=1.5.2 # TODO: remove
click>=7.1.2  # core
eclipse-sumo>=1.12.0  # TODO: move
gym>=0.17.3,<=0.19.0 # TODO: remove
gymnasium==0.27.0 # TODO: move
numpy>=1.19.5,<1.24.0  # core
pandas>=1.3.4  # TODO: remove
psutil>=5.4.8 # core
pybullet>=3,<4.0 # TODO: move
rich>=11.2.0 # TODO: remove
Rtree>=0.9.7 # TODO: move to waymo, argoverse, opendrive extras
sh>=1.14.2 # TODO: remove
shapely>=2.0.0 # core, possibly remove later because of performance
tableprint>=0.9.1 # core
trimesh==3.9.29  # core
visdom>=0.1.8.9 # TODO: move
yattag>=1.14.0 # TODO: move to sstudio extra
cloudpickle>=1.3.0,<=2.1.0 # TODO: remove, due to package conflicts
tornado>=5.1.1 # TODO: move to envision extra
websocket-client>=1.2.1 # TODO: move to envision extra
ijson>=3.1.4 # TODO: move to envision extra
matplotlib>=3.2.2 # TODO: move
protobuf>=3.17.3, <4.0.0 # TODO: move to waymo extra, and remove remote agent code
PyYAML>=3.13 # core
twisted>=21.7.0 # TODO: remove
# scipy # TODO: remove 
```

PR #2074, opened 2023-06-29
Remove gym dependency

saulfield:
More progress towards #1981.

- Fix `make test`
- Change all remaining references of `gym` to `gymnasium`
- Remove `hiway-v0` env
- Remove gym v21 compatibility wrapper
- Remove gym dependency
- Made gymnasium dependency optional


saulfield:
All the examples appear to be working with `gymnasium` aside from `drive` and `platoon`. There appears to be an issue with `stable_baselines3` (see the failing CI test). I tried using v2.0.0 which added support for `gymnasium`, but the error is still happening. I've made this a draft PR since I think we should just do this all in one go, instead of leaving `gym` as a dependency for the rl examples, which would require some additional code changes.

Adaickalavan:
PR #2075 updates the `drive` and `platoon` examples to use Stable Baselines3 v2.0.0 and gymnasium.

smarts/env/gymnasium/wrappers/api_reversion.py
```diff
@@ -21,7 +21,7 @@
 # THE SOFTWARE.
 from typing import Any, Dict, SupportsFloat, Tuple
 
-import gym
+import gymnasium as gym
 
 
 class Api021Reversion(gym.Wrapper):
```
Adaickalavan, line 27: + I guess we cannot change gym to gymnasium here in the compatibility wrapper, as it is meant to support subsequent layers which use gym interface.
+ We may consider removing this wrapper entirely and let the users build their own compatibility wrapper or use compatibility wrappers provided by RL libraries of their choice.

Gamenot: That might be reasonable, although this is just a primitive wrapper. It does not do the full conversion between the two since it still passes up the `gymnasium` spaces which act like `gym` spaces but are not. It is fragile because changes to the `gymnasium.Space` interface would break this wrapper in the future.

saulfield: I have removed it.

Adaickalavan (approved):

Gamenot (approved):

setup.cfg
```diff
@@ -84,18 +78,19 @@ envision =
...
     opendrive2lanelet>=1.2.1
     Rtree>=0.9.7
-remote_agent = 
-    grpcio==1.32.0
 rllib = 
     opencv-python==4.1.2.30
     opencv-python-headless==4.1.2.30
     ray[rllib]~=2.5.0
     tensorflow-probability
+    tensorflow>=2.12.0
```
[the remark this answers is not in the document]
saulfield: I remember adding it to get a test to pass. I'll try removing it.

PR #2075, opened 2023-06-29
Upgrade SB3 to v2.0.0 in Drive and VehicleFollowing RL examples

Adaickalavan:
In Drive and VehicleFollowing RL examples
+ upgrade Stables Baselines3 v1.7.0 to v2.0.0
+ replace gym with gymnasium


saulfield (approved):
Good work! I think this will fix the remaining issues with #2074.

PR #2079, opened 2023-07-06
Add hydra configured example

Gamenot:
This example can be run like this:

`python examples/control.py episodes=1 num_agents=4 env_cfg.params.scenarios="[scenarios/sumo/loop, scenarios/sumo/figure_eight]"`

Alternatively, configuration files can be used to fill the parameters such as if you look at the temporary `click`+`hydra` integration example.


setup.cfg
```diff
@@ -77,6 +74,8 @@ envision =
     tornado>=5.1.1
     websocket-client>=1.2.1
     ijson>=3.1.4
+examples = 
+    hydra-core
```
Adaickalavan, line 78: Need to update list of examples and instructions at `SMARTS/docs` and `SMARTS/README.md` files.

examples/tools/primitive_agents.py
Adaickalavan, line 1: Can we move and merge these agents to `SMARTS/zoo/policies` ?

Gamenot: I found importing `SMARTS/zoo` to be slightly complicated but it might be most reasonable to do that, leaving one local example to show how it works. Although, I am actually not sure how useful these primitives are generally, I would also like to have ML agents available for the base experiments.

Do we have any readily available?

Adaickalavan: I am planning to create one scl-zoo-installable ML zoo agent by simply creating a packaged agent from trained `examples/RL/drive/inference` code.

Gamenot: Ok, I think that should go into this PR. Or into a mandatory follow-up.

PR #2080, opened 2023-07-07
Update panda3d for python 3.10

Gamenot:
This allows `smarts[camera_obs]` to work on `python3.10` (and hopefully 3.11).


Gamenot:
Noted problems: 

1. `pybullet` blocks using `smarts` in `3.11`.
2. `pybullet` has to compile from source for `3.10` which is slow.

PR #2112, opened 2023-11-16
Allow random routes to vary across traffic groups.

### LINKED ISSUE #2108: [Help Request] Randomize scenario
### High Level Description

I have a same demand with #1423, i wish my traffic flow's `RandomRoute()` are different in each episode instead of a fixing one.
```
1. Instantitate the smarts env with a list of scenarios, instead of a single scenario, to cycle through various maps on each episode reset call.
2. Create a list of traffic flows in scenario.py, instead of a single traffic flow, to cycle through the various traffic patterns on each episode reset call.
```
the second point i have made it, but about the first point, I only have a single scenario-4lane, and don't know how to cycle it. so i just tried to copy my scenario a few times to make it a list, obviously it failed.

### Version

1.4.0

### Operating System

_No response_

### Problems

_No response_

Gamenot:
Closes #2108

### DIFF
--- a/smarts/sstudio/genscenario.py
+++ b/smarts/sstudio/genscenario.py
@@ -34,6 +34,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import cloudpickle
+import numpy as np
 import yaml
 
 import smarts.core
@@ -245,8 +246,9 @@ def gen_scenario(
         db_conn, scenario.traffic, artifact_paths, obj_hash, map_needs_rebuild
     ):
         with timeit("traffic", logger.info):
-            for name, traffic in scenario.traffic.items():
-                derived_seed = random.randint(-0b111111111111111, 0b111111111111111)
+
+            for iteration, (name, traffic) in enumerate(scenario.traffic.items()):
+                derived_seed = seed + iteration
                 gen_traffic(
                     scenario=scenario_dir,
                     traffic=traffic,
--- a/smarts/sstudio/genscenario.py
+++ b/smarts/sstudio/genscenario.py
@@ -27,6 +27,7 @@ import itertools
 import logging
 import os
 import pickle
+import random
 import sqlite3
 from dataclasses import dataclass, replace
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


smarts/sstudio/genscenario.py
```diff
@@ -243,11 +246,12 @@ def gen_scenario(
     ):
         with timeit("traffic", logger.info):
             for name, traffic in scenario.traffic.items():
+                derived_seed = random.randint(-0b111111111111111, 0b111111111111111)
```
Adaickalavan, line 249: 1. Does the use of `random.randint()` here have any determinism issues between diferent runs?
2. If different seeds are required, perhaps we can fix them as `derived_seed = global_seed + loop_iteration_number`.
3. Instead of binary inputs to `random.randint()`, we can simply use integer inputs.

Gamenot: 1. `random.randint()` does not cause determinism issues because it is pseudo-random.
2. That could also work.
3. If I take suggestion 2 this point would be already handled.

smarts/sstudio/genscenario.py
```diff
@@ -27,14 +27,17 @@
...
 import os
 import pickle
+import random
 import sqlite3
 from dataclasses import dataclass, replace
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import cloudpickle
+import numpy as np
```
Adaickalavan, line 37: `np` seems unused.
```suggestion
```

smarts/sstudio/genscenario.py
```diff
@@ -27,14 +27,17 @@
 import logging
 import os
 import pickle
+import random
```
Adaickalavan, line 30: `random` seems unused.
```suggestion
```

Adaickalavan (approved):

smarts/sstudio/genscenario.py
```diff
@@ -242,12 +246,14 @@ def gen_scenario(
...
-            for name, traffic in scenario.traffic.items():
+
+            for iteration, (name, traffic) in enumerate(scenario.traffic.items()):
+                derived_seed = seed + iteration
                 gen_traffic(
                     scenario=scenario_dir,
                     traffic=traffic,
                     name=name,
-                    seed=seed,
+                    seed=derived_seed,
```
Adaickalavan, line 256: Perhaps we should wrap the seed here (at the point where it is introduced) instead of later where it is used in `generator.plan_and_save()`.

Gamenot: I disagree, if we wrap it at this level then the implementation will affect how we use this method.

PR #2140, opened 2024-01-20
Improve TraCI shutdown.

### LINKED ISSUE #2139: [Help Request] traci
### High Level Description

I have noticed i have the same problem with #2127,  so i update the latest smarts version. but the problems still exist.
problem 1 :
```
Could not connect to TraCI server at localhost:59573 [Errno 111] Connection refused
 Retrying in 0.05 seconds
Could not connect to TraCI server at localhost:59573 [Errno 111] Connection refused
 Retrying in 0.05 seconds
Could not connect to TraCI server at localhost:59573 [Errno 111] Connection refused
 Retrying in 0.05 seconds
```
in the beginning, the traci will tried to connect to different ports. however, after running 10 hours, the traci only tried to connect to the same port with constant failure, so my code stucked, i have to rerun my code.

problem 2 :
this problem is like the #2127, because of 
```
  File "./smart_master/SMARTS/smarts/core/sumo_traffic_simulation.py", line 239, in _initialize_traci_conn
    self._traci_conn.setOrder(0)
TypeError: 'NoneType' object is not callable
```
so my smarts can't reset successfully, i update the code, but this problem still exist occasionally.

and can i **ture off this traci warning?**  90% of my console output is traci warning, I can't find the info i really need. thank you very much

### Version

the latest v

### Operating System

ubuntu

### Problems

_No response_

Gamenot:
This should be a general improvement which will help prevent `sumo|sumo-gui` processes from leaking or becoming zombie processes.

Fixes #2139


Gamenot:
The server now manages to get through a large scale stress test (60k instances, 32 parallel) without without errors on the client or server side.

Previously there would be errors in roughly 3/1000 instances.

PR #2141, opened 2024-02-12
Refactor RL platoon example

Adaickalavan:
Use common observation utility functions in the `e11_platoon` example.


Gamenot (approved):
LGTM

PR #2165, opened 2024-11-07
Fixed chassis colissions

dduro2020:
BoxChassis has been added to a group of collisions to make them work.

BoxChassis after fix:
Screenshot from 2024-11-07 02-20-16


Gamenot:
Sorry for the very late reply. Thank you for your contribution.

